PR #227, opened 2021-07-15
Revert "Update deps, podfile to support building with xcode12.5, Update ReadMe"

DominicSherman:
Reverts cardstack/cardwallet#219


cball (commented):
Before we merge, can we:
* double check this on a local device
* send the build to testflight to confirm

topwebtek7 (approved):

PR #235, opened 2021-07-19
[CS-1289] handle keys

SparkleFaerieCoder:
Moves setup of api keys from repo

- [x] Completes #CS-1289


DominicSherman (approved):
Looks good to me! We can kick off a release with these changes to test

PR #243, opened 2021-07-21
[CS-1332] remove ability to tap prepaid card

SparkleFaerieCoder:


DominicSherman (approved):

DominicSherman:
Is there a lint for the unused onPress variable now?

PR #244, opened 2021-07-22
Temporarily rename Merchant Name into Mandello

tintinthong:
### Description

Rename hardcoded "Merchant Name" into "Mandello". Mandello represents a merchant store name. As part of the beta program, we want to show a watch address so that users can view account history from the perspective of a merchant. This is a temporary fix and is intended to be reverted post beta. 

### Checklist

- [ ] All UI changes have been tested on a small device

### Screenshots

IMG_4286
IMG_4285


DominicSherman (approved):

tintinthong:
Based on info, the merchant portion will be excluded due to app causing some crashes.

PR #246, opened 2021-07-22
Don't set Gnosis data to empty array on error and add sentry logging

DominicSherman:
Screen Shot 2021-07-22 at 09 38 13

### DIFF
--- a/src/redux/data.js
+++ b/src/redux/data.js
@@ -313,14 +313,16 @@ export const addressAssetsReceived = (
   const prepaidCards = get(message, 'payload.prepaidCards', []);
   const merchantSafes = get(message, 'payload.merchantSafes', []);
 
-  dispatch({
-    payload: {
-      depots,
-      prepaidCards,
-      merchantSafes,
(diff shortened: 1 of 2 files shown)


src/redux/data.js
```diff
@@ -313,14 +313,16 @@ export const addressAssetsReceived = (
   const prepaidCards = get(message, 'payload.prepaidCards', []);
   const merchantSafes = get(message, 'payload.merchantSafes', []);
```
lukemelia: Don't these `get` calls with the empty array fallback argument guarantee that these values will be truthy?

DominicSherman: @lukemelia My understanding was that these would only fallback to empty array if these are `undefined`, whereas on error these values will be dispatched as `null`. In my testing it seemed like that was the case (empty arrays were _not_ getting dispatched on error), but I can also have these fallback to null just to be safe.

topwebtek7 (approved):
LGTM!

PR #247, opened 2021-07-22
Fix transaction caching issues

DominicSherman:
* We were returning sokol data on xdai rather than empty data (there is no deployed subgraph for xdai)
* We were not resetting sections data for transactions if account was null

### DIFF
--- a/cardstack/src/services/transaction-service.ts
+++ b/cardstack/src/services/transaction-service.ts
@@ -105,6 +105,8 @@ const useSokolTransactions = () => {
         }
 
         setLoading(false);
+      } else if (data?.account === null) {
+        setSections([]);
       }
     };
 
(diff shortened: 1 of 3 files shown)


cardstack/src/services/transaction-service.ts
```diff
@@ -105,6 +105,8 @@ const useSokolTransactions = () => {
         }
 
         setLoading(false);
+      } else if (data?.account === null) {
```
topwebtek7, line 108: @DominicSherman it could be undefined, no? what about change like this?
```
} else if (!data?.account) {
```

topwebtek7 (approved):

cardstack/src/services/transaction-service.ts
DominicSherman: We want to check specifically for `null` since `undefined` would signal a loading state.

PR #249, opened 2021-07-22
Add custom handling for hub authentication requests

DominicSherman:
image


cardstack/src/hooks/transaction-confirmation/use-message.ts
```diff
@@ -2,11 +2,11 @@ import { useRouteParams } from './use-route-params';
 
 export const useMessage = () => {
   const {
-    transactionDetails: {
-      payload: { params },
-    },
+    transactionDetails: { payload },
   } = useRouteParams();
 
+  const { params } = payload;
```
SparkleFaerieCoder: Nice refactor 👍

cardstack/src/hooks/transaction-confirmation/use-message.ts
```diff
@@ -2,11 +2,11 @@ import { useRouteParams } from './use-route-params';
...
+
   const message = params[1].message;
```
SparkleFaerieCoder: Unless you would prefer that distinction, would it be possible to refactor this to reuse the same code between `use-message` and `use-primary-type `and  pass a  type param (`primaryType`, `message`) to grab the specified key? I think we could do something like  `params[1][type]`.

SparkleFaerieCoder (approved):
Nothing blocking, just one comment on duplicative code.

cardstack/src/hooks/transaction-confirmation/use-message.ts
DominicSherman: Sure, I think we could just return the whole params[1] object and then allow it to pull off whatever as needed 👍 I'll update

SparkleFaerieCoder: @DominicSherman I like that even better 🎉

PR #253, opened 2021-07-26
Fix prepaidcards issuername and add did-resolver function

topwebtek7:
image


cardstack/src/types/PrepaidCardType.ts
```diff
@@ -9,4 +9,5 @@ export interface PrepaidCardType {
   type: string;
   reloadable: boolean;
   transferrable: boolean;
+  cardCustomization: PrepaidCardCustomization;
```
DominicSherman: Since this field is optional, we should probably add a `?` here

DominicSherman (approved):

PR #257, opened 2021-07-27
Fix GoogleServices output path

danibonilha:
### Description

The build was falling because GoogleServices was being exported in the root instead of inside ios folder


DominicSherman (approved):
Thanks! Let's try to kick off another build once this is merged

PR #259, opened 2021-07-27
[CS-809] Add activity data to full screen depots

DominicSherman:
Screen Shot 2021-07-27 at 12 58 03

### DIFF
--- a/cardstack/src/screens/DepotScreen.tsx
+++ b/cardstack/src/screens/DepotScreen.tsx
@@ -1,11 +1,17 @@
-import React, { useState } from 'react';
 import { getConstantByNetwork } from '@cardstack/cardpay-sdk';
 import { useRoute } from '@react-navigation/native';
-import { Linking, ScrollView, StatusBar } from 'react-native';
-import isNativeStackAvailable from '../../../src/helpers/isNativeStackAvailable';
+import React, { useState } from 'react';
+import {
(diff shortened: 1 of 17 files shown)


cardstack/src/screens/DepotScreen.tsx
```diff
@@ -219,18 +229,64 @@ const Balances = ({ tokens }: BalancesProps) => {
...
+              refreshing={refetchLoading}
+              onRefresh={refetch}
+            />
+          }
+          onEndReached={onEndReached}
+          onEndReachedThreshold={1}
+          style={{ width: '100%' }}
+        />
+      )}
+    </Container>
```
danibonilha, line 282: Performance wise, it's best to extract the anonymous functions and objects to useMemos and useCallbacks, but taking in consideration the release term we can worry about this later.

danibonilha (approved):
LGTM

cardstack/src/screens/DepotScreen.tsx
DominicSherman: In general I only utilize these when I see over rendering or performance issues, based on the thoughts expressed in this article: https://kentcdodds.com/blog/usememo-and-usecallback/

If you have any additional thoughts though and want to have a discussion on it to set a standard I'm totally up for that!

PR #260, opened 2021-07-27
[CS 1359]: Update QR icon and add text

danibonilha:
### Description
Updates QR icon svg, adds "Scan QR" text beside the icon and align vertically Avatar and icon on center instead of bottom

- [x] Completes #(CS-1359)

### Screenshots
Before/After alignment:
Screen Shot 2021-07-27 at 17 48 47

### DIFF
--- a/cardstack/src/components/Icon/custom-icons/qr-code.tsx
+++ b/cardstack/src/components/Icon/custom-icons/qr-code.tsx
@@ -1,53 +1,42 @@
 import * as React from 'react';
-import Svg, { SvgProps, G, Path, Rect } from 'react-native-svg';
+import Svg, { SvgProps, Path, Circle } from 'react-native-svg';
 import { palette } from '@cardstack/theme';
 
 function SvgComponent(props: SvgProps) {
   return (
     <Svg
-      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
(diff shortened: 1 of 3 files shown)


SparkleFaerieCoder (approved):
Nice job!

cardstack/src/components/Icon/custom-icons/qr-code.tsx
```diff
@@ -1,53 +1,42 @@
...
+        data-name="Ellipse 1300"
+        cx={3}
+        cy={3}
+        r={3}
+        transform="translate(7 7)"
+        fill={props.color || palette.tealLight}
+      />
     </Svg>
   );
 }
```
danibonilha, line 42: @DominicSherman what are the benefits of setting up svgs like this instead of using `SvgXml `? Also are we required to have the `xmlns` prop for some reason?

DominicSherman: These icons are the result of copying the SVG code from XD and then pasting it into the React Native SVG converter application. I'm not sure what the benefits are but it's been working for us 🤷‍♂️ 

I'm not sure if the `xmlns` prop is necessary, I believe it broke something when we removed but if it doesn't then go ahead and remove

PR #262, opened 2021-07-28
Update prepaidcard customization with dynamic values

topwebtek7:
Screen Shot 2021-07-29 at 7 58 39 PM


cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -21,14 +22,41 @@ import {
...
+  <Text
+    {...props}
+    style={{
+      textShadowColor: '#ffffff',
+      textShadowOffset: {
+        width: 0,
+        height: 1,
+      },
+      textShadowRadius: 0,
+    }}
```
danibonilha: Since these style properties are "hardcoded" we can create this outside the component using StyleSheet to avoid recreating this object every-time the component re-renders

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -152,50 +181,156 @@ export const PrepaidCard = (props: PrepaidCardProps) => {
...
-        id="Union_18"
-        data-name="Union 18"
-        d="M 0 164.992 v -0.127 H 0 V 0 H 139.563 s 13.162 0.132 24.094 12.362 s 15.768 15.605 15.768 15.605 s 7.3 8.09 22.43 8.452 H 411 l -0.064 128.572 Z"
-        fill="#fff"
+const GradientBackground = ({ cardCustomization }: CardGradientProps) => {
+  let gradientValues: Array<string> = [];
+
+  const hasGradient = cardCustomization?.background.startsWith(
+    'linear-gradient'
+  );
```
lukemelia, line 195: Currently we support solid colors and linear gradients that have exactly 2 color stops. We should do more validation on `cardCustomization.background` so that the type of input we support is more explicit and we can error gracefully and emit a useful error if we are presented with a background that does not fit into one of these categories.

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -152,50 +181,156 @@ export const PrepaidCard = (props: PrepaidCardProps) => {
...
-      id="Bottom_platter"
-      data-name="Bottom platter"
-      transform="translate(0 71)"
-    >
-      <Path
-        id="Union_18"
-        data-name="Union 18"
-        d="M 0 164.992 v -0.127 H 0 V 0 H 139.563 s 13.162 0.132 24.094 12.362 s 15.768 15.605 15.768 15.605 s 7.3 8.09 22.43 8.452 H 411 l -0.064 128.572 Z"
-        fill="#fff"
+const GradientBackground = ({ cardCustomization }: CardGradientProps) => {
```
lukemelia: CustomizableBackground` would be a better name for this component, since it only sometimes renders a gradient now.

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -152,50 +181,156 @@ export const PrepaidCard = (props: PrepaidCardProps) => {
...
+        patternColor={cardCustomization?.patternColor}
+      />
+      <G
+        id="Bottom_platter"
+        data-name="Bottom platter"
+        transform="translate(0 71)"
+      >
+        <Path
+          id="Union_18"
+          data-name="Union 18"
```
lukemelia: I know that these `data-name` attributes are not new in this PR, but do they have any purpose? Similarly, does this path need an `id` attribute or can it be removed?

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -21,14 +22,41 @@ import {
...
+  ScrollView,
+  Text,
+  TextProps,
+} from '@cardstack/components';
+
+const TextOverGrad = (props: TextProps) => (
+  <Text
+    {...props}
+    style={{
+      textShadowColor: '#ffffff',
```
lukemelia: @backspace @Aierie what rule did we decide on with respect to the text shadow color on the prepaid card? Is it just the opposite of the text color?

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -152,50 +181,156 @@ export const PrepaidCard = (props: PrepaidCardProps) => {
...
       >
-        <Text size="xs" weight="extraBold">
-          {issuer}
-        </Text>
+        <TextOverGrad
+          size="xs"
+          weight="extraBold"
+          color={cardCustomization?.textColor as ColorTypes}
+        >
+          {cardCustomization?.issuerName || '...'}
```
lukemelia: People seem confused by `...`. Let's try making it `Unknown` if the issuerName doesn't exist.

lukemelia (changes requested):

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
Aierie: Use pattern color as text shadow color, design has mentioned that we can assume that text shadow color will always match the pattern color.

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -152,50 +181,156 @@ export const PrepaidCard = (props: PrepaidCardProps) => {
...
+    'linear-gradient'
+  );
+
+  if (hasGradient) {
+    gradientValues =
+      cardCustomization?.background
+        .substring(16, cardCustomization?.background.length - 1)
+        .split(',')
+        .map((value: string) => value.trim()) || [];
+  }
```
danibonilha: We could memoize this, so it won't be recomputed on every new render. One other thing what 16 means ? Maybe assign this value to a const could make clear that clear

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -152,50 +181,156 @@ export const PrepaidCard = (props: PrepaidCardProps) => {
...
+        await fetch(
+          uri.startsWith('http') ? uri : `https://app.cardstack.com${uri}`
+        )
+      ).text();
+
+      setPattern(response);
+    } catch {
+      setPattern(null);
+    }
+  };
```
danibonilha, line 281: We could wrap this in a useCallback to keep the reference and only update if uri prop changes

package.json
```diff
@@ -291,6 +291,7 @@
     "@types/react-native-vector-icons": "^6.4.6",
     "@types/react-test-renderer": "^16.9.3",
     "@types/semver": "^7.3.6",
+    "@types/styled-components": "^5.1.11",
```
danibonilha, line 294: I believe we can remove this dependency

danibonilha (dismissed):
Overall lgtm, just some comments for improvements, but I would wait for reviews from people that are in the project longer

package.json
topwebtek7: I think we can keep it as it annoys with styled-components types when checking rainbow components

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
topwebtek7: basically it is being called only once inside useEffect without deps, so will be updated only when component mounted, so I don't believe we need useCallback here..

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
topwebtek7: @lukemelia all current gradient types are all 2 color stops and so worked ok with tests.., can we merge it for now and push a new testflight today, and get back to it later?

lukemelia: Sure, please create a Linear issue to cover this in Current Sprint

lukemelia (approved):

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -21,14 +22,42 @@ import {
...
+} from '@cardstack/components';
+
+const styles = StyleSheet.create({
+  TextOverGrad: {
+    textShadowColor: '#ffffff',
+    textShadowOffset: {
+      width: 0,
+      height: 1,
+    },
+    textShadowRadius: 0,
```
lukemelia, line 39: @topwebtek7 did you see @Aierie's note about using the patternColor for the text shadow?

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -152,50 +182,152 @@ export const PrepaidCard = (props: PrepaidCardProps) => {
...
+  const hasGradient = cardCustomization?.background.startsWith(
+    'linear-gradient'
+  );
+
+  const gradientValues = useMemo(() => {
+    if (!hasGradient) return [];
+
+    return (
+      cardCustomization?.background
+        .substring(16, cardCustomization?.background.length - 1)
```
lukemelia: As @danibonilha suggested, let's add a named constant for this `16` to be more expressive.

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -21,14 +22,42 @@ import {
...
+  Container,
+  Icon,
+  ScrollView,
+  Text,
+  TextProps,
+} from '@cardstack/components';
+
+const styles = StyleSheet.create({
+  TextOverGrad: {
+    textShadowColor: '#ffffff',
```
DominicSherman: If possible, could we see if we can take advantage of [restyle](https://github.com/Shopify/restyle) for styling? It might not support this in which case this fine 👍

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
topwebtek7: done

PR #264, opened 2021-07-28
[CS-1333, CS-1334] Fix sending on layer 2

DominicSherman:


src/redux/gas.js
```diff
@@ -148,7 +149,7 @@ export const gasPricesStartPolling = () => async (dispatch, getState) => {
             type: GAS_PRICES_SUCCESS,
           });
         } else {
-          const apiBaseUrl = getConstantByNetwork('apiBaseUrl', network);
+          const apiBaseUrl = getConstantByNetwork('apiBaseUrl', Network.xdai);
```
SparkleFaerieCoder, line 152: From what I recall, just setting this still fixes the gas fee to be USD. Is that what's wanted, or should be it converted to Wei?

DominicSherman: This returns the gas price in wei and there is already handling around this conversion to USD or any other currency

SparkleFaerieCoder: @kdowlin Do you recall what the issue was? We changed the network on Layer 2 to xDai like above, but something didn't add up right.

kdowlin: Currently it is meant to convert to an as converted in USD. As long as the on converted to amount is greater that 0.00 (which I believe it should be close to 0.03USD on xDai and Sokol equiv) that works fine to show it as USD.

DominicSherman: @kdowlin @SparkleFaerieCoder 

image

kdowlin: That works for now.

SparkleFaerieCoder (approved):

PR #274, opened 2021-07-30
Fix no way to unhide if all cards are hidden in prepaid, balance

topwebtek7:
Screen Shot 2021-07-30 at 8 55 58 PM


DominicSherman (approved):

DominicSherman:
Only thing to check is if we want the count shown next to the `Prepaid Cards` header to show all prepaid cards or just visible ones?

lukemelia:
Is this something that we could write a test for?

PR #275, opened 2021-07-30
[CS-1404] - Update release commands

danibonilha:
### Description

Updates commands to use same from package.json to keep it consistent and  adds cache for Pods folder making the release build 13 min faster.


lukemelia (approved):
🎁  Yay for 13 minutes less waiting!

ios/Podfile.lock
```diff
@@ -1038,4 +1038,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 888ba22051801541c4ba2dbecd4a0736f9bead9d
 
-COCOAPODS: 1.10.1
+COCOAPODS: 1.9.1
```
topwebtek7, line 1041: @danibonilha do we need this change in this pr? why need to back to pod 1.9.1?

topwebtek7 (approved):
LGTM

ios/Podfile.lock
danibonilha: The cocoapods in gemfile is 1.9.1, and in order to cache we are using the Podfile.lock, if the gemfile cocoapods and Podfile.lock divert the cache won't work.

topwebtek7: I thought pod version should be 1.10.1 and gemfile is not updated.. somehow let's look at it together today which version would be

PR #278, opened 2021-08-02
[CS-1279] Update ImportSeed screen description dynamic based on selected chain

topwebtek7:


src/screens/ImportSeedPhraseSheet.js
```diff
@@ -106,8 +106,23 @@ const Sheet = styled(Column).attrs({
...
+      return 'Enter seed phrase, private key, Ethereum address, or ENS name';
+    case networkTypes.sokol:
+      return 'Enter seed phrase, private key, or Sokol address';
+    case networkTypes.xdai:
+      return 'Enter seed phrase, private key, or xDai address';
+    default:
+      return 'Enter seed phrase, private key, Ethereum address, or ENS name';
+  }
+};
+
```
danibonilha, line 123: Can we add some tests to this ? Maybe it could an object/keys

topwebtek7: I think it's better than object/keys in terms of exception handling and being used in this file only, so not sure if needs test with this only. though we will discuss regarding writing tests in today standup

danibonilha (approved):

PR #282, opened 2021-08-03
[CS-1438, CS-1457] Fix svg rendering on prepaid card for smaller pattern, fix issues with invalid did format

topwebtek7:
image


DominicSherman (approved):
Maybe we can come back and refactor this component to extract some of the logic out so it doesn't get too big?

topwebtek7:
> Maybe we can come back and refactor this component to extract some of the logic out so it doesn't get too big?

I did some refactoring in prepaid card component and fixed possible such bugs as well. so it would be all good now.

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -198,29 +199,32 @@ const CustomizableBackground = ({
   cardCustomization,
   isEditing,
 }: CardGradientProps) => {
-  // ToDo: add more validation here, it supports gradients with only 2 color stops atm
-  const hasGradient = cardCustomization?.background.startsWith(
+  const hasGradient = !!cardCustomization?.background?.startsWith(
```
lukemelia, line 202: @topwebtek7 Suggestion for a refactor here. Introduce a util function: `parseLinearGradient(candidateBackgroundValue: string) => LinearGradientInfo | undefined`. It would handle parsing the value into the constituent parts of the linear gradient, or if it is not a linear gradient, or not one in the format we support, it would return `undefined`. That would move a lot of the complex and conditional logic out of this file and keep things cleaner.

topwebtek7: @lukemelia ok, that's good idea! will do that in a separate pr

PR #283, opened 2021-08-03
[CS-1394, CS-1408] Add prepaid card details modal

DominicSherman:
Screen Shot 2021-08-03 at 13 41 56
Screen Shot 2021-08-03 at 12 45 26

### DIFF
--- a/cardstack/src/screens/PrepaidCardModal.tsx
+++ b/cardstack/src/screens/PrepaidCardModal.tsx
@@ -1,0 +1,79 @@
+import { useRoute } from '@react-navigation/core';
+import React from 'react';
+import { SectionList } from 'react-native';
+import { TransactionListLoading } from '../components/TransactionList/TransactionListLoading';
+import {
+  CenteredContainer,
+  Container,
+  PrepaidCard,
+  PrepaidCardProps,
+  SafeAreaView,
+  SheetHandle,
+  Text,
+  TransactionItem,
+} from '@cardstack/components';
+import { usePrepaidCardTransactions } from '@cardstack/services';
+
+const PrepaidCardModal = () => {
+  const {
+    params: { prepaidCardProps },
+  } = useRoute() as { params: { prepaidCardProps: PrepaidCardProps } };
+
+  const { sections } = usePrepaidCardTransactions(prepaidCardProps.address);
+
+  return (
+    <SafeAreaView backgroundColor="black" flex={1} width="100%">
+      <PrepaidCard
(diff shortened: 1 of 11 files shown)


cardstack/src/services/transaction-service.ts
```diff
@@ -180,10 +189,101 @@ const useSokolTransactions = (safeAddress?: string) => {
   };
 };
 
-export const useTransactions = (safeAddress?: string) => {
+export const usePrepaidCardTransactions = (prepaidCardAddress: string) => {
```
DominicSherman, line 192: I'd like to come back and refactor this a bit more. It's very difficult to share much between these but I'd like to break the other one out into separate hooks rather than shared and reuse whatever I can across all three

cardstack/src/services/transaction-service.ts
```diff
@@ -180,10 +189,101 @@ const useSokolTransactions = (safeAddress?: string) => {
...
+  const client = getApolloClient(network);
+
+  const { data, error } = useGetPrepaidCardHistoryDataQuery({
+    client,
+    variables: {
+      address: prepaidCardAddress,
+    },
+  });
+
+  if (error) {
```
DominicSherman, line 218: I believe we have a ticket for handling error screens on transaction activity data, as of now we are just logging it

kdowlin: Adding a ticket CS-1474

topwebtek7 (approved):
LGTM!

cardstack/src/screens/PrepaidCardModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+        minHeight="100%"
+        backgroundColor="white"
+        flexDirection="column"
+        width="100%"
+      >
+        <SheetHandle />
+        <Container width="100%" paddingVertical={4} paddingHorizontal={5}>
+          <Text size="medium">Recent Activity</Text>
+        </Container>
+        {!sections ? (
```
danibonilha, line 45: If sections is an array, shouldn't we validate on length just in case we have an empty array?

danibonilha (approved):
LGTM, just a small comment about sections validation

cardstack/src/screens/PrepaidCardModal.tsx
DominicSherman: This null check is to handle loading, but if it has been set to an empty array then we have a `ListEmptyComponent` within the section list that handles that scenario.

PR #284, opened 2021-08-03
[CS-1406] Fix sending of assets in Layer 2

DominicSherman:
We were previously dividing by 10 based off what Rainbow does, but we don't want to do that because it was miscalculating and giving less than 1 gwei as our gas price. Also updated the fallback so we should always have a gas price of at least 1

https://user-images.githubusercontent.com/17347720/128089338-f68eca99-ccf0-483d-b08f-4c80a0b958a2.mp4

Screen Shot 2021-08-03 at 14 32 38


kdowlin (approved):
LGTM

PR #286, opened 2021-08-04
[CS-1194] Fix sending of SPOA

DominicSherman:
Need to handle instances of hard coding to check for `ETH` rather than the given native token for the network

### DIFF
--- a/src/handlers/web3.js
+++ b/src/handlers/web3.js
@@ -22,7 +22,7 @@ import AssetTypes from '../helpers/assetTypes';
 import NetworkTypes from '../helpers/networkTypes';
 import smartContractMethods from '../references/smartcontract-methods.json';
 import { ethereumUtils } from '../utils';
-import { isLayer1 } from '@cardstack/utils';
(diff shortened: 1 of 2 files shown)


src/handlers/web3.js
```diff
@@ -311,8 +311,9 @@ export const getTransferTokenTransaction = async transaction => {
  * @param {Object} transaction { asset, from, to, amount, gasPrice }
```
topwebtek7, line 311: @DominicSherman could you update this as well cause params are changed?

src/handlers/web3.js
```diff
@@ -381,6 +382,7 @@ export const getDataForNftTransfer = (from, to, asset) => {
  */
```
topwebtek7, line 384: same for this as well

topwebtek7 (approved):
left a couple small comments, others looks good to me

src/handlers/web3.js
DominicSherman: Updated

DominicSherman: Updated

PR #291, opened 2021-08-05
Handle ellipses on issuer names if too long on prepaid card

DominicSherman:
Screen Shot 2021-08-04 at 15 10 48


danibonilha (approved):
LGTM

PR #292, opened 2021-08-05
[CS-1514]: Update firebase cloud functions paths 

danibonilha:
### Description

The new firebase functions doesn't allow names with dash, so after deploying the new versions the dashed names needed to be updated to camelCase , took the opportunity to create a own model for firebase functions this way we have everything in the same place. 

- [x] Completes #(CS-1514)

PS: Was unable to check is the tests are passing, I suspected it has something to do with the path alias we are using with ts


topwebtek7 (approved):
LGTM

PR #293, opened 2021-08-06
[CS-1417]: Disabling automatic fontscale

eshinkawa:
### Description

Fixing automatic fontscalling when font size is changed through accessibility options

### Checklist

### DIFF
--- a/cardstack/src/components/Text/Text.ts
+++ b/cardstack/src/components/Text/Text.ts
@@ -17,6 +17,9 @@ export type TextProps = ShopifyTextProps<Theme> &
   };
 
 const BaseText = createText<Theme>();
+BaseText.defaultProps = {
+  allowFontScaling: false,
+};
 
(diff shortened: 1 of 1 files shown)


cardstack/src/components/Text/Text.ts
```diff
@@ -17,6 +17,8 @@ export type TextProps = ShopifyTextProps<Theme> &
   };
 
 const BaseText = createText<Theme>();
+BaseText.defaultProps = BaseText.defaultProps || {};
+BaseText.defaultProps.allowFontScaling = false;
```
danibonilha: How about ?
```suggestion
BaseText.defaultProps = {
  allowFontScaling: false,
};
```

This way we don't need to check for undefined and give a default empty object

danibonilha (approved):
LGTM, just a small suggestion. Also let's just make sure that the legacy `Text.js` is handling this as well. (I assume it is)

PR #294, opened 2021-08-06
[WIP] Fix WC issue with different accounts on a wallet - [CS-1286]

topwebtek7:


src/model/wallet.ts
```diff
@@ -565,7 +587,7 @@ export const createWallet = async (
           () =>
             Alert.alert(
               'Oops!',
-              'Looks like you already imported this account!'
+              'Looks like you already imported this wallet!'
```
kdowlin: @topwebtek7 We want to leverage the word Account for Card Wallet. This is what we are calling a wallet for end users.

src/model/wallet.ts
```diff
@@ -1170,7 +1197,7 @@ export const loadSeedPhraseAndMigrateIfNeeded = async (
         seedPhrase = get(migratedSecrets, 'seedphrase', null);
       } else {
         logger.sentry('Migrated flag was set but there is no key!', id);
-        captureMessage('Missing seed for account');
+        captureMessage('Missing seed for wallet');
```
kdowlin, line 1200: revert to account

src/model/wallet.ts
```diff
@@ -1199,7 +1226,7 @@ export const loadSeedPhraseAndMigrateIfNeeded = async (
         logger.sentry('got seed succesfully');
       } else {
         captureMessage(
-          'Missing seed for account - (Key exists but value isnt valid)!'
+          'Missing seed for wallet - (Key exists but value isnt valid)!'
```
kdowlin: revert to account

src/model/wallet.ts
kdowlin: We should change this to 'Missing secret phrase for account'

src/model/wallet.ts
topwebtek7: ok, got it.

topwebtek7:
Refer to https://linear.app/cardstack/issue/CS-1325/bug-if-i-connect-my-wallet-to-web-dapp-using-wallet-connect-but-open
https://linear.app/cardstack/issue/CS-1286/account-changes-can-cause-inconsistent-ui-in-workflows

src/entities/index.ts
```diff
@@ -9,3 +9,4 @@ export type {
   UniswapSubgraphAsset,
   RawUniswapSubgraphAsset,
 } from './tokens';
+export type EthereumAddress = string;
```
danibonilha, line 12: nit: I find odd creating a type that's is just a string, although I understand for readability purposes, would make sense to have an interface, since we are always using with an address keyword

```ts
export interface EthereumAddress { 
     address: string;
}
```
Feel free to disconsider

src/helpers/signingWallet.ts
```diff
@@ -0,0 +1,128 @@
...
+import {
+  signingWalletAddress,
+  signingWallet as signingWalletKeychain,
+} from '../utils/keychainConstants';
+import { EthereumAddress } from '@rainbow-me/entities';
+import AesEncryptor from '@rainbow-me/handlers/aesEncryption';
+import { addHexPrefix } from '@rainbow-me/handlers/web3';
+import { ethereumUtils, logger } from '@rainbow-me/utils';
+
+export async function getPublicKeyOfTheSigningWalletAndCreateWalletIfNeeded(): Promise<EthereumAddress> {
```
danibonilha: This function is always call inside if's which makes clear that it's been called ifNeeded, maybe we can shrink the name a little bit

src/helpers/signingWallet.ts
```diff
@@ -0,0 +1,128 @@
+import { verifyMessage, Wallet } from '@ethersproject/wallet';
+import { generateMnemonic } from 'bip39';
+import { default as LibWallet } from 'ethereumjs-wallet';
+// @ts-ignore
```
danibonilha: What is this ts-ignore for ?

src/helpers/signingWallet.ts
```diff
@@ -0,0 +1,128 @@
...
+  signingWallet as signingWalletKeychain,
+} from '../utils/keychainConstants';
+import { EthereumAddress } from '@rainbow-me/entities';
+import AesEncryptor from '@rainbow-me/handlers/aesEncryption';
+import { addHexPrefix } from '@rainbow-me/handlers/web3';
+import { ethereumUtils, logger } from '@rainbow-me/utils';
+
+export async function getPublicKeyOfTheSigningWalletAndCreateWalletIfNeeded(): Promise<EthereumAddress> {
+  let alreadyExistingWallet = await loadString(signingWalletAddress);
+
```
danibonilha: I find kind of dangerous reassigning a value to return, where we could mistakenly  change this value, maybe this could a `const`, and in case `typeof alreadyExistingWallet` is different from `string` we early `return address`, and if not we `return alreadyExistingWallet` as default outside the if

src/helpers/signingWallet.ts
```diff
@@ -0,0 +1,128 @@
...
+    );
+
+    await saveString(signingWalletAddress, address, publicAccessControlOptions);
+    alreadyExistingWallet = address;
+  }
+  logger.log('Signing wallet already existing');
+  return alreadyExistingWallet;
+}
+
+export async function getSignatureForSigningWalletAndCreateSignatureIfNeeded(
```
danibonilha: It seems this function is never called, where is it being used ? Is it for a future feature ? if so could you add a TODO this way we don't forget it's going to be used

src/helpers/signingWallet.ts
```diff
@@ -0,0 +1,128 @@
...
+      alreadyExistingEncodedSignature
+    );
+    if (
+      address ===
+      verifyMessage(publicKeyForTheSigningWallet, decryptedSignature)
+    ) {
+      return decryptedSignature;
+    } else {
+      logger.log('Signature does not match. Creating a new one.');
+      alreadyExistingEncodedSignature = null;
```
danibonilha: Why do we need to reset this to null ?

src/helpers/signingWallet.ts
```diff
@@ -0,0 +1,128 @@
...
+      return createSignature(address);
+    }
+  } else {
+    return createSignature(address);
+  }
```
danibonilha: Also, if when matching a certain condition we have a `return decryptedSignature`, do we need those `else`s ?  If we remove the else's and do a default `return createSignature(address)` does it have the same expected behavior ?

src/helpers/signingWallet.ts
```diff
@@ -0,0 +1,128 @@
...
+): Promise<string> {
+  const encryptedPrivateKeyOfTheSigningWallet = await loadString(
+    signingWalletKeychain,
+    publicAccessControlOptions
+  );
+  const encryptor = new AesEncryptor();
+  const decryptedPrivateKeyOfTheSigningWallet = await encryptor.decrypt(
+    RAINBOW_MASTER_KEY,
+    encryptedPrivateKeyOfTheSigningWallet
+  );
```
danibonilha: nit: Do we need a `new AesEncryptor()` for each method ?  Maybe we can abstract that, or have a encryptor at the top

src/helpers/signingWallet.ts
```diff
@@ -0,0 +1,128 @@
...
+      logger.log('Signature does not match. Creating a new one.');
+      alreadyExistingEncodedSignature = null;
+      return createSignature(address);
+    }
+  } else {
+    return createSignature(address);
+  }
+}
+
+export async function signWithSigningWallet(
```
danibonilha: it seems this function is not being used also

src/model/wallet.ts
```diff
@@ -339,18 +357,22 @@ export const signPersonalMessage = async (
...
+      existingWallet || (await loadWallet(undefined, true, provider));
     if (!wallet) return null;
     try {
       const pkeyBuffer = toBuffer(addHexPrefix(wallet.privateKey));
       let parsedData = message;
-      if (typeof message === 'string') {
-        parsedData = JSON.parse(message);
-      }
+      try {
+        parsedData = typeof message === 'string' && JSON.parse(message);
```
danibonilha: is it okay for `parsedData` to be` boolean` ? bc if message is not a string it will assign false to `parsedData`

src/model/wallet.ts
```diff
@@ -233,10 +240,12 @@ export const loadWallet = async (): Promise<null | Wallet> => {
 export const sendTransaction = async ({
   transaction,
   existingWallet,
+  provider,
 }: TransactionRequestParam): Promise<null | Transaction> => {
   try {
     logger.sentry('about to send transaction', transaction);
-    const wallet = existingWallet || (await loadWallet());
+    const wallet =
+      existingWallet || (await loadWallet(undefined, true, provider));
```
danibonilha, line 248: nit: if `loadWallet` received object params would be easier to understand what we are passing, bc just by taking a look at it we would be able to see `{ address: undefined,  showErrorIfNotLoaded: true, provider }`  or even better we wouldn't need to pass the address at all

danibonilha (commented):
I raised some concerns which I think need to be addressed before approval, also how can we test this ? Could you add a .gif of the working flow please?

src/model/wallet.ts
topwebtek7: yeah nice catch!

src/model/wallet.ts
kdowlin: I may be wrong here, but I think we want to be using signingWallet here - because that should be defined as the wallet that you've chosen as the initial wallet to connect, vs existingWallet which will be the last active wallet shown in the application?

topwebtek7:
@kdowlin did you have a chance to check with the build from this branch? do you think we can merge it or wait till wc 2.0 release?

kdowlin:
@topwebtek7 I think we wait until WC2.0. When I tested the build it did not function as anticipated.

lukemelia:
What is our plan for this PR?

topwebtek7:
> What is our plan for this PR?

we're gonna update this pr again when walletconnect v2.0 is released with significant fix that we're experiencing now with multiple accounts

danibonilha:
Can we close this PR ? We can keep the branch though, but probably lots of thing will change anyway

topwebtek7:
> Can we close this PR ? We can keep the branch though, but probably lots of thing will change anyway

Yes agreed. we will recall this when WC 2.0 released

PR #297, opened 2021-08-09
Claim revenue functionality

DominicSherman:
image

### DIFF
--- a/cardstack/src/screens/MerchantScreen.tsx
+++ b/cardstack/src/screens/MerchantScreen.tsx
@@ -56,14 +56,26 @@ export default function MerchantScreen() {
   );
 }
 
+const useMerchantSafe = () => {
+  const {
+    params: {
+      merchantSafe: { address },
+    },
+  } = useRoute<RouteType>();
+
+  const merchantSafes = useRainbowSelector(state => state.data.merchantSafes);
+
+  const merchantSafe = merchantSafes.find(
+    safe => safe.address === address
+  ) as MerchantSafeType;
+
+  return merchantSafe;
+};
+
 const Header = () => {
   const { goBack } = useNavigation();
 
-  const {
-    params: { merchantSafe },
-  } = useRoute<RouteType>();
-
-  const { address } = merchantSafe;
+  const { address } = useMerchantSafe();
   const network = useRainbowSelector(state => state.settings.network);
   const blockExplorer = getConstantByNetwork('blockExplorer', network);
 
@@ -150,9 +162,7 @@ const MerchantInfo = () => (
 );
 
 const LifetimeEarningsSection = () => {
-  const {
-    params: { merchantSafe },
-  } = useRoute<RouteType>();
+  const merchantSafe = useMerchantSafe();
 
   const { navigate } = useNavigation();
 
@@ -193,9 +203,7 @@ const LifetimeEarningsSection = () => {
 };
 
 const UnclaimedRevenueSection = () => {
-  const {
-    params: { merchantSafe },
-  } = useRoute<RouteType>();
+  const merchantSafe = useMerchantSafe();
 
   const { navigate } = useNavigation();
 
@@ -230,9 +238,7 @@ const UnclaimedRevenueSection = () => {
 };
 
 const AvailableBalancesSection = () => {
-  const {
-    params: { merchantSafe },
-  } = useRoute<RouteType>();
+  const merchantSafe = useMerchantSafe();
 
   const { tokens } = merchantSafe;
 
--- a/src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
+++ b/src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
@@ -1,5 +1,10 @@
-import React, { useEffect } from 'react';
+import { getSDK } from '@cardstack/cardpay-sdk';
+import HDWalletProvider from 'parity-hdwallet-provider';
+import React, { useCallback, useEffect, useState } from 'react';
 import CoinIcon from 'react-coin-icon';
+import Web3 from 'web3';
+import { getSeedPhrase } from '../../../src/model/wallet';
+import { ethereumUtils } from '../../../src/utils';
 import { SlackSheet } from '../sheet';
 import {
   Button,
@@ -8,7 +13,11 @@ import {
   Text,
 } from '@cardstack/components';
 import { MerchantSafeType, TokenType } from '@cardstack/types';
+import { web3ProviderSdk } from '@rainbow-me/handlers/web3';
+import { useWallets } from '@rainbow-me/hooks';
 import { useNavigation } from '@rainbow-me/navigation';
+import { useRainbowSelector } from '@rainbow-me/redux/hooks';
+import logger from 'logger';
 
 const CHART_HEIGHT = 600;
 
@@ -17,6 +26,9 @@ export default function UnclaimedRevenueExpandedState(props: {
 }) {
   const merchantSafe = props.asset;
   const { setOptions } = useNavigation();
+  const [loading, setLoading] = useState(false);
+  const { selectedWallet } = useWallets();
+  const network = useRainbowSelector(state => state.settings.network);
 
   useEffect(() => {
     setOptions({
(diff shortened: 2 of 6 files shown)


src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -26,6 +38,38 @@ export default function UnclaimedRevenueExpandedState(props: {
...
+        providerOrUrl: web3ProviderSdk,
+      });
+      const web3 = new Web3(hdProvider);
+      const revenuePool = await getSDK('RevenuePool', web3);
+
+      const promises = revenueBalances.map(token =>
+        revenuePool.claim(
+          merchantSafe.address,
+          token.tokenAddress,
+          Web3.utils.toWei(token.balance.amount)
```
habdelra, line 77: This is gonna cause the transaction to fail because you won't have any tokens left to pay for gas. You'll need to estimate how much gas is going to cost and then deduct that from this amount. To do that I think you'll need a new API from the SDK. why don't you open a ticket for that and I'll work on it.

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -26,6 +38,47 @@ export default function UnclaimedRevenueExpandedState(props: {
...
+          token.tokenAddress,
+          Web3.utils.toWei(
+            (Number(token.balance.amount) - Number(gasEstimate)).toString()
+          )
+        );
+      });
+
+      await Promise.all(promises);
+    } catch (error) {
+      logger.sentry('Error claiming revenue', error);
```
topwebtek7, line 93: I think we need a proper error handling here so can show users why it failed..
basically, I think also we need to hide claim button when there's no valid unclaimed amount

habdelra (approved):
This looks good 👍

habdelra:
oops I didn't mean to press the "re-request review" button sorry about that.

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -26,6 +38,47 @@ export default function UnclaimedRevenueExpandedState(props: {
...
+        providerOrUrl: web3ProviderSdk,
+      });
+      const web3 = new Web3(hdProvider);
+      const revenuePool = await getSDK('RevenuePool', web3);
+
+      const promises = revenueBalances.map(async token => {
+        const gasEstimate = await revenuePool.claimGasEstimate(
+          merchantSafe.address,
+          token.tokenAddress,
+          // divide amount by 2 for estimate since we can't estimate the full amount and the amoutn doesn't affect the gas price
```
danibonilha: nit: amoutn typo

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -26,6 +38,47 @@ export default function UnclaimedRevenueExpandedState(props: {
...
+          Web3.utils.toWei((Number(token.balance.amount) / 2).toString())
```
habdelra: oh wait--actually this is a big int. let use BigNumber to do the math here. something like:
```js
Web3.utils.toWei(BigNumber.from(token.balance.amount).div(BigNumber.from("2")).toString());
```

danibonilha (approved):
LGTM

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
danibonilha: Will we need to do this on other places ? If so it would be nice to create a util for that

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
habdelra: right now this ability to estimate is very specific to claiming tokens. if you anticipate that you'll need to claim tokens in other parts of the app then yes you'll probably want to refactor. if this is the only place that tokens are claimed then i'm not sure how much utility that provides.

As we see the need to provide gas estimates for other safe related activities we can look into how we could generalize the safe txn estimation capabilities in the SDK, which would probably guide how you want to share the capability better in the app. Estimation, though, does require the specific params unique to each API so the  method signature for the gas estimation of paying a merchant, for instance, would look very different for the method signature for claiming revenue. So I'm not sure right now how best to generalize this. I think we'll need to encounter it a couple more times before we can get a good handle on how best to generalize it in the SDK.

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
habdelra: right now this ability to estimate is very specific to claiming tokens. if you anticipate that you'll need to claim tokens in other parts of the app then yes you'll probably want to refactor. if this is the only place that tokens are claimed then i'm not sure how much utility that provides.

As we see the need to provide gas estimates for other safe related activities we can look into how we could generalize the safe txn estimation capabilities in the SDK, which would probably guide how you want to share the capability better in the app. Estimation, though, does require the specific params unique to each API so the method signature for the gas estimation of paying a merchant, for instance, would look very different for the method signature for claiming revenue. So I'm not sure right now how best to generalize this. I think we'll need to encounter it a couple more times before we can get a good handle on how best to generalize it in the SDK.

cardstack/src/screens/MerchantScreen.tsx
```diff
@@ -56,14 +56,26 @@ export default function MerchantScreen() {
...
+const useMerchantSafe = () => {
   const {
-    params: { merchantSafe },
+    params: {
+      merchantSafe: { address },
+    },
   } = useRoute<RouteType>();
 
-  const { address } = merchantSafe;
+  const merchantSafes = useRainbowSelector(state => state.data.merchantSafes);
```
DominicSherman, line 66: We need to switch these to pull the merchant safe out of redux rather than params so that when we update assets in the background this will update as well

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -8,15 +14,26 @@ import {
...
+import { useRainbowSelector } from '@rainbow-me/redux/hooks';
+import logger from 'logger';
 
 const CHART_HEIGHT = 600;
 
 export default function UnclaimedRevenueExpandedState(props: {
   asset: MerchantSafeType;
 }) {
-  const merchantSafe = props.asset;
+  const merchantSafes = useRainbowSelector(state => state.data.merchantSafes);
```
DominicSherman, line 29: Same thing here

src/redux/fallbackExplorer.js
```diff
@@ -331,15 +331,258 @@ const fetchAssetBalances = async (tokens, address, network) => {
   }
 };
 
+export const fetchAssetsBalancesAndPrices = async () => {
```
DominicSherman, line 343: This had to get moved out to be consumed by the claim all function so we can refresh all the data after claiming revenue. Ideally we'd be able to break this all out into smaller pieces so we could refresh only the merchant safe data but that is too large of an undertaking for this pr. cc @danibonilha

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -26,6 +43,59 @@ export default function UnclaimedRevenueExpandedState(props: {
...
+        await revenuePool.claim(
+          merchantSafe.address,
+          token.tokenAddress,
+          claimAmount
+        );
+      });
+
+      await Promise.all(promises);
+
+      await fetchAssetsBalancesAndPrices();
```
DominicSherman, line 91: refresh balances so UI updates

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
DominicSherman: https://linear.app/cardstack/issue/CS-1568/add-error-alert-dropdowns-for-failures-on-api-calls

Ran into issues trying to implement in this PR, so added a ticket and we'll cover this in a follow-up

topwebtek7 (approved):
LGTM

PR #298, opened 2021-08-09
Refactor transactions for scalability

DominicSherman:
Previously I was hoping to be able to share more across different locations where we were pulling transactions. It's looking like almost every instance will need its own query so breaking them each out into their own hooks with a helper for what can be shared.


cardstack/src/utils/sorting-utils.ts
```diff
@@ -0,0 +1,6 @@
+export const sortByTime = (a: any, b: any) => {
+  const timeA = Number(a.timestamp || a.minedAt || a.createdAt);
+  const timeB = Number(b.timestamp || b.minedAt || b.createdAt);
```
danibonilha, line 3: the params are typed as any but we use properties of an object, shouldn't we type that ?

cardstack/src/hooks/transactions/use-depot-transactions.tsx
```diff
@@ -0,0 +1,60 @@
...
+    isFetchingMore,
+    onEndReached: () => {
+      if (!isFetchingMore && fetchMore) {
+        fetchMore({
+          variables: {
+            skip: transactionsCount,
+          },
+        });
+      }
+    },
```
danibonilha: nit: if this will have always have the same behavior, we could have a hook that returns a useCallback and wraps this, or we could just wrap this in a useCallback itself to avoid losing reference

cardstack/src/hooks/transactions/use-depot-transactions.tsx
```diff
@@ -0,0 +1,60 @@
...
+    refetch,
+    refetchLoading: networkStatus === NetworkStatus.refetch,
+    sections: sections,
```
danibonilha: nit: could use shorthand

cardstack/src/hooks/transactions/use-prepaid-card-transactions.tsx
```diff
@@ -0,0 +1,98 @@
...
+        setSections([]);
+      }
+    };
+
+    setSectionsData();
+  }, [currencyConversionRates, nativeCurrency, accountAddress, data]);
+
+  return {
+    sections: sections,
+  };
```
danibonilha, line 49: ditto

cardstack/src/hooks/transactions/use-depot-transactions.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const client = getApolloClient(network);
+
+  const {
+    data,
+    networkStatus,
+    fetchMore,
+    refetch,
+    error,
+  } = useGetSafeTransactionHistoryDataQuery({
```
topwebtek7: @DominicSherman well don't you think we need to use a separate query that returns only depot transactions? e.g. like GetPrepaidCardHistoryData

cardstack/src/hooks/transactions/use-transaction-sections.tsx
```diff
@@ -0,0 +1,83 @@
...
+          logger.log('Error setting sections data', e);
+        }
+
+        setLoading(false);
+      } else if (isEmpty) {
+        setSections([]);
+      }
+    };
+
+    setSectionsData();
```
danibonilha, line 82: This piece is exactly the same as the other one the only difference are the transactions, is there a way to reuse this ?

cardstack/src/hooks/transactions/use-depot-transactions.tsx
DominicSherman: I can rename this

cardstack/src/utils/sorting-utils.ts
DominicSherman: The result from the group by that uses this has a very weird type which made this difficult to type correctly. if you wanted to help me debug maybe we could figure out how to add types for this?

topwebtek7 (approved):
Good refactoring! 👍🏼

cardstack/src/utils/sorting-utils.ts
danibonilha: Maybe we could cast `a.data[0] as SomeType` , but I won't block you're PR over this since I noticed we have multiple places where the transaction typing is a bit out of place, we can refactor that later

danibonilha (approved):

PR #299, opened 2021-08-10
[CS-1496]: Wallet account name and account names UI fixes

eshinkawa:
### Description

Account name and wallet name with ellipsis and proper formatting

- [X] Completes #1496 #1561

### Checklist

### Screenshots

Aug-10-2021 12-54-15

### DIFF
--- a/src/components/change-wallet/AddressRow.js
+++ b/src/components/change-wallet/AddressRow.js
@@ -8,6 +8,7 @@ import { ContactAvatar } from '../contacts';
 import ImageAvatar from '../contacts/ImageAvatar';
 import { TruncatedAddress } from '../text';
 import { Container, Icon, Text } from '@cardstack/components';
+import { screenWidth } from '@cardstack/utils';
 import networkInfo from '@rainbow-me/helpers/networkInfo';
 import { useAccountSettings } from '@rainbow-me/hooks';
 import { abbreviations } from '@rainbow-me/utils';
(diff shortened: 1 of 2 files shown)


src/components/change-wallet/AddressRow.js
```diff
@@ -84,9 +85,14 @@ export default function AddressRow({ data, editMode, onPress, onEditWallet }) {
                 value={label || ens || `${index + 1}`}
               />
             )}
-            <Container margin={1}>
+            <Container
+              margin={1}
+              width={isReadOnly ? screenWidth * 0.44 : screenWidth * 0.65}
```
topwebtek7: @eshinkawa basically 0.44, 0.65 does not make sense.. could we define constants for these with meaningful name?

src/components/profile/ProfileMasthead.js
```diff
@@ -191,8 +192,20 @@ export default function ProfileMasthead({
         onPress={handlePressAvatar}
       />
       <ButtonPressAnimation onPress={handlePressChangeWallet}>
-        <Container alignItems="center" flexDirection="row">
-          <Text color="white" fontSize={24} fontWeight="600" marginRight={1}>
+        <Container
+          alignItems="center"
+          flexDirection="row"
+          justifyContent="center"
+          width={screenWidth * 0.85}
```
topwebtek7: same here. I think it's better to test in smaller screen as well

topwebtek7 (approved):
a couple comments. Looks good to me except that.

PR #301, opened 2021-08-10
Fix pricing and charts for depot tokens

DominicSherman:
Screen Shot 2021-08-10 at 12 23 33

### DIFF
--- a/src/redux/fallbackExplorer.js
+++ b/src/redux/fallbackExplorer.js
@@ -1,4 +1,8 @@
-import { delay, getConstantByNetwork } from '@cardstack/cardpay-sdk';
+import {
+  convertAmountToNativeDisplay,
+  delay,
+  getConstantByNetwork,
+} from '@cardstack/cardpay-sdk';
 import { BigNumber } from '@ethersproject/bignumber';
 import { Contract } from '@ethersproject/contracts';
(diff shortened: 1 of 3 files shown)


src/redux/fallbackExplorer.js
```diff
@@ -540,10 +556,25 @@ export const fallbackExplorerInit = () => async (dispatch, getState) => {
...
+                const token = depot.tokens[j];
+
+                if (toLower(token.coingecko_id) === toLower(coingeckoId)) {
+                  depots[i].tokens[j].chartPrices = chartData[coingeckoId];
+                } else if (token.coingecko_id === null) {
+                  depots[i].tokens[j].chartPrices = null;
+                }
+              }
+            }
+          }
```
danibonilha, line 577: Do we need to loop everything over again just to add the chartData ? Since for `depots` native price and coingecko_id,  we are using the same `coingecko_id` condition, maybe inside of it we could just check if `chartData` and add what we need, this way we save one loop = )

danibonilha (approved):
just a suggestion , overall lgtm

src/redux/fallbackExplorer.js
DominicSherman: After discussing with @danibonilha, we'll fix this among other things in a follow-up refactor pr

PR #302, opened 2021-08-11
[CS-1576] Refactor transaction confirmation decoding to use strategies

DominicSherman:
Screen Shot 2021-08-11 at 12 58 47
Screen Shot 2021-08-11 at 12 59 26
Screen Shot 2021-08-11 at 13 03 07
Screen Shot 2021-08-11 at 13 03 52
Screen Shot 2021-08-11 at 13 04 50
Screen Shot 2021-08-11 at 13 05 27
Screen Shot 2021-08-11 at 13 06 03
Screen Shot 2021-08-11 at 13 06 15
Screen Shot 2021-08-11 at 13 08 23


cardstack/src/transaction-confirmation-strategies/transfer-prepaid-card-1-strategy.ts
```diff
@@ -0,0 +1,36 @@
+import { TransferPrepaidCard1DecodedData } from '../types/transaction-confirmation-types';
+import { BaseStrategy } from './base-strategy';
+import { decodeParameters } from './decoding-utils';
+import { TransactionConfirmationType } from '@cardstack/types';
+
+const TRANSFER_PREFIX = '0xe318b52b';
```
lukemelia: This seems dicey. Should the SDK own this magic hex value?

lukemelia (approved):
This came out so beautifully, it brings tears to my eyes. 😿  OK, maybe not tears, but it is really satisfying in its elegance. Left one comment I think we should address, otherwise good to go.

cardstack/src/transaction-confirmation-strategies/transfer-prepaid-card-1-strategy.ts
habdelra: instead of the magic string, use this instead: 
```js
import Web3 from 'web3';
Web3.eth.abi.encodeFunctionSignature('swapOwner(address,address,address)'); 
```
from my solidity repl, i see this: 
```sh
truffle(xdai)> web3.eth.abi.encodeFunctionSignature('swapOwner(address,address,address)')
'0xe318b52b'
```
that way it's no longer magic.

habdelra: (there is probably an ethers.js way of doing this too if you guys are using that)

PR #305, opened 2021-08-12
[CS-1592] Refactor transactions mapping to use strategies

topwebtek7:
Simulator Screen Shot - iPhone 12 Pro Max - 2021-08-13 at 03 47 51

Simulator Screen Shot - iPhone 12 Pro Max - 2021-08-13 at 03 47 22


cardstack/src/transaction-strategies/context.ts
```diff
@@ -0,0 +1,66 @@
...
+  PrepaidCardCreationStrategy,
+  PrepaidCardPaymentStrategy,
+  PrepaidCardTransferStrategy,
+  BridgeToLayer1EventStrategy,
+  BridgeToLayer2EventStrategy,
+  MerchantCreationStrategy,
+  ERC20TokenStrategy,
+];
+
+export class TransactionContext {
```
lukemelia: Consider renaming to something like TransactionMappingContext, and corresponding directories as well. I think the name needs a verb, to describe the action it is responsible for. A comment in this file describing the purpose would be helpful, too.

cardstack/src/transaction-strategies/context.ts
```diff
@@ -0,0 +1,66 @@
...
+            });
+
+            if (strategy.handlesTransaction()) {
+              const mappedTransaction = await strategy.mapTransaction();
+
+              return mappedTransaction;
+            }
+          }
+
+          return null;
```
lukemelia: Do we expect some transactions to have no strategy that handles them? If not, should we log an error to Sentry here?

lukemelia (changes requested):
Nice application of the strategy pattern 👍  Left a couple of suggestions.

cardstack/src/transaction-strategies/context.ts
topwebtek7: Updated with more meaningful names and corresponding directories, added sentry logs when no strategy that handles transaction

cardstack/src/transaction-strategies/transaction-mapping-strategy-types/bridge-to-layer1-strategy.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    if (bridgeToLayer1Events?.[0]) {
+      return true;
+    }
+
+    return false;
+  }
+
+  async mapTransaction(): Promise<DepotBridgedLayer1TransactionType | null> {
+    const bridgeToLayer1Events = this.transaction.bridgeToLayer1Events?.[0];
```
DominicSherman: Should this be `bridgeToLayer1Event` since it's a single item?

cardstack/src/transaction-strategies/transaction-mapping-strategy-types/bridge-to-layer2-strategy.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    if (bridgeToLayer2Events?.[0]) {
+      return true;
+    }
+
+    return false;
+  }
+
+  async mapTransaction(): Promise<DepotBridgedLayer2TransactionType | null> {
+    const bridgeToLayer2Events = this.transaction.bridgeToLayer2Events?.[0];
```
DominicSherman: Same thing here

cardstack/src/transaction-strategies/transaction-mapping-strategy-types/erc20-token-strategy.ts
```diff
@@ -0,0 +1,101 @@
...
+  TransactionStatus,
+} from '@cardstack/types';
+import { fetchHistoricalPrice } from '@cardstack/services';
+import { TokenTransferFragment } from '@cardstack/graphql';
+
+export class ERC20TokenStrategy extends BaseStrategy {
+  handlesTransaction(): boolean {
+    const { tokenTransfers } = this.transaction;
+
+    if (tokenTransfers?.length) {
```
DominicSherman: One thing we need to be sure of is that there aren't other fields on here. I _think_ most transactions have token transfers on them. We'll either need to make sure we check this last or add prelim checks to ensure this isn't any other transaction type

tsconfig.json
```diff
@@ -104,6 +104,12 @@
       "@cardstack/hooks": [
         "cardstack/src/hooks"
       ],
+      "@cardstack/transaction-strategies": [
+        "cardstack/src/transaction-strategies"
```
DominicSherman: Should we update this directory to be `transaction-mapping-strategies`?

DominicSherman (approved):
Just a few comments, looks good!

cardstack/src/transaction-strategies/transaction-mapping-strategy-types/erc20-token-strategy.ts
topwebtek7: yeah, good point. will make sure to check token transfer at the end

cardstack/src/transaction-strategies/context.ts
topwebtek7: @lukemelia updated directories and names, added helpful comments as well

lukemelia (approved):
👍

PR #306, opened 2021-08-13
[CS-1555] Update address format in wallet list

topwebtek7:
- Update address format 0xXX…XXXXXX to 0xXX…XXXX
- Replaced old addresses with new TruncatedAddress component in cardstack

image


danibonilha (commented):
LGTM, but shouldn't we also edit when a account doesn't have a name ?  and check other places for 4,6
image
Taking this in consideration to keep consistent is there a away to have a util where we have this 4,4 as default ?

topwebtek7:
@kdowlin do you think all address need to be 0xXX…XXXX format?

kdowlin:
> @kdowlin do you think all address need to be 0xXX…XXXX format?

@topwebtek7 Yes let's make this a global update. Good catch @danibonilha

kdowlin:
@topwebtek7 I just spoke with Chris - we need to do a global change to 0xXXXX...XXXX.

topwebtek7:
@kdowlin changed to 0xXXXX...XXXX. please check above screenshot

kdowlin:
This looks good from a UI perspective.

src/components/change-wallet/AddressRow.js
```diff
@@ -97,12 +96,7 @@ export default function AddressRow({ data, editMode, onPress, onEditWallet }) {
                   {cleanedUpLabel || ens}
                 </Text>
               ) : (
-                <TruncatedAddress
-                  address={address}
-                  firstSectionLength={4}
-                  fontWeight="600"
-                  truncationLength={6}
-                />
```
danibonilha, line 105: We are removing the props here, but we are never formatting the `address` is this the expected behavior ?  Because we only format the `accountSubLabel` for layer 2

src/hooks/useAccountProfile.js
```diff
@@ -40,10 +40,10 @@ export default function useAccountProfile() {
 
   const accountName = removeFirstEmojiFromString(
     network === networkTypes.mainnet
-      ? label || accountENS || address(accountAddress, 6, 4)
+      ? label || accountENS || getAddressPreview(accountAddress)
       : label === accountENS
-      ? address(accountAddress, 6, 4)
-      : label || address(accountAddress, 6, 4)
+      ? getAddressPreview(accountAddress)
+      : label || getAddressPreview(accountAddress)
```
danibonilha, line 46: nit: can we make this condition more readable ?

src/components/text/TruncatedAddress.js
```diff
@@ -11,11 +12,13 @@ export default function TruncatedAddress({
...
-            truncationLength,
-            firstSectionLength
-          )
+        ? truncationLength || firstSectionLength
+          ? abbreviations.formatAddressForDisplay(
+              address,
+              truncationLength,
+              firstSectionLength
+            )
+          : getAddressPreview(address)
```
danibonilha: On multiple places the `truncationLength`  and `firstSectionLength` prop was removed, so if we want to have the same behavior for all `TruncatedAddress` component, shouldn't we use `getAddressPreview` as default and remove these props that are not being used anymore ?

src/components/change-wallet/AddressRow.js
topwebtek7: firstSectionLength={4}
truncationLength={6}
these are removed cause same as default values and it should be formated as well to make it as your attached screenshot in case there's no specified cardname

src/components/text/TruncatedAddress.js
topwebtek7: we use getAddressPreview as default and use other format only when there's special needs with truncationLength, firstSectionLength params, was able to find only couple places that use different format in rainbow code, not completely sure if we're using that part in our app

danibonilha: It would be nice to check if we will ever need to use other address format, if not, we can totally remove this since we don't need to keep rainbow code we don't use

danibonilha (approved):
LGTM 🚀

PR #307, opened 2021-08-13
[CS-1558] Update support email address, add alias for constants path

topwebtek7:
Update support email to appfeedback@cardstack.com, add contstants alias path


danibonilha (approved):
LGTM

PR #308, opened 2021-08-13
[CS-1331] - Adds the ability to send tokens from depot to EOA or another depot

danibonilha:
### Description

Adds the ability to send tokens from depot to EOA or another depot.
This allows to transfer the token without handling the gas estimate, this is not the final code, just a workaround to make both flows work conditionally, the next step will be split the logic into a SendScreen to handle EOA's  and SendScreenDepot to handle depot, this is so we can release the ability to transfer in the next build, so if there's no blocker to this then improvements to the code will be handled on different PR's

- [x] Completes #(CS-1331)

### Screenshots

output

### DIFF
--- a/cardstack/src/models/safes-providers.ts
+++ b/cardstack/src/models/safes-providers.ts
@@ -1,0 +1,55 @@
+import { getSDK } from '@cardstack/cardpay-sdk';
+import HDWalletProvider from 'parity-hdwallet-provider';
+import Web3 from 'web3';
+import { getSeedPhrase, RainbowWallet } from '@rainbow-me/model/wallet';
+import logger from 'logger';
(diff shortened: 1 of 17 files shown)


cardstack/src/models/safes-providers.ts
```diff
@@ -0,0 +1,55 @@
...
+    return hdProvider;
+  } catch (e) {
+    logger.error('Unable to getSeedPhrase', e);
+  }
+};
+
+export const getSafesInstance = async (
+  signedProviderParams?: SignedProviderParams
+) => {
+  let web3Prodiver = web3ProviderSdk;
```
DominicSherman: nit: `web3Provider`

DominicSherman (approved):
LGTM! 🚀

PR #310, opened 2021-08-16
[CS-1571] Redesign QR scanning screen

topwebtek7:
Updated QR scanning screen

image


kdowlin:
UI looks good.

cardstack/src/components/Sheet/Sheet.tsx
```diff
@@ -10,6 +11,7 @@ export const Sheet = ({
   borderRadius = 39,
   children,
   hideHandle = false,
+  style,
```
DominicSherman: We could just extend the `ContainerProps` and pass those through that we still are using the restyle library for our styling

src/components/walletconnect-list/WalletConnectExplainerItem.js
```diff
@@ -1,35 +1,31 @@
...
-            lineHeight="loose"
-            size="smedium"
-          >
-            {content}
-          </Text>
-        </FlexItem>
-        {children && <FlexItem marginTop={8}>{children}</FlexItem>}
-      </Column>
-    </Row>
+    <Column
```
DominicSherman: Could we refactor this to use `Container` and `Text` from `@cardstack/components`?

topwebtek7: @DominicSherman refactored using `Container` and `Text` from `@cardstack/components`

DominicSherman (approved):

PR #311, opened 2021-08-16
[CS-1538] Lifetime earnings chart

DominicSherman:
Screen Shot 2021-08-16 at 11 04 07

https://user-images.githubusercontent.com/17347720/129609387-829fbaad-8df6-4624-a00f-002c8424ea2d.mp4


cardstack/src/screens/MerchantScreen.tsx
```diff
@@ -191,12 +202,28 @@ const LifetimeEarningsSection = () => {
...
+          <Container alignItems="center" justifyContent="center" width="100%">
+            <ChartPath
+              data={{ points: data, smoothingStrategy: 'bezier' }}
+              gestureEnabled={false}
+              height={125}
+              stroke={palette.tealDark}
+              strokeLinecap="round"
+              strokeLinejoin="round"
+              strokeWidth={3.5}
+              width={screenWidth - 40}
```
danibonilha: How did you came up with 40 ? Would it make more sense to use some kind of percentage, to deal with all screen sizes?

danibonilha (approved):
LGTM, just a small comment, also could you document/comment the date accumulations, timestamps stuff, so it's easier for us if we need to get back to it, thanks. Oh and let's make sure the PR verify is passing before merging

cardstack/src/screens/MerchantScreen.tsx
DominicSherman: The padding we pass to the container is 5, and that padding is multiplied by 4 with our theme, so the horizontal padding of the container is 20 pixels each side. I can extract this to make it more readable 👍

PR #312, opened 2021-08-16
Update cardpay-sdk and did-resolver to 0.21.0 and use new name for ExchangeRate SDK

lukemelia:
This is part of CS-1582. Note that I have not verified this PR by running the app so I'd appreciate it if a reviewer could do that.


cardstack/src/services/exchange-rate-service.ts
```diff
@@ -18,9 +18,9 @@ export const getNativeBalance = async (props: {
   // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
   // @ts-ignore
```
topwebtek7, line 19: we really don't need these above 2 commentted lines, I think

topwebtek7 (approved):
Tried a build on this branch and played on simulator, all looks good to me!

cardstack/src/services/exchange-rate-service.ts
lukemelia: 👍  I'll remove them in a separate PR

PR #313, opened 2021-08-17
[CS-1553] Fix bug with transaction history and new prepaid card rendering

topwebtek7:
- Fix transactions history cache issue that didn't render new transaction after added new card unless restart app
- Fix card svg rendering updating RN-SVG key prop


eshinkawa (approved):

eshinkawa:
LGTM

cardstack/src/components/PrepaidCard/PrepaidCard.tsx
```diff
@@ -70,6 +70,7 @@ export interface PrepaidCardProps extends PrepaidCardType, ContainerProps {
 interface CardGradientProps {
```
lukemelia, line 70: This interface name no longer makes sense. Let's change it.

topwebtek7: ok, will do in a separate pr

PR #314, opened 2021-08-17
Fix issue when rendering new card confirmation sheet

topwebtek7:
- fixed an issue that I noticed when test adding new card(depot's undefined)


cardstack/src/components/TransactionConfirmationSheet/IssuePrepaidCardDisplay.tsx
```diff
@@ -48,7 +48,7 @@ const FromSection = ({ tokenAddress }: { tokenAddress: string }) => {
   const depots = useRainbowSelector(state => state.data.depots);
   const depot = depots[0];
 
-  const token = depot.tokens.find(
+  const token = (depot?.tokens || []).find(
     t => t.tokenAddress.toLowerCase() === tokenAddress.toLowerCase()
   );
```
danibonilha: We have a util called `getDepotTokenByAddress` maybe we could handle the empty case there too and replace those `find` with the util in other places that is being used .

eshinkawa (approved):

PR #316, opened 2021-08-17
[CS-1479] Merchant transactions updates

DominicSherman:
- [x] Add transactions to life time earnings sheet
- [x] Update unclaimed revenue transactions to properly display claimed and earned transactions

This is a bit complicated because we are handling the same transaction (prepaid card payment) 3 different ways, soon to be 4.

If we are viewing from the lifetime earnings sheet, we want to show it as an earned spend transaction

Screen Shot 2021-08-17 at 11 48 52

If we are viewing it from the unclaimed revenue sheet, we want to show it as an earned tokens transaction

Screen Shot 2021-08-17 at 11 49 11

In a future ticket, we will have to handle showing it as both on an EOA screen, only if the prepaid card payment target is one of our merchant safes. 

image

We also still want to show valid prepaid card payments correctly on the EOA screen.

image

Open to suggestions on a better way to implement this, but for now I have split out the handling of an earned spend transaction and earned revenue transaction into separate strategies and I pass the strategies I want to use when pulling the transactions. This allows us to map the transaction correctly based on where we are using it and display correctly.

### DIFF
--- a/cardstack/src/components/Transactions/MerchantEarnedRevenueTransaction.tsx
+++ b/cardstack/src/components/Transactions/MerchantEarnedRevenueTransaction.tsx
@@ -1,22 +1,21 @@
 import React from 'react';
 
-import { SafeHeader } from '../SafeHeader';
 import {
   TransactionBase,
   TransactionBaseCustomizationProps,
 } from './TransactionBase';
-import { MerchantClaimType } from '@cardstack/types';
-import { CoinIcon } from '@cardstack/components';
+import { CoinIcon, SafeHeader } from '@cardstack/components';
+import { MerchantEarnedRevenueTransactionType } from '@cardstack/types';
 
-export interface MerchantClaimTransactionProps
+export interface MerchantEarnRevenueTransactionProps
   extends TransactionBaseCustomizationProps {
-  item: MerchantClaimType;
+  item: MerchantEarnedRevenueTransactionType;
 }
 
-export const MerchantClaimTransaction = ({
+export const MerchantEarnedRevenueTransaction = ({
   item,
   ...props
-}: MerchantClaimTransactionProps) => {
+}: MerchantEarnRevenueTransactionProps) => {
   return (
     <TransactionBase
       {...props}
@@ -26,9 +25,9 @@ export const MerchantClaimTransaction = ({
       Header={
         <SafeHeader address={item.address} rightText="MERCHANT NAME" small />
       }
-      primaryText={`- ${item.balance.display}`}
+      primaryText={`+ ${item.balance.display}`}
       statusIconName="arrow-down"
-      statusText="Claimed"
+      statusText="Earned"
       subText={item.native.display}
       transactionHash={item.transactionHash}
     />
--- a/cardstack/src/components/Transactions/MerchantEarnedSpendTransaction.tsx
+++ b/cardstack/src/components/Transactions/MerchantEarnedSpendTransaction.tsx
@@ -1,0 +1,34 @@
+import React from 'react';
+
+import { Icon } from '../Icon';
+import {
+  TransactionBase,
+  TransactionBaseCustomizationProps,
+} from './TransactionBase';
+import { MerchantEarnedSpendTransactionType } from '@cardstack/types';
+import { SafeHeader } from '@cardstack/components';
+
+export interface MerchantEarnSpendTransactionProps
+  extends TransactionBaseCustomizationProps {
+  item: MerchantEarnedSpendTransactionType;
+}
+
+export const MerchantEarnedSpendTransaction = ({
+  item,
+  ...props
+}: MerchantEarnSpendTransactionProps) => {
+  return (
+    <TransactionBase
+      {...props}
+      CoinIcon={<Icon name="spend" />}
+      Header={
+        <SafeHeader address={item.address} rightText="MERCHANT NAME" small />
+      }
+      primaryText={`+ ${item.spendBalanceDisplay}`}
+      statusIconName="arrow-down"
+      statusText="Earned"
+      subText={item.nativeBalanceDisplay}
+      transactionHash={item.transactionHash}
+    />
+  );
+};
--- a/cardstack/src/components/Transactions/TransactionItem.tsx
+++ b/cardstack/src/components/Transactions/TransactionItem.tsx
@@ -1,15 +1,18 @@
 import React from 'react';
-import { ERC20Transaction } from './ERC20Transaction';
+
+import { DepotBridgedLayer1Transaction } from './DepotBridgedLayer1Transaction';
 import { DepotBridgedLayer2Transaction } from './DepotBridgedLayer2Transaction';
-import { MerchantCreationTransaction } from './MerchantCreationTransaction';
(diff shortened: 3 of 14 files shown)


cardstack/src/components/Transactions/MerchantEarnedSpendTransaction.tsx
```diff
@@ -0,0 +1,34 @@
...
+      CoinIcon={<Icon name="spend" />}
+      Header={
+        <SafeHeader address={item.address} rightText="MERCHANT NAME" small />
+      }
+      primaryText={`+ ${item.spendBalanceDisplay}`}
+      statusIconName="arrow-down"
+      statusText="Earned"
+      subText={item.nativeBalanceDisplay}
+      transactionHash={item.transactionHash}
+    />
```
danibonilha, line 32: Wondering if it makes any sense to reuse this piece, since it's almost the same across the components

cardstack/src/components/Transactions/MerchantEarnedRevenueTransaction.tsx
```diff
@@ -0,0 +1,35 @@
...
+  ...props
+}: MerchantEarnRevenueTransactionProps) => {
+  return (
+    <TransactionBase
+      {...props}
+      CoinIcon={
+        <CoinIcon address={item.token.address} symbol={item.token.symbol} />
+      }
+      Header={
+        <SafeHeader address={item.address} rightText="MERCHANT NAME" small />
```
danibonilha, line 26: on rightText we have "MERCHANT NAME" hardcoded, I suppose it will eventually come from be right ? So could we get the value from a prop and if its empty we default to `MERCHANT NAME`

cardstack/src/components/Transactions/TransactionItem.tsx
```diff
@@ -38,6 +42,10 @@ export const TransactionItem = (props: TransactionItemProps) => {
     return <MerchantCreationTransaction {...props} item={item} />;
   } else if (item.type === TransactionTypes.MERCHANT_CLAIM) {
     return <MerchantClaimTransaction {...props} item={item} />;
+  } else if (item.type === TransactionTypes.MERCHANT_EARNED_REVENUE) {
+    return <MerchantEarnedRevenueTransaction {...props} item={item} />;
+  } else if (item.type === TransactionTypes.MERCHANT_EARNED_SPEND) {
+    return <MerchantEarnedSpendTransaction {...props} item={item} />;
```
danibonilha: nit: maybe a switch/case could make it a little less crowded

cardstack/src/transaction-mapping-strategies/transaction-mapping-strategy-types/merchant-earned-revenue-strategy.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (
+      prepaidCardPayments?.[0] &&
+      this.merchantSafeAddress &&
+      prepaidCardPayments[0].merchantSafe?.id === this.merchantSafeAddress
+    ) {
+      return true;
+    }
+
+    return false;
+  }
```
danibonilha: couldn't we just return the condition ?  Or maybe assign to a const idk like, `canHandleTrasaction` and then return this

cardstack/src/transaction-mapping-strategies/transaction-mapping-strategy-types/merchant-earned-revenue-strategy.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const symbol = prepaidCardPaymentTransaction.issuingToken.symbol || '';
+    const amount = prepaidCardPaymentTransaction.issuingTokenAmount;
+
+    const nativeBalance = await getNativeBalance({
+      symbol,
+      balance: amount,
+      nativeCurrency: this.nativeCurrency,
+      currencyConversionRates: this.currencyConversionRates,
+    });
```
danibonilha, line 39: should we wrap in a try/catch ?

cardstack/src/transaction-mapping-strategies/transaction-mapping-strategy-types/merchant-earned-spend-strategy.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (
+      prepaidCardPayments?.[0] &&
+      this.merchantSafeAddress &&
+      prepaidCardPayments[0].merchantSafe?.id === this.merchantSafeAddress
+    ) {
+      return true;
+    }
+
+    return false;
+  }
```
danibonilha: nit: wondering if Revenue and Spend could share this

cardstack/src/types/transaction-types.ts
```diff
@@ -129,6 +131,35 @@ export interface PrepaidCardPaymentTransactionType {
...
+export interface MerchantEarnedRevenueTransactionType {
+  address: string;
+  balance: {
+    amount: string;
+    display: string;
+  };
+  native: {
+    amount: string;
+    display: string;
+  };
```
danibonilha: I think we have some of these types defined elsewhere, we could reuse those instead of redeclaring

src/components/expanded-state/LifetimeEarningsExpandedState.tsx
```diff
@@ -135,13 +137,65 @@ const ChartSection = ({ address }: { address: string }) => {
...
+              >
+                <Text color="blueText" size="medium">
+                  {title}
+                </Text>
+              </Container>
+            )}
+            sections={sections}
+            style={{ width: '100%' }}
+          />
+        )}
```
danibonilha, line 189: For lists we should avoid as much as possible anonymous functions and make use of the memoization hooks, to avoid performance issues

cardstack/src/components/Transactions/MerchantEarnedSpendTransaction.tsx
DominicSherman: Not sure the cleanest way to do that, do you have any suggestions?

cardstack/src/components/Transactions/MerchantEarnedRevenueTransaction.tsx
DominicSherman: We'll be updating all of the merchant transactions soon to pull this data from off-chain so at that point all of these instances will be fixed

cardstack/src/transaction-mapping-strategies/transaction-mapping-strategy-types/merchant-earned-revenue-strategy.ts
DominicSherman: Adding one a level higher to be able to properly handle an error

cardstack/src/transaction-mapping-strategies/transaction-mapping-strategy-types/merchant-earned-spend-strategy.ts
DominicSherman: It could, but since these are the only two using this logic I'd rather keep it separate for scalability and if one needs to tweak a bit. If you want me to extract to a helper function I can though

danibonilha (approved):
LGTM

PR #321, opened 2021-08-18
[CS-1628] Add merchant revenue transactions to EOA screen

DominicSherman:
Handling merchant revenue transactions on EOA screen. A little bit tricky because a single transaction could be both a prepaid card payment and a merchant revenue transaction, if the EOA owns both the prepaid card and the merchant safe. Added a map for transaction types that may have additional handlers so that we can break out a single transaction type into multiple in the UI. If anyone has suggestions for a better way to handle this I am open to hearing them. Also, looking at the description of PR #316 for more information on how we are handle merchant revenue transactions differently based on location

Screen Shot 2021-08-18 at 09 45 54

Also validated that it properly handles earning revenue from a prepaid card the EOA does not own and sending a prepaid card payment to a merchant safe we don't own. This entailed updating the prepaid card payment strategy to check that the prepaid card payment is from a prepaid card we own, otherwise previously it wouldn't have shown every merchant payment as a prepaid card payment transaction as well

Screen Shot 2021-08-18 at 11 49 29

### DIFF
--- a/cardstack/src/transaction-mapping-strategies/transaction-mapping-strategy-types/prepaid-card-payment-strategy.ts
+++ b/cardstack/src/transaction-mapping-strategies/transaction-mapping-strategy-types/prepaid-card-payment-strategy.ts
@@ -9,11 +9,12 @@ export class PrepaidCardPaymentStrategy extends BaseStrategy {
   handlesTransaction(): boolean {
     const { prepaidCardPayments } = this.transaction;
 
-    if (prepaidCardPayments?.[0]) {
-      return true;
-    }
-
-    return false;
+    return Boolean(
+      prepaidCardPayments?.[0] &&
+        this.prepaidCardAddresses.includes(
+          prepaidCardPayments[0].prepaidCard.id
+        )
+    );
   }
 
   async mapTransaction(): Promise<PrepaidCardPaymentTransactionType | null> {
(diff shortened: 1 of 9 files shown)


cardstack/src/transaction-mapping-strategies/transaction-mapping-strategy-types/prepaid-card-payment-strategy.ts
```diff
@@ -9,11 +9,12 @@ export class PrepaidCardPaymentStrategy extends BaseStrategy {
...
 
-    if (prepaidCardPayments?.[0]) {
-      return true;
-    }
-
-    return false;
+    return Boolean(
+      prepaidCardPayments?.[0] &&
+        this.prepaidCardAddresses.includes(
+          prepaidCardPayments[0].prepaidCard.id
```
topwebtek7, line 15: @DominicSherman hmm why we need this filter? do we need to filter with current prepaid cards when displaying prepaidcard payment transactions?

DominicSherman: Yeah, if the payment is a payment to a merchant that is owned by our EOA, it will show as a prepaid card payment. We only want to display it as a prepaid card payment in the UI to the user if this payment was from a prepaid card that we own

topwebtek7: ok, makes sense 👍🏼

topwebtek7 (approved):

PR #323, opened 2021-08-18
[CS-1623] - Fix disappearing safes assets on error

danibonilha:
### Description

- Refactoring of fetchBalancesAndAssets from fallback explorer in order to have a more consistent approach and be able to test it.
- Fixed the bug on error, using try/catch
- Added tests and fixed current one, using relative paths and not testing components, enables us to test every .ts file, as long as the modules are imported with relatives paths.
- Added testing step on PR verify 

image

- [x] Completes #(CS-1623)


cardstack/src/services/firebase-cloud-functions.ts
```diff
@@ -15,7 +13,7 @@ const createCloudFuncGetRequest = async (name: CloudFuncNames) => {
 
     return await response.json();
   } catch (e) {
-    logger.error(`get${name}`, e);
+    console.error(`get${name}`, e);
```
DominicSherman: Should we log this to sentry?

DominicSherman (approved):
Looks good to me!

PR #329, opened 2021-08-20
[CS-1400] Fix signing request is not translating issue, some refactoring with web3 ws reconnect bahavior

topwebtek7:
### Description
Reason - When app was in locked/background state for some mins and back to active, web3 websocket disconnected and does not reconnect automatically, so it breaks many parts of our app, like signing requests' translating, fetching assets, etc and had to restart app to fix it.
Solution - Fixed it by running settingsLoadNetwork() when app state changed back to active from background, and added ws reconnect behavior when it's not connected, refactored with such changes.

- [x] Completes #1400

### Checklist

- [x] checked on real device

### DIFF
--- a/cardstack/src/transaction-confirmation-strategies/decoding-utils.ts
+++ b/cardstack/src/transaction-confirmation-strategies/decoding-utils.ts
@@ -1,12 +1,11 @@
 import Web3 from 'web3';
-import { web3ProviderSdk } from '@rainbow-me/handlers/web3';
 import logger from 'logger';
 
 export const safeDecodeParameters = <T>(
   params: object[],
   data: string
 ): T | null => {
-  const web3 = new Web3(web3ProviderSdk as any);
+  const web3 = new Web3();
 
   try {
     const result = web3.eth.abi.decodeParameters(params, data) as T;
--- a/src/handlers/web3.js
+++ b/src/handlers/web3.js
@@ -17,6 +17,7 @@ import { parseEther } from '@ethersproject/units';
 import UnstoppableResolution from '@unstoppabledomains/resolution';
 import { get, startsWith } from 'lodash';
 import Web3 from 'web3';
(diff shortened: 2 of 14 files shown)


cardstack/src/transaction-confirmation-strategies/decoding-utils.ts
```diff
@@ -1,12 +1,11 @@
 import Web3 from 'web3';
-import { web3ProviderSdk } from '@rainbow-me/handlers/web3';
 import logger from 'logger';
 
 export const safeDecodeParameters = <T>(
   params: object[],
   data: string
 ): T | null => {
-  const web3 = new Web3(web3ProviderSdk as any);
+  const web3 = new Web3();
```
danibonilha, line 8: Does this one do not need a `web3ProviderSdk` ?

src/handlers/web3.js
```diff
@@ -52,6 +53,29 @@ export const web3SetHttpProvider = async network => {
   return web3Provider.ready;
 };
 
+export const getWeb3ProviderSdk = async () => {
+  if (web3ProviderSdk && web3ProviderSdk.connected) {
+    return web3ProviderSdk;
+  }
+
+  if (web3ProviderSdk && !web3ProviderSdk.connected) {
```
danibonilha, line 61: nit: Could this be an else ?

topwebtek7: no, web3ProviderSdk could be `Null` so we need to initialize with web3SetHttpProvider in that case

cardstack/src/transaction-confirmation-strategies/decoding-utils.ts
topwebtek7: no need a `web3ProviderSdk` as we just use `decodeParameters` function only here with this web3 variable

eshinkawa (approved):

PR #330, opened 2021-08-20
[CS-1572] - Temp disabling pagination to fix transactions loop

danibonilha:
### Description
The merging of the cache was creating duplicated entries on transactions, since this is affecting the majority of users, we are disabling the pagination to have it work, because it defaults to 100 transactions anyways, so it shouldn't affect many users, an upcoming PR will have both things working, for now, we are prioritizing fixing repeated transaction


eshinkawa (approved):
LGTM

PR #332, opened 2021-08-23
[CS-1684]: Fix crash on mainnet

danibonilha:
### Description

Sometimes we don't have depots yet, since we don't use then on mainnet, it may trigger a crash, that's why we need to optional chain it `depots?.[0]` and the other potential problem, is that the fetchBalancesAndPrices function, creates multiple setimout functions to recursively call it, and it was never getting cleanup, so I added a couple of those where needed, there's still stuff to improve though but hopefully non-crashable.


eshinkawa (approved):
nice LGTM

PR #333, opened 2021-08-24
Dynamic avatar color, name and first letter from merchant information JSON

eshinkawa:
- Dynamic merchant info with avatar color and name
- important: INFO_DID in order to retrieve merchant information is hardcoded for now (TODO change later)

### Description

- [X] Completes #1525 #1534 #1536

### Checklist

### Screenshots

Screen Shot 2021-08-24 at 17 52 54

Screen Shot 2021-08-24 at 17 51 52

### DIFF
--- a/cardstack/src/components/MerchantSafe/MerchantSafe.tsx
+++ b/cardstack/src/components/MerchantSafe/MerchantSafe.tsx
@@ -60,6 +60,7 @@ export const MerchantInfo = (props: MerchantSafeProps) => (
           color={props.merchantInfo?.color}
           size="medium"
           value={props.merchantInfo?.name}
+          textColor={props.merchantInfo?.['text-color']}
         />
       ) : (
         <Icon name="user" />
--- a/cardstack/src/components/MerchantSafe/MerchantSafe.tsx
+++ b/cardstack/src/components/MerchantSafe/MerchantSafe.tsx
@@ -11,7 +11,7 @@ import {
   Touchable,
 } from '@cardstack/components';
 import { MerchantInformation, MerchantSafeType } from '@cardstack/types';
-import { convertSpendForBalanceDisplay, screenWidth } from '@cardstack/utils';
+import { convertSpendForBalanceDisplay } from '@cardstack/utils';
 import { useNavigation } from '@rainbow-me/navigation';
 import Routes from '@rainbow-me/routes';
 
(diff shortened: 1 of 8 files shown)


cardstack/src/screens/MerchantScreen.tsx
```diff
@@ -19,6 +20,7 @@ import { MerchantSafeType, TokenType } from '@cardstack/types';
 import {
   convertSpendForBalanceDisplay,
   getAddressPreview,
+  screenWidth,
```
topwebtek7: screenWidth is defined already twice in this file

cardstack/src/services/gnosis-service.ts
```diff
@@ -87,9 +90,26 @@ export const fetchGnosisSafes = async (address: string) => {
       })
     );
 
+    const extendedMerchants = await Promise.all(
+      merchantSafes.map(async (merchantSafe: MerchantSafe) => {
+        // TODO CHANGE HARDCODED VARIABLE
+        const INFO_DID =
+          'did:cardstack:1m8oR2jV2aKZqQXttNzhB2vga2fe8168b71700c5';
```
topwebtek7, line 97: It's still using hardcoded did, should update with merchantSafe.merchantDID, i think.

kdowlin: For the time being I think we have to hard code - until the backend team gets it a bit further along. We can swap it once they have the merchant creation + DID working as expected.

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -57,9 +61,25 @@ const MerchantInfo = () => (
...
+        marginLeft={4}
+        justifyContent="center"
+        width={CONTAINER_WIDTH}
+      >
+        <Text weight="bold" ellipsizeMode="tail" numberOfLines={1}>
+          {props.merchantInfo?.name || ''}
+        </Text>
         <Text variant="subText">Merchant Account</Text>
       </Container>
     </Container>
```
danibonilha, line 80: I don't think we need to use `screenWidth` to handle these cases

danibonilha (commented):
Good job on the did stuff, but there are some design inconsistencies in this flow, that we can take this opportunity to address.
- The tokenIcons are way bigger then the merchant avatar for example
- We don't need to use screenWidth to handle ellipses 
- We should keep the horizontal padding consistent within the component, keeping it in the parent, so we don't need to set on every child.

I did some changes on the `MerchantSafe` based on your branch, you can check it out [in this draft PR](https://github.com/cardstack/cardwallet/pull/336), there are examples with consistent padding, token icon size and no need for `screenWidth`, feel free to cherry-pick if you'd like, or merge into your branch.
Take this consideration into `MerchantScreen` also.

danibonilha (approved):
LGTM

PR #334, opened 2021-08-24
[CS-1681] - Fix tx pagination and add refresh on focus

danibonilha:
### Description

- Fixes pagination issue to retrieve all transactions 
- Fixes infinite loading while fetching more
- Adds ability to refresh onFocus, so after a tx is complete the user can see it updating // unfortunately it's a bit unstable sometimes it updates other times doesn't, but still it's getting the most up-to-date info 

Context: For every query we were creating a new apollo client, since we have just two possible uris I'm handling it with apolloLink split, this way we can have only one client and based on the query context we can change. Also decreased the transactions page size since 100 is too high if we have pagination.

- [x] Completes #(CS-1681)

### Screenshots

https://user-images.githubusercontent.com/20520102/130699340-c50ea779-6866-476d-8ed6-8014d8246c50.mp4

### DIFF
--- a/cardstack/src/components/TransactionList/TransactionList.tsx
+++ b/cardstack/src/components/TransactionList/TransactionList.tsx
@@ -1,5 +1,10 @@
-import React from 'react';
-import { RefreshControl, SectionList, ActivityIndicator } from 'react-native';
+import React, { memo, useCallback, useEffect } from 'react';
+import {
+  RefreshControl,
+  SectionList,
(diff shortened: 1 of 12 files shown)


eshinkawa:
Great, LGTM

eshinkawa (approved):

topwebtek7 (commented):
LGTM

topwebtek7 (approved):

PR #335, opened 2021-08-25
[CS-1693] - Update toast color with opacity

danibonilha:
### Description

Adds opacity to toast background, so the user can see whats underneath

- [x] Completes #(CS-1693)

### Screenshots

image


topwebtek7 (approved):
LGTM!

PR #336, opened 2021-08-25
[Fix]: Merchant card icons and padding

danibonilha:
### Screenshots

image

image


cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -75,20 +69,20 @@ export const MerchantInfo = (props: MerchantSafeProps) => (
         flexDirection="column"
         marginLeft={4}
         justifyContent="center"
-        width={CONTAINER_WIDTH}
+        width="85%"
```
eshinkawa, line 72: great! i did not do it because on previous cases that i used it wasn't working for some reason, but if it does then it is definitely better

eshinkawa (approved):

PR #337, opened 2021-08-25
[Refactor]: Extract cardstack's routes to it's own navigation 

danibonilha:
### Description

For my current task I need to create a new screen, and  since we were creating everything just for the iOS stack, I took the liberty to extract all cardstack  routes to one place, this way we can use across android and ios, without duplicating code. 
From now on, we can just add the screens into the `Screens` object, it's straight forward but if you have any doubts lmk.


topwebtek7 (approved):
Looks good

PR #340, opened 2021-08-27
[CS-1621] - Adds max button and gas fee on send tokens from depot

danibonilha:
### Description

Adds max button and gas fee on send tokens from depot, 
Extracted some rainbow's logic in order to reuse the component, the commit are pretty self explanatory, the plan is to add some tests when we get some time, hopefully I didn't break anything on both flows, will be adding a .gif on monday, just to have more time to validate all the scenarios.

- [x] Completes #(CS-1621)

### Screenshots

send-max

### DIFF
--- a/cardstack/src/components/Button/Button.tsx
+++ b/cardstack/src/components/Button/Button.tsx
@@ -58,9 +58,6 @@ export const Button = ({
   onPress,
   ...props
 }: ButtonProps) => {
-  const width = useVariantValue('buttonVariants', 'width', props.variant);
-  const maxWidth = useVariantValue('buttonVariants', 'maxWidth', props.variant);
-
   const textStyle = useVariantValue(
     'buttonVariants',
     'textStyle',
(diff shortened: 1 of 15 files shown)


cardstack/src/components/Button/Button.tsx
```diff
@@ -119,8 +116,7 @@ export const Button = ({
           position="absolute"
           height="100%"
           zIndex={1}
-          width={width}
-          maxWidth={maxWidth}
+          width="100%"
```
topwebtek7, line 119: @danibonilha did you confirm if this change does not break any existing buttons?

topwebtek7 (approved):
Looks good to me. just left a comment
And will you add tests for max button behavior and sending tokens from depot?

eshinkawa:
LGTM

eshinkawa (approved):

cardstack/src/components/Button/Button.tsx
danibonilha: Yes, on the buttons I could find, I didn't find any other disabled button, so I'm guessing this change only affects this flow

danibonilha:
> Looks good to me. just left a comment
> And will you add tests for max button behavior and sending tokens from depot?

Will be adding tests, once merchant flow is complete

PR #341, opened 2021-08-30
[CS-1484] Added request payment qr code page, Fix a bug with spend calculation

topwebtek7:
### Description

Display QRCode of merchant payment request URL

- [x] Completes #CS-1484
- [x] Completes #CS-1712

### Checklist

### Screenshots

Simulator Screen Shot - iPhone 8 - 2021-08-31 at 03 07 10
Simulator Screen Shot - iPhone 12 Pro Max - 2021-08-31 at 03 06 45


cardstack/src/screens/ShowQRCodeModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+        downsampleFactor={10}
+        reducedTransparencyFallbackColor="black"
+        overlayColor={colors.overlayGray}
+        style={{
+          position: 'absolute',
+          top: 0,
+          left: 0,
+          bottom: 0,
+          right: 0,
+        }}
```
danibonilha: we can just use `style={StyleSheet.absoluteFill}`

https://reactnative.dev/docs/stylesheet#absolutefill

cardstack/src/screens/ShowQRCodeModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const {
+    params: { value, amountInSpend, amountWithSymbol, name },
+  } = useRoute() as { params: ShowQRCodeModalParamTypes };
+
+  const { width } = useDimensions();
+  const BlurViewWrapper = ios ? VibrancyView : BlurView;
+
+  return (
+    <SafeAreaView flex={1} width="100%" backgroundColor="transparent">
+      <BlurViewWrapper
```
danibonilha: Why are we using this `BlurViewWrapper` if it's not actually blurring anything, we can still see what's underneath the modal, also it causes the navigation to be weird, because when you drag down to dismiss there's like a line 
drag

cardstack/src/screens/ShowQRCodeModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+        justifyContent="center"
+        backgroundColor="transparent"
+      >
+        <Container
+          width="100%"
+          height="auto"
+          alignItems="center"
+          backgroundColor="transparent"
+          paddingBottom={15}
+        >
```
danibonilha, line 51: We can use flex containers here, this way we don't need to set width, multiple paddings, and transparent backgrounds

cardstack/src/screens/ShowQRCodeModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+            >
+              <QRCode
+                size={width - 112}
+                value={value}
+                logoMargin={12}
+                logoBorderRadius={6}
+                logo={CardstackColorLogo}
+                quietZone={4}
+              />
+            </Container>
```
danibonilha: I'm guessing we will have other QR codes in the app, so I'd suggest we should build a wrapper and a default size, so everytime we need a QR code we can just use the same component in order to have the same design

cardstack/src/screens/ShowQRCodeModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+                letterSpacing={0.39}
+                color="white"
+                fontWeight="bold"
+              >
+                {amountWithSymbol}
+              </Text>
+              <Text size="xs" color="underlineGray" fontWeight="600">
+                {`§${amountInSpend} SPEND`}
+              </Text>
+              <Text
```
danibonilha: I believe in of the meetings, Chris said USD and SPEND should have the same fontsize so one doesn't seem more important than the other

cardstack/src/screens/index.ts
```diff
@@ -5,3 +5,4 @@ export { default as TransactionConfirmation } from './TransactionConfirmation';
 export { default as CopyAddressSheet } from './CopyAddressSheet';
 export { default as PrepaidCardModal } from './PrepaidCardModal';
 export { default as SendSheetDepot } from './SendSheetDepot';
+export { default as ShowQRCodeModal } from './ShowQRCodeModal';
```
danibonilha, line 8: Will this QRCodeModal be global? Or will be it used just for the merchant pay ? If it's global it should take the bottom text, if it's just for the merchant flow, the name of the component should reflect that, so it could be something like `MerchantShowQRCodeModal`

src/navigation/Routes.ios.js
```diff
@@ -321,6 +322,23 @@ function NativeStackNavigator() {
...
+          transparentCard: true,
+          onAppear: null,
+          topOffset: 0,
+          ignoreBottomOffset: true,
+          cornerRadius: 0,
+          customStack: true,
+          springDamping: 1,
+          transitionDuration: 0.2,
+        }}
+      />
```
danibonilha, line 330: Could you add a comment here, saying we need to add the screen in the native stack in other to use the native modal behavior ? Just so if we eventually try to remove some of this stack we don't forget this one is important to us

danibonilha (commented):
Couple of things we need to tackle
- Use same fontSizes for USD and SPEND
- QRCode wrapper design 
- Check the Blur behavior and drag animation 
- Screen naming 
- Using flex and percentages to deal with QRCode screen, I opened a [draft PR](https://github.com/cardstack/cardwallet/pull/342 ) with an example of what I'm talking about, this way we can even have one `<Container />` less  =)

cardstack/src/screens/ShowQRCodeModal.tsx
topwebtek7: Yeah will update when get provided new designs for that

cardstack/src/screens/index.ts
topwebtek7: I made this component to use as global component so can be used in other screens as well

cardstack/src/utils/merchant-utils.ts
```diff
@@ -42,10 +42,13 @@ export const fetchMerchantInfoFromDID = async (
...
   merchantSafeID: string,
-  amountInSpend: number,
-  network = 'sokol'
+  amount: number,
+  network = 'sokol',
+  currency = 'SPD'
 ) => {
-  return `https://wallet.cardstack.com/pay/${network}/${merchantSafeID}?amount=${amountInSpend}`;
+  // https://wallet.cardstack.com/pay/[sokol|xdai]/[merchant-safe-id]?amount=[amount-in-specified-currency]&currency=[3-letter-symbol]
+  return `https://wallet.cardstack.com/pay/${network}/${merchantSafeID}?amount=${amount}&currency=${currency}`;
```
danibonilha, line 53: Should we have this comment here or is it just a left over ?

danibonilha (approved):
small comment, overall LGTM, if you could update the screenshots, it would be great = )

cardstack/src/utils/merchant-utils.ts
topwebtek7: It's just left over, will get rid of it when confirm and finalize it

PR #346, opened 2021-09-01
[CS-1757] - Fix infinite re-rendering loop

danibonilha:
### Description
The condition to set sections was based only when we had transactions, in the prepaidCard modal, we get `transactions` from the query, then pass forward to `useTransactionSections` , this hook, checks if transactions exist and update the state, so once the state is updated, it triggers a re-render in the parent, which checks for the transactions again and pass thru `useTransactionSections` over and over, even though the transactions are the same. This PR changes this behavior so it will only set the state again if the amount of current transactions is different from the previous amount avoiding this infinite loop behavior.

- [x] Completes #(CS-1757)

### Screenshots

Before, we can see how the JS performance drops to 10 and stays low: 
Simulator Screen Recording - iPhone 11 - 2021-09-01 at 15 22 12

After fix, it drops to 42 and quickly  gets back to 60:
Simulator Screen Recording - iPhone 11 - 2021-09-01 at 15 21 17

### DIFF
--- a/cardstack/src/hooks/transactions/use-transaction-sections.tsx
+++ b/cardstack/src/hooks/transactions/use-transaction-sections.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { groupBy } from 'lodash';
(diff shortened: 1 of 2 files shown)


topwebtek7 (approved):
LGTM!

PR #347, opened 2021-09-02
[CS-1481] Update merchant available balances detail

topwebtek7:
### Description

- Added balances section list in merchant available balances expanded state
- Remove shadows in coin icons

- [x] Completes #CS-1481
- [x] Completes #CS-1629
- [x] Completes #CS-1686

### Checklist

### Screenshots

simulator_screenshot_8077C320-807F-4DBC-8A88-54774731CF12

### DIFF
--- a/cardstack/src/components/CoinIcon/CoinIcon.tsx
+++ b/cardstack/src/components/CoinIcon/CoinIcon.tsx
@@ -15,7 +15,7 @@ export const CoinIcon = ({
   address,
   symbol,
   size = 40,
-  shadowColor = 'transparent',
+  shadowColor = 'transparent', // 'transparent' is to remove drop shadow on Token icons
 }: CoinIconProps) => {
   const network = useRainbowSelector(state => state.settings.network);
   const forceFallback = !symbol && !isNativeToken(symbol || '', network);
--- a/cardstack/src/components/CoinIcon/CoinIcon.tsx
+++ b/cardstack/src/components/CoinIcon/CoinIcon.tsx
@@ -8,9 +8,15 @@ interface CoinIconProps {
   address?: string;
   symbol?: string | null;
   size?: number;
+  shadowColor?: string;
 }
 
-export const CoinIcon = ({ address, symbol, size = 40 }: CoinIconProps) => {
+export const CoinIcon = ({
+  address,
+  symbol,
+  size = 40,
+  shadowColor = 'transparent',
+}: CoinIconProps) => {
   const network = useRainbowSelector(state => state.settings.network);
   const forceFallback = !symbol && !isNativeToken(symbol || '', network);
 
@@ -21,6 +27,7 @@ export const CoinIcon = ({ address, symbol, size = 40 }: CoinIconProps) => {
       size={size}
       forceFallback={forceFallback}
       fallbackRenderer={CoinIconFallback}
+      shadowColor={shadowColor}
     />
   );
 };
--- a/src/components/expanded-state/AvailableBalancesExpandedState.tsx
+++ b/src/components/expanded-state/AvailableBalancesExpandedState.tsx
@@ -1,21 +1,38 @@
+import { add, convertAmountToNativeDisplay } from '@cardstack/cardpay-sdk';
+import { get } from 'lodash';
 import React, { useEffect, useState } from 'react';
 import { SlackSheet } from '../sheet';
 import {
+  AssetList,
+  AssetListSectionItem,
   Container,
   HorizontalDivider,
   Text,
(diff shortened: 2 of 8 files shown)


src/components/expanded-state/AvailableBalancesExpandedState.tsx
```diff
@@ -1,23 +1,40 @@
...
+} from '@rainbow-me/redux/hooks';
 import Routes from '@rainbow-me/routes';
 
 const CHART_HEIGHT = 650;
 
 enum Tabs {
-  BALANCES = 'Balances',
+  Assets = 'Assets',
   ACTIVITIES = 'Activities',
 }
```
danibonilha: Let's either use  all uppercase letters or just the first for both tabs enum key

src/components/expanded-state/AvailableBalancesExpandedState.tsx
```diff
@@ -37,14 +54,18 @@ export default function AvailableBalancesExpandedState(
   return (
     <>
       {/* @ts-ignore */}
```
danibonilha: I know this is a left of from other task, but instead of using `ts-ignore` we can just add undefined to  the` SlackSheet` destructed props, then ts won't complain the property doesn't exist 

```tsx
export default function SlackSheet({
  additionalTopPadding = false,
  backgroundColor = undefined,
  borderRadius = 30,
  children = undefined,
  contentHeight = undefined,
  deferredHeight = false,
  hideHandle = false,
  renderHeader = undefined,
  renderFooter = undefined,
  scrollEnabled = true,
  discoverSheet = undefined,
  hasKeyboard = false,
  ...props
}) {}
```

src/components/expanded-state/AvailableBalancesExpandedState.tsx
```diff
@@ -37,14 +54,18 @@ export default function AvailableBalancesExpandedState(
...
-      <SlackSheet bottomInset={42} height="100%" scrollEnabled>
+      <SlackSheet
+        additionalTopPadding={android}
```
danibonilha: Do we know how is the behavior on android already ? Also if we won't to check if it's android or not, I'd suggest we create a device utils on our side, without using these global variables as we should move away from `js` and we should depend less on this. 

```ts
import { Platform } from 'react-native';

const Device = {
  isIOS: Platform.OS === 'ios',
  isAndroid: Platform.OS === 'android',
};

export { Device };

```

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -105,7 +105,11 @@ export default function UnclaimedRevenueExpandedState(props: {
   return (
     <>
       {/* @ts-ignore */}
-      <SlackSheet bottomInset={42} height="100%" scrollEnabled>
+      <SlackSheet
+        additionalTopPadding={android}
+        height="100%"
+        scrollEnabled={false}
+      >
```
danibonilha: ditto

src/components/expanded-state/AvailableBalancesExpandedState.tsx
```diff
@@ -97,36 +112,86 @@ const TabHeader = ({ tab, selectedTab, setSelectedTab }: TabHeaderProps) => {
...
+    tokenBalance: token.balance.display,
+    nativeBalance: token.native.balance.display,
+    onPress: () =>
+      navigate(Routes.EXPANDED_ASSET_SHEET, {
+        asset: token,
+        type: 'token',
+      }),
+    key: token.tokenAddress,
+    tokenBalanceFontSize: 'largeBalance',
+  }));
```
danibonilha: we can wrap in a useMemo

cardstack/src/components/CoinIcon/CoinIcon.tsx
```diff
@@ -8,9 +8,15 @@ interface CoinIconProps {
...
   size?: number;
+  shadowColor?: string;
 }
 
-export const CoinIcon = ({ address, symbol, size = 40 }: CoinIconProps) => {
+export const CoinIcon = ({
+  address,
+  symbol,
+  size = 40,
+  shadowColor = 'transparent', // 'transparent' is to remove drop shadow on Token icons
```
danibonilha, line 18: nice 💯

src/components/expanded-state/AvailableBalancesExpandedState.tsx
topwebtek7: oh good catch!

danibonilha (approved):
LGTM

PR #348, opened 2021-09-02
[CS-1720] [CS-1651] - Show merchant earned/claimed Transaction Account Details

eshinkawa:
### Description

- [X] Completes #CS-1720 #CS-1651

### Checklist

### Screenshots

Sep-06-2021 11-09-48

image

image

### DIFF
--- a/cardstack/src/components/Transactions/MerchantClaimTransaction.tsx
+++ b/cardstack/src/components/Transactions/MerchantClaimTransaction.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import { useNavigation } from '@react-navigation/core';
 import { SafeHeader } from '../SafeHeader';
 import {
   TransactionBase,
@@ -7,6 +8,7 @@ import {
 } from './TransactionBase';
 import { MerchantClaimType } from '@cardstack/types';
 import { CoinIcon } from '@cardstack/components';
+import Routes from '@rainbow-me/routes';
 
 export interface MerchantClaimTransactionProps
   extends TransactionBaseCustomizationProps {
@@ -17,6 +19,14 @@ export const MerchantClaimTransaction = ({
   item,
   ...props
 }: MerchantClaimTransactionProps) => {
+  const { navigate } = useNavigation();
+
+  const onPressTransaction = (assetProps: any) =>
+    navigate(Routes.EXPANDED_ASSET_SHEET, {
+      asset: { ...assetProps },
+      type: 'merchantTransaction',
+    });
+
   return (
     <TransactionBase
       {...props}
@@ -33,6 +43,7 @@ export const MerchantClaimTransaction = ({
       statusText="Claimed"
       subText={item.native.display}
       transactionHash={item.transactionHash}
+      toTransactionDetails={onPressTransaction}
     />
   );
 };
--- a/cardstack/src/components/Transactions/MerchantClaimTransaction.tsx
+++ b/cardstack/src/components/Transactions/MerchantClaimTransaction.tsx
@@ -1,10 +1,11 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 
 import { useNavigation } from '@react-navigation/core';
 import { SafeHeader } from '../SafeHeader';
 import {
   TransactionBase,
   TransactionBaseCustomizationProps,
+  TransactionBaseProps,
 } from './TransactionBase';
 import { MerchantClaimType } from '@cardstack/types';
 import { CoinIcon } from '@cardstack/components';
@@ -21,11 +22,14 @@ export const MerchantClaimTransaction = ({
 }: MerchantClaimTransactionProps) => {
   const { navigate } = useNavigation();
 
-  const onPressTransaction = (assetProps: any) =>
-    navigate(Routes.EXPANDED_ASSET_SHEET, {
-      asset: { ...assetProps },
-      type: 'merchantTransaction',
-    });
+  const onPressTransaction = useCallback(
+    (assetProps: TransactionBaseProps) =>
+      navigate(Routes.EXPANDED_ASSET_SHEET, {
+        asset: { ...assetProps },
+        type: 'merchantTransaction',
+      }),
+    [navigate]
+  );
 
   return (
     <TransactionBase
@@ -43,7 +47,7 @@ export const MerchantClaimTransaction = ({
       statusText="Claimed"
       subText={item.native.display}
       transactionHash={item.transactionHash}
-      toTransactionDetails={onPressTransaction}
+      onPressTransaction={onPressTransaction}
     />
   );
 };
--- a/cardstack/src/components/Transactions/TransactionBase.tsx
+++ b/cardstack/src/components/Transactions/TransactionBase.tsx
@@ -1,5 +1,5 @@
 import { getConstantByNetwork } from '@cardstack/cardpay-sdk';
-import React from 'react';
+import React, { useCallback } from 'react';
 import { Linking } from 'react-native';
 import { Icon, IconName, IconProps } from '../Icon';
 import { ContainerProps } from '../Container';
@@ -31,7 +31,7 @@ export interface TransactionBaseProps
   topText?: string;
   transactionHash: string;
   isFullWidth?: boolean;
-  toTransactionDetails?: any;
+  onPressTransaction?: (props: TransactionBaseProps) => void;
 }
 
 export const TransactionBase = (props: TransactionBaseProps) => {
@@ -41,7 +41,7 @@ export const TransactionBase = (props: TransactionBaseProps) => {
     transactionHash,
     includeBorder,
     isFullWidth,
-    toTransactionDetails,
+    onPressTransaction,
   } = props;
 
   const network = useRainbowSelector(state => state.settings.network);
@@ -49,19 +49,29 @@ export const TransactionBase = (props: TransactionBaseProps) => {
   const blockExplorerName = isLayer1(network) ? 'Etherscan' : 'Blockscout';
   const normalizedHash = normalizeTxHash(transactionHash);
 
-  const onPressBlockscout = () => {
-    showActionSheetWithOptions(
-      {
-        options: [`View on ${blockExplorerName}`, 'Cancel'],
-        cancelButtonIndex: 1,
-      },
-      (buttonIndex: number) => {
-        if (buttonIndex === 0) {
-          Linking.openURL(`${blockExplorer}/tx/${normalizedHash}`);
+  const handleOnPressTransaction = useCallback(() => {
+    const onPressBlockscout = () => {
+      showActionSheetWithOptions(
+        {
+          options: [`View on ${blockExplorerName}`, 'Cancel'],
+          cancelButtonIndex: 1,
+        },
+        (buttonIndex: number) => {
+          if (buttonIndex === 0) {
+            Linking.openURL(`${blockExplorer}/tx/${normalizedHash}`);
+          }
         }
-      }
-    );
-  };
+      );
+    };
+
+    onPressTransaction ? onPressTransaction(props) : onPressBlockscout;
+  }, [
+    blockExplorer,
+    blockExplorerName,
+    normalizedHash,
+    onPressTransaction,
+    props,
+  ]);
 
   return (
     <Container
@@ -72,11 +82,7 @@ export const TransactionBase = (props: TransactionBaseProps) => {
       <Touchable
         width="100%"
         testID="inventory-card"
-        onPress={
-          toTransactionDetails
-            ? () => toTransactionDetails(props)
-            : onPressBlockscout
-        }
+        onPress={handleOnPressTransaction}
       >
         <Container
           backgroundColor="white"
@@ -126,11 +132,6 @@ export const TransactionRow = ({
   hasBottomDivider = false,
   ...props
 }: TransactionRowProps) => {
-  const {
-    section: { data },
-    index,
-  }: any = props;
-
   return (
     <Container
       flexDirection="row"
@@ -141,19 +142,6 @@ export const TransactionRow = ({
       {...props}
     >
       <Container flexDirection="column" width="100%">
-        {data[index]?.address ? (
-          <Container flexDirection="row" width="50%" paddingBottom={4}>
-            <Text
-              color="blueText"
-              fontFamily="RobotoMono-Regular"
-              ellipsizeMode="middle"
-              fontSize={13}
-              numberOfLines={1}
-            >
-              {data[index]?.address}
-            </Text>
-          </Container>
-        ) : null}
         <Container
           alignItems="center"
           flexDirection="row"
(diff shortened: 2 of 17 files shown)


eshinkawa:
I apologize for the amount of code, but they tackle two tickets

eshinkawa:
I apologize for the amount of code, but they tackle two tickets and they're dependent of each other. I tried to separate them in different commits because it would be easier to understand the dev flow

cardstack/src/components/Transactions/TransactionBase.tsx
```diff
@@ -63,7 +72,11 @@ export const TransactionBase = (props: TransactionBaseProps) => {
       <Touchable
         width="100%"
         testID="inventory-card"
-        onPress={onPressTransaction}
+        onPress={
+          toTransactionDetails
+            ? () => toTransactionDetails(props)
+            : onPressBlockscout
+        }
```
danibonilha, line 79: Let's keep the `onPressTransaction` function, wrap it with an useCallback and handle the condition based on `toTransactionDetails` there

danibonilha: Let's add a  `handleOnPressTransaction` function, wrap it with an useCallback and handle if we have the `onPressTransation` function from props to be called, otherwise we default to `showBlockscout`

cardstack/src/components/Transactions/TransactionBase.tsx
```diff
@@ -108,8 +123,14 @@ export const TransactionRow = ({
...
   primaryText,
   subText,
+  hasBottomDivider = false,
   ...props
 }: TransactionRowProps) => {
+  const {
+    section: { data },
+    index,
+  }: any = props;
+
```
danibonilha: We should properly type this

cardstack/src/components/Transactions/TransactionBase.tsx
```diff
@@ -119,41 +140,61 @@ export const TransactionRow = ({
...
+              color="blueText"
+              fontFamily="RobotoMono-Regular"
+              ellipsizeMode="middle"
+              fontSize={13}
+              numberOfLines={1}
+            >
+              {data[index]?.address}
+            </Text>
+          </Container>
+        ) : null}
```
danibonilha: Actually I don't think we should check for address to render this, `PrepairdCard`, for example, have an address, and since this is used on different places, it affects others components, as you can see is double showing the address: 


We should probably use the `Header` prop from `TransactionBase`, 
also the address needs to be obfuscated, we use `getAddressPreview` to handle that

cardstack/src/components/Transactions/TransactionBase.tsx
```diff
@@ -30,17 +31,25 @@ interface TransactionBaseProps extends TransactionBaseCustomizationProps {
   topText?: string;
   transactionHash: string;
   isFullWidth?: boolean;
+  toTransactionDetails?: any;
```
danibonilha: isn't this a function?  If we don't know the prop types we should probably use a generic type, also maybe this should be called  `onPressTransaction`, because what's is navigating to transaction details, is your function in the parent, so the tx item doesn't need to know what happens onPress

cardstack/src/utils/__tests__/merchant-utils.test.ts
```diff
@@ -74,4 +80,83 @@ describe('Merchant utils', () => {
...
+          null,
+          'USD',
+          0.49999999,
+          currencyConversionRates,
+          'DAI'
+        )
+      ).toBeNull();
+    });
+  });
 });
```
danibonilha, line 136: 👏

cardstack/src/utils/merchant-utils.ts
```diff
@@ -105,3 +112,95 @@ export const shareRequestPaymentLink = (
     options
   );
 };
+
+export function getMerchantClaimTransactionDetails(
+  merchantClaimTransaction: any,
```
danibonilha: Don't we have a type for that ?

cardstack/src/utils/merchant-utils.ts
```diff
@@ -105,3 +112,95 @@ export const shareRequestPaymentLink = (
...
+  nativeCurrency = 'USD',
+  address?: string
+): any {
```
danibonilha: let's avoid using any as much as possible, in expectation for when we are dealing with ts to js stuff, if is something really hard to figure, we can spend some time together in it, or add a TODO to properly type

cardstack/src/types/transaction-types.ts
```diff
@@ -139,6 +145,7 @@ export interface MerchantEarnedRevenueTransactionType {
   timestamp: number;
   type: TransactionTypes.MERCHANT_EARNED_REVENUE;
   transactionHash: string;
+  transaction?: any;
```
danibonilha: ditto

cardstack/src/utils/merchant-utils.ts
```diff
@@ -105,3 +112,95 @@ export const shareRequestPaymentLink = (
...
+  if (!merchantClaimTransaction || !address) return null;
```
danibonilha: Should we return `undefined` instead of `null`, since `transaction` is typed as a optional, which means that, it's either that type or `undefined`

cardstack/src/utils/merchant-utils.ts
```diff
@@ -105,3 +112,95 @@ export const shareRequestPaymentLink = (
...
+export function getMerchantEarnedTransactionDetails(
+  prepaidCardPaymentTransaction: any,
+  nativeCurrency = 'USD',
+  nativeBalance: number,
+  currencyConversionRates: {
+    [key: string]: number;
+  },
+  symbol: string
+): any {
+  if (!prepaidCardPaymentTransaction) return null;
```
danibonilha: ditto

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -0,0 +1,165 @@
...
+      <Button
+        marginBottom={12}
+        onPress={() => Linking.openURL(`${blockExplorer}/tx/${normalizedHash}`)}
+        variant="smallWhite"
+        width="100%"
+      >
+        View on Blockscout
+      </Button>
+    );
+  };
```
danibonilha: We can move those components outside the main component, at least the ones that receives params which will make the code cleaner and we will avoid recreating those components on every re-render or we can wrap in useMemos

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -0,0 +1,165 @@
...
+        />
+      </Container>
+    );
+  };
+
+  const BlockscoutButton = () => {
+    return (
+      <Button
+        marginBottom={12}
+        onPress={() => Linking.openURL(`${blockExplorer}/tx/${normalizedHash}`)}
```
danibonilha: Let's avoid using anonymous functions on render, if we create a component for this, we can even use the normalizedHash only there to keep the context

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -0,0 +1,165 @@
...
+        />
+        <ItemDetail
+          description={'REVENUE \nCOLLECTED'}
+          value={props.asset.primaryText}
+        />
+        <ItemDetail
+          description={'PROTOCOL FEE \n(0.5%)'}
+          subValue={transactionData?.protocolFeeUsd}
+          value={transactionData?.protocolFee}
+        />
```
danibonilha: nit: Thinking about how many ItemDetail we are using, maybe we can have a data structure, were we just loop over and render a itemDetail

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -0,0 +1,165 @@
...
+                <EarnedTransaction />
+              )}
+            </Container>
+          </Container>
+          <BlockscoutButton />
+        </Container>
+      </SlackSheet>
+    </>
+  );
+}
```
danibonilha, line 225: We can't actually see the BlockscoutButton, and if we try to scroll up, the modal get's back down, and we can't click on it

cardstack/src/components/Transactions/MerchantClaimTransaction.tsx
```diff
@@ -17,6 +19,14 @@ export const MerchantClaimTransaction = ({
...
   ...props
 }: MerchantClaimTransactionProps) => {
+  const { navigate } = useNavigation();
+
+  const onPressTransaction = (assetProps: any) =>
+    navigate(Routes.EXPANDED_ASSET_SHEET, {
+      asset: { ...assetProps },
+      type: 'merchantTransaction',
+    });
+
```
danibonilha, line 33: We should double check this behavior, because while navigating to Transaction details, we cannot get back to the activities list, I think we should have a modal on top the other, or a back button, whatever it's easier, I noticed rainbow has it's own useNavigation, not sure though if it impacts this

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { TransactionRow } from '@cardstack/components/Transactions/TransactionBase';
+import { normalizeTxHash } from '@cardstack/utils';
+import { useRainbowSelector } from '@rainbow-me/redux/hooks';
+
+interface ItemDetail {
+  description?: string;
+  value?: string;
+  subValue?: string;
+  symbol?: string;
+}
```
danibonilha, line 28: Should everything be optional ? If we don't have any of those properties, why render itemDetail ? at least description should be required

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+const CLAIMED_STATUS = 'Claimed';
+
+export default function MerchantTransactionExpandedState(props: any) {
```
danibonilha: ditto

danibonilha (changes requested):

cardstack/src/components/Transactions/TransactionBase.tsx
eshinkawa: Right! This was a nice to have for the ticket, but it will need a bigger discussion, so I am adding a TODO for it

src/components/expanded-state/MerchantTransactionExpandedState.tsx
eshinkawa: yes, I talked to Kiel about, since this gonna impact other screens so I will keep the way things currently are for the next discussions!

cardstack/src/components/Transactions/TransactionBase.tsx
```diff
@@ -63,7 +72,11 @@ export const TransactionBase = (props: TransactionBaseProps) => {
       <Touchable
         width="100%"
         testID="inventory-card"
-        onPress={onPressTransaction}
+        onPress={
+          onPressTransaction
+            ? () => onPressTransaction(props)
+            : onPressBlockscout
+        }
```
danibonilha: sorry I think I didn't explained my idea correctly, here I'm suggesting that we have:
```tsx
const handleOnPressTransaction = useCallback(() => {
         onPressTransaction ? onPressTransaction(props) : onPressBlockscout
} ,[onPressTransaction])
```
then we use this handleOnPress, on this Touchable onPress

cardstack/src/components/Transactions/TransactionBase.tsx
```diff
@@ -119,41 +135,64 @@ export const TransactionRow = ({
...
+              fontFamily="RobotoMono-Regular"
+              ellipsizeMode="middle"
+              fontSize={13}
+              numberOfLines={1}
+            >
+              {data[index]?.address}
+            </Text>
+          </Container>
+        ) : null
+        */}
```
danibonilha: we should remove this entirely and take `TransactionBase` as an example, if we want to create a new component that will have a specific header, and specific footer, we can just do something like it and if this is just for the merchant flow, it doesn't need to be related to TransactionBase

cardstack/src/utils/merchant-utils.ts
```diff
@@ -105,3 +121,96 @@ export const shareRequestPaymentLink = (
...
+      nativeCurrency,
+      currencyConversionRates
+    ).nativeBalanceDisplay,
+  };
+}
+
+export const blockScoutLink = (network: string, normalizedHash: string) => {
+  const blockExplorer = getConstantByNetwork('blockExplorer', network);
+  return Linking.openURL(`${blockExplorer}/tx/${normalizedHash}`);
+};
```
danibonilha: I think we can move this to `BlockscoutButton`

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -0,0 +1,216 @@
...
+              {props.asset.statusText === CLAIMED_STATUS ? (
+                <ClaimedTransaction {...transactionData} />
+              ) : (
+                <EarnedTransaction {...earnedTxnData} />
+              )}
+            </Container>
+          </Container>
+          <BlockscoutButton
+            blockScoutLink={() => blockScoutLink(network, normalizedHash)}
+          />
```
danibonilha: if we move the link inside this Button the only prop we need to pass is a transaction

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -0,0 +1,216 @@
...
+    <Button
+      marginBottom={12}
+      onPress={blockScoutLink}
+      variant="smallWhite"
+      width="100%"
+    >
+      View on Blockscout
+    </Button>
+  );
+};
```
danibonilha:
```suggestion
const BlockscoutButton = ({
  transactionHash,
}: {
  transactionHash: string;
}) => {
const onPress = useCallback(() => {
  const blockExplorer = getConstantByNetwork('blockExplorer', network);
  const normalizedHash = normalizeTxHash(transactionHash);

  Linking.openURL(`${blockExplorer}/tx/${normalizedHash}`);
}, [transactionHash]);  

  return (
    <Button
      marginBottom={12}
      onPress={onPress}
      variant="smallWhite"
      width="100%"
    >
      View on Blockscout
    </Button>
  );
};
```

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -185,3 +185,11 @@ const TokenItem = ({ token }: { token: TokenType }) => {
...
   );
 };
+
+const ListEmptyComponent = () => (
+  <CenteredContainer flex={1} height={100} width="100%">
+    <Text color="grayText" textAlign="center">
+      No activity Data
+    </Text>
+  </CenteredContainer>
+);
```
danibonilha: I believe we need to revert this change, because it makes sense to use `<ListEmptyComponent text="No activity" />`since the component already exists

danibonilha (approved):
If the scroll is working properly, we can ship it  🚀

PR #349, opened 2021-09-02
[CS-1704] - Add uni links, url schemes and remove rb leftovers

danibonilha:
### Description

Adds new universal links and remove rainbow's one, also changes icloud backup path to be related to cardwallet.

- [x] Completes #(CS-1704)

### Screenshots

deeplink

### DIFF
--- a/ios/Rainbow/Info.plist
+++ b/ios/Rainbow/Info.plist
@@ -26,17 +26,7 @@
 			<key>CFBundleTypeRole</key>
 			<string>Editor</string>
 			<key>CFBundleURLName</key>
-			<string>me.rainbow</string>
-			<key>CFBundleURLSchemes</key>
-			<array>
-				<string>rainbow</string>
-			</array>
-		</dict>
-		<dict>
-			<key>CFBundleTypeRole</key>
-			<string>Editor</string>
-			<key>CFBundleURLName</key>
-			<string>Bundle ID</string>
+			<string>com.cardstack.cardpay</string>
 			<key>CFBundleURLSchemes</key>
 			<array>
 				<string>com.cardstack.cardpay</string>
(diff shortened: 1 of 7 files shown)


ios/Rainbow/Info.plist
```diff
@@ -23,30 +23,27 @@
...
 			</array>
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>CodePushDeploymentKey</key>
 	<string>$(CODEPUSH_KEY)</string>
+	<key>FirebaseDynamicLinksCustomDomains</key>
+	<array>
+		<string>https://pay.cardstack.com/request</string>
```
lukemelia, line 44: I don't think we need/want this

ios/Rainbow/Rainbow.entitlements
```diff
@@ -6,14 +6,10 @@
...
 	<array>
-		<string>applinks:rnbwapp.com</string>
-		<string>applinks:rnbw.app</string>
-		<string>applinks:rnbwappdotcom.app.link</string>
-		<string>applinks:rnbwappdotcom-alternate.app.link</string>
-		<string>applinks:rainbowdotme.app.link</string>
-		<string>applinks:rainbowdotme-alternate.app.link</string>
-		<string>webcredentials:rainbow.me</string>
 		<string>applinks:cardpay.page.link</string>
+		<string>webcredentials:cardstack.cardpay</string>
```
lukemelia: I think this should be `cardstack.com` because the purpose of webcredentials is to allow sharing of secrets between iOS app and safari website.

lukemelia (changes requested):

ios/Rainbow/Info.plist
danibonilha: This was updated automatically on xcode, I left it there just in case we need to test something with dynamic-links, but once we get the universal links working we can remove all fb dynamic links references in a different task, also we are using this somewhere in the scanner.

lukemelia (approved):
👍

PR #350, opened 2021-09-02
[CS-1772] - Allow only appropriate activities while sharing a link

danibonilha:
### Description

The ShareSheet allowed to copy the shared full text, which may confuse the user while copying this assuming it's a link but then it has a bunch of other texts, since we have a specific button to copy the link we can safely remove the copy option from the ShareSheet, using `excludedActivityTypes`  also added a couple of new tests

- [x] Completes #(CS-1772)
- [x] Completes #(CS-1109)
- [x] Completes #(CS-1728)
- [x] Completes #(CS-1702)

### Screenshots

image

### DIFF
--- a/cardstack/src/utils/merchant-utils.ts
+++ b/cardstack/src/utils/merchant-utils.ts
@@ -1,5 +1,8 @@
 import { Resolver } from 'did-resolver';
 import { getResolver } from '@cardstack/did-resolver';
+import { Share } from 'react-native';
+import { getAddressPreview } from './formatting-utils';
+import { Device } from './device';
 import { MerchantRevenueEventFragment } from '@cardstack/graphql';
 import { MerchantInformation } from '@cardstack/types';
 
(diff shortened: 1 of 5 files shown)


eshinkawa:
LGTM

eshinkawa (approved):

cardstack/src/utils/merchant-utils.ts
```diff
@@ -53,13 +56,41 @@ export const fetchMerchantInfoFromDID = async (
   }
 };
 
-// ToDo: Add test once merchant flow finished
 export const generateMerchantPaymentUrl = (
   merchantSafeID: string,
   amount: number,
```
topwebtek7: @danibonilha here amount should be optional and we should consider url without amount param as well. it's from recently updated design :-) 
I know it's not your change but could you update in this pr?

topwebtek7 (approved):
LGTM!

PR #352, opened 2021-09-03
Upgrade reanimated and fix animation issues

topwebtek7:
### Description

I experienced bad animations and sloppy in section list rendering, and able to fix it by upgrading `react-native-animated` to latest version, but had to use `@rainbow-me/animated-charts` rather than using `src/react-native-animated-charts` inside project.
I think we need to push a new build in this project and confirm if this change does not break any existing animations

Sep-07-2021 23-52-02


danibonilha (changes requested):
It's crashing when we tap on assets in the main screen, also inside depots, basically every place that is using ChartExpandedState

scripts/postinstall.sh
```diff
@@ -60,3 +60,7 @@ else
   sentry-cli --version
   echo "sentry-cli is already installed. Skipping" 
 fi
+
+# Remove CocoaAsyncSocket in react-native-tcp to fix duplication issue
+rm -rf node_modules/react-native-tcp/ios/CocoaAsyncSocket
+rm -rf node_modules/react-native-udp/ios/CocoaAsyncSocket
```
danibonilha: EOF

topwebtek7:
> It's crashing when we tap on assets in the main screen, also inside depots, basically every place that is using ChartExpandedState

Resolved this crash issue

danibonilha (changes requested):
It's crashing still =/ . if you open a chart, close it and then open another one, it crashes, plus the chart is a bit funky while sliding through values

topwebtek7:
> It's crashing still =/ . if you open a chart, close it and then open another one, it crashes, plus the chart is a bit funky while sliding through values

Resolved crash issue. It looks ok to be merged for me but please review again

danibonilha (commented):
Send flow crashed

image

Seems related to useTimingTransition

src/components/animations/RoundButtonSizeToggler.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import Animated, { SpringUtils } from 'react-native-reanimated';
-import { bin, useSpringTransition } from 'react-native-redash';
+import { bin, useSpring as useSpringTransition } from 'react-native-redash';
```
danibonilha: If the API has changed, maybe it's better to use the new method name instead of casting to the old value, so we won't forget to remove all old name's api, f.e we can search for `useSpringTransition`  in the whole project and make sure we got rid of all occurrences, bc the search will have 0 results, or we just keep using the old api, at the `v1` path

topwebtek7: @danibonilha resolved crash issue, please check it again

PR #354, opened 2021-09-06
[CS-1740] - Update merchant pay tx request

danibonilha:
### Description

- Creates `MerchantSectionCard` which contains the avatar and merchant name, and receives children so we can reuse this gray card across the pay flow.
- Adds react testing hooks.
- Shares merchantInfoDID logic between components and add tests to it.

For now the currency is defaulting to spend because we don't have the currency selector yet 

The trigger to this pay merchant request should be added in a different PR

- [x] Completes #(CS-1740)

### Checklist

### Screenshots

Screen Shot 2021-09-06 at 18 56 48


topwebtek7 (approved):
LGTM!

kdowlin:
@danibonilha I think you tagged the wrong ticket - pretty sure this is CS-1740 not CS-1742.

PR #358, opened 2021-09-08
[CS-1793 ]- Fix merchant name on QR code

danibonilha:
### Description

The Merchant name disappeared from QR code screen, because we were using multiple sources of truth to get the merchantSafe, in one component we were using from navigation params, others, rainbow selections, in order to avoid inconsistencies I refactor the components to receive the merchantSafe as props, as well to reuse the same navigation logic, also refactor some components  that had the same layout, giving the ability to reuse them.
Added the native currency to the amount, updated the spend fontSize to be the same as native currency and handled bigger merchant names.

- [x] Completes #(CS-1780)

### Checklist

### Screenshots

image


topwebtek7 (approved):
Good refactoring. LGTM! 👍🏼

PR #359, opened 2021-09-08
[CS-1790] - Fix crashing on merchant tx

danibonilha:
### Description

The merchant transactions array have a different structure from the others transactions, which means that it doens't have the `transaction` key in it, so while checking for prev and current with `transactions?.[0].transaction` for the merchant tx, it was causing a crash, the solution for now, is to add a diff behavior if we are handling merchant transactions, but the goal is to add tests, and properly type this flow, since it's use across the whole application, and on every change, it's possible to break something else 

### Screenshots

crash

### DIFF
--- a/cardstack/src/hooks/transactions/use-transaction-sections.tsx
+++ b/cardstack/src/hooks/transactions/use-transaction-sections.tsx
@@ -25,6 +25,7 @@ interface UseTransactionSectionsProps {
   fetchMore?: (props: any) => void;
   merchantSafeAddress?: string;
   transactionStrategies?: TransactionMappingStrategy[];
+  isMerchantTransaction?: boolean;
 }
 
 export const useTransactionSections = ({
@@ -35,6 +36,7 @@ export const useTransactionSections = ({
   fetchMore,
   merchantSafeAddress,
   transactionStrategies,
+  isMerchantTransaction = false,
 }: UseTransactionSectionsProps) => {
   const [sections, setSections] = useState<any[]>([]);
   const [loading, setLoading] = useState(false);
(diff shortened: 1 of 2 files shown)


eshinkawa (approved):

cardstack/src/hooks/transactions/use-transaction-sections.tsx
```diff
@@ -56,12 +58,17 @@ export const useTransactionSections = ({
...
 
-  const prevLastTransaction = usePrevious(transactions?.[0]?.transaction.id);
-  const currentLastTransaction = transactions?.[0]?.transaction.id;
+  const prevLastTransaction = usePrevious(transactions?.[0]?.transaction?.id);
+  const currentLastTransaction = transactions?.[0]?.transaction?.id;
+
+  // Quick workaround to have merchant tx working
+  // TODO: refactor and add tests
+  const shouldUpdate =
+    prevLastTransaction !== currentLastTransaction || isMerchantTransaction;
```
topwebtek7, line 67: @danibonilha just curious, why it renders if merchant transaction even though same transaction id?

danibonilha: All the transactions, have the transaction key in it, and then an id, but merchantTransactions don't, they are like 
```ts
transactions:  {
  [{ id }]
}
```

and. the other transactions are: 

```ts
transactions:  {
  [{  transaction: { id }]
}
```
The right way here is to refactor and use the correct type, so we get always the same object to work with, but in order to do that we need more time, so for now, we are checking only for merchant, we could try to handle previous merchant id also to improve this, but it's probably not worth right now and the previous behavior was to update the tx every render anyways, so this is already kind of a improvement

PR #360, opened 2021-09-08
Update card-pay SDK to 0.23.0

lukemelia:
The SDK version has a signature change for blockchain-mutating SDK methods. There was only spot affected in cardwallet.


lukemelia:
Looks like there is an incompatibility between typescript 4.2.3 (monorepo) and typescript 3.9.7 (cardwallet). I will look into resolving this.

danibonilha (approved):
LGTM

PR #363, opened 2021-09-08
Update to handle new encoded params in signing screen for v0.8.0 of protocol [CS-1729]

habdelra:
This is a change to accommodate new encoded parameters in the wallet connect payload that are included in v0.8.0 of the card protocol for create prepaid card signing requests and split prepaid card signing requests specifically. For these 2 actions, the newly created prepaid cards will be placed into a market contract. These new parameters capture that market contract address (and we need to include them in order for decoding to work properly). 

Hold off on merging this until the new v0.8.0 contracts have been deployed. I'll send a message to the team when it is time for this PR to be merged (it will likely be accompanied by another PR to bump the SDK version)


topwebtek7 (approved):
LGTM! 🚢

PR #365, opened 2021-09-09
[CS-1738] [CS-1727] Add enter amount for merchant payment screen

topwebtek7:
### Description

- Added enter amount for merchant payment screen and integrated with deep linking and confirmation screen. 
- Added currency selection behavior
- Refactored cardstack routes with main routes and global routes like modal

- [x] Completes #CS-1738
- [x] Completes #CS-1727

### Checklist

### Screenshots

merchantpaymentflow

### DIFF
--- a/cardstack/src/components/CurrencySelection/CurrencySelection.tsx
+++ b/cardstack/src/components/CurrencySelection/CurrencySelection.tsx
@@ -16,16 +16,21 @@ const SPDCurrency = {
   symbol: '§',
 };
 
-export const CurrencySelection = () => {
+export const CurrencySelection = ({
+  onChange,
+}: {
+  onChange: (selectedCurrency: string) => void;
+}) => {
   const { paymentChangeCurrency, currency } = usePayment();
 
   const onSelectCurrency = useCallback(
     (selectedCurrency: string) => {
       if (currency !== selectedCurrency) {
         paymentChangeCurrency(selectedCurrency);
+        onChange(selectedCurrency);
       }
     },
-    [currency, paymentChangeCurrency]
+    [currency, paymentChangeCurrency, onChange]
   );
 
   const currencyListItems = [
--- a/cardstack/src/components/CurrencySelection/CurrencySelection.tsx
+++ b/cardstack/src/components/CurrencySelection/CurrencySelection.tsx
@@ -1,27 +1,12 @@
 import analytics from '@segment/analytics-react-native';
-import { isNil } from 'lodash';
 import React, { useCallback } from 'react';
-import { CoinIcon, EmojiText, RadioList, RadioListItem } from '../';
+import { RadioList } from '../';
 import { useAccountSettings } from '@rainbow-me/hooks';
 import { supportedNativeCurrencies } from '@rainbow-me/references';
 
-const renderCurrencyIcon = (currency, emojiName) => {
-  if (!currency) return null;
-  if (!isNil(emojiName)) return <EmojiText name={'flag_' + emojiName} />;
+// ToDo: Refactor RadioListItem with ability to show icon like i.e. renderCurrencyIcon
 
-  return <CoinIcon address={currency} size={23} symbol={currency} />;
-};
-
-const CurrencyListItem = ({ currency, emojiName, label, ...item }) => (
-  <RadioListItem
-    {...item}
-    icon={renderCurrencyIcon(currency, emojiName)}
-    label={`${label}`}
-    value={currency}
-  />
-);
-
-const CurrencySection = () => {
+export const CurrencySelection = () => {
   const { nativeCurrency, settingsChangeNativeCurrency } = useAccountSettings();
 
   const onSelectCurrency = useCallback(
@@ -49,13 +34,5 @@ const CurrencySection = () => {
     }))
     .filter(({ title }) => title !== 'ETH');
 
-  return (
-    <RadioList
-      items={currencyListItems}
-      onChange={onSelectCurrency}
-      // value={nativeCurrency}
-    />
-  );
+  return <RadioList items={currencyListItems} onChange={onSelectCurrency} />;
 };
-
-export default CurrencySection;
--- a/cardstack/src/components/CurrencySelection/CurrencySelection.tsx
+++ b/cardstack/src/components/CurrencySelection/CurrencySelection.tsx
@@ -35,7 +35,9 @@ export const CurrencySelection = () => {
     .map(({ currency: nativeCurrency, label, ...item }, index) => ({
       ...item,
       disabled: false,
-      label: `${label} (${nativeCurrency})`,
+      label: `${label} (${
+        nativeCurrency === 'SPD' ? '§1 = 0.01 USD' : nativeCurrency
+      })`,
       key: index,
       index: index,
       value: nativeCurrency,
--- a/cardstack/src/components/CurrencySelection/CurrencySelection.tsx
+++ b/cardstack/src/components/CurrencySelection/CurrencySelection.tsx
@@ -1,0 +1,61 @@
+import analytics from '@segment/analytics-react-native';
+import { isNil } from 'lodash';
+import React, { useCallback } from 'react';
+import { CoinIcon, EmojiText, RadioList, RadioListItem } from '../';
+import { useAccountSettings } from '@rainbow-me/hooks';
+import { supportedNativeCurrencies } from '@rainbow-me/references';
+
+const renderCurrencyIcon = (currency, emojiName) => {
+  if (!currency) return null;
+  if (!isNil(emojiName)) return <EmojiText name={'flag_' + emojiName} />;
+
+  return <CoinIcon address={currency} size={23} symbol={currency} />;
+};
+
+const CurrencyListItem = ({ currency, emojiName, label, ...item }) => (
+  <RadioListItem
+    {...item}
+    icon={renderCurrencyIcon(currency, emojiName)}
+    label={`${label}`}
+    value={currency}
+  />
+);
+
+const CurrencySection = () => {
+  const { nativeCurrency, settingsChangeNativeCurrency } = useAccountSettings();
+
+  const onSelectCurrency = useCallback(
+    currency => {
+      settingsChangeNativeCurrency(currency);
+      analytics.track('Changed native currency', { currency });
+    },
+    [settingsChangeNativeCurrency]
+  );
+
+  const currencyListItems = Object.values(supportedNativeCurrencies)
+    .map(({ currency, ...item }, index) => ({
+      title: currency,
+      data: [
+        {
+          ...item,
+          disabled: false,
+          label: currency,
+          key: index,
+          value: currency,
+          default: !!currency,
+          selected: currency === nativeCurrency,
+        },
+      ],
+    }))
+    .filter(({ title }) => title !== 'ETH');
+
+  return (
+    <RadioList
+      items={currencyListItems}
+      onChange={onSelectCurrency}
+      // value={nativeCurrency}
+    />
+  );
+};
+
+export default CurrencySection;
--- a/cardstack/src/components/CurrencySelection/CurrencySelection.tsx
+++ b/cardstack/src/components/CurrencySelection/CurrencySelection.tsx
@@ -1,38 +1,52 @@
-import analytics from '@segment/analytics-react-native';
 import React, { useCallback } from 'react';
 import { RadioList } from '../';
-import { useAccountSettings } from '@rainbow-me/hooks';
+import usePayment from '@cardstack/redux/hooks/usePayment';
 import { supportedNativeCurrencies } from '@rainbow-me/references';
 
-// ToDo: Refactor RadioListItem with ability to show icon like i.e. renderCurrencyIcon
+const SPDCurrency = {
+  alignment: 'left',
+  assetLimit: 1,
+  currency: 'SPD',
+  decimals: 2,
+  emojiName: '§',
+  label: 'SPEND',
+  mask: '[099999999999]{.}[00]',
+  placeholder: '0.00',
+  smallThreshold: 1,
+  symbol: '§',
+};
 
 export const CurrencySelection = () => {
-  const { nativeCurrency, settingsChangeNativeCurrency } = useAccountSettings();
+  const { paymentChangeCurrency, currency } = usePayment();
 
   const onSelectCurrency = useCallback(
-    currency => {
-      settingsChangeNativeCurrency(currency);
-      analytics.track('Changed native currency', { currency });
+    (selectedCurrency: string) => {
+      if (currency !== selectedCurrency) {
+        paymentChangeCurrency(selectedCurrency);
+      }
     },
-    [settingsChangeNativeCurrency]
+    [currency, paymentChangeCurrency]
   );
 
-  const currencyListItems = Object.values(supportedNativeCurrencies)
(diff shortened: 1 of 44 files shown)


cardstack/src/components/Icon/custom-icons/doubleCaret.tsx
```diff
@@ -0,0 +1,29 @@
...
+        strokeWidth={2}
+      >
+        <Path data-name="Path 8523" d="M8.914 4.75L5.164 1l-3.75 3.75" />
+        <Path data-name="Path 8524" d="M1.414 12.25L5.164 16l3.75-3.75" />
+      </G>
+    </Svg>
+  );
+}
+
+export default DoubleCaretSvgComponent;
```
danibonilha, line 29: nit: FYI, for svgs that does not require any modification we can use the Svgxml components and just render with the svg string, for example: https://github.com/react-native-svg/react-native-svg#use-with-xml-strings, this way we don't need to convert anything

cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const openCurrencySelectionModal = useCallback(() => {
+      navigate(Routes.CURRENCY_SELECTION_MODAL);
+    }, [navigate]);
+
+    return (
+      <Container
+        flexDirection="row"
+        width="100%"
+        alignItems="center"
+        justifyContent="center"
```
danibonilha: We have the CenteredContainer

cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -0,0 +1,121 @@
...
+      >
+        {hasCurrencySymbol && (
+          <Text
+            color={inputValue ? 'black' : 'underlineGray'}
+            fontWeight="bold"
+            paddingRight={1}
+            paddingTop={1}
+            size="largeBalance"
+          >
+            {(supportedNativeCurrencies as any)[nativeCurrency].symbol}
```
danibonilha: Wondering if instead of using this as a json we should either transform it to a object, or handle this as we would handle an api response

cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -0,0 +1,121 @@
...
+            alignSelf="stretch"
+            autoCapitalize="none"
+            autoCorrect={false}
+            autoFocus
+            color="black"
+            fontSize={30}
+            fontWeight="bold"
+            keyboardType="numeric"
+            maxLength={(inputValue || '').length + 2} // just to avoid possible flicker issue
+            multiline
```
danibonilha, line 75: Do we want multiple for this ?

cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -0,0 +1,121 @@
...
+            testID="RequestPaymentInput"
+            value={inputValue}
+            zIndex={1}
+          />
+        </Container>
+        {setCurrency ? (
+          <Touchable
+            onPress={() => {
+              openCurrencySelectionModal();
+            }}
```
danibonilha:
```suggestion
            onPress={openCurrencySelectionModal}
```

cardstack/src/components/Text/EmojiText.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return isString(name) ? emojisMap.get(normalizeName(name)) : null;
+};
+
+export type EmojiTextProps = {
+  name: string;
+} & TextProps;
+
+export const EmojiText = ({ children, name, ...props }: EmojiTextProps) => {
+  return <Text {...props}>{children || getEmoji(name)}</Text>;
+};
```
danibonilha, line 29: What is this for ? it seems it's not being used anywhere

cardstack/src/screens/PayMerchant.tsx
```diff
@@ -0,0 +1,175 @@
...
+        borderBottomWidth={1}
+        borderBottomColor="borderBlue"
+        paddingBottom={1}
+        hasCurrencySymbol={false}
+        nativeCurrency={selectedCurrency}
+        inputValue={inputValue}
+        setInputValue={setInputValue}
+        setCurrency={setCurrency}
+      />
+      {inputValue && (
```
danibonilha: This needs to be cast to boolean, otherwise it will crash

cardstack/src/navigation/hooks.tsx
```diff
@@ -1,21 +1,27 @@
...
 // Not a big fan of returning components inside hooks,
 // but react-navigation does not allow components other than Screen
 
-export const useCardstackScreens = (Stack: any) => {
-  const screens = Object.entries(
-    Screens
-  ).map(([name, { component, options }]) => (
+const getScreens = (
+  routes: Record<string, string>,
+  screens: Record<string, { component: any; options?: StackNavigationOptions }>,
```
danibonilha, line 11: I believe we have a type for this in the screens file

cardstack/src/screens/PayMerchant.tsx
```diff
@@ -0,0 +1,175 @@
...
+}
+
+const CustomAmountBody = ({
+  merchantInfoDID,
+  goToConfirm,
+  inputValue,
+  setInputValue,
+  isLoading,
+}: CustomAmountBodyProps) => {
+  const [selectedCurrency, setCurrency] = useState<string>('SPD');
```
danibonilha: Should this default to what we receive from the link ? Also the supported currencies json doesn't have SPD how are we planning to deal with that ?

cardstack/src/screens/CurrencySelectionGlobalModal.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import CurrencySection from '@rainbow-me/components/settings-menu/CurrencySection';
+import { Container, SheetHandle } from '@cardstack/components';
+
+const CurrencySelectionGlobalModal = () => {
+  return (
+    <Container
+      flex={1}
+      width="100%"
```
danibonilha: we don't need this

cardstack/src/screens/CurrencySelectionGlobalModal.tsx
```diff
@@ -0,0 +1,30 @@
...
+    >
+      <Container
+        width="100%"
+        borderRadius={20}
+        height="50%"
+        backgroundColor="white"
+        overflow="scroll"
+        paddingBottom={10}
+      >
+        <Container width="100%" alignItems="center" padding={5}>
```
danibonilha, line 26: don't need the width also

cardstack/src/screens/PayMerchant.tsx
```diff
@@ -0,0 +1,175 @@
...
+        <TransactionConfirmationSheet
+          data={{
+            ...data,
+            spendAmount: localCurrencyToAbsNum(`${inputValue || 0}`),
+          }}
+          onCancel={goBack}
+          onConfirm={onConfirm}
+          onConfirmLoading={isLoadingTx}
+          loading={isLoading}
+        />
```
danibonilha: I think we need to align with design, but this should be it's own screen, imagine I entered an amount hit next, and then I realized, oh the amount is wrong, how can I go back to the enter amount screen if now it's this confirmation screen ? I'll need to get the link again, and start the whole flow

cardstack/src/screens/PayMerchant.tsx
```diff
@@ -0,0 +1,175 @@
...
+      ) : (
+        <Container
+          flex={1}
+          alignItems="center"
+          marginTop={4}
+          borderTopRightRadius={20}
+          borderTopLeftRadius={20}
+          backgroundColor="white"
+          width="100%"
```
danibonilha: don't need width

cardstack/src/screens/PayMerchant.tsx
```diff
@@ -0,0 +1,175 @@
...
+          <Text>Next</Text>
+        </Button>
+      </Container>
+    </Container>
+  );
+};
+
+interface AmountInputSectionProps {
+  inputValue: string | undefined;
+  setInputValue: (_val: string | undefined) => void;
```
danibonilha, line 184: we could probably have an interface with this and extends thru the components

cardstack/src/components/CurrencySelection/CurrencySelection.tsx
```diff
@@ -0,0 +1,38 @@
...
+export const CurrencySelection = () => {
+  const { nativeCurrency, settingsChangeNativeCurrency } = useAccountSettings();
+
+  const onSelectCurrency = useCallback(
+    currency => {
+      settingsChangeNativeCurrency(currency);
+      analytics.track('Changed native currency', { currency });
+    },
+    [settingsChangeNativeCurrency]
+  );
```
danibonilha: We shouldn't be using nativeCurrency here, because even if I want to change the currency for this specific payment, it doesn't mean that I want to change for the whole application

danibonilha (changes requested):
It's not working as expected,
- changing the currency, changes the entire app currency and it doesn't reflect in the input currency button, it's always SPD
- the component has some unnecessary re-renders, which refetches merchantInfo and re-checks for no prepaid cards.
- Clicking next goes to the confirmation and we have no way to get back
- Deleting the input value crashes the app
- Using this as a native stack if the app is closed and the click on the link to open it up we get stuck on this screen, without any merchant info, nor screens at the bottom

cardstack/src/components/Input/InputAmount.tsx
topwebtek7: yes we need according to the design

cardstack/src/components/CurrencySelection/CurrencySelection.tsx
topwebtek7: yea, was trying to find a workaround for it.. as we change currency in another modal, it's not easy to change for this specific payment

cardstack/src/components/Icon/custom-icons/doubleCaret.tsx
topwebtek7: we need to update size so created svg component

cardstack/src/components/Input/InputAmount.tsx
topwebtek7: we will fetch this from sdk in next change, worked on it with Michael to be consistent with dapp

cardstack/src/screens/PayMerchant.tsx
topwebtek7: there's edit amount button according to design update and I thought this flow would be more easy to go back and edit amount and re-confirm.

cardstack/src/components/Icon/custom-icons/doubleCaret.tsx
danibonilha: But we can change size with that too, when I said modifications, I meant, conditionally rendering a <Path> for example, or animating an svg component, but this way is fine to, just thought you'd like to know there's more than one way =)

cardstack/src/redux/payment.ts
```diff
@@ -0,0 +1,52 @@
...
+  switch (action.type) {
+    case PAYMENT_UPDATE_CURRENCY_SUCCESS:
+      return {
+        ...state,
+        currency: action.payload,
+      };
+    default:
+      return state;
+  }
+};
```
danibonilha, line 52: We are talking about moving away from redux, I don't think understand the need to create this, since the currency selection screen can just receive a callback and return the selected currency, even if we need to share it globally  we could use context instead of redux, but taking into consideration the value of the feature, we can let this pass and revisit this implementation to remove redux.

cardstack/src/screens/PayMerchant.tsx
```diff
@@ -0,0 +1,219 @@
...
+        merchantInfoDID={merchantInfoDID}
+        goToConfirm={() => setConfirmScreen(true)}
+        inputValue={inputValue}
+        setInputValue={setInputValue}
+        isLoading={props.loading}
+        nativeCurrency={nativeCurrency || 'SPD'}
+      />
+    </Container>
+  );
+});
```
danibonilha: nit: we could wrap the anonymous functions with useCallback

cardstack/src/screens/PayMerchant.tsx
```diff
@@ -0,0 +1,219 @@
...
+        >
+          <AmountInputSection
+            inputValue={inputValue}
+            setInputValue={setInputValue}
+            nativeCurrency={nativeCurrency}
+          />
+        </MerchantSectionCard>
+      </Container>
+      <Container alignItems="center" flex={1}>
+        <Button onPress={goToConfirm}>
```
danibonilha: nit: This is called `goToConfirm` but it actually doesn't do that, it does what the parent says, which is go to confirm, so this probably should be called `onNextPress`, or `onButtonPress` and in the parent, the callback could be goToConfirm and setConfirmScreen

src/navigation/Routes.ios.js
```diff
@@ -137,7 +139,7 @@ function AddCashFlowNavigator() {
 
 function MainNavigator() {
   const initialRoute = useContext(InitialRouteContext);
-  const cardstackScreens = useCardstackScreens(Stack);
+  const cardstackMainScreens = useCardstackMainScreens(Stack);
```
danibonilha, line 142: This change needs to be applied to `Routes.android too`, search for `useCardstackScreens`

cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
```diff
@@ -34,9 +31,9 @@ const handleAlertError = (message: string) => {
   Alert.alert(`Oops!`, message);
 };
 
-const usePaymentMerchantUniversalLink = () => {
+export const usePaymentMerchantUniversalLink = () => {
   const {
-    params: { merchantAddress, amount = '0', network },
+    params: { merchantAddress, amount = '0', network, currency },
   } = useRoute<RouteType>();
```
danibonilha, line 37: Since we are not navigating and this is being extract from the navigation, the first value from the link remains, for example, if my link says it's 300 spend, and I change in `PayMerchant` to 2 spend, the UI will say 2, but the request will be made with 300 spend because it never changed.
So we need to change this implementation, we could either move this from `usePaymentMerchantUniversalLink` and use in `PayMerchant` or in `PayMerchant` once we change the amount or currency, we update the nav params with it.

danibonilha (changes requested):
Couple of things are still not working as expected: 
Major: The request is done with the link amount instead of the edited amount, the comment have more context.
Minor: Selection a currency in the modal only works on second tap.

Since we are implementing the currency selector in this PR, we should probably change the way the merchant link is being created, it's defaulting to Spend, so if the user changes the currency in the request payment, when the other user tries to pay the currency will be wrong, and it will be confusing

cardstack/src/redux/payment.ts
topwebtek7: Personally we can optimize current redux as current redux is not well structured, but i don't think using context is not a great idea.. yeah we can revisit and discuss this later

topwebtek7:
@danibonilha can you review?

danibonilha (approved):
The generated merchant url is not being handled yet, but we can probably do this on a different PR, since we will get from sdk.

topwebtek7:
> The generated merchant url is not being handled yet, but we can probably do this on a different PR, since we will get from sdk.

Yep we have another ticket for that and planning to do that next week monday

PR #366, opened 2021-09-09
[CS-1791] - Add merchant name to tx components

danibonilha:
### Description

- Adds the merchant name on transaction items
- Fixes claimed transaction to show merchantSafe address instead of txHash 
- Handles huge name merchant names on tx items 
- Refactor a couple of huge conditions to object literals for better readability and maintenance, using obj literals for those cases, ensures that if we add a new type of transaction we will not forget to handle it properly in the display, because .ts will complain about missing key. 

PS: I know `useMerchantInfoFromDID` file name does not follow the dash pattern, but I created this way so we can start to move from the dash pattern, since it's easier for to find camelCase files.

- [x] Completes #(CS-1790)

### Checklist

### Screenshots

Screen Shot 2021-09-09 at 14 15 34
Screen Shot 2021-09-09 at 12 13 44


cardstack/src/components/Transactions/TransactionItem.tsx
```diff
@@ -20,41 +20,38 @@ interface TransactionItemProps extends TransactionBaseCustomizationProps {
...
-      return (
-        <MerchantEarnedSpendAndRevenueTransaction {...props} item={item} />
-      );
-    case TransactionTypes.ERC_20:
-      return <ERC20Transaction {...props} item={item} />;
-    default:
-      return null;
-  }
+  const Transaction =
+    transactionItemMap[item.type as TxTypesWithoutReneveueEvent] || null;
```
eshinkawa, line 54: good stuff

eshinkawa (approved):

PR #369, opened 2021-09-13
[Fix]: Keychain types and platform globals

danibonilha:
### Description

Besides the sdk "errors" the newer version os TS got a couple of more errors, this PR solves it, there were custom methods untyped on .ts files and some platform globals missing.

image

~~If we figure any implementation to fix the sdk, I can add to this PR~~
Also it removes the check files step, from the PR, this way we can guarantee that it doesn't matter what was changed, we are still type safe
Bumps the sdk if TS fixes and locks web3 to 1.3.6 to match SDK versions

### DIFF
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -8,32 +8,13 @@ on:
 jobs:
   check:
     outputs:
-      run_verify: ${{ steps.check_files.outputs.run_verify }}
+      run_verify: true
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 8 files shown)


topwebtek7 (commented):
Looks good to me! only one thing is - can you get rid of globalVars in cardstack/.eslintrc.js as you updated ios & android global vars with custom utils so we don't need that in eslint config

topwebtek7 (approved):
LGTM! 🚢

eshinkawa (approved):

PR #371, opened 2021-09-14
[CS-1725] [CS-1788]  - Modal Scrolling / Payment Details Screen

eshinkawa:
### Description

- Recent activity and unclaimed revenues modals were not allowing the user to dismiss them. There's a property on Slacksheet component that needs to be passed by navigation via `setOptions` in order to set their height (really weird, but that's how it works) 

- Adding the  Transaction details screen on Transactions EOA ([reference to the designs](https://linear.app/cardstack/issue/CS-1725/transaction-detail-from-eoa-transaction-screen))

- [X] Completes #(1725) (1788)

### Checklist

### Screenshots
Sep-14-2021 17-56-36-paymt-confirmation
Sep-14-2021 17-55-46-scrolling

### DIFF
--- a/src/components/expanded-state/PaymentConfirmationExpandedState.tsx
+++ b/src/components/expanded-state/PaymentConfirmationExpandedState.tsx
@@ -1,0 +1,159 @@
+import React, { useEffect } from 'react';
+import { SlackSheet } from '../sheet';
+import { BlockscoutButton, Container, Text } from '@cardstack/components';
+import MerchantSectionCard from '@cardstack/components/TransactionConfirmationSheet/displays/components/sections/MerchantSectionCard';
+import {
+  TransactionRow,
+  TransactionRowProps,
+} from '@cardstack/components/Transactions/TransactionBase';
+import { PrepaidCardPaymentTransactionType } from '@cardstack/types';
+import { dateFormatter, screenHeight } from '@cardstack/utils';
+import { ContactAvatar } from '@rainbow-me/components/contacts';
+import { useNavigation } from '@rainbow-me/navigation';
+import { useRainbowSelector } from '@rainbow-me/redux/hooks';
+
+interface PaymentConfirmationExpandedStateProps {
+  asset: Asset;
+  type: string;
+}
+
+interface Asset extends TransactionRowProps {
+  CoinIcon: JSX.Element;
+  Header: any;
+  includeBorder: boolean;
+  index: number;
+  isFullWidth: boolean;
+  primaryText: string;
+  section: Section;
+  statusText: string;
+  subText: string;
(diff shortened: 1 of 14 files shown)


src/components/expanded-state/PaymentConfirmationExpandedState.tsx
```diff
@@ -0,0 +1,215 @@
...
+    <Button
+      marginBottom={12}
+      onPress={onPress}
+      variant="smallWhite"
+      width="100%"
+    >
+      View on Blockscout
+    </Button>
+  );
+};
```
danibonilha: Wondering if this should be a reusable component, since it's the same as MerchantExpandedState

src/components/expanded-state/PaymentConfirmationExpandedState.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+            <Container>
+              <Text
+                ellipsizeMode="tail"
+                fontSize={20}
+                marginBottom={4}
+                numberOfLines={1}
+                weight="extraBold"
+              >
+                {name || ''}
```
danibonilha: We have`MerchantSectionCard`  that has this layout and takes childreb, I think we can use here, and a great addition to this component would be to fallback to `<Icon name="user" size={80} />` is no info

danibonilha (commented):
looking good, just small suggestions

eshinkawa:
nice @danibonilha was able to decrease the code amount in the app

src/components/expanded-state/PaymentConfirmationExpandedState.tsx
```diff
@@ -0,0 +1,159 @@
...
+  textColor,
+  name = null,
+  info,
+  isTimestamp = false,
+}: {
+  title: string;
+  color?: string;
+  textColor?: string;
+  info: any;
+  name?: any;
```
danibonilha: Why any ? Aren't those strings ?

src/components/expanded-state/PaymentConfirmationExpandedState.tsx
```diff
@@ -0,0 +1,159 @@
...
+  isTimestamp?: boolean;
+}) => {
+  return (
+    <Container marginBottom={6} paddingHorizontal={6}>
+      <Text color="blueText" fontSize={13} marginBottom={2} weight="extraBold">
+        {title}
+      </Text>
+      {name && (
```
danibonilha: Let's not forget to cast string and numbers values to boolean, otherwise we may have crashes, since react will try to render the string without being wrapped in  <Text>

danibonilha (approved):
LGTM, just a small fix missing

PR #372, opened 2021-09-15
[CS-1722] - Initial test structure, send tokens

danibonilha:
### Description

This is the initial structure to start testing the Send Depot tokens flow, here we have example of mocks and how to use the renderHooks testing library, this hook is huge so it's missing coverage, since we are prioritizing other flows I just wanted to have the structure up so any of us can write tests for this, which maybe it's nice because then other person can understand the send flow well too.

### DIFF
--- a/cardstack/src/screens/SendSheetDepot/SendSheetDepot.tsx
+++ b/cardstack/src/screens/SendSheetDepot/SendSheetDepot.tsx
@@ -1,0 +1,11 @@
+import React from 'react';
+import { useSendSheetDepotScreen } from './useSendSheetDepotScreen';
+import SendSheet from '@rainbow-me/components/send/SendSheet';
+
+const SendSheetDepot = () => {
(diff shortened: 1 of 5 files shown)


eshinkawa:
LGTM (ran locally, looks fine)

question: does this work for strategies too? I tried to write tests for strategies, but I started to get circular deps errors again

eshinkawa (approved):

danibonilha:
> LGTM (ran locally, looks fine)
> 
> 
> 
> question: does this work for strategies too? I tried to write tests for strategies, but I started to get circular deps errors again
> 
> 

It should, you can see that most you external things are mocked and I'm importing avoiding index's files, of course if it's not a hook we don't need to use the renderHooks libraries, one thing to found out which import is giving errors, is to comment each one, this way you know which work and which doesn't and you can property mock.

PR #373, opened 2021-09-15
[CS-1835] - Fix prepaidcard signing request layout

danibonilha:
### Description

We don't actually need to decode marketAddress and using it was breaking the layout, so we decided to remove

- [x] Completes #(CS-1835)

### Screenshots

image


eshinkawa (approved):
🚢

habdelra (approved):
👍

PR #374, opened 2021-09-16
Adding unit tests for Merchant Earned Spend and Prepaid Card Payment strategies

eshinkawa:
### Description

Following tests made by Dani, I thought it was important to start unit testing strategies, since they carry some important data flows across the app.

I wrote some for two strategies and was able to understand how we can mock some deps. I hope it serves as a `template` in case someone needs to change some of'em.

### Screenshots

image


danibonilha (approved):
👏

PR #375, opened 2021-09-16
[CS-1669] Handle merchant qr code using WC area QR scanner

topwebtek7:
Implemented handling merchant qr code using WC area QR scanner

### Description

- [x] Completes #CS-1669

### Screenshots

Sep-16-2021 21-35-29

### DIFF
--- a/src/hooks/useScanner.js
+++ b/src/hooks/useScanner.js
@@ -154,7 +154,8 @@ export default function useScanner(enabled) {
       }
       if (isValidMerchantPaymentUrl(deeplink)) {
         const params = parseMerchantPaymentDeepLinkUrl(deeplink);
-        return navigate(Routes.PAY_MERCHANT_LINK, params);
+        console.log(deeplink, params);
+        return navigate(Routes.PAY_MERCHANT, params);
       }
       return handleScanInvalid(data);
     },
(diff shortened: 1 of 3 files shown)


src/hooks/useScanner.js
```diff
@@ -150,6 +152,10 @@ export default function useScanner(enabled) {
         haptics.notificationSuccess();
         return handleDeepLink(deeplink.split('?link=')[1]);
       }
+      if (isValidMerchantPaymentUrl(deeplink)) {
+        const params = parseMerchantPaymentDeepLinkUrl(deeplink);
+        return navigate(Routes.PAY_MERCHANT, params);
+      }
```
danibonilha, line 158: We actually don't need to parse parameters , we can just replace the navigate action, with the [useLink](https://reactnavigation.org/docs/use-link-to/) hook from react navigation
```ts
// hook
const navigateWithLink = useLinkTo();
// then call

navigateWithLink(deeplink)

```
Then we can guarantee our link is working properly

topwebtek7: @danibonilha I've tried but it did not work using deeplink like above
so I used navigate and it works smoothly

danibonilha: Alright, let's ship the way it is and we can circle back while restructuring the navigation later

danibonilha (approved):

PR #376, opened 2021-09-17
[CS-1848 1883] - Rounding unclaimed revenue / Dismiss modal bug

eshinkawa:
### Description

1) Rounding unclaimed revenue needs an extra parameter (buffer) to round to `0.00`. That should only happen if the value is < 0.001

2) Dismiss modal is buggy because redux was updating the async state update on FallBackExplorer (
```
store.dispatch(addressAssetsReceived(
```
and that disables dismissing the <Slacksheet /> modal  if it's on fullscreen state

The fix was to use `useMemo` in order to avoid re-renders and, hence, incompatibilities with the modal

That was a bug that was happening to `ALL` modals. 

If you want to reproduce it, (you can see it on Testflight 282):
- start the app
- go to any modal right away 
- pull up to make it fullscreen and wait
- screen will blink (re-render) and won't be able to dismiss the modal

### Checklist

- [X] Completes #(1848) #(1883)

### Screenshots

image

image

Sep-16-2021 22-49-20
Sep-16-2021 22-54-21

### DIFF
--- a/cardstack/src/services/gnosis-service.ts
+++ b/cardstack/src/services/gnosis-service.ts
@@ -119,10 +119,18 @@ export const getTokensWithPrice = async (
           ? usdBalance
           : currencyConversionRates[nativeCurrency] * usdBalance;
 
+      const isAmountDust = nativeBalance < 0.01;
+
+      //decimal places formatting for residual crypto values
+      const bufferValue = isAmountDust ? 0 : undefined;
       return {
         ...tokenItem,
         balance: {
-          ...convertRawAmountToBalance(tokenItem.balance, tokenItem.token),
+          ...convertRawAmountToBalance(
+            tokenItem.balance,
+            tokenItem.token,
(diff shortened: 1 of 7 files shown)


danibonilha (approved):
Looks good, just one thing I'd like to confirm, the dust amount is being applied to all tokens, not just unclaimed one, that is the expected behavior right ? If so, 🚀 .  One suggestion  is, wrapping SlackSheet with React.memo would probably apply the same behavior on the modal, then we wouldn’t  need to add useMemo on return of every component which uses SlackSheet and next modals would be handled too, but there's lots of weird stuff going on so we can take a look later 😅

danibonilha:
Just thought about something, if the token is 0.00 maybe we should disable the claim all button

kdowlin:
> Just thought about something, if the token is 0.00 maybe we should disable the claim all button

I think this idea.

PR #379, opened 2021-09-20
[CS-1855] - Fix HDProvider poller crashing bug 

danibonilha:
### Description

After setting a signed provider with HdProviderWallet, poller processes were building up, for every web3 instance we had, new pollers were created (we suspect), also we were having troubles with the websocket disconnecting once app went to background or inactive state.
This PR refactors the websocket reconnection, creates singleton instances, for web3 and providers and makes sure after using HdProvider to kill the engine.  
Hopefully this is enough to avoid the PollingBlockerTracker crash, I tested the payment request flow, clamming, sending tokens, refreshing tx, let the app sit idle for a while and it seems it's working fine, I was not able to reproduce the crash again.

- [x] Completes #(CS-1855)


eshinkawa (commented):
the previous code doesn't seem to make sense indeed
was it creating an instance (**new Web3()**) for every function call?

eshinkawa (approved):

PR #380, opened 2021-09-21
[CS-1890] - Refactor pay merchant to get card customization

danibonilha:
### Description
 
This PR refactors PayMerchant screen to allow card customization on receipt.

- Extracts MediumPrepaidCard into it's own component along side with Inner and Bottom reusable pieces
- Rises up selected Prepaidcard state to be handled by parent, removing all other 3 intermediate states by adding a callback that retrieves the selectedItem
- UI fixes, removing bottom line on last card item, fix transaction request sticky header layout
- Switching between steps was re-rendering and flickering the cards so this is also handling  memoizing and removes anonymous function in order to prevent flickering.
- Adds transition animation between steps.

Ps: Changing input amount to a value which is not sufficient for some cards, still needs to be handled, and we probably can do this, while handling minimum and max input amount.

- [x] Completes #(CS-1890)

### Checklist

### Screenshots
prepaid-selection


topwebtek7 (approved):
Good refactoring! 🚢

PR #381, opened 2021-09-22
[CS-1874] [CS-1856] Dismiss drilled down modal without dismissing previous one

eshinkawa:
### Description

- 1874 - Bug: Transactions related to receiving payments show a (-) and need to show a (+)

- 1856 - Modals now are opened in different stances - created a new Stack for drilled down sheets so it won't be a re-render of the same modal from previous screen)

- [X] Completes #(1874) #(1856)

### Checklist

### Screenshots

ezgif-7-f2113712e61a


danibonilha (approved):
lgtm

PR #382, opened 2021-09-22
[CS-1721] [CS-1891] Add test coverage for merchant payment flow functions, services and Round Spend value

topwebtek7:
### Description

- Add test coverage for usePaymentMerchantUniversalLink, currency util funcs, inputAmount(will need to recall this after test's setup properly)
- Round spend amount to nearest whole spend

- [x] Completes #CS-1721
- [x] Completes #CS-1891

### DIFF
--- a/cardstack/src/hooks/merchant/__tests__/usePaymentMerchantUniversalLink.test.ts
+++ b/cardstack/src/hooks/merchant/__tests__/usePaymentMerchantUniversalLink.test.ts
@@ -1,0 +1,127 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { useRoute } from '@react-navigation/native';
+import { updatedData } from '../../../../test/helpers/mocks/dataMocks';
+import { usePaymentMerchantUniversalLink } from '../usePaymentMerchantUniversalLink';
+import { useRainbowSelector } from '@rainbow-me/redux/hooks';
+import { getSafeData } from '@cardstack/services';
+
+const mockedDID = {
+  did: 'did',
+  name: 'Merchant Name',
+  slug: 'Merchant slug',
+  color: 'blue',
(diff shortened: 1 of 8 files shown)


eshinkawa (approved):
Great! Gonna be useful for tests are similar to the future 🚢

danibonilha:
With the changes on #380 we can add tests for the good part of flow

cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
```diff
@@ -37,6 +37,12 @@ export const usePaymentMerchantUniversalLink = () => {
     params: { merchantAddress, amount = '0', network, currency },
   } = useRoute<RouteType>();
 
+  const networkName: Network = ['sokol', 'xdai'].includes(network)
+    ? network
+    : Network.sokol;
```
danibonilha, line 42: I wonder which case we wouldn't have the network and if so, we shouldn't really default to 'sokol', because once we move from beta, the default need to 'xdai'

cardstack/src/hooks/merchant/__tests__/usePaymentMerchantUniversalLink.test.ts
```diff
@@ -0,0 +1,139 @@
...
+      params: {
+        merchantAddress: '0x7bAeEbbd7Fd1f41f3DA69A08f8E053C8CCBb592b',
+      },
+    }));
+
+    const { result } = renderHook(() => usePaymentMerchantUniversalLink());
+
+    expect(result.current.data.spendAmount).toBe(0);
+  });
+});
```
danibonilha, line 139: Shouldn't we test the onConfirm function

PR #383, opened 2021-09-22
[CS-1913] - Removes Segment

danibonilha:
### Description

In order to upgrade RN we are first removing unused packages starting with Segment, which rainbows uses to track analytics and we will be not using the same tool, so this PR removes it and also deletes the bitrise script since we use github actions. 

It shouldn't have any impact on the way the app works, I tested the flows, and everything seems good, if you can pull the branch, rebuild it to make sure everything is working it would be great.

- [x] Completes #(CS-1913)


topwebtek7 (approved):
I also experienced same issue in rn-reanimate upgrade and thought it is not needed and you were fast! 👍🏼  looks no critical changes in this pr and all looks good to me 🚢

PR #384, opened 2021-09-22
[CS-1914] - Remove Branch.io

danibonilha:
### Description

To handle notifications and deeplinks rainbows uses [Branch.io](https://branch.io/) which we don't plan to use it, part of the native implementation  on iOS had already been removed as part of deeplinking task and this PR cleans up leftovers

- [x] Completes #(CS-1914])


eshinkawa (approved):
nice 🚢

PR #387, opened 2021-09-23
[CS-1876] - Fix tx update after switching accounts

danibonilha:
### Description

Fixes the tx update after switching accounts, the validation with current/prev values was missing a case, when both values are undefined so once a account had no values, switching to one that have it wouldn't update because undefined === undefined, also remove the auto refresh, because for some reason I don't understand when we open the modal to switch between accounts it sets isFocused to true, which triggers the refresh, and we don't want that while the user is switching account because you may get the other account transactions. 
Also added a reconnect when the ws ends to see if we get less errors related to that.

- [x] Completes #(Linear ticket)
tx-update

### DIFF
--- a/cardstack/src/components/TransactionList/TransactionList.tsx
+++ b/cardstack/src/components/TransactionList/TransactionList.tsx
@@ -44,13 +44,13 @@ export const TransactionList = memo(
     }, [refetch]);
 
     useEffect(() => {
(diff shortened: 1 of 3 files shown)


eshinkawa (approved):

eshinkawa:
🚢

PR #388, opened 2021-09-24
[CS-1838][CS-1900] - Handle deep link when app is closed and update info

danibonilha:
### Description
We have nested navigators and although the previous config seemed correct because it was working when app was open, once it's closed it doesn't know where to go, plus when it's closed we don't have the prepaidCards and stuff ready yet when we go to the screen, so this PR:

- Opens app on right screen when it's closed 
- Adds loading on deep link cold start, to get redux data, there's a workaround with fake loading (open to suggestions if anyone has a better idea), because on first render prepaidcards are empty
- With the right link path, now we can use the `Linking.openUrl` to handle the QRcode
- Added hepatics when QR code is recognized
Update
- Handles single prepaid card case, if user has only one card it goes to InputScreen directly
- 

- [x] Completes #(CS-1838)
- [x] Completes #(CS-1900)

### Screenshots

Open QR code app closed:
pay-merchant-close
Open a different QR code:
pay-merchant-update
Single card: 
single-card

### DIFF
--- a/cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
+++ b/cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
@@ -13,7 +13,7 @@ import { getSafeData } from '@cardstack/services';
 import { useWorker } from '@cardstack/utils';
 import { Network } from '@rainbow-me/helpers/networkTypes';
 import { useRainbowSelector } from '@rainbow-me/redux/hooks';
-import { useWallets } from '@rainbow-me/hooks';
+import { useAssetListData, useWallets } from '@rainbow-me/hooks';
 import Web3Instance from '@cardstack/models/web3-instance';
 import HDProvider from '@cardstack/models/hd-provider';
 
@@ -51,6 +51,8 @@ export const usePaymentMerchantUniversalLink = () => {
     state => state.data.prepaidCards
   );
 
+  const { isLoadingAssets } = useAssetListData();
+
   const { selectedWallet } = useWallets();
 
   const spendAmount = parseFloat(amount);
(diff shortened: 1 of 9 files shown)


cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
```diff
@@ -126,5 +137,12 @@ export const usePaymentMerchantUniversalLink = () => {
     [infoDID, spendAmount, merchantAddress, currencyName]
   );
 
-  return { prepaidCards, goBack, onConfirm, isLoadingTx, isLoading, data };
+  return {
+    prepaidCards,
+    goBack,
+    onConfirm,
+    isLoadingTx,
+    isLoading: isLoading && fakeLoading,
```
topwebtek7: not sure but does it should be like `isLoading || fakeLoading`?

cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
```diff
@@ -51,6 +51,15 @@ export const usePaymentMerchantUniversalLink = () => {
...
   );
 
+  // Workaround to wait redux data update when app is closed
+  const [fakeLoading, setFakeLoading] = useState<boolean>(!prepaidCards.length);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setFakeLoading(false);
+    }, 2000);
+  }, []);
```
topwebtek7: I think there's no guarantee prepaidcards can be loaded in 2 seconds.. 
IMO, we can watch `isLoadingAssets` change from false to true if no prepaidcards
`isLoadingAssets=true` means fetching prepaidCards are finished.

we can get isLoadingAssets like below.
```
const { isLoadingAssets } = useAssetListData();
```

topwebtek7 (changes requested):
Overall this PR's looks good to me and great job! 👍🏼 
Just small comment and suggestion

cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
danibonilha: You're right, didn't know we had a loading prop for that, I thought this prop was always loading because of fallbackExplorer, thanks

eshinkawa (approved):

topwebtek7 (approved):
LGTM! 🚢

PR #389, opened 2021-09-24
[CS-1764] [CS-1763] Request payment confirmation skip amount

topwebtek7:
### Description
- Add skip amount feature in request payment
- Update edit amount screen UI
- Update request payment confirmation UI

- [x] Completes #CS-1763
- [x] Completes #CS-1763

### Checklist

### Screenshots

Sep-24-2021 21-18-18


cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -46,68 +53,88 @@ export const InputAmount = memo(
...
+          <Text
+            size="xxs"
+            textAlign="center"
+            fontWeight="bold"
+            marginBottom={2}
+            textTransform="uppercase"
+            textShadowOffset={{
+              width: 0,
+              height: 1,
+            }}
```
danibonilha: nit: this obj can be created outside component

cardstack/src/screens/PaymentRequest/PaymentRequestExpandedState.tsx
```diff
@@ -0,0 +1,182 @@
...
+          />
+          <SpendAmount
+            formattedAmount={inputValue}
+            nativeCurrencyRate={currencyConversionRate}
+            textCenter
+          />
+        </Container>
+      ) : (
+        <>
+          {amountInSpend && amountInSpend > 0 ? (
```
danibonilha, line 121: Isn't just  `amountInSpend > 0` enough ?

cardstack/src/screens/PaymentRequest/RequestPaymentConfirmation.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const paymentRequestLink = useMemo(
+    () =>
+      generateMerchantPaymentUrl({
+        merchantSafeID: address,
+        amount: amountInSpend,
+        network,
+        currency: 'SPD',
+      }),
+    [address, amountInSpend, network]
+  );
```
danibonilha, line 51: We do have the currency selector now, so we should probably handle this, which maybe already handled, because we get the native amount if asked and convert to spend and then on the payment flow, we do the opposite, if you can confirm the bahvior is correctly we should remove this comment

cardstack/src/screens/PaymentRequest/PaymentRequestExpandedState.tsx
```diff
@@ -0,0 +1,182 @@
...
+            amountInSpend={amountInSpend}
+            amountWithSymbol={amountWithSymbol}
+            merchantInfo={merchantInfo}
+          />
+        </>
+      )}
+    </SlackSheet>
+  );
+};
+
```
danibonilha, line 181: Moving to cardstack, nice 💯

cardstack/src/screens/PaymentRequest/RequestPaymentConfirmation.tsx
```diff
@@ -0,0 +1,220 @@
...
+  return (
+    <Container
+      alignSelf="flex-end"
+      backgroundColor="white"
+      bottom={isTallPhone ? 70 : 0}
+      paddingBottom={5}
+      paddingTop={3}
+      style={shadow.buildAsObject(0, -1, 2, 'rgba(0, 0, 0, 0.25)', 1)}
+      width="100%"
+    >
```
danibonilha, line 207: It seems this divider is based  on the scroll I guess? because we don't have a divider here [zpl](https://zpl.io/25mDO6Q) but we do have it here [zpl](https://zpl.io/2jdQDEp), it would be nice to check with the designer

danibonilha (approved):
Looking good, just small comments

cardstack/src/screens/PaymentRequest/RequestPaymentConfirmation.tsx
topwebtek7: yes working on it in next pr now

cardstack/src/screens/PaymentRequest/RequestPaymentConfirmation.tsx
topwebtek7: i think it's different state that has amount or not.. but will confirm with designer

PR #391, opened 2021-09-27
[CS-1901] Add prepaid card balance update mechanism after pay merchant

topwebtek7:
### Description

Updated prepaid card's faceValue after pay merchant success using data redux action so user can see prepaidcard balance update instantly and then re-fetched assets on the other side to update charts, etc.

- [x] Completes #CS-1901

### Checklist

### Screenshots

Sep-27-2021 21-08-08


danibonilha (approved):
LGTM 🚀

PR #392, opened 2021-09-27
[CS-1945] - Bump Reanimated

danibonilha:
### Description

This PR updates Reanimated and Redash to a version that supports RN latest one, used some rainbows PR's to change as little as possible from current implementation, tried to get rid of embedded charts library, but the external package it's not with the most updated version, so some crashes and weird behaviors were happening, that's why I decided to change the code in the embedded lib to match the updates. I tested some flows, send, charts and didn't experience any issues, if this is looking good we can probably close #352, I will trigger a build from this branch so @kdowlin can validate.

- [x] Completes #(CS-1945)


topwebtek7 (approved):
LGTM! 🚢 
Tested on my device and all works smoothly for me

PR #393, opened 2021-09-28
[CS-1946] - Fix linter issues 

danibonilha:
### Description

This PR adds linter verification for the whole project, consequently we needed to fix all the issues, previously we were only checking the cardstack folder. Also removed some unused files alongside with the charts library example, which was an example within our project we don't use. I tested the flows that I significantly changed but it's worth another pair of eyes.

PS: Linter is taking a long time to run, we probably need to deal with some cache on CI

- [x] Completes #(CS-1946)


topwebtek7 (approved):
LGTM! 🚢

eshinkawa (approved):
🚢

PR #396, opened 2021-09-29
[CS-1865] Buy Prepaid Card UI - Logic not fully implemented

eshinkawa:
### Description

This PR is related **only** to the Buy Prepaid Card UI. 

Part of the functionalities logic has been partially implemented:

Documentation link: https://linear.app/cardstack/issue/CS-1865/build-prepaid-card-purchase-ui

1) useAuthToken - WIP Test
2) useCustodial Wallet - WIP Test
3) Inventory Hook - WIP Test
4) Reservation Hook  - WIP Test

For the next sprint:

1) Add missing tests
2) Add Wyre integration
3) Add rest of the internal wallet logic (on the doc link)

https://user-images.githubusercontent.com/8547776/135201886-41914f7b-fa42-497c-9a11-2ff804222455.mov

- [X] Completes #(1865)

### Checklist


kdowlin:
@eshinkawa I want to confirm we didn't remove any of the existing functionality path for Add Cash for mainnet. This should be all net new path work?

cardstack/src/hooks/prepaid-card/useAuthToken.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const { callback: getAuthToken, error, isLoading } = useWorker(async () => {
+    const web3 = await Web3Instance.get({ selectedWallet, network });
+    const authAPI = await getSDK('HubAuth', web3, hubURL);
+    setAuthToken(await authAPI.authenticate());
+    const isValid = await authAPI.checkValidAuth(authToken);
+
+    if (!isValid) {
+      setAuthToken(await authAPI.authenticate());
+    }
```
danibonilha: What is the goal here ?

cardstack/src/hooks/prepaid-card/useAuthToken.ts
```diff
@@ -0,0 +1,40 @@
+import { useEffect, useState } from 'react';
+import { getSDK } from '@cardstack/cardpay-sdk';
+import Web3Instance from '@cardstack/models/web3-instance';
+import useWallets from '@rainbow-me/hooks/useWallets';
+import { useRainbowSelector } from '@rainbow-me/redux/hooks';
+import { Network } from '@rainbow-me/networkTypes';
+import { useWorker } from '@cardstack/utils';
+import logger from 'logger';
+
+export const useAuthToken = (hubURL?: string) => {
```
danibonilha: `hubURL` is optional but without it we can't really do much, so maybe it should be required

cardstack/src/hooks/prepaid-card/usePrepaidCardReservation.ts
```diff
@@ -0,0 +1,57 @@
...
+        },
+      }),
+    });
+
+    if (results.ok) {
+      const result = await results.json();
+      setReservationData(result.data);
+    }
+  }, [sku]);
+
```
danibonilha: We have `axios` installed on the project, maybe it's 'easier' to create a instance/model, and add all the hub api calls there, to keep it organized and just make the call here with everything formatted. I'm thinking if we simplify these calls, we probably don't need a hook for each, we can have probably just one that retrieves the info, unless of course we are planning to reuse some of these hooks separately

cardstack/src/screens/BuyPrepaidCard.tsx
```diff
@@ -1,76 +1,230 @@
...
-      screen: Routes.ADD_CASH_SCREEN_NAVIGATOR,
-    });
-  };
+const Shimmer = () => {
+  return (
+    <Container>
+      <Skeleton light height={100} width="100%" marginTop={3} />
+    </Container>
+  );
+};
```
danibonilha: nit: can omit return

cardstack/src/screens/BuyPrepaidCard.tsx
```diff
@@ -1,76 +1,230 @@
...
-      height="100%"
+    <Button
+      borderColor={isSelected ? 'buttonPrimaryBorder' : 'buttonSecondaryBorder'}
+      variant={isSelected ? 'squareSelected' : 'square'}
       width="100%"
-      flex={1}
-      padding={5}
+      marginRight={4}
+      marginBottom={3}
+      borderRadius={4}
```
danibonilha: I believe the borderRadius is bigger, around 10 I guess

cardstack/src/screens/BuyPrepaidCard.tsx
```diff
@@ -1,76 +1,230 @@
...
+  );
+
+  return (
+    <Container backgroundColor="backgroundBlue" flex={1}>
+      <Container marginTop={18} padding={4}>
+        <TopContent />
+      </Container>
+      <ScrollView padding={5} flex={10}>
+        <Container width="100%" marginBottom={8}>
+          <Subtitle text="ENTER AMOUNT" />
```
danibonilha: Is it enter or choose amount ?

cardstack/src/screens/BuyPrepaidCard.tsx
```diff
@@ -1,76 +1,230 @@
...
+                      issuerName: 'Cardstack',
+                      patternColor: 'white',
+                      patternUrl:
+                        'https://app.cardstack.com/images/prepaid-card-customizations/pattern-5.svg',
+                      textColor: 'black',
+                    }
+              }
+            />
+          </Container>
+        ) : null}
```
danibonilha: I believe the prepaid card is the Medium one and not the "normal", because based on design the size is smaller, also I'm sure the behavior while pressing on the prepaidcard is correct, isn't it supposed to be just a UI component without functionality ? Currently it's opening the card activities and we can copy the address.

cardstack/src/screens/BuyPrepaidCard.tsx
```diff
@@ -1,76 +1,230 @@
...
+        {card ? (
+          <Container width="100%" marginBottom={16}>
+            <Subtitle text="PREVIEW" />
+            <PrepaidCard
+              networkName={network}
+              nativeCurrency={nativeCurrency}
+              currencyConversionRates={currencyConversionRates}
+              address={card?.attributes['issuing-token-address']}
+              issuer="Cardstack"
+              issuingToken={card?.attributes['issuing-token-symbol']}
```
danibonilha: With the model suggestion using axios and stuff, since we would have the values formatted the way we want, we wouldn't need to use string dash as key

cardstack/src/screens/BuyPrepaidCard.tsx
```diff
@@ -1,76 +1,230 @@
...
+                  : {
+                      background:
+                        'linear-gradient(139.27deg, #00ebe5 34%, #c3fc33 70%)',
+                      issuerName: 'Cardstack',
+                      patternColor: 'white',
+                      patternUrl:
+                        'https://app.cardstack.com/images/prepaid-card-customizations/pattern-5.svg',
+                      textColor: 'black',
+                    }
+              }
```
danibonilha: This conditional object could be outside the component

cardstack/src/screens/BuyPrepaidCard.tsx
```diff
@@ -1,76 +1,230 @@
...
+  return (
+    <Container backgroundColor="backgroundBlue" flex={1}>
+      <Container marginTop={18} padding={4}>
+        <TopContent />
+      </Container>
+      <ScrollView padding={5} flex={10}>
+        <Container width="100%" marginBottom={8}>
+          <Subtitle text="ENTER AMOUNT" />
+          {isLoading ? (
+            <Shimmer />
```
danibonilha: nit: This shimmer, looks like a "bug" I think it's because it's rendering the whole width, maybee if we have two blocks like cardContent, it would be best, but anyways, we can align this with design

cardstack/src/screens/BuyPrepaidCard.tsx
```diff
@@ -1,76 +1,230 @@
...
+        modifiedCards.filter((cardItem: Inventory) => cardItem.isSelected)[0]
+      );
+    },
+    [inventoryData, setInventoryData]
+  );
+
+  const renderItem = useCallback(
+    ({ item, index }: { item: Inventory; index: number }) => (
+      <CardContent
+        onPress={() => onSelectCard(item, index)}
```
danibonilha, line 185: CardContent could have a `handleOnPress` for example, this way we can just pass item to card and retrieve on the callback

src/components/profile/ProfileMasthead.js
```diff
@@ -208,11 +208,9 @@ export default function ProfileMasthead({
...
         </Container>
       </ButtonPressAnimation>
-      {addCashAvailable && (
-        <Container marginTop={4}>
-          <Button onPress={handlePressAddCash}>Add Funds</Button>
-        </Container>
-      )}
+      <Container marginTop={4}>
+        <Button onPress={handlePressAddCash}>Add Funds</Button>
+      </Container>
```
danibonilha, line 227: I think we should keep the mainnet add funds behavior, which means in order to decide where to navigate, we need to check the network

danibonilha: Also the shimmering is overlapping with the copy address button
image

danibonilha:
What is the navigation expected behavior ? This screen should be like a modal ? or come from the side? If so we need a back button or something

src/components/profile/ProfileMasthead.js
```diff
@@ -144,14 +143,14 @@ export default function ProfileMasthead({
...
-  const handlePressAddCash = useCallback(() => {
+  const handleBuyPrepaidCard = useCallback(() => {
     if (isDamaged) {
       showWalletErrorAlert();
       return;
     }
 
     if (ios) {
-      navigate(Routes.ADD_CASH_FLOW);
+      navigate(Routes.BUY_PREPAID_CARD);
```
danibonilha: We need to keep the Add funds behavior on mainnet, We can't just replace this button with BuyPrepaidCard, we need to conditionally choose which button to show and where to navigate to, based on the network, if it's mainnet, it should keep the previous behavior

cardstack/test/jest-setup.js
```diff
@@ -98,5 +98,193 @@ jest.mock('@react-navigation/material-top-tabs', () => ({
...
+
+jest.mock('../../src/references/migratedTokens.json', () => ({}));
+
+jest.mock('../../src/redux/gas', () => ({
+  default: jest.fn(),
+}));
+
+jest.mock('../../src/redux/keyboardHeight', () => ({
+  default: jest.fn(),
+}));
```
danibonilha: Not sure about those relative path mocks in the global setup, because this won't work globally, it's the specific path for the file you are testing, besides if we don't use index based imports on our files we probably don't need to mock all those things, as for the global libraries it seems good

src/components/profile/ProfileMasthead.js
```diff
@@ -144,14 +149,18 @@ export default function ProfileMasthead({
...
       return;
     }
 
     if (ios) {
-      navigate(Routes.ADD_CASH_FLOW);
+      if (isLayer1(network)) {
+        navigate(Routes.ADD_CASH_FLOW);
+      } else {
+        navigate(Routes.BUY_PREPAID_CARD);
+      }
```
danibonilha, line 163: need to add `network` as a dep of the usecallback

danibonilha (approved):
Approving to unblock feature, there are a couple of UI, behaviors and logic inconsistencies which we will need to revisit.

PR #398, opened 2021-09-29
[CS-1840] - Upgrade react native

danibonilha:
### Description

This PRs bumps RN version to 65.1, updating required packages and removing unused ones, there's a warning about virtualized nested lists happening still, but it's just the warning it has not impact the way the app works, from what I tested most of the things are looking fine, but I'd appreciate if you could test on your end, the build and the app, I'll still validate some other things on my end(That's why it's a draft) but the sooner we try to break this, the better.

Update:

- There was an issue that would freeze the app if any alerts were shown, I patched RN to fixed that.
- `Keyboard` does not extends the `NativeEventListener` anymore, thus it doesn't have a `removeListener` method
- Removed outdated types, react and rn provides their own
- Not sure if it's because the new version but had to disable `include_bitcode` in order to have the release build working, this doesn't impact anything, this property it's used if apple for example updates the arch or something if it has the bitcode it can autoupdate, not sure those if it even makes sense for react-native projects.
- Also RN Dev Menu on simulator, stopped showing, but we can use the Shake feature on the simulator to trigger that
- Build now works on Xcode 12.5 

PS: Did try to add the redux plugin to the project, but ran into some issues, we can figure this later

- [x] Completes (#CS-1840)


danibonilha:
~~Just found a crash related to the keyboard, I'm fixing it up~~
Fixed it

eshinkawa (approved):
rebased with my current branch, everything works, great job! 🚢

topwebtek7 (approved):
Tested 310 and all works ok to me. @danibonilha Good job! 🚢

PR #400, opened 2021-10-04
Refactor merchant payment flow to leverage sdk currencies list and util functions

topwebtek7:
### Description

Added SPD to sdk's currencies list and refactored currencies selection using sdk's currencies list and refactored merchant payment request flow as well while fixing bug tickets

- [x] Completes #CS-1897
- [x] Completes #CS-2036
- [x] Completes #CS-1906

### Checklist


cardstack/src/components/CurrencySelection/CurrencySelection.tsx
```diff
@@ -1,54 +1,39 @@
...
-  const currencyListItems = [
-    SPDCurrency,
-    ...Object.values(supportedNativeCurrencies),
-  ]
-    .map(({ currency: nativeCurrency, label, ...item }, index) => ({
+  const currencyListItems = Object.values(supportedNativeCurrencies)
+    .filter(
+      ({ currency }) =>
+        currency !== 'ETH' && currency !== 'DAI' && currency !== 'CARD'
+    )
```
danibonilha: I'm wondering if it makes for nativeCurrencies to have ETH, DAI and CARD, meaning, we probably should have one that are native currencies only, and another one that have the crypto ones, this way we can either merge those, or use then separately, then we don't need to filter on the app, the sdk could return  all supported  `nativeCurrencies`, `cryptoCurrencies` , `allCurrencies` for example.

cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -58,10 +57,7 @@ export const InputAmount = memo(
       );
     }, [nativeCurrency, navigate]);
 
-    const selectedCurrency =
-      nativeCurrency === 'SPD'
-        ? SPDCurrency
-        : (supportedNativeCurrencies as any)[nativeCurrency];
+    const selectedCurrency = (supportedNativeCurrencies as any)[nativeCurrency];
```
danibonilha: We should type the currencies correctly, so we don't need to use any, I've opened a PR with the suggestion 
https://github.com/cardstack/cardstack/pull/2198

cardstack/src/screens/PaymentRequest/PaymentRequestExpandedState.tsx
```diff
@@ -66,19 +79,25 @@ const PaymentRequestExpandedState = (props: { asset: MerchantSafeType }) => {
     </Container>
   );
 
-  const currencyConversionRate =
-    nativeCurrency === 'SPD' ? 100 : currencyConversionRates[nativeCurrency];
+  const amountInNum = formattedCurrencyToAbsNum(inputValue);
```
danibonilha, line 82: We could default this to zero, so we won't need to double check `amountInNum && amountInNum > 0`

danibonilha (approved):
It look goods, I just can't seem to figure why they had `ETH` in the supportedNative currencies, if they filter out on currency selector, not sure if this could have some impact on mainnet

PR #402, opened 2021-10-04
[CS-2041] - Add error boundary and error screen on ws error

danibonilha:
### Description

- Creates a ErrorFallback component 
- Adds error boundary with ErrorFallback to catch any JS exception and allows the user to restart the app
- Adds an ErrorFallbackScreen so we can conditionally navigate to the screen in case of an error with webSocket
- Enables option to fetch on background on iOS to see if it helps when app is inactive to keep the connection alive
- Add a new condition to create a new socket, in case it's disconnected.
- Clean up some unused ios references .
- Handles error on gnosis fetch fails
- Handles error on sendDepot, because if the converter or safes call fails it means the ws is probably with some problems, and without them we can't actually send anything.

- [x] Completes #(CS-2041)

### Checklist

### Screenshots

// Websocket error

Screen Shot 2021-10-01 at 19 22 56
>

// General error
Screen Shot 2021-10-04 at 18 35 19

### DIFF
--- a/cardstack/src/models/web3-instance.ts
+++ b/cardstack/src/models/web3-instance.ts
@@ -23,6 +23,4 @@ const Web3Instance = {
   },
 };
 
-Object.freeze(Web3Instance);
-
 export default Web3Instance;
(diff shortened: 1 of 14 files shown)


cardstack/src/models/web3-instance.ts
```diff
@@ -23,6 +23,4 @@ const Web3Instance = {
   },
 };
 
-Object.freeze(Web3Instance);
```
eshinkawa, line 26: I was going to ask you, why were we doing that

eshinkawa (approved):

cardstack/src/models/web3-instance.ts
danibonilha: This is to make the object immutable, to avoid anyone reassigning the property, usually with singleton we use this, alongside with a getter and a setter, I'm. not sure why it was not working properly, maybe bc the getter is also a setter, but anyways, if we are careful, it shouldn't make much difference

PR #404, opened 2021-10-05
[CS-1918] - Enforce light-content statusbar

danibonilha:
### Description

 Rainbow seems to have support for dark/light mode, but our app defaults to light mode, when most of the screens are actually dark, I thought about setting the mode to dark, but was not sure the impact it might have across the whole application, so as discussed in the last mobile meeting, we are temp removing the conditions, if we want to support dark/light mode in the near future we can just reference this PR, or get rid of the whole theme.

- [x] Completes #(CS-1918)

Screen Shot 2021-10-04 at 18 35 19

Screen Shot 2021-10-04 at 18 35 19

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -13,3 +13,7 @@
 ### Screenshots
 
 <!-- Screenshots or animated GIFs included here -->
+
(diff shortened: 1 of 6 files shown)


topwebtek7 (approved):
LGTM! 🚢

PR #405, opened 2021-10-05
[CS-2060] - Update minimum version screen to match design

danibonilha:
### Description

Update minimum version screen to match design, didn't use SVG because this background img is too complex, and some properties were broken.

- [x] Completes #(CS-2060)

### Checklist

### Screenshots

image


topwebtek7 (approved):
Looks good to me! 
Just a note, It's better to display current app version and minimumVersion in UI, IMO. I think it's more design related discussion and we can discuss later in a meeting

kdowlin:
@danibonilha Do you want to add current version to this screen. I'm not sure we want to necessarily add the updated version. But maybe on the bottom add the current version?

danibonilha:
> @danibonilha Do you want to add current version to this screen. I'm not sure we want to necessarily add the updated version. But maybe on the bottom add the current version?

We can, but since we have that background, I think we should talk to Gardella first, so I'll merge this one and we can revisit this together with the maintenance screen.

PR #407, opened 2021-10-06
[CS-1919]: Only ask for backup when user has safes/assets

danibonilha:
### Description

The backupSheet was appearing every time the user started the app regardless if the user had something or not, this PRs checks if it has safes or assets before prompting the sheet.

Chores:
 - Adds redux-flipper plugin, it has a flag to enable or not, because it's still not working properly, if we have a wallet, it just keeps loading forever and it triggers a navigation error, so we can't use it's full capacity yet, but it works before creating an account, which is already more helpful than it was before, hopefully while migrating the navigation we might be able to fix this.
 - Add Async-storage flipper plugin

- [x] Completes #(CS-1919)


eshinkawa (approved):
LGTM

topwebtek7 (approved):
LGTM! :shipit:

PR #413, opened 2021-10-08
[CS-2095]- Bumps cardpay-sdk to support xdai

danibonilha:
### Description

Bumps cardpay-sdk to support xdai

- [x] Completes #(CS-2095)


topwebtek7 (approved):
:shipit:

PR #414, opened 2021-10-12
[CS-1908] Post signing transaction pending screen

topwebtek7:
### Description

Added global loading overlay component so can be called in other places 

- [x] Completes #CS-1908

### Checklist

### Screenshots

Oct-13-2021 00-28-41


cardstack/src/components/LoadingOverlay/LoadingOverlay.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { Container, CenteredContainer, Text } from '@cardstack/components';
+import { Device } from '@cardstack/utils';
+import { colors } from '@cardstack/theme';
+import ActivityIndicator from '@rainbow-me/components/ActivityIndicator';
```
danibonilha: Can we use react-native's one ?

cardstack/src/components/LoadingOverlay/PortalConsumer.tsx
```diff
@@ -0,0 +1,35 @@
...
+    setComponent,
+    inProcess,
+    processTitle,
+    processSubTitle,
+  ]);
+
+  return null;
+}
+
+export default PortalConsumer;
```
danibonilha: What does the wallet state have to do with the pay transaction load?

danibonilha (dismissed):
I don't think we need to have a provider to handle this tx loading state, also we are mixing the wallet import stuff, with the pay merchant tx stuff, the `IsImporting` state is not working properly anymore, I don't see why we need to track this on a global state, couldn't we do something like the `ERROR_FALLBACK_SCREEN` ? Where we have this as a route and just navigate to it ? I did some kind of draft [here](https://github.com/cardstack/cardwallet/tree/post-signing-load-navigation), and it seems to have worked, let me know if you encounter any pitfalls with this implementation, also there are still some typing errors on the draft that needs to be fixed.

cardstack/src/components/LoadingOverlay/LoadingOverlay.tsx
topwebtek7: animation's quite different with react-native's ActivityIndicator

cardstack/src/components/LoadingOverlay/PortalConsumer.tsx
topwebtek7: transaction load has nothing to do with wallet state loading, it's different behavior and existed before

topwebtek7:
> I don't think we need to have a provider to handle this tx loading state, also we are mixing the wallet import stuff, with the pay merchant tx stuff, the `IsImporting` state is not working properly anymore, I don't see why we need to track this on a global state, couldn't we do something like the `ERROR_FALLBACK_SCREEN` ? Where we have this as a route and just navigate to it ? I did some kind of draft [here](https://github.com/cardstack/cardwallet/tree/post-signing-load-navigation), and it seems to have worked, let me know if you encounter any pitfalls with this implementation, also there are still some typing errors on the draft that needs to be fixed.

I don't agree with you. first was needed to block user ability to exit the pending screen until the transaction is completed, and I still believe it's more easy approach and easy to customize loading animation than navigate to another screen and back again. also it's easy to call in any places.

danibonilha:
> > I don't think we need to have a provider to handle this tx loading state, also we are mixing the wallet import stuff, with the pay merchant tx stuff, the `IsImporting` state is not working properly anymore, I don't see why we need to track this on a global state, couldn't we do something like the `ERROR_FALLBACK_SCREEN` ? Where we have this as a route and just navigate to it ? I did some kind of draft [here](https://github.com/cardstack/cardwallet/tree/post-signing-load-navigation), and it seems to have worked, let me know if you encounter any pitfalls with this implementation, also there are still some typing errors on the draft that needs to be fixed.
> 
> I don't agree with you. first was needed to block user ability to exit the pending screen until the transaction is completed, and I still believe it's more easy approach and easy to customize loading animation than navigate to another screen and back again. also it's easy to call in any places.

The `gestureEnabled: false` nav option, blocks the user ability to exit until the tx is completed. Which scenarios can you think that using navigation would be harder ? How come is it an easier approach if we have multiple "magic" states and we are mixing the PortalConsumer, with wallet import and processing stuff ? If we want to customize it to use on different places we can even have a hook that shows the overlay and then dismisses it then we don't even need redux, if you still think that we need this as a global state, I'd suggest that we take advantage of React context api, to handle the overlay, which should be separated from the current `PortalConsumer` implementation, we should be moving away from outdated libraries dependencies and this way we are creating more dependencies.

topwebtek7:
> > > I don't think we need to have a provider to handle this tx loading state, also we are mixing the wallet import stuff, with the pay merchant tx stuff, the `IsImporting` state is not working properly anymore, I don't see why we need to track this on a global state, couldn't we do something like the `ERROR_FALLBACK_SCREEN` ? Where we have this as a route and just navigate to it ? I did some kind of draft [here](https://github.com/cardstack/cardwallet/tree/post-signing-load-navigation), and it seems to have worked, let me know if you encounter any pitfalls with this implementation, also there are still some typing errors on the draft that needs to be fixed.
> > 
> > 
> > I don't agree with you. first was needed to block user ability to exit the pending screen until the transaction is completed, and I still believe it's more easy approach and easy to customize loading animation than navigate to another screen and back again. also it's easy to call in any places.
> 
> The `gestureEnabled: false` nav option, blocks the user ability to exit until the tx is completed. Which scenarios can you think that using navigation would be harder ? How come is it an easier approach if we have multiple "magic" states and we are mixing the PortalConsumer, with wallet import and processing stuff ? If we want to customize it to use on different places we can even have a hook that shows the overlay and then dismisses it then we don't even need redux, if you still think that we need this as a global state, I'd suggest that we take advantage of React context api, to handle the overlay, which should be separated from the current `PortalConsumer` implementation, we should be moving away from outdated libraries dependencies and this way we are creating more dependencies.

I agree that should move away from old library like `PortalConsumer` but we definitely need globalState and don't understand why you think don't need existing redux structure and creating a new react context api's better?? basically tried already with react context but didn't work for some reason for me.. and I think using redux's easier for me and more organized.
anyway I'm updating this pr

danibonilha:
> > > > I don't think we need to have a provider to handle this tx loading state, also we are mixing the wallet import stuff, with the pay merchant tx stuff, the `IsImporting` state is not working properly anymore, I don't see why we need to track this on a global state, couldn't we do something like the `ERROR_FALLBACK_SCREEN` ? Where we have this as a route and just navigate to it ? I did some kind of draft [here](https://github.com/cardstack/cardwallet/tree/post-signing-load-navigation), and it seems to have worked, let me know if you encounter any pitfalls with this implementation, also there are still some typing errors on the draft that needs to be fixed.
> > > 
> > > 
> > > I don't agree with you. first was needed to block user ability to exit the pending screen until the transaction is completed, and I still believe it's more easy approach and easy to customize loading animation than navigate to another screen and back again. also it's easy to call in any places.
> > 
> > 
> > The `gestureEnabled: false` nav option, blocks the user ability to exit until the tx is completed. Which scenarios can you think that using navigation would be harder ? How come is it an easier approach if we have multiple "magic" states and we are mixing the PortalConsumer, with wallet import and processing stuff ? If we want to customize it to use on different places we can even have a hook that shows the overlay and then dismisses it then we don't even need redux, if you still think that we need this as a global state, I'd suggest that we take advantage of React context api, to handle the overlay, which should be separated from the current `PortalConsumer` implementation, we should be moving away from outdated libraries dependencies and this way we are creating more dependencies.
> 
> I agree that should move away from old library like `PortalConsumer` but we definitely need globalState and don't understand why you think don't need existing redux structure and creating a new react context api's better?? basically tried already with react context but didn't work for some reason for me.. and I think using redux's easier for me and more organized. anyway I'm updating this pr

Redux makes everything accessible from any place, which is confusing and can lead to mistakes, it's really hard to track where things are being changed and why (specifically for us, since we don't have the ability to debug properly), we definitely need global state, but some, can be scoped, and with the currently redux implementation we can't do that, by using the context api for example we can make sure that a specific components tree has access to information, but the rest won't, once we find a provider we can check all the logic that's happening, with redux we need to go to 2,3 files to try to understand what's going on, besides the performance issue to be re-rendering a bunch of things multiple times. If you think the way it is, is a must have, just make sure the importing wallet state is working, and we can talk as a team about the future of redux in the project, to either partially or totally adopt it, or a migration plan.

eshinkawa:
Moving away from Rainbow stuff is on the roadmap (and as Kiel mentioned, same repo). Given that, I think we should pursue 
a better global state implementations than Redux.

Right we are dealing with lots of data sources which is always confusing to me:
1) GraphQL data
2) External API calls, for eg. Wyre
3) Internal API calls for eg. Hub
4) Navigation params
5) Local storage
6) Redux

We could try to create consistent context providers that we could use with **useContext()** and deal with all these data flows (when needed of course)? TS would help a lot on that sense with proper interfaces and more predictable IO.

Mixing our logic with Rainbow's logic gonna cause problems on the long run and if we keep using Rainbow's Redux logic we are bound to create more and more tight coupling to their project.

lukemelia:
Sounds like we need to align as a team in our direction around this so everyone knows what the goal to iterate towards is. Good topic for our next mobile meeting.

topwebtek7:
Yeah I also think it would be a good topic for next mobile meeting

danibonilha:
So we need to make a final decision for this PR, my proposal is this one -> [Comparing Changes](https://github.com/cardstack/cardwallet/compare/post-signing-transaction-pending-screen-cs-1908...post-signing-load-navigation?expand=1) 
- React Navigation has it's own state, so we can track based on a specific screen. ie LoadingOverlay.
- We are showing an UI component and it can be a route, it's clear that something is appearing or not on the screen, in contrast to having a global component sibling with the AppContainer to show an overlay.
- We have a hook to make this behavior reusable, so anywhere we can just call `useLoadingOverlay` with any title and subtitle, in any situations, not just related to Payment for example, we can also use on buy/send flows.
- We don't need to change rainbow's Portal, importing wallet code, which makes sure the current behavior is not affected.
- We don't need to track any redux actions because we are already controlling what is happening thru navigation state.
- It's clearer what show/dismiss overlay does, rather than, paymentProcessing/paymentDone.

@lukemelia @eshinkawa @topwebtek7 let me know your thoughts on this approach, or if we rather keep the way it is right now.

topwebtek7:
> So we need to make a final decision for this PR, my proposal is this one -> [Comparing Changes](https://github.com/cardstack/cardwallet/compare/post-signing-transaction-pending-screen-cs-1908...post-signing-load-navigation?expand=1)
> 
> * React Navigation has it's own state, so we can track based on a specific screen. ie LoadingOverlay.
> * We are showing an UI component and it can be a route, it's clear that something is appearing or not on the screen, in contrast to having a global component sibling with the AppContainer to show an overlay.
> * We have a hook to make this behavior reusable, so anywhere we can just call `useLoadingOverlay` with any title and subtitle, in any situations, not just related to Payment for example, we can also use on buy/send flows.
> * We don't need to change rainbow's Portal, importing wallet code, which makes sure the current behavior is not affected.
> * We don't need to track any redux actions because we are already controlling what is happening thru navigation state.
> * It's clearer what show/dismiss overlay does, rather than, paymentProcessing/paymentDone.
> 
> @lukemelia @eshinkawa @topwebtek7 let me know your thoughts on this approach, or if we rather keep the way it is right now.

@danibonilha Yeah, it's true that your approach is a easy way to go, but.. we can't use that LoadingOverlay route screen on some screens like importing wallet as need to `goback()` and navigate to that screen, and then `goback` twice is really weird behavior, also don't wanna use separate loading behaviors with same thing. also I believe we might have cases that needs to navigate to the other screen in the back side of processing animation(user can see those navigates through transparent background while processing) like payment transaction or whatever. That's why I determined to go with global processing animation behavior and it's global state, so we use redux to manage the state, and I think we can keep the way it is for now.. we have design meeting on Tuesday and will discuss new navigation stuff, after that will need to refactor all navigations, get rid of rainbow's old navigation structure, Portal, etc. so we can recall cleaning it up without Portal at that time, IMO.

topwebtek7:
@danibonilha @eshinkawa  It's updated as discussed and please review this pr again

kdowlin:
@topwebtek7 I think you meant @eshinkawa right?

topwebtek7:
> @topwebtek7 I think you meant @eshinkawa right?

ah yes. sorry for that

src/screens/ImportSeedPhraseSheet.js
```diff
@@ -324,6 +324,7 @@ export default function ImportSeedPhraseSheet() {
                     screen: Routes.WALLET_SCREEN,
                   });
                 } else {
+                  goBack();
                   navigate(Routes.WALLET_SCREEN, { initialized: true });
```
danibonilha, line 328: No need to change this one, because I'm handling on the wallets creation PR =)

cardstack/src/redux/appState.ts
```diff
@@ -1,20 +1,25 @@
...
-// -- Reducer ----------------------------------------- //
-const INITIAL_STATE = {
-  walletReady: false,
-};
-
-export default (state = INITIAL_STATE, action) => {
+export default (state = INITIAL_STATE, action: AnyAction) => {
   switch (action.type) {
     case APP_STATE_UPDATE:
       return { ...state, ...action.payload };
```
danibonilha, line 25: Personally I don't feel we needed to move this to cardstack's folder in this PR, but I guess it's okay

danibonilha (approved):

PR #415, opened 2021-10-12
[CS-2105] - Fix testsnet toggle crash

danibonilha:
### Description

Fixes testnet toggle crash and keep the state of last selected network, this is almost a workaround, because the underlying problem is actually inside the `RadioList` component, it has a useEffect that's being triggered with undefined values sometimes, but since it's being used by other components, we need to investigate more carefully.

- [x] Completes #(CS-2105)

### Screenshots

Simulator Screen Recording - iPhone 11 - 2021-10-12 at 16 09 37

### DIFF
--- a/src/components/settings-menu/NetworkSection.js
+++ b/src/components/settings-menu/NetworkSection.js
@@ -16,7 +16,6 @@ import {
 } from '../../redux/settings';
 
 import { Checkbox, Container, RadioList, Text } from '@cardstack/components';
-import networkTypes from '@rainbow-me/helpers/networkTypes';
 
 const networks = values(networkInfo);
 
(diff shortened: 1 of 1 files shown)


src/components/settings-menu/NetworkSection.js
```diff
@@ -83,15 +92,10 @@ const NetworkSection = () => {
         <Checkbox
           isSelected={showTestnets}
           label="Show testnets"
-          onPress={() => {
-            if (networkSelected.isTestnet && showTestnets) {
-              onNetworkChange(networkTypes.mainnet);
-            }
```
topwebtek7, line 89: When you are on testnet and uncheck ShowTestnet makes nothing is selected visually.. isn't it a bit weird
maybe we need a discussion what's better user experience here

topwebtek7 (approved):
Looks good to me and left a comment for a concern

src/components/settings-menu/NetworkSection.js
danibonilha: I agree, I guess this was the default behavior before, then it seemed a bit weird and this was added, but now users found confusing and the ticket specified it should keep the last selected anyway  🤷🏻‍♀️

PR #416, opened 2021-10-14
[CS-1821] Add min validation in payment request and pay merchant flow

topwebtek7:
### Description

Added minimum amount validation in payment request and pay merchant flow and did some refactoring with currency utils consuming sdk util functions as noticed some decimal amount difference with dapp

- [x] Completes #CS-1821

### Checklist

### Screenshots

image

### DIFF
--- a/cardstack/src/screens/PayMerchant/PayMerchant.tsx
+++ b/cardstack/src/screens/PayMerchant/PayMerchant.tsx
@@ -1,8 +1,17 @@
 import React, { memo } from 'react';
 import { ActivityIndicator } from 'react-native';
 import { NativeCurrency } from '@cardstack/cardpay-sdk/sdk/currencies';
+import {
+  convertAmountAndPriceToNativeDisplay,
+  spendToUsd,
+} from '@cardstack/cardpay-sdk';
 import ChoosePrepaidCard from './ChoosePrepaidCard';
 import { usePayMerchant, PAY_STEP } from './usePayMerchant';
+import {
+  AmountInNativeCurrency,
+  MIN_SPEND_AMOUNT,
+  useAmountConvertHelper,
+} from '@cardstack/screen/PaymentRequest/helper';
 import MerchantSectionCard from '@cardstack/components/TransactionConfirmationSheet/displays/components/sections/MerchantSectionCard';
 import {
   SafeAreaView,
@@ -16,11 +25,6 @@ import {
   CenteredContainer,
 } from '@cardstack/components';
 import { MerchantInformation } from '@cardstack/types';
-import {
-  convertSpendForBalanceDisplay,
-  formattedCurrencyToAbsNum,
-  nativeCurrencyToSpend,
-} from '@cardstack/utils';
 import { useNativeCurrencyAndConversionRates } from '@rainbow-me/redux/hooks';
 
 const PayMerchant = memo(() => {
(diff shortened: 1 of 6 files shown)


kdowlin:
This UI looks super nice - nicely done Brian

cardstack/src/screens/PayMerchant/PayMerchant.tsx
```diff
@@ -108,48 +120,71 @@ const CustomAmountBody = memo(
...
+                setInputValue={setInputValue}
+                amountInAnotherCurrency={amountInAnotherCurrency.display}
+                nativeCurrency={nativeCurrency}
+                isInvalid={isInvalid}
+                currencyConversionRates={currencyConversionRates}
+              />
+            </MerchantSectionCard>
+          </Container>
+          <Container alignItems="center" flex={1} paddingHorizontal={5}>
+            <Button onPress={onNextPress} disabled={!amountInNum || isInvalid}>
```
danibonilha: ~~`isInvalid` is based on `amountInNum > 0` right ? So do we need to check if `!amountInNum` ? If so , couldn't we check that on the `isInvalid` const too?~~
I think understood why use `!amountInNum || isInvalid`, but I needed to check the function in order to get that, so what do you think of having a const inside the hook, something like `canSubmit` this way we just enable the buttons based on this, although if the user types 0, it should be consider invalid, unless s/he presses skip amount 🤔  so probably it can be in the `isInvalid` const I guess

cardstack/src/screens/PaymentRequest/PaymentRequestExpandedState.tsx
```diff
@@ -63,10 +61,22 @@ const PaymentRequestExpandedState = (props: { asset: MerchantSafeType }) => {
...
+    nativeCurrency,
+    accountNativeCurrency,
+    currencyConversionRates
+  );
+
   const EditFooter = () => (
     <Container paddingHorizontal={5}>
       <Button
-        disabled={!inputValue}
+        disabled={!amountInNum || isInvalid}
```
danibonilha: ditto

cardstack/src/screens/PayMerchant/PayMerchant.tsx
```diff
@@ -164,21 +199,29 @@ const AmountInputSection = memo(
...
+            color="red"
+            marginTop={1}
+          >{`minimum ${MIN_SPEND_AMOUNT} spend (${
+            convertAmountAndPriceToNativeDisplay(
+              spendToUsd(MIN_SPEND_AMOUNT) || 0,
+              currencyConversionRates[nativeCurrency],
+              nativeCurrency
+            ).display
+          })`}</Text>
+        ) : null}
```
danibonilha: nit: This piece is the same as the one used on RequestPayment, maybe we can have a component to keep consistent in case we need to change something eventually, but we can do it later if you prefer

cardstack/src/screens/PaymentRequest/helper.tsx
```diff
@@ -36,16 +48,56 @@ export const RequestPaymentMerchantInfo = ({
...
+          inputValue,
+          currencyConversionRates[inputNativeCurrency],
+          true
+        );
+
+  // input amount should be more than MIN_SPEND_AMOUNT (50)
+  const isInvalid =
+    amountInNum > 0 &&
+    (inputNativeCurrency === NativeCurrency.SPD
+      ? amountInNum
```
danibonilha: should we cast this ? Since the return is supposed to be boolean ?

danibonilha (commented):
Lookin really good, just one small concern

cardstack/src/screens/PayMerchant/PayMerchant.tsx
topwebtek7: ok, good point. if types 0, can't submit as button disabled with `!amountInNum || isInvalid`

danibonilha (approved):
🚀

PR #417, opened 2021-10-14
[Chore]:  Create reduxLogger middleware

danibonilha:
### Description

As I mentioned, using redux-flipper, and trying to log `store.getState()` with `JSON.stringify`  was breaking the redux debug middleware, because of circular deps !? I guess, even using `resolveCyclic` so I decide to create our own logger, this way we can log all redux reducers keys, or specify the keys we want to not flood the entire log.

To get all the reducers assign an empty array to `singleKeys`:
```ts
 const singleKeys: string[] = [];
```
To get specify reducers, just add them to the array: 

```ts
const singleKeys: string[] = ['wallets', 'appState'];
```
and set `enableReduxLogger` to `true``

The keys need to be one of the reducers name, we could try and type this, but I don't think it's worth, since it's pretty straight forward and just for debug purposes.

I'm putting it up so it can help with any redux debugging issue we may face. =)

### DIFF
--- a/src/redux/store.ts
+++ b/src/redux/store.ts
@@ -4,15 +4,42 @@ import thunk from 'redux-thunk';
 import reducers from './reducers';
 
 const enableReduxFlipper = false;
+const enableReduxLogger = false;
 
-const middlewares = [thunk];
+// Ignoring types, bc it's just a debug helper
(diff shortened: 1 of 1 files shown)


eshinkawa (approved):
Looks good, can we have a walkthrough next week about redux debugging?

PR #419, opened 2021-10-15
[CS-1978] Update useMerchant so it redirects to edit amount if not amount specified

topwebtek7:
### Description

- [x] Completes #CS-1978

### Checklist


danibonilha (approved):
🚀

PR #421, opened 2021-10-15
Tweak docs based on Luke's dev env setup experience

lukemelia:


ios/Podfile.lock
```diff
@@ -918,7 +918,7 @@ SPEC CHECKSUMS:
   Flipper-RSocket: d9d9ade67cbecf6ac10730304bf5607266dd2541
   FlipperKit: aec2d931adeee48a07bab1ea8bcc8a6bb87dfce4
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
-  glog: 5337263514dd6f09803962437687240c5dc39aa4
+  glog: 40a13f7840415b9a77023fbcae0f1e6f43192af3
   GoogleDataTransport: 85fd18ff3019bb85d3f2c551d04c481dedf71fc9
   GoogleUtilities: 8de2a97a17e15b6b98e38e8770e2d129a57c0040
   libevent: 4049cae6c81cdb3654a443be001fb9bdceff7913
```
danibonilha, line 924: Wondering why the podfile.lock got updated 🤔, did you ran pod install outside the yarn commands?

.ruby-version
```diff
@@ -0,0 +1 @@
+2.6.8
```
danibonilha, line 1: Maybe we should also lock in the gemfile? Just considering that are some commands where we can override the setup from `.ruby-version`

lukemelia: Updated ✅

ios/Podfile.lock
lukemelia: Good question. I just verified by reverting the Podfile.lock changes and re-running `yarn install-pods` and it still ends up updated. Do you think I should exclude the change from this commit?

danibonilha: I think it's fine, it may have something to do with the ruby version I'm guessing

danibonilha (approved):
looks good, thanks!

PR #422, opened 2021-10-15
[CS-1852]: Update Merchant Background Header to Avatar Color

danibonilha:
### Description

Updates MerchantSafeHeader background to Avatar Color

PS: We are not handling the header in the merchant screen, bc it requires to treat the StatusBar appearance which is not worth right now.

- [x] Completes #(CS-1852)

### Screenshots

Screenshot


topwebtek7 (approved):
LGTM! 🚢

PR #426, opened 2021-10-19
[CS-2154] Fix qrcode display bug in small screen, update qrcode with payment request deeplink

topwebtek7:
### Description

Replaced QRCode generation with new library(easyqrcode-react-native, not a great library but most close RN library to the dapp's qrcode style) and fixed a bug that generates QR Code with payment request web link.

- [x] Completes #CS-2154

### Checklist

### Screenshots

Screen Shot 2021-10-19 at 8 02 17 PM


danibonilha (approved):
looks good, maybe its missing a bit more of padding between the code and the QrCode wrapper, but it could be just my eyes

PR #427, opened 2021-10-19
[CS-2089] Filter card inventory by issuer address

eshinkawa:
### Description

- [X] Completes #(2089)

Fixes:
- Card inventory now being properly filtered
- Card customization styles are now being properly applied
- API calls refactor now using UseWorker hook

### Checklist

### Screenshots

CS-2089

### DIFF
--- a/cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
+++ b/cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
@@ -1,4 +1,5 @@
 import { useCallback, useEffect, useState } from 'react';
+import { getAddressByNetwork } from '@cardstack/cardpay-sdk';
 import { Alert } from '@rainbow-me/components/alerts';
 import {
   getOrderId,
@@ -27,6 +28,7 @@ import {
 import {
   fetchCardCustomizationFromDID,
   getNativeBalanceFromSpend,
+  useWorker,
 } from '@cardstack/utils';
 import { PrepaidCardCustomization } from '@cardstack/types';
 
@@ -37,7 +39,7 @@ const getHubUrl = (network: Network) =>
   network === Network.xdai ? HUB_URL_PROD : HUB_URL_STAGING;
 
 interface CardAttrs extends InventoryAttrs {
-  customizationDID?: PrepaidCardCustomization;
+  customizationDID?: PrepaidCardCustomization | null;
 }
 export interface Card {
   id: string;
@@ -64,7 +66,6 @@ export default function useBuyPrepaidCard() {
   const [inventoryData, setInventoryData] = useState<Inventory[]>();
   const [sku, setSku] = useState<string>('');
   const [isLoading, setIsLoading] = useState<boolean>(false);
-  const [isInventoryLoading, setIsInventoryLoading] = useState<boolean>(false);
   const [wyreOrderId, setWyreOrderId] = useState<string>('');
 
   const [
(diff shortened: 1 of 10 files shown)


cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
```diff
@@ -133,25 +138,25 @@ export default function useBuyPrepaidCard() {
...
 
+      const formattedCard = modifiedCards
+        .filter((cardItem: Card) => cardItem.isSelected)
+        .map((carditem: Card) => {
+          return {
+            ...carditem,
+            customizationDID,
+          };
+        })[0].attributes;
+
```
danibonilha, line 156: can you add a comment why we need to do this ?

cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
```diff
@@ -133,25 +138,25 @@ export default function useBuyPrepaidCard() {
...
-        modifiedCards
-          .filter((cardItem: Card) => cardItem.isSelected)
-          .map((carditem: Card) => {
-            return {
-              ...carditem,
-              customizationDid,
-            };
-          })[0].attributes
-      );
+      setCard({ ...formattedCard, customizationDID });
```
danibonilha, line 158: we add customizationDID in the formattedCard, and then, here again, is this intentional ?

cardstack/src/screens/BuyPrepaidCard/Components.tsx
```diff
@@ -27,11 +27,13 @@ export const TopContent = () => {
...
-  borderRadius: 10,
-  width: '100%',
+const baseStyles = (isEven: boolean) => {
+  return {
+    marginRight: 4,
+    marginBottom: 3,
+    borderRadius: 10,
+    width: isEven ? '67.4%' : '100%',
+  };
 };
```
danibonilha: We shouldn't need to track the quantity to get the components wrapped right, also, we should probably ask design what is the expected behavior when we have an odd number of inventory, I feel that filling the whole width is better: 

I did some layout adjustments to achieve this: 
image

if the expected behavior is to have all items with same width, then just add `maxWitdh: '40%'` on CardContent

Feel free to drop the commit if it doesn't match the behavior

danibonilha (approved):
Looks good!

cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
eshinkawa: **Card** component only needs `CardAttrs` types which is an extension of `InventoryAttrs` plus the `CustomizationDID`

cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
eshinkawa: Same answer above

We can use Card type but I wanted to be more specific here about what we were using

edit: got now what you were saying, removing it

cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
```diff
@@ -93,35 +94,39 @@ export default function useBuyPrepaidCard() {
...
+        buttons: [{ text: 'Okay' }],
+        message: 'Error loading inventory',
+      });
+    }
+  }, [inventoryError]);
 
-    getCustodialWalletData();
+  const { callback: getCustodialWalletData } = useWorker(async () => {
+    const data = await getCustodialWallet(hubURL, authToken);
+    setCustodialWalletData(data);
```
topwebtek7: basically as far I understood, useWorker is for tracking errors, loading behavior but we didn't use those here.. so IMO we don't need to use `useWorker` as it might be called multiple times.. (for me, called twice)

PR #429, opened 2021-10-20
[Fix]: Disabled buttons layout

danibonilha:
### Description

Some buttons need to be wrapped in a Container to work properly and others don't, so I'm creating the Conditional wrapper to handle those cases, tested most places that have disabled buttons, but it might exist on some flow I don't know about.

Simulator Screen Recording - iPhone 11 - 2021-10-20 at 14 47 42
.


cardstack/src/components/Button/Button.tsx
```diff
@@ -54,6 +67,7 @@ export const Button = ({
   disabled,
   iconProps,
   iconPosition = 'left',
+  wrapper = 'container',
```
eshinkawa, line 70: yeah that looks good

eshinkawa (approved):

PR #433, opened 2021-10-21
[CS-2200] Update prepaid card transfer transaction faceValue with original value

topwebtek7:
### Description

Updated subgraph query and used original faceValue(creation.spendAmount) when prepaid transferred(basically they make them when transfer) to make it work properly on both issuer and receiver accounts.

- [x] Completes #CS-2200

### Checklist

### Screenshots

Screen Shot 2021-10-22 at 2 17 01 AM


danibonilha (approved):
🚀

PR #434, opened 2021-10-21
[CS-2220][CS-2222]: Update PrepaidCards section on purchase

danibonilha:
### Description

This updates the prepaidcards section with the new purchase one, once the tx is done successfully we navigate the user to the wallet screen and auto scroll to the PrepaidCard section so the user can see it's new card. 
Fixes `Show Secret Recovery Phrase`  button UI.
Fixes refresh, the previous implementation was just dispatching an action to redux, without waiting the promise to be resolved, we don't actually need to call fallbackExplorerInit when we want to update just the assets, so now we are calling `fetchAssetsBalancesAndPrices` directly which is enough to wait for the updated items.

- [x] Completes #(CS-2222)
- [x] Completes #(CS-2220)

### Screenshots

https://user-images.githubusercontent.com/20520102/138346545-642599c6-48ec-4f56-b164-4c1ee6d9f35e.mov

### DIFF
--- a/cardstack/src/components/AssetList/AssetList.tsx
+++ b/cardstack/src/components/AssetList/AssetList.tsx
@@ -1,7 +1,8 @@
-import React, { useState, useCallback } from 'react';
+import React, { useState, useCallback, useEffect, createRef } from 'react';
 import { LayoutAnimation, RefreshControl, SectionList } from 'react-native';
 import { BackgroundColorProps, ColorProps } from '@shopify/restyle';
 import { getConstantByNetwork } from '@cardstack/cardpay-sdk';
+import { useRoute } from '@react-navigation/core';
 import AddFundsInterstitial from '../../../../src/components/AddFundsInterstitial';
 import ButtonPressAnimation from '../../../../src/components/animations/ButtonPressAnimation';
 import { Theme } from '../../theme';
@@ -26,6 +27,7 @@ import showWalletErrorAlert from '@rainbow-me/helpers/support';
 import { useNavigation } from '@rainbow-me/navigation';
 import Routes from '@rainbow-me/routes';
 import { showActionSheetWithOptions } from '@rainbow-me/utils';
+import logger from 'logger';
 
 interface HeaderItem {
   title: string;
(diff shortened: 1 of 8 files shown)


src/redux/data.js
```diff
@@ -663,6 +663,16 @@ export const updateRefetchSavings = fetch => dispatch =>
...
+export const addNewPrepaidCard = newPrepaidCard => (dispatch, getState) => {
+  const { prepaidCards } = getState().data;
+  const updatedCards = [newPrepaidCard, ...prepaidCards];
+
+  dispatch({
+    payload: { prepaidCards: updatedCards },
+    type: DATA_UPDATE_PREPAIDCARDS,
+  });
+};
+
```
danibonilha, line 675: For now, I'm adding it to the data file, because it's were the prepaidcard, depots, etc are added, "it's faster",  but I'm planning to create a new data structure soon to handle this. So it's not definitive.

cardstack/src/components/AssetList/AssetList.tsx
```diff
@@ -96,6 +114,26 @@ export const AssetList = (props: AssetListProps) => {
...
+        animated: false,
+        sectionIndex: prepaidCardSectionIndex,
+        itemIndex: 0,
+      });
+
+      // Set to false so it won't update on assetsRefresh
+      setTimeout(() => {
+        setParams({ scrollToPrepaidCardsSection: false });
+      }, 2500);
+    }
```
topwebtek7, line 134: what does this 2500ms mean?

cardstack/src/components/AssetList/AssetList.tsx
```diff
@@ -64,11 +66,27 @@ interface AssetListProps
...
+    scrollToPrepaidCardsSection?: boolean;
+  };
+  key: string;
+  name: string;
+}
+
+// We need to pass this prop if the section to scrollTo is not on viewport
+const onScrollToIndexFailed = () => {
+  logger.log('onScrollToIndexFailed');
+};
```
topwebtek7, line 80: Just curious what it means.. is it for debug purpose only?

topwebtek7 (approved):
Looks good! 🚀 and just a few comments

cardstack/src/components/AssetList/AssetList.tsx
danibonilha: It's the approx. time it takes the screen to navigate and scroll to the prepaidcard section, so we wait this action to happen, and when it does we set the scrollToPrepaidCardsSection to false to be it's initial state

cardstack/src/components/AssetList/AssetList.tsx
danibonilha: We need to pass this prop if the section to scrollTo is not on viewport, it's in the react native docs section:  https://reactnative.dev/docs/sectionlist#scrolltolocation, it could be an empty function too, it doesn't really matter

PR #435, opened 2021-10-22
[CS-2188][CS-2140] - Fix sdk transactions on derived accounts

danibonilha:
### Description

If we don't pass the `from` params, the sdk assumes it's the first account of the wallet, which is not always the case, since with HD wallets we can derivate multiple account in the same wallet, so in order to have the txs sent to the right address, we needed to add this.
I'm also adding the initial tests on `gnosis-services`, because we normalize a bunch of data but didn't had any tests

To test it, you can create a fresh wallet that automatically creates an account, which will be the primary one, then manuallt create a new account this one will be derived from your wallet, just buy a ppcard, this should appear on the right account.

- [x] Completes #(CS-2188)
- [x] Completes #(CS-2140)

### Screenshots

Screenshot


cardstack/src/services/__tests__/gnosis-service.test.ts
```diff
@@ -0,0 +1,41 @@
+import { fetchGnosisSafes } from '../gnosis-service';
```
eshinkawa, line 1: 👏

eshinkawa: this function is huge

eshinkawa (approved):

PR #436, opened 2021-10-22
[CS-2226] - Fix initial layer1 purchase screen

danibonilha:
### Description

There were lots of conditionals happen in the "component" which we don't even use anymore, so i did a bit of refactoring while fixing the bug 

### Checklist

### Screenshots

Screenshot


topwebtek7 (approved):
Looks good but why don't we move AddFundsInterstitial.js to cardstack's components?

danibonilha:
> Looks good but why don't we move AddFundsInterstitial.js to cardstack's components?

It's a rainbow's component really, hopefully when we exclude the network limitations we can remove this completely.

PR #438, opened 2021-10-22
[CS-2076] - Use loadingOverlay on card purchase, handle no orderData

danibonilha:
### Description

- Add default message to useLoadingOverlay
- Add loadingOverlay to prepaidcard purchase
- Handles error, case orderData is empty
- Reduces poller time to see if tx happens faster

- [x] Completes #(CS-2076)

### Screenshots

https://user-images.githubusercontent.com/20520102/138499455-f328972f-e522-4d58-b51a-fb1adae0f889.mov


eshinkawa (approved):
🚢

PR #441, opened 2021-10-25
[CS-2221] Add converted fiat currency to the choose prepaid card screen

topwebtek7:
### Description

Added converted fiat currency to the choose prepaid card screen and updated qrcode generation to speedup

- [x] Completes #CS-2221

### Checklist

### Screenshots

Screen Shot 2021-10-25 at 8 42 03 PM


eshinkawa (approved):
Looks good

PR #442, opened 2021-10-25
[CS-1917] - Refactor welcome screen

danibonilha:
### Description

Refactored  `WelcomeScreen` to typescript  and added a new structure to create screens, where we can have a folder which contains the component in a `.tsx` file, the relative hook to the screen, the strings used in the screen and the tests directory where we can add tests for the screen hook. Let me know what you guys think.

image

- [x] Completes #(CS-1917)

### Checklist

### Screenshots

image


cardstack/src/screens/WelcomeScreen/strings.ts
```diff
@@ -0,0 +1,6 @@
+export const strings = {
```
eshinkawa, line 1: Should we start having this approach for strings? Just asking because I like it but we haven't been using that constantly

eshinkawa (approved):

cardstack/src/screens/WelcomeScreen/strings.ts
danibonilha: I think so, if everyone is comfortable with this approach, it will make things easier when we start to i18n

PR #443, opened 2021-10-26
[CS-2129][CS-2237] Show prop id on error / Buy card txn timeout handling

eshinkawa:
### Description

- [X] Completes #(2129) #(2237) 

### Checklist

### Screenshots
image

### DIFF
--- a/cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
+++ b/cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
@@ -25,6 +25,7 @@ import {
   Inventory,
   InventoryAttrs,
   makeReservation,
+  ReservationData,
   updateOrder,
 } from '@cardstack/services';
 import {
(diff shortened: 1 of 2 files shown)


topwebtek7 (approved):
LGTM 🚢

cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
```diff
@@ -107,17 +108,21 @@ export default function useBuyPrepaidCard() {
...
+    const startTime = new Date().getTime();
+
     const orderStatusPolling = setInterval(async () => {
+      const currentTime = new Date().getTime();
+
       if (wyreOrderId) {
         const orderData = await getOrder(hubURL, authToken, wyreOrderId);
 
-        if (!orderData) {
+        if (!orderData || currentTime - startTime > 60000) {
```
danibonilha, line 119: Would a setTimeout not work for this case? If this is the only way, can we give more meaningful names to what we are doing ?

eshinkawa: I added a better naming for this. I came with that implementation because it had better readability, but we can change it later on

PR #446, opened 2021-10-26
[CS-2244]: Add xDai support

danibonilha:
### Description

Adds suppot to xDai removing sokol only conditions 

~~status: on hold to be able to test wc~~

Will test wc later

- [x] Completes #(CS-2244)


kdowlin (approved):
approved to test

PR #448, opened 2021-10-27
[CS-2247]: Add loading per inventory item

danibonilha:
### Description
It's defined that we are only going to show 4 inventory options, so we can safely add the loading per item, this PR adds the loading skeleton for each inventory item and also disables the press if item is SoldOut

- [x] Completes #(CS-2247)

### Checklist

### Screenshots

Simulator Screen Recording - iPhone 11 - 2021-10-27 at 10 52 33


eshinkawa (approved):
🚢

topwebtek7 (approved):
LGTM! 🚢

PR #450, opened 2021-10-27
[CS-2239]: Handle websocket disconnections

danibonilha:
### Description

Attempt to reduce the web3 ws disconnects, we will let the error go silent and will check if it's connected when we need a web3Instance, if not, we set a new provider.
Resets the HDProvider after getting the auth token to avoid the PollingTracker crash, also added documentation  on `WebInstance` usage to prevent forgetting this in the future.

PS: Still testing some edge cases to feel comfortable merging it, if you could test too, it would be great

- [x] Completes #(CS-2239)

### DIFF
--- a/cardstack/src/models/web3-provider.ts
+++ b/cardstack/src/models/web3-provider.ts
@@ -25,8 +25,8 @@ const Web3WsProvider = {
         clientConfig: {
           keepalive: true,
           keepaliveInterval: 60000,
-          maxReceivedFrameSize: 200000000, // 2MiB
-          maxReceivedMessageSize: 100000000, // 10MiB
+          maxReceivedFrameSize: 100000000,
+          maxReceivedMessageSize: 100000000,
         },
       });
 
--- a/cardstack/src/models/web3-provider.ts
+++ b/cardstack/src/models/web3-provider.ts
@@ -5,18 +5,8 @@ import { WebsocketProvider } from 'web3-core';
 import { getNetwork } from '@rainbow-me/handlers/localstorage/globalSettings';
 import { Network } from '@rainbow-me/helpers/networkTypes';
 import logger from 'logger';
-import { Navigation } from '@rainbow-me/navigation';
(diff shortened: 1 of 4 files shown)


cardstack/src/models/web3-provider.ts
```diff
@@ -29,32 +19,29 @@ const Web3WsProvider = {
...
           auto: true,
           delay: 1000,
+          onTimeout: true,
           maxAttempts: 10,
         },
         clientConfig: {
           keepalive: true,
           keepaliveInterval: 60000,
+          maxReceivedFrameSize: 200000000, // 2MiB
+          maxReceivedMessageSize: 100000000, // 10MiB
```
eshinkawa: what would that be?

eshinkawa (approved):
i've been navigating for a couple of minutes in the app now and still haven't got the `uh oh` message

cardstack/src/models/web3-provider.ts
danibonilha: It increases the request result allowed size, in case it's larger

danibonilha:
> i've been navigating for a couple of minutes in the app now and still haven't got the `uh oh` message

For the disconnect case you won't actually see the error screen, it's silent, the goal is to test if even after the ws disconnects we are still able to use the functionalities in the app, because it should actually set a new provider once we try to use any web3 instance.

topwebtek7:
> > i've been navigating for a couple of minutes in the app now and still haven't got the `uh oh` message
> 
> For the disconnect case you won't actually see the error screen, it's silent, the goal is to test if even after the ws disconnects we are still able to use the functionalities in the app, because it should actually set a new provider once we try to use any web3 instance.

I can see this error([Error: CONNECTION ERROR: The connection got closed with the close code `1001` and the following reason string `Stream end encountered`]) in logs after some minutes of idle state but looks we're able to use the functionalities

topwebtek7 (approved):
🚢

PR #456, opened 2021-10-28
[CS-2311]: Revert to show only local time on payment confirmation

danibonilha:
### Description

- [x] Completes #(CS-2311)

### Screenshots

Screen Shot 2021-10-28 at 16 30 05

### DIFF
--- a/cardstack/src/utils/date-utils.ts
+++ b/cardstack/src/utils/date-utils.ts
@@ -1,4 +1,4 @@
-import { format, subMinutes } from 'date-fns';
+import { format } from 'date-fns';
 
 export type Units = 'days' | 'hours';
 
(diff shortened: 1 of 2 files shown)


cardstack/src/utils/date-utils.ts
```diff
@@ -116,18 +116,11 @@ export const groupAccumulations = (
...
   delimitation = '\n'
 ) => {
-  let timestamp = new Date(ts * 1000);
-
-  if (isLocalTime) {
-    // get local time based on timezone offset
-    const offsetInMinites = new Date().getTimezoneOffset();
-    timestamp = subMinutes(timestamp, offsetInMinites);
-  }
+  const timestamp = new Date(ts * 1000);
```
topwebtek7, line 123: @danibonilha just curious.. did you confirm if it shows local time? I thought it's utc time?

topwebtek7 (commented):
looks good

cardstack/src/utils/date-utils.ts
danibonilha: Yes, it was actually always showing the local time, but we did had an hardcoded string saying it was UTC before, I made a merchant payment at 4.29PM BRT and it's reflected in the screenshoot.

topwebtek7: Cool. hmm can we check this confirmation screen in other timezone device to confirm if it's actual local time for different users as well?

eshinkawa (approved):

topwebtek7 (approved):
🚀

PR #458, opened 2021-10-28
[CS-2315] - Add version to mobile app starting page

eshinkawa:
### Description

- [X] Completes #(2315)

### Checklist

### Screenshots
Screen Shot 2021-10-28 at 19 05 14

### DIFF
--- a/cardstack/src/screens/WelcomeScreen/WelcomeScreen.tsx
+++ b/cardstack/src/screens/WelcomeScreen/WelcomeScreen.tsx
@@ -8,6 +8,7 @@ import {
   Icon,
   Text,
 } from '@cardstack/components';
+import { useAppVersion } from '@rainbow-me/hooks';
 
 const layouts = {
   iconSize: 100,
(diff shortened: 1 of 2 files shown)


cardstack/src/screens/WelcomeScreen/WelcomeScreen.tsx
```diff
@@ -35,15 +42,20 @@ const WelcomeScreen = () => {
...
+        flex={0.5}
+        justifyContent="space-around"
+        marginBottom={2}
+      >
         <Button onPress={onCreateWallet}>{strings.newAccBtn}</Button>
         <Button onPress={onAddExistingWallet} variant="primaryWhite">
           {strings.existingAccBtn}
         </Button>
       </CenteredContainer>
+      <AppVersionStamp textStyleProps={bottomTextStyleProps} />
```
danibonilha: I think this should a simple text, instead of using this component, since the styles and behavior are different, and it's also missing the `Version` keyword 

image

danibonilha (changes requested):

cardstack/src/screens/WelcomeScreen/WelcomeScreen.tsx
eshinkawa: @danibonilha you're right I overcomplicated this

danibonilha (approved):

PR #459, opened 2021-10-29
 [CS-2321]: Enable send token from merchant address

danibonilha:
### Description

This PR passes the merchant safe address forward to the send flow so we can transfer tokens from a merchant safe, now that we have the ability to transfer from depot and from merchants, we should probably rename the whole flow to `SendSheetSafes`  or something, since we have some refactor to do across these flows we can do this later.

- [x] Completes #(CS-2321)

### Screenshots

Merchant safe address:
Screenshot 

Blockscout Transfer: 
Screenshot

### DIFF
--- a/cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.test.ts
+++ b/cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.test.ts
@@ -5,6 +5,7 @@ import { useSendSheetDepotScreen } from './useSendSheetDepotScreen';
 import { useAccountAssets } from '@rainbow-me/hooks';
 import { getSafesInstance } from '@cardstack/models/safes-providers';
 import { getUsdConverter } from '@cardstack/services/exchange-rate-service';
+import { reshapeSingleDepotTokenToAsset } from '@cardstack/utils';
 
 jest.mock('@cardstack/utils/device', () => ({ Device: { isAndroid: false } }));
 jest.mock('@rainbow-me/navigation/Navigation', () => ({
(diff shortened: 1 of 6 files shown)


cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.test.ts
```diff
@@ -71,14 +72,20 @@ describe('useSendSheetDepotScreen', () => {
     expect(result.current.selected).toBeUndefined();
   });
 
-  it('should return empty array if not depot is found without error', async () => {
+  it('should return the current reshapped token from params if no depot is found', async () => {
```
lukemelia:
```suggestion
  it('should return the current reshaped token from params if no depot is found', async () => {
```

eshinkawa (approved):

cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
```diff
@@ -52,17 +52,31 @@ export const useSendSheetDepotScreen = () => {
     depots: [depot],
   } = useAccountAssets();
 
-  const depotAssets = useMemo(
-    () => (depot ? reshapeDepotTokensToAssets(depot) : []),
-    [depot]
+  const reshapedAsset = useMemo(
+    () =>
+      params?.asset?.tokenAddress
```
lukemelia, line 57: Future task: understand whether we can ever have an asset without a tokenAddress. If no, simplify this.

lukemelia (approved):

PR #462, opened 2021-11-02
[CS-2228] Resolve gas calc issues with transaction in EOA, Refactor importing web3provider

topwebtek7:
### Description

- Sending tokens in EOA balances were failed sometimes cause of Web3Provider is disconnected cause of ws issue.
- Added `getWeb3Provider` in '@rainbow-me/handlers/web3' which reconnects when WS is not connected
- Resolved an issue with gas calc in EOA balances
 
- [x] Completes #CS-2228

### Checklist

### Screenshots

Nov-02-2021 05-01-22

### DIFF
--- a/src/handlers/web3.js
+++ b/src/handlers/web3.js
@@ -47,6 +47,24 @@ export const web3SetHttpProvider = async network => {
   return web3Provider.ready;
 };
 
+/**
+ * @desc returns connected web3Provider, reconnect when it's not connected
+ * @param {String} network
+ */
+
+export const getWeb3Provider = async network => {
+  let wsConnected = web3Provider.provider?.connected;
+
+  // check websocket state and reconnect if disconnected
+  while (!wsConnected) {
+    await web3SetHttpProvider(network);
+    wsConnected = web3Provider.provider?.connected;
+    logger.log('ws restarted', wsConnected, network);
+  }
+
+  return web3Provider;
+};
+
 export const sendRpcCall = async payload =>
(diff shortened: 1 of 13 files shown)


src/handlers/web3.js
```diff
@@ -47,8 +48,29 @@ export const web3SetHttpProvider = async network => {
...
 };
 
-export const sendRpcCall = async payload =>
-  web3Provider.send(payload.method, payload.params);
+/**
+ * @desc returns connected web3Provider, reconnect when it's not connected
+ * @param {String} network
+ */
+
+export const getWeb3Provider = async (network = undefined) => {
```
danibonilha: Can we change this name to reflect that it's Ether's web3 provider ? Just so we don't get confused about our web3 provider and rainbow's one, which uses ether's one

src/handlers/web3.js
```diff
@@ -47,8 +48,29 @@ export const web3SetHttpProvider = async network => {
...
+  let wsConnected = web3Provider.provider?.connected;
+
+  // check websocket state and reconnect if disconnected
+  while (!wsConnected) {
+    const currentNetwork = network || (await getNetwork());
+    await web3SetHttpProvider(currentNetwork);
```
danibonilha: This one too

src/handlers/web3.js
topwebtek7: @danibonilha Cool, changed names

danibonilha (approved):
Looks good!

PR #464, opened 2021-11-03
[CS-2386][CS-2384] - Update claim strategy

eshinkawa:
### Description

- Changes on Merchant Safe 
- Avoid re-rendering on onClaim press button which disabled the user to dismiss the share sheet

- [X] Completes #2386 #2379 #2384

### Checklist

### Screenshots

image

image

image

### DIFF
--- a/cardstack/src/components/Transactions/Merchant/MerchantClaimTransaction.tsx
+++ b/cardstack/src/components/Transactions/Merchant/MerchantClaimTransaction.tsx
@@ -11,6 +11,7 @@ import { MerchantClaimType } from '@cardstack/types';
 import { CoinIcon } from '@cardstack/components';
 import Routes from '@rainbow-me/routes';
 import { useMerchantInfoFromDID } from '@cardstack/hooks/merchant/useMerchantInfoFromDID';
+import { ClaimStatuses, getClaimProps } from '@cardstack/utils';
 
 export interface MerchantClaimTransactionProps
   extends TransactionBaseCustomizationProps {
@@ -27,11 +28,16 @@ export const MerchantClaimTransaction = ({
   const onPressTransaction = useCallback(
     (assetProps: TransactionBaseProps) =>
       navigate(Routes.EXPANDED_ASSET_SHEET_DRILL, {
-        asset: { ...assetProps },
+        asset: {
+          ...assetProps,
+          claimStatus: item.claimStatus || ClaimStatuses.DEPOSITED,
+        },
         type: 'merchantTransaction',
       }),
-    [navigate]
+    [item.claimStatus, navigate]
   );
+
+  const claimProps = getClaimProps(item.claimStatus);
 
   return (
     <TransactionBase
@@ -50,9 +56,9 @@ export const MerchantClaimTransaction = ({
           />
         ) : null
       }
-      primaryText={`- ${item.balance.display}`}
-      statusIconName="arrow-up"
-      statusText="Claimed"
+      primaryText={`${claimProps.sign} ${item.balance.display}`}
+      statusIconName={claimProps.icon}
+      statusText={claimProps.text}
       subText={item.native.display}
       transactionHash={item.transactionHash}
       onPressTransaction={onPressTransaction}
--- a/cardstack/src/utils/merchant-utils.ts
+++ b/cardstack/src/utils/merchant-utils.ts
@@ -20,6 +20,19 @@ import {
 } from '@cardstack/types';
 import { convertSpendForBalanceDisplay } from '@cardstack/utils/cardpay-utils';
 import { getNativeBalance } from '@cardstack/services';
+import { IconName } from '@cardstack/components';
+
+export const ClaimedStatus = {
+  CLAIMED_TEXT: `Claimed from \nAvailable Revenue`,
+  DEPOSITED_TEXT: `Deposited into \nAccount`,
+};
+
+export enum ClaimStatuses {
+  DEPOSITED = 'deposited',
+  CLAIMED = 'claimed',
+}
+
+export type ClaimStatusTypes = ClaimStatuses;
 
 export const merchantRevenueEventsToTransactions = (
   revenueEvents: MerchantRevenueEventFragment[]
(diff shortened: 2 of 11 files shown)


src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -58,7 +58,8 @@ interface ClaimedTransactionProps extends MerchantClaimTypeTxn {
   txRowProps: Asset;
 }
 
-const CLAIMED_STATUS = 'Claimed';
+const CLAIMED_TEXT = `Claimed from \nAvailable Revenue`;
+const DEPOSITED_TEXT = `Deposited into \nAccount`;
```
danibonilha: Isn't this the same from `MerchantClaimTransaction` ? and if we are using this to validate what we should show, this should be shared, although I think the best way would be to have an enum to match the type of the asset to match the condition

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -136,7 +136,7 @@ export default function UnclaimedRevenueExpandedState(props: {
         </Container>
       </SlackSheet>
     ),
-    [isDust, loading, onClaimAll, props.asset, revenueBalances]
+    [isDust, onClaimAll, props.asset, revenueBalances]
```
danibonilha: Is `loading` removed on purpose ? If so why, and can you add a comment ?

cardstack/src/components/Transactions/TransactionItem.tsx
```diff
@@ -18,6 +18,7 @@ import { TransactionType, TransactionTypes } from '@cardstack/types';
 interface TransactionItemProps extends TransactionBaseCustomizationProps {
   item: TransactionType;
   isFullWidth?: boolean;
+  isClaimedFromAvailableRevenue?: boolean;
```
danibonilha: The txItem shouldn't really know whether it's from available revenue, or if it's a claimed tx, we could have a custom primaryText,  statusText etc, and a default one, so the default would be `deposited` etc and the custom one we can pass in the `UnclaimedRevenueExpandedState`

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -171,7 +171,12 @@ const Activities = ({ address }: { address: string }) => {
             />
           }
           renderItem={props => (
-            <TransactionItem {...props} includeBorder isFullWidth />
+            <TransactionItem
+              {...props}
+              includeBorder
+              isClaimedFromAvailableRevenue
+              isFullWidth
+            />
```
danibonilha: Here we could pass the custom props

cardstack/src/components/Transactions/Merchant/MerchantClaimTransaction.tsx
```diff
@@ -50,9 +53,15 @@ export const MerchantClaimTransaction = ({
...
-      statusIconName="arrow-up"
-      statusText="Claimed"
+      primaryText={`${props.isClaimedFromAvailableRevenue ? '-' : '+'} ${
+        item.balance.display
+      }`}
+      statusIconName={
+        props.isClaimedFromAvailableRevenue ? 'arrow-up' : 'arrow-down'
+      }
+      statusText={
+        props.isClaimedFromAvailableRevenue ? CLAIMED_TEXT : DEPOSITED_TEXT
```
danibonilha: I guess this change means that any MerchantClaim Tx in the transactions list, will be treated as Deposit, is this the expected behavior ?

eshinkawa: Yes, that's right

cardstack/src/utils/merchant-utils.ts
```diff
@@ -20,6 +20,12 @@ import {
 } from '@cardstack/types';
 import { convertSpendForBalanceDisplay } from '@cardstack/utils/cardpay-utils';
 import { getNativeBalance } from '@cardstack/services';
+import { IconName } from '@cardstack/components';
+
+export enum ClaimedStatus {
+  CLAIMED_TEXT = `Claimed from \nAvailable Revenue`,
+  DEPOSITED_TEXT = `Deposited into \nAccount`,
+}
```
danibonilha, line 33: nit: This could be just a strings object, doesn't need to be an enum, I talked about the enum while referring to the status itself, like, 'deposited' | 'claimed'

cardstack/src/utils/merchant-utils.ts
```diff
@@ -223,3 +229,26 @@ export async function getMerchantEarnedTransactionDetails(
     netEarnedNative: convertAmountToNativeDisplay(netValue, nativeCurrency),
   };
 }
+
+export const getClaimProps = (status = 'deposited') => {
```
danibonilha: the status should be typed as the object keys(`'deposited, claimed'`), otherwise we can pass `'ble'` to this which will not behave as expected

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -222,7 +221,9 @@ export default function MerchantTransactionExpandedState(
             overflow="scroll"
             paddingHorizontal={2}
           >
-            {props.asset.statusText === CLAIMED_STATUS ? (
+            {Object.values<string>(ClaimedStatus).includes(
+              props.asset.statusText
```
danibonilha: The only to check it it's a ClaimedTransaction is from the `statusText`?  isn't it possible to use the claimStatus or so ?

cardstack/src/components/Transactions/Merchant/MerchantClaimTransaction.tsx
```diff
@@ -50,9 +51,11 @@ export const MerchantClaimTransaction = ({
...
         ) : null
       }
-      primaryText={`- ${item.balance.display}`}
-      statusIconName="arrow-up"
-      statusText="Claimed"
+      primaryText={`${getClaimProps(item.claimStatus).sign} ${
+        item.balance.display
+      }`}
+      statusIconName={getClaimProps(item.claimStatus).icon}
+      statusText={getClaimProps(item.claimStatus).text}
```
danibonilha: nit: we could probably destruct this to avoid rewriting `getClaimProps(item.claimStatus)` 3 times

src/components/expanded-state/MerchantTransactionExpandedState.tsx
eshinkawa: added a claimStatus

danibonilha (approved):

PR #465, opened 2021-11-03
[CS-2363] Implement pending screen when claiming revenue

topwebtek7:
### Description

- [x] Completes #CS-2363

### Checklist

### Screenshots

Nov-03-2021 20-09-39


eshinkawa (dismissed):

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -54,6 +61,7 @@ export default function UnclaimedRevenueExpandedState(props: {
 
   const onClaimAll = useCallback(async () => {
     setLoading(true);
```
danibonilha: Should we still track the loading state if now we are using a overlay to handle that ?  We can remove from the button too

eshinkawa (approved):

danibonilha (approved):

PR #466, opened 2021-11-03
[CS-2373]: Get balances from oracle for tokens w/o coingekoid

danibonilha:
### Description

This is kind of a nightmare because the assets structure changes on multiple places, but this was the "better" way to handle this case, and hopefully we can change this data structure to have one source of truth. Also I'm handling the case the price is zero on the chart, so it won't show any value at the top

### Screenshots

Screenshot 

Screenshot

### DIFF
--- a/cardstack/src/helpers/fallbackExplorerHelper.ts
+++ b/cardstack/src/helpers/fallbackExplorerHelper.ts
@@ -9,11 +9,13 @@ import { isNativeToken } from '@cardstack/utils/cardpay-utils';
 import {
   AssetType,
   AssetWithNativeType,
+  CurrencyConversionRates,
   DepotType,
   PrepaidCardType,
   TokenType,
 } from '@cardstack/types';
 import { Network } from '@rainbow-me/helpers/networkTypes';
+import { getNativeBalance } from '@cardstack/services';
 
 interface Prices {
   [key: string]: {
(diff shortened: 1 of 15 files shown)


eshinkawa:
looks good overall with some minor observations, did not approve because CI is failing

cardstack/src/helpers/fallbackExplorerHelper.ts
```diff
@@ -9,11 +9,13 @@ import { isNativeToken } from '@cardstack/utils/cardpay-utils';
 import {
   AssetType,
   AssetWithNativeType,
+  CurrencyConversionRates,
   DepotType,
   PrepaidCardType,
   TokenType,
 } from '@cardstack/types';
 import { Network } from '@rainbow-me/helpers/networkTypes';
+import { getNativeBalance } from '@cardstack/services';
```
topwebtek7: can we change this service name `getNativeBalance` to be more make sense like `getNativeBalanceFromOracle`, as there are similar names cardstack utils as well.

topwebtek7 (commented):
Looks good to me and left just a comment

topwebtek7 (approved):

cardstack/src/helpers/fallbackExplorerHelper.ts
topwebtek7: image

we need to update depo send sheet UI for tokens without balances as has too much space but I think we can fix it in a separate ticket

danibonilha: Agreed, and it's a different issue, since it's totally related to UI 👍

PR #467, opened 2021-11-03
fix: pass DAI as default parameter on reserving wyre order

eshinkawa:
### Description

- [ ] Completes #(Linear ticket)

### Checklist

- [ ] All UI changes have been tested on a small device

### DIFF
--- a/cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
+++ b/cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
@@ -13,7 +13,6 @@ import {
   showApplePayRequest,
 } from '@cardstack/utils/wyre-utils';
 import useAccountSettings from '@rainbow-me/hooks/useAccountSettings';
-import { getTokenMetadata } from '@rainbow-me/utils';
 import logger from 'logger';
 import { useNativeCurrencyAndConversionRates } from '@rainbow-me/redux/hooks';
 import { useAuthToken } from '@cardstack/hooks';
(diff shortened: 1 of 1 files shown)


kdowlin (approved):
lgtm

lukemelia (approved):
Consider adding a comment sharing the context of why we are adding this hard-coded string

habdelra:
@lukemelia yeah, the background is that eventually this service will be responsible for the dest currency used in the wyre order request https://linear.app/cardstack/issue/CS-2194/introduce-service-for-prepaid-card-wyre-pricing

cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
```diff
@@ -253,7 +253,7 @@ export default function useBuyPrepaidCard() {
   const onPurchase = useCallback(
     async ({ address, value, depositAddress, sourceCurrency }) => {
       const metadata = getTokenMetadata(address);
```
habdelra: like I was mentioning in discord--this is problematic because this is giving you information about layer 2 tokens, but wyre only understands layer 1 tokens. I would honestly not even use this at all.

cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
```diff
@@ -253,7 +253,7 @@ export default function useBuyPrepaidCard() {
...
-      const currency = metadata?.symbol || '';
+      const currency = metadata?.symbol || 'DAI';
```
habdelra: better if this is just:
```js
const currency = 'DAI'; // TODO this will come from a service in #CS-2194
```

PR #469, opened 2021-11-04
[CS-2403]: Get price from oracle for CPXD tokens

danibonilha:
### Description

We should always get the prices of CPXD token from oracle, so while calling `fetchHistoricalPrice` we can check if it's a CPXD token, if so we call the oracle. In order to do that I noticed we had a couple of deps on the way we store currencyConversionRates and we just drill down this param to `getNativeBalanceFromOracle` that's why I created the `cachedCurrencyConversionRates`, scooped in the `exchange-rate` so we can leverage this info without passing down on multiple components and services, also this provided the opportunity to move this fetch from initialilizeWallet to fallbackExplorer. In `fetchHistoricalPrice` I don't understand why the token prices were being stored in the local storage, since it can be something volatile, and change everyday, so I removed the storage cache, because it would fetch just upon installation and then never again if you don't cleanup the data.

PS: there's a delay after changing the currency and updating the txs I'll be supporting this on the next PR 

- [x] Completes #(CS-2403)

### Screenshots

image


cardstack/src/services/exchange-rate-service.ts
```diff
@@ -1,15 +1,59 @@
...
+// Currency conversions
+const CurrencyConsts = {
+  fixerApi: 'https://data.fixer.io/api',
+  usdToSpendRate: 100,
+  currencySymbols: Object.keys(nativeCurrencies),
+};
+
+let cachedCurrencyConversionRates: CurrencyConversionRates | null = null;
+
+export const getCurrencyConversionsRates = async () => {
```
eshinkawa, line 20: could we add a test for this, just for comprehension purporses?

eshinkawa (approved):
LGTM with an observation

PR #471, opened 2021-11-04
[CS-2377] Update transaction processing behaviors, refetch assets after success

topwebtek7:
### Description

- Added overlay loading for sending transactions from depot and from EOA
- Moved SendSheetEOA screen route to cardstack routes to show send screen below of overlay loading
- Redirected to wallet screen and refetched assets to reflect balance changes(it redirected to profile screen before but weirdly transactions are never updated even after refetched until assets are refetched.. need to investigate why)

- [x] Completes #CS-2377

### Checklist

### Screenshots

Screenshot
Screenshot

### DIFF
--- a/cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
+++ b/cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
@@ -28,7 +28,7 @@ import { DepotAsset, TokenType } from '@cardstack/types';
 import { getUsdConverter } from '@cardstack/services/exchange-rate-service';
 import HDProvider from '@cardstack/models/hd-provider';
 import { useWorker } from '@cardstack/utils/hooks-utilities';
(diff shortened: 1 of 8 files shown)


eshinkawa (approved):
🚢

cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
```diff
@@ -296,30 +297,43 @@ export const useSendSheetDepotScreen = () => {
...
         Alert({
           message: errorMessage,
-          title: 'An error ocurred',
+          title:
+            'An error has occurred, could not send. If this persists, please contact support@cardstack.com',
         });
+      } else {
+        Alert(
+          'An error has occurred, could not send. If this persists, please contact support@cardstack.com'
+        );
```
danibonilha: really nit: but we could probably has this as a const and reuse it

danibonilha (approved):

PR #472, opened 2021-11-04
[CS-2385][CS-2238][CS-2419] - Token symbol layout 

eshinkawa:
### Description

- text changes / layout fix

- [X] Completes #CS-2385 #CS-2238 #CS-2419

### Checklist

### Screenshots

image
image
image

### DIFF
--- a/src/components/send/SendAssetFormField.js
+++ b/src/components/send/SendAssetFormField.js
@@ -8,7 +8,7 @@ export default function SendAssetFormField({
   autoFocus,
   format,
   label,
-  labelMaxLength = 6,
+  labelMaxLength = 9,
   mask,
   onChange,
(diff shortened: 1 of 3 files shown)


src/components/send/SendAssetFormField.js
```diff
@@ -8,7 +8,7 @@ export default function SendAssetFormField({
   autoFocus,
   format,
   label,
-  labelMaxLength = 6,
+  labelMaxLength = 9,
```
danibonilha, line 11: I'm not sure this is the right behavior, should the fontSize be smaller ?

danibonilha (approved):
Talked with @kdowlin, it seems the requirements change and this approach is fine! so 🚢

PR #476, opened 2021-11-08
Be resilient to wallet-connect peerMeta with no icons

lukemelia:
A peerMeta with no icons can occur when connecting to the dapp running on localhost in development mode


danibonilha (approved):
LGTM, the only suggestion I'd give is to add some description on the PR so we know the context if we ever need to return to this issue, something like, why it was happening or whatever =)

PR #477, opened 2021-11-08
Convert useWalletConnectConnections hook and sortList helper to TypeScript

lukemelia:
- [x] First merge #476

Since we established through bug fix in #476 that the useWalletConnections hook is in use, this PR moves it to the cardstack/ source tree and converts it to TypeScript, along with the sortList helper that it depends on.


eshinkawa (approved):

topwebtek7 (approved):
LGTM! 🚀

PR #478, opened 2021-11-08
[CS-2450] - Fix no fiat balance crashes

danibonilha:
### Description

Sometimes we may get a Timeout or in any other case may not be able to  retrieve the prices from SDK, so the fiat balance will be zero, if we don't get the price we can't calculate the balance amount. This PR checks wether we have prices and balance amount in order to hide or show the information.

PS: Maybe the layout on send flow can be improve to keep the input field fixed on top, we can discuss it later, just didn't think it was worth right now.

- [x] Completes #(CS-2450)

### Checklist

### Screenshots

Simulator Screen Recording - iPhone 11 - 2021-11-08 at 16 35 59

### DIFF
--- a/src/components/coin-row/SendCoinRow.js
+++ b/src/components/coin-row/SendCoinRow.js
@@ -14,14 +14,18 @@ import { CenteredContainer } from '@cardstack/components';
 const isTinyPhone = deviceUtils.dimensions.height <= 568;
 const selectedHeight = isTinyPhone ? 62 : 78;
 
-const BottomRow = ({ balance, native, nativeCurrencySymbol }) => {
+const BottomRow = ({ balance, native }) => {
   const { colors } = useTheme();
-  const fiatValue =
-    get(native, 'balance.display') || `${nativeCurrencySymbol}0.00`;
+
+  const hasNativeBalance = !!parseFloat(native?.balance?.amount);
+
+  const fiatValue = hasNativeBalance
+    ? `≈ ${get(native, 'balance.display')}`
+    : '';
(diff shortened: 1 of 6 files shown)


eshinkawa (approved):
🚢

src/components/coin-row/SendCoinRow.js
```diff
@@ -14,14 +14,18 @@ import { CenteredContainer } from '@cardstack/components';
...
-const BottomRow = ({ balance, native, nativeCurrencySymbol }) => {
+const BottomRow = ({ balance, native }) => {
   const { colors } = useTheme();
-  const fiatValue =
-    get(native, 'balance.display') || `${nativeCurrencySymbol}0.00`;
+
+  const hasNativeBalance = !!parseFloat(native?.balance?.amount);
+
+  const fiatValue = hasNativeBalance
+    ? `≈ ${get(native, 'balance.display')}`
```
lukemelia, line 23: Is it safe to hard-code `$` here? Should this be a currency symbol based on the user's device setting?

danibonilha: The balance display already handles the currency symbol based on the user's device setting

danibonilha: amount is just, well the amount, and display is the formatted display value e.g. `$ 1.00 USD` 
example:
image

lukemelia: Whoops, misintepreted the `${...}` syntax 🤦

PR #479, opened 2021-11-08
Reorganize tests for consistency

lukemelia:
- Each source directory with tests has a `__tests__` subdirectory containing test files for the code in that directory

Open questions: 

- [x] There is one remaining tests in the /cardstack/tests/ directory: TransactionCoinRow.test.tsx. This contains tests for a component that is in the "rainbow" source tree. What should we do with it?
- [x] The /cardstack/tests/ directory also contains jest-setup.js and test-utils/*. Where should these files live?


topwebtek7 (approved):
Very good refactoring! 👍🏼 
IMO, we'll need to do refactoring tests for rainbow components later, and can move jest-setup.js and test-utils/* to cardstack/src

danibonilha (commented):
Thanks! As far as the `TransactionCoinRow.test`, it's a bit misleading because it's not actually testing the `TransactionCoinRow`, it's testing the `ERC20Transaction` component, which is a cardstack one, so we can rename this file and move it to the components one.
For the test-utils and jest setup, I guess we could have a test-utils folder inside src where they can live

lukemelia:
Thanks for the feedback! Updated.

danibonilha (approved):
LGTM

PR #483, opened 2021-11-09
[CS-2400] Add discord link in settings panel

topwebtek7:
### Description

- [x] Completes #CS-2400

### Checklist

### Screenshots

image


lukemelia (approved):
Looks good. +1 for eliminating a rainbow reference too.

cardstack/src/constants.ts
```diff
@@ -3,3 +3,11 @@ export const TRANSACTION_PAGE_SIZE = 100; // Temp increase the page size
...
 export const SEND_TRANSACTION_ERROR_MESSAGE =
   'An error has occurred, could not send. If this persists, please contact support@cardstack.com';
+export const SettingsExternalURLs = {
+  cardstackHomepage: 'https://cardstack.com/',
+  review:
+    'itms-apps://itunes.apple.com/us/app/appName/id1457119021?mt=8&action=write-review',
+  twitterDeepLink: 'twitter://user?screen_name=cardstack',
+  twitterWebUrl: 'https://twitter.com/cardstack',
+  discordInviteLink: 'https://discord.gg/NYtJY5Fx',
+};
```
danibonilha, line 15: I think we are starting to mix contexts by extracting scoped things into this constants file, this is like a strings object which is only used in the settings section, so why does it need to live in the global constants files ?

danibonilha (approved):
looks good, just one small comment

cardstack/src/constants.ts
topwebtek7: it can be used in our files as well and also easy to find & update later.

PR #484, opened 2021-11-09
[CS-2449]: Fix fiat conversion on send from depot

danibonilha:
### Description
This PR adds the conversion to native currency on gas estimation and nativeAmount input on the send flow for safes

- [x] Completes #(CS-2449)

### Screenshots

Simulator Screen Recording - iPhone 11 - 2021-11-09 at 10 21 49

### DIFF
--- a/cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
+++ b/cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
@@ -4,7 +4,7 @@ import {
 } from '@cardstack/cardpay-sdk';
 import { useRoute } from '@react-navigation/native';
 import { captureException } from '@sentry/react-native';
-import { get, isEmpty, isString } from 'lodash';
+import { isEmpty, isString } from 'lodash';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import Web3 from 'web3';
 import BN from 'bn.js';
@@ -30,6 +30,7 @@ import { getUsdConverter } from '@cardstack/services/exchange-rate-service';
 import HDProvider from '@cardstack/models/hd-provider';
 import { useWorker } from '@cardstack/utils/hooks-utilities';
 import { MainRoutes, useLoadingOverlay } from '@cardstack/navigation';
+import { useNativeCurrencyAndConversionRates } from '@rainbow-me/redux/hooks';
 
 interface RouteType {
   params: { asset: TokenType; safeAddress?: string };
@@ -91,7 +92,21 @@ export const useSendSheetDepotScreen = () => {
   const isValidAddress = useSendAddressValidation(recipient);
   const [isAuthorizing, setIsAuthorizing] = useState(false);
 
-  const { accountAddress, nativeCurrency, network } = useAccountSettings();
+  const { accountAddress, network } = useAccountSettings();
+
+  const [
+    nativeCurrency,
+    currencyConversionRates,
+  ] = useNativeCurrencyAndConversionRates();
+
+  const getNativeCurrencyAmount = useCallback(
+    (amount: string) => {
+      const usdConvertedAmount = usdConverter.current?.(amount) || 0;
+
+      return currencyConversionRates[nativeCurrency] * usdConvertedAmount;
+    },
+    [currencyConversionRates, nativeCurrency]
+  );
 
   // Gas Estimates
   const [gasEstimatedFee, setGasEstimatedFee] = useState(0);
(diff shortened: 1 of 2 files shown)


cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
```diff
@@ -91,7 +93,24 @@ export const useSendSheetDepotScreen = () => {
...
+    (amount: string) => {
+      const isUSD = nativeCurrency === NativeCurrency.USD;
+      const usdConvertedAmount = usdConverter.current?.(amount) || 0;
+
+      return isUSD
+        ? usdConvertedAmount
+        : currencyConversionRates[nativeCurrency] * usdConvertedAmount;
+    },
+    [currencyConversionRates, nativeCurrency]
+  );
```
topwebtek7, line 109: nit: do we really need to check if it's USD currency? `currencyConversionRates[NativeCurrency.USD]` is always 1, so I think don't need even `getNativeCurrencyAmount` function..

suggestion: 
```
const nativyCurrencyGasFee = (usdConverter.current?.(amount) || 0) * currencyConversionRates[nativeCurrency];
```
basically I'm not convinced enough why use `usdConverter` useRef here

cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
```diff
@@ -110,9 +129,9 @@ export const useSendSheetDepotScreen = () => {
           amountWei
         )) || '0';
 
-      const gasFeeInUsd = usdConverter.current?.(gasEstimate) || 0;
+      const nativyCurrencyGasFee = getNativeCurrencyAmount(gasEstimate) || 0;
```
topwebtek7: ^^
`const nativyCurrencyGasFee = (usdConverter.current?.(amount) || 0) * currencyConversionRates[nativeCurrency];
`

cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
```diff
@@ -196,10 +216,15 @@ export const useSendSheetDepotScreen = () => {
...
+
+      const nativyCurrencyAssetAmount = hasAmount
+        ? getNativeCurrencyAmount(Web3.utils.toWei(assetAmount))
+        : 0;
 
       try {
-        const nativeAmount = assetAmount.length
-          ? usdConverter.current?.(Web3.utils.toWei(assetAmount)).toString()
+        const nativeAmount = hasAmount
+          ? nativyCurrencyAssetAmount.toString()
```
eshinkawa: is it `natify`?

cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
```diff
@@ -196,10 +216,15 @@ export const useSendSheetDepotScreen = () => {
   const updateAssetAmount = useCallback(
     async newAssetAmount => {
       const assetAmount: string = newAssetAmount.replace(/[^0-9.]/g, '');
+      const hasAmount = !!assetAmount.length;
+
+      const nativyCurrencyAssetAmount = hasAmount
```
topwebtek7: maybe typo?

topwebtek7 (approved):
looks good. just a couple comments

eshinkawa (approved):

cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
danibonilha: The usdConverter is a converter based on the token which we fetch from the sdk, we store in the ref because it doesn't need to trigger a re-render, `useRef` works as instance variables, also it's best to get the converter once, than hit the sdk every time to get the token value, on gasEstimate, nativeInput and tokenInput.

About checking USD, you're right we probably don't need it, but we do need to keep the `getNativeCurrencyAmount` function because it makes the code cleaner and consistent, it doesn't make much sense to keep repeating `(usdConverter.current?.(amount) || 0) * currencyConversionRates[nativeCurrency]` if we can have a function for that.

topwebtek7: ah makes sense now. thanks!

btw I think ```const nativeCurrencyGasFee = (usdConverter.current?.(amount) || 0) * currencyConversionRates[nativeCurrency];``` is easy enough to understand.. just wanted to suggest to have less function/variables definitions as much as we can

PR #485, opened 2021-11-09
[CS-2424]: Chop off cpxd token to get hist price

danibonilha:
### Description
We don't track historical prices in the sdk, but since cpxd token are based on mainnet we can remove `.CPXD` and use the token symbol to get the historical price. Now we can see that for the same amount of CARD for example `0.10` we have a different conversion amount. 

-  [x] Completes #(CS-2424)

### Screenshots

Screenshot

### DIFF
--- a/cardstack/src/services/historical-pricing-service.ts
+++ b/cardstack/src/services/historical-pricing-service.ts
@@ -1,8 +1,6 @@
 import { CRYPTOCOMPARE_API_KEY } from 'react-native-dotenv';
-import { getNativeBalanceFromOracle } from './exchange-rate-service';
 import logger from 'logger';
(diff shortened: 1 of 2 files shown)


topwebtek7 (approved):
🚢

lukemelia (commented):
For tokens that don't have a .CPXD suffix, it looks like we will send them as-is to the cryptocompare API. Is that our intention, or are there conditions they need to meet to be suitable for use with cryptocompare?

danibonilha:
> For tokens that don't have a .CPXD suffix, it looks like we will send them as-is to the cryptocompare API. Is that our intention, or are there conditions they need to meet to be suitable for use with cryptocompare?

Yes, we are keeping the previous behavior for non-cpxd, which is to send as is without any conditions

PR #486, opened 2021-11-09
[CS-2422] - Update assets after claim all

eshinkawa:
### Description

Update merchant safe after claiming all.

- [X] Completes #2422

### Checklist

### Screenshots

Nov-09-2021 12-27-18

### DIFF
--- a/cardstack/src/screens/MerchantScreen.tsx
+++ b/cardstack/src/screens/MerchantScreen.tsx
@@ -33,7 +33,10 @@ import {
 } from '@cardstack/utils';
 import { ChartPath } from '@rainbow-me/animated-charts';
 import { useNavigation } from '@rainbow-me/navigation';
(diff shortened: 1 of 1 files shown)


lukemelia:
I would love to understand this part of the code base better. It is hard to understand what architecture we have that makes it difficult to refresh a balance after a claim completes.

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -115,6 +118,7 @@ export default function UnclaimedRevenueExpandedState(props: {
   }, [
     accountAddress,
     dismissLoadingOverlay,
+    goBack,
     merchantSafe.address,
     network,
     revenueBalances,
```
danibonilha: Well, it seems the unclaimedRevenue it's using the merchantSafes from redux, if we use it also in the MerchantScreen we should be able to update it

topwebtek7 (approved):
LGTM1

PR #489, opened 2021-11-10
[CS-2442] - Update wallet app to use wyre price hub service

eshinkawa:
### Description

So we had `destCurrency` hardcoded as prop. Now with the Wyre price service we get this info from it, and also the `source-currency-price` which is used to convert currency

- [X] Completes #2442

### Checklist

### DIFF
--- a/cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
+++ b/cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
@@ -22,17 +22,15 @@ import {
   getHubUrl,
   getInventories,
   getOrder,
+  getWyrePrice,
   Inventory,
   InventoryAttrs,
   makeReservation,
   ReservationData,
   updateOrder,
+  WyrePriceData,
 } from '@cardstack/services';
-import {
-  fetchCardCustomizationFromDID,
-  getNativeBalanceFromSpend,
-  useWorker,
-} from '@cardstack/utils';
+import { fetchCardCustomizationFromDID, useWorker } from '@cardstack/utils';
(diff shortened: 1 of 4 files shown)


cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
```diff
@@ -389,7 +413,8 @@ export default function useBuyPrepaidCard() {
       wyreOrderIdData = await onPurchase({
         value: amount.toString(),
         depositAddress: custodialWalletData?.attributes['deposit-address'],
-        sourceCurrency: nativeCurrency,
+        sourceCurrency: nativeCurrency || card?.['source-currency'],
+        destCurrency: card?.['dest-currency'],
```
danibonilha: Should it default to DAI ?

danibonilha (commented):
Did a purchase went thru after this change ?  on xDai  or was it tested on sokol ? There's no .gif, that's why I'm asking

cardstack/src/hooks/prepaid-card/useBuyPrepaidCard.ts
eshinkawa: hassan: `for now that is ok. but when/if we are able to buy prepaid cards in the future backed by different tokens, it should just throw an error`

eshinkawa:
> Did a purchase went thru after this change ? on xDai or was it tested on sokol ? There's no .gif, that's why I'm asking

Tested on Sokol

https://user-images.githubusercontent.com/8547776/141178508-933eac22-b26f-4de5-ab3f-eed9c3da5af7.mov

danibonilha (approved):

PR #490, opened 2021-11-10
[Fix]: Handle test warnings 

danibonilha:
### Description

There were multiple act errors on the test, because sometimes while calling a async function and not waiting properly the component was updating after the test had run, so instead of waiting for the next update, we actually need to wait for the promise to resolve, for more information about this type of issue, there's a great Kent's doc [here](https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning). There was also some annoying  console.warn happening because an unmocked package, so that's fixed too. Now we can read the test results better, without garbage.


eshinkawa (approved):
Great! Ran locally and it's way  better to read logs 🚢

topwebtek7 (approved):
lgtm 👍🏼

PR #491, opened 2021-11-10
[CS-2476]: Update merchant copy to business

danibonilha:
### Description

This removes the merchant text from Scan QR code and I used the opportunity to look for other merchant references and update it alongside


cardstack/src/components/TransactionConfirmationSheet/displays/ClaimRevenueDisplay.tsx
```diff
@@ -87,12 +87,12 @@ const ToSection = ({ merchantSafe }: { merchantSafe: string }) => {
...
       <Container paddingHorizontal={3} marginTop={4} flexDirection="column">
         <Text size="xxs" color="blueText" marginLeft={9}>
-          MERCHANT SAFE
+          BUSINESS SAFE
         </Text>
         <Container flexDirection="row" alignItems="center">
           <Icon name="user" />
           <Text size="small" weight="extraBold" style={{ marginLeft: 6 }}>
-            Merchant Name
+            Business
```
danibonilha, line 95: This one I think we will need to use the merchant info in the future, but I guess we don't use this flow yet, so let's wait

eshinkawa (approved):
looks good

PR #492, opened 2021-11-11
[CS-2482][CS-2506] - Earnings history sticky sheet / Blockscout Link

eshinkawa:
### Description

- Share sheet with sticky experience. We set height to 0.85 as initial HEIGHT for both Transactions / Lifetime Earnings, that makes it easier for the user to snap it to the top, specially in larger devices (Kiel tested it) 

- User setting with link to Blockscout wallet address

### Checklist

### Screenshots

image

https://user-images.githubusercontent.com/8547776/141212994-3915ea10-b395-4d46-a92a-8b9b8f3179ad.mp4

### DIFF
--- a/src/components/expanded-state/LifetimeEarningsExpandedState.tsx
+++ b/src/components/expanded-state/LifetimeEarningsExpandedState.tsx
@@ -18,15 +18,15 @@ import {
 } from '@cardstack/hooks';
 import { palette } from '@cardstack/theme';
 import { MerchantSafeType } from '@cardstack/types';
-import { convertSpendForBalanceDisplay } from '@cardstack/utils';
+import { convertSpendForBalanceDisplay, screenHeight } from '@cardstack/utils';
 import { sectionStyle } from '@cardstack/utils/layouts';
 import { ChartPath } from '@rainbow-me/animated-charts';
 import { useDimensions } from '@rainbow-me/hooks';
 import { useNavigation } from '@rainbow-me/navigation';
 import { useRainbowSelector } from '@rainbow-me/redux/hooks';
 
-const CHART_HEIGHT = 200;
-const HEIGHT = CHART_HEIGHT + 400;
+const CHART_HEIGHT = screenHeight * 0.25;
+const HEIGHT = screenHeight * 0.85;
 
 const useMerchantSafe = (address: string) => {
   const merchantSafes = useRainbowSelector(state => state.data.merchantSafes);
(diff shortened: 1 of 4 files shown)


src/components/expanded-state/LifetimeEarningsExpandedState.tsx
```diff
@@ -18,15 +18,15 @@ import {
...
 import { sectionStyle } from '@cardstack/utils/layouts';
 import { ChartPath } from '@rainbow-me/animated-charts';
 import { useDimensions } from '@rainbow-me/hooks';
 import { useNavigation } from '@rainbow-me/navigation';
 import { useRainbowSelector } from '@rainbow-me/redux/hooks';
 
-const CHART_HEIGHT = 200;
-const HEIGHT = CHART_HEIGHT + 400;
+const CHART_HEIGHT = screenHeight * 0.25;
+const HEIGHT = screenHeight * 0.85;
```
danibonilha, line 29: if this is a pattern we should be able to reuse we could either extract this to "global"const, or if we need to call the setOptions on every expandedState, we could write a hook with this 
```ts
  useEffect(() => {
    setOptions({
      longFormHeight: HEIGHT,
    });
  }, [setOptions]);
```

But it's just something nice to thing about, hopefully we will get rid of the ExpandedState anyways, so wouldn't worry too much about it

src/components/settings-menu/SettingsSection.js
```diff
@@ -175,6 +181,14 @@ export default function SettingsSection({
             {networkInfo?.[network]?.name}
           </ListItemArrowGroup>
         </ListItem>
+        <ListItem
+          icon={<Icon color="settingsTeal" name="eye" />}
+          label="View on Blockscout"
+          onPress={onPressBlockscout}
+          testID="network-section"
+        >
+          <ListItemArrowGroup />
```
danibonilha: Not sure we need the arrow here, from UX perspective it usually means there's a flow starting inside the app from this button, which is not the case for this one , the buttons that currently triggers an action outside the app don't have any arrows.

danibonilha (approved):
Looks good, just a couple of small comments

src/components/settings-menu/SettingsSection.js
eshinkawa: Yeah, that was @kdowlin opinion, Kiel WDYT?

kdowlin: @eshinkawa Dani makes a good point we can remove.

PR #493, opened 2021-11-11
[CS-2491] Fix reading qrcode currency when use camera outside app

topwebtek7:
### Description

Fixed payment currency overrides happened when opens cardwallet app after read qrcode using camera app

- [x] Completes #CS-2491

### Checklist

### Screenshots

Nov-11-2021 14-06-22


danibonilha (approved):
🚀

PR #494, opened 2021-11-11
[CS-2475]: Initial fixes for Android build

danibonilha:
### Description

Makes android at least buildable and without crashes, it's still slow and with multiple UI/UX issues, but at least it's alive

### Screenshots

Screenshot 

Screenshot

### DIFF
--- a/src/screens/QRScannerScreen.js
+++ b/src/screens/QRScannerScreen.js
@@ -2,7 +2,6 @@ import { useIsFocused } from '@react-navigation/native';
 import React, { useCallback, useEffect, useState } from 'react';
 import { FlatList } from 'react-native';
 import styled from 'styled-components';
-import { DiscoverSheet } from '../components/discover-sheet';
 import { BackButton, Header, HeaderHeight } from '../components/header';
 
 import {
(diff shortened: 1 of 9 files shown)


src/screens/QRScannerScreen.js
```diff
@@ -87,47 +82,41 @@ const QRScannerScreen = () => {
         </CameraDimmer>
 
         <Container bottom={0} position="absolute" width="100%">
-          {discoverSheetAvailable ? (
-            android ? (
-              <DiscoverSheet />
-            ) : null
```
topwebtek7, line 93: @danibonilha just for more context, looks like rainbow used DiscoverSheet for android.. does Sheet handles for android as well? 
anyway will give a try android build on my side shortly

eshinkawa: trying to run on my device as well

danibonilha: DiscoverSheet is a feature that shows assets in the QR code page, searchBar etc, the one on iOS had already been removed, it's an experimental  feature which was never used, this is the sheet from rainbow app:

image

eshinkawa: @danibonilha was able to run after adding this to `build.gradle`
https://stackoverflow.com/a/69050529/2716525

eshinkawa:
@danibonilha was only able to run after adding this to `build.gradle`
https://stackoverflow.com/a/69050529/2716525

eshinkawa (approved):

danibonilha:
> https://stackoverflow.com/a/69050529/2716525

I did not run into this issue, did you run `yarn setup` before trying to build ? Or it could a m1 thing 🤔 , I'd recommend to clean `node_modules`, and all build folders, then `yarn setup && yarn android`

src/screens/QRScannerScreen.js
topwebtek7: good job. I was able to run android

PR #495, opened 2021-11-11
[CS-2480, CS-2516] Implement redirect to transactions screen after payment confirmation, Remove link in QR screen

topwebtek7:
### Description

- [x] Completes #CS-2480
- [x] Completes #CS-2516

### Checklist

### Screenshots

Nov-11-2021 22-27-02

### DIFF
--- a/cardstack/src/components/TransactionList/TransactionList.tsx
+++ b/cardstack/src/components/TransactionList/TransactionList.tsx
@@ -59,11 +59,7 @@ export const TransactionList = memo(({ Header }: TransactionListProps) => {
     refetch && refetch();
   }, [refetch]);
 
-  useFocusEffect(
-    useCallback(() => {
-      onRefresh();
-    }, [onRefresh])
-  );
+  useFocusEffect(onRefresh);
 
   if (isLoadingTransactions) {
     return (
--- a/cardstack/src/components/TransactionList/TransactionList.tsx
+++ b/cardstack/src/components/TransactionList/TransactionList.tsx
@@ -1,4 +1,5 @@
-import React, { memo, useCallback, useEffect } from 'react';
+import React, { memo, useCallback } from 'react';
+import { useFocusEffect } from '@react-navigation/native';
 import {
   RefreshControl,
   SectionList,
@@ -20,7 +21,6 @@ import {
 interface TransactionListProps {
   Header: JSX.Element;
   accountAddress: string;
-  isFocused: boolean;
 }
 
 const styles = StyleSheet.create({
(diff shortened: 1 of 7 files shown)


cardstack/src/hooks/transactions/use-transaction-sections.tsx
```diff
@@ -70,9 +70,7 @@ export const useTransactionSections = ({
   const isPagination =
     prevLastestTx === currentLastestTx && currentTxLength > prevTxLength;
 
-  const isNewtx =
-    prevLastestTx !== currentLastestTx && prevTxLength !== currentTxLength;
-
```
topwebtek7, line 75: Updated as was a bug when transactions were as much as pagination limit

cardstack/src/screens/PayMerchant/usePayMerchant.ts
```diff
@@ -127,12 +125,9 @@ export const usePayMerchant = () => {
...
       const timestamp = await getBlockTimestamp(receipt.blockNumber);
 
-      // Dismisses PayMerchantModal
-      dismissLoadingOverlay();
-
-      if (canGoBack()) {
-        goBack();
-      }
+      // Navigate to Transaction screen. somtimes useIsFocused does not work, so added this hack
+      setParams({ focused: false });
```
danibonilha: This is a little odd, since [setParams](https://reactnavigation.org/docs/navigation-prop/#setparams), sets the param on the current route, which is `PayMerchant` at the time we call this, we probably need to investigate is it's being properly set. Would using [useFocusEffect](https://reactnavigation.org/docs/use-focus-effect/) on TransactionList maybe help ?

topwebtek7: tried with useFocusEffect but still happens.. sometime when you try to go to transactions screen from wallet assets screen, useFocusEffect or useIsFocused is not called.. maybe will be fixed after react-navigation upgrade or navigation re-design. anyway will update with useFocusEffect for now

cardstack/src/components/TransactionList/TransactionList.tsx
```diff
@@ -45,69 +45,67 @@ const renderSectionHeader = ({
...
-          {Header}
-          <TransactionListLoading />
-        </ScrollView>
-      );
-    }
+  useFocusEffect(
+    useCallback(() => {
+      onRefresh();
+    }, [onRefresh])
+  );
```
danibonilha: Do we need to wrap on a useCallback if `onRefresh` is already a useCallback function ?

danibonilha (approved):

PR #497, opened 2021-11-11
[CS-2409]: Fix external wc signing request

danibonilha:
### Description

The `params[1]` might not exist, it seems on external signing requests the data comes from the first item in the array, rather than the second one, example: 

External: 
```
    "params": [
        {
            "gas": "0x14313",
            "from": "0xd6f3f565e207a4e4b1b2e51f1a86d26d3dbf5811",
            "to": "0x26f2319fbb44772e0ed58fb7c99cf8da59e2b5be",
            "data": "0x095ea7b30000000000000000000000001b02da8cb0d097eb8d57a175b88c7d8b47997506ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
            "gasLimit": "0x14313"
        }
    ]
```

CardPay: 
```
    "params": [
        "0xd6F3F565E207A4e4B1b2E51F1A86d26D3DBf5811",
        {
            "types": {
                "EIP712Domain": [...]
             }...
       }
       ...
    ]
```

They are handled differently already, and there's a place that checks if message exists or not, that's why I'm adding an empty state, so it won't blowup all the destructing. I noticed there was a hook that used the same `params[1]`  so I combined that in the payloadParams one. 

There's an issue we can see in the message request, it's showing `gasPrice: NaN`,  I investigated a bit, was hoping to fix in this PR but its a little complex, so we'll add a new ticket to handle that separately, also tapping on the black price button crashes the app, anyways, all different issue.

### Screenshots

Screenshot


eshinkawa (approved):
🚢

PR #498, opened 2021-11-11
[CS-2509] - Apple Pay button should not be clickable if no card is selected

eshinkawa:
### Description

Disabling Apple Button if no card is selected or Inventory data is empty

- [X] Completes #2509

### Checklist

### Screenshots

Nov-15-2021 11-42-00

### DIFF
--- a/cardstack/src/screens/BuyPrepaidCard/BuyPrepaidCard.tsx
+++ b/cardstack/src/screens/BuyPrepaidCard/BuyPrepaidCard.tsx
@@ -68,6 +68,11 @@ const BuyPrepaidCard = () => {
     [navigate]
   );
 
+  const isDisabled =
+    card?.quantity === 0 ||
+    inventoryData?.length === 0 ||
+    inventoryData?.filter(item => item?.isSelected).length === 0;
+
   return (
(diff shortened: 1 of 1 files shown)


cardstack/src/screens/BuyPrepaidCard/BuyPrepaidCard.tsx
```diff
@@ -90,7 +95,7 @@ const BuyPrepaidCard = () => {
             justifyContent="center"
           >
             <ApplePayButton
-              disabled={card?.quantity === 0}
+              disabled={isDisabled}
```
danibonilha: Maybe we should hide the button ? Or handled the disabled state in the UI, it looks like a bug when the button seems tapeable but it doesn't trigger any action, like giving a warning that the user should select a product, so in this case maybe hiding  it's best, and in case there's no inventory, we could have a text, saying it's not available, try again later in the place of inventories, @kdowlin can come up with a better text

eshinkawa: agreed

topwebtek7 (approved):
LGTM! 🚢

PR #499, opened 2021-11-12
[CS-2396, CS-2470] Upgrade walletconnect and add patches, fix qrcode reader blackout

topwebtek7:
### Description

- Upgraded WC from 1.6.1 to 1.6.6, no big changes but 1.6.6 has a patch fix for issues that we experienced when bridge tokens. Tested but sorry, couldn't record this fix on cardwallet.
- Also refactored QRCode scanner to fix qrcode reader blackout issues. Note: we need to upgrade react-native-camera library to v4 to improve capability, I think and can be done in a separate ticket.

- [x] Completes #CS-2396
- [x] Completes #CS-2470

### Checklist

### Screenshots

Nov-13-2021 02-08-30


ios/Podfile.lock
```diff
@@ -1014,4 +1014,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 3436c19c0f20f9a28c7fac2bed2e8606e6195857
 
-COCOAPODS: 1.10.1
+COCOAPODS: 1.11.2
```
danibonilha, line 1017: This is different from the version set on gemfile

src/components/qrcode-scanner/QRCodeScanner.js
```diff
@@ -74,8 +89,11 @@ export default function QRCodeScanner({
           top={contentPositionTop}
           width="100%"
         >
-          {showErrorMessage && <ErrorText error="Error mounting camera" />}
-          {showCrosshair && <QRCodeScannerCrosshair />}
+          {error ? (
+            <ErrorText error="Error mounting camera" />
+          ) : (
+            !isInitialized && <QRCodeScannerCrosshair />
+          )}
```
danibonilha: It seems there's something wrong with this condition, since `QRCodeScannerCrosshair` is never shown, shouldn’t it be `isInitialized` ?

topwebtek7: good catch. fixed it

eshinkawa (approved):
🚢

PR #500, opened 2021-11-15
Delete code from Rainbow that is no longer in use

lukemelia:
In the process of finding the code that is responsible for Mainnet NFT display, I came across lots of code that is no longer in use.


topwebtek7 (approved):
Awesome cleaning up in this PR! I've tested and no breaks, looks all works fine 💯

kdowlin:
this has passed my manual testing.

PR #501, opened 2021-11-15
Rename "unique tokens" to "collectibles" in code base 

lukemelia:
### Description

Much of the language in the code around collectibles/NFTs used the term "unique token" which doesn't appear in the UI and is not expressive. This PR standardizes on "collectibles" which is how the current UI refers to NFTs.

### Checklist

- [x] First merge https://github.com/cardstack/cardwallet/pull/500


topwebtek7 (approved):
This looks good to me! 🚀

src/redux/reducers.js
```diff
@@ -35,6 +34,7 @@ export default combineReducers({
   appState,
   coingecko,
   charts,
+  collectibles,
```
topwebtek7: I think it's better to have a short description so can understands easily what this reducer meant for.

lukemelia: ✅

PR #503, opened 2021-11-15
[CS-2466] Update pay merchant error messages, invalid network handling

topwebtek7:
### Description

Added handling for network check, selected prepaid card has not enough balance and updated error messages in pay merchant flow.

- [x] Completes #CS-2466

### Checklist

### Screenshots

Screenshot
Screen Shot 2021-11-15 at 9 35 18 PM

### DIFF
--- a/cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
+++ b/cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
@@ -1,6 +1,6 @@
 import { useEffect, useMemo, useState } from 'react';
 import { useNavigation, useRoute } from '@react-navigation/native';
-import { Alert } from 'react-native';
+import { Alert, AlertButton, InteractionManager } from 'react-native';
 import { getSDK, MerchantSafe } from '@cardstack/cardpay-sdk';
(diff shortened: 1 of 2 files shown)


cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
```diff
@@ -140,7 +141,10 @@ export const usePaymentMerchantUniversalLink = () => {
   useEffect(() => {
     if (error) {
       dismissLoadingOverlay();
-      handleAlertError('Something unexpected happened!');
+      logger.sentry('Merchat Payment failed!', error);
+      handleAlertError(
+        'Something unexpected happened! Please try again. If this error persists please contact support@cardstack.com'
```
eshinkawa, line 164: nit: it would be good to move all this error messages to constants file

eshinkawa (approved):
left some comments, but looks good overall

cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
topwebtek7: I think we can have repeatedly using global conts in constants file.. it's just only used here, so it's ok for now in my opinion

PR #506, opened 2021-11-15
[CS-2504] - Update Merchant Balances to look like depot balance

eshinkawa:
### Description

Merchant balances are confusing to end users and not clear they can be drilled into; Update Merchant Balances to look like depot balance.

I moved the **< BalanceSection />** component to the component folder so it would be reusable for both Depot and Merchant screens.

- [X] Completes #2504

### Checklist

### Screenshots

Nov-16-2021 11-31-08

### DIFF
--- a/cardstack/src/components/BalanceSection/BalanceSection.tsx
+++ b/cardstack/src/components/BalanceSection/BalanceSection.tsx
@@ -1,0 +1,63 @@
+import React, { useCallback, useMemo } from 'react';
+import { ScrollView } from 'react-native';
+import { TokenType } from '@cardstack/types';
+import Routes from '@rainbow-me/routes';
+import { Container, Text, TokenBalance } from '@cardstack/components';
+import { useNavigation } from '@rainbow-me/navigation';
+
+interface BalancesProps {
+  tokens: TokenType[];
+  navProps: any;
+}
+
+export const BalanceSection = ({ tokens, navProps }: BalancesProps) => {
+  const { navigate } = useNavigation();
+
+  const onPress = useCallback(
+    (token: TokenType) => () => {
+      navigate(Routes.EXPANDED_ASSET_SHEET_DRILL, {
+        asset: token,
+        type: 'token',
+        ...navProps,
+      });
+    },
+    [navProps, navigate]
+  );
+
+  const renderTokens = useMemo(
+    () =>
+      tokens.map(token => (
+        <TokenBalance
+          key={token.tokenAddress}
+          address={token.tokenAddress}
+          includeBorder
+          marginHorizontal={5}
+          nativeBalance={token.native.balance.display}
+          onPress={onPress(token)}
+          tokenBalance={token.balance.display}
+          tokenSymbol={token.token.symbol}
+          zIndex={1}
+        />
+      )),
+    [onPress, tokens]
+  );
+
+  return (
+    <ScrollView>
+      <Container
+        paddingHorizontal={5}
+        paddingBottom={3}
+        marginTop={7}
+        flexDirection="row"
+      >
+        <Text size="medium" marginRight={2}>
+          Balances
+        </Text>
+        <Text size="medium" color="tealDark">
+          {tokens.length}
+        </Text>
+      </Container>
+      {renderTokens}
+    </ScrollView>
+  );
+};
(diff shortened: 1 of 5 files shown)


lukemelia (commented):
How about a little test coverage for this newly moved component?

src/components/expanded-state/AvailableBalancesExpandedState.tsx
```diff
@@ -61,6 +50,8 @@ export default function AvailableBalancesExpandedState(
 ) {
   const { setOptions } = useNavigation();
   const [selectedTab, setSelectedTab] = useState<Tabs>(Tabs.ASSETS);
+  const depots = useRainbowSelector(state => state.data.depots);
+  const tokensData = depots?.[0].tokens;
```
danibonilha: We should have the same UI as depots on merchant balances, but not the same information, we need to use the tokens balances from the selected merchant safe and not from depot. Merchant, PrepaidCard and Depot they are all different types of safes and should not mix its contexts.

cardstack/src/components/Balances/Balances.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { TokenType } from '@cardstack/types';
+import Routes from '@rainbow-me/routes';
+import { Container, Text, TokenBalance } from '@cardstack/components';
+import { useNavigation } from '@rainbow-me/navigation';
+
+interface BalancesProps {
+  tokens: TokenType[];
+}
+
+export const Balances = ({ tokens }: BalancesProps) => {
```
danibonilha: We could rename this to BalanceSection or so, only Balances does not tell much

danibonilha (changes requested):
The wrong information is being displayed, more info on my comment

src/components/expanded-state/AvailableBalancesExpandedState.tsx
```diff
@@ -122,93 +118,6 @@ const TabHeader = ({ tab, selectedTab, setSelectedTab }: TabHeaderProps) => {
...
-
-  const assets = useMemo(
-    () =>
-      sortedByTokenBalanceAmount(tokens).map((token: TokenType) => ({
-        tokenSymbol: token.token.symbol,
-        tokenBalance: token.balance.display,
-        nativeBalance: token.native.balance.display,
-        onPress: () =>
-          navigate(Routes.EXPANDED_ASSET_SHEET, {
-            safeAddress,
```
danibonilha, line 179: Here we are passing the `safeAddress` as a nav param, so we can get the merchantSafea address in order to send the tokens from merchant, this is missing in the new BalancesSection, thus while trying to send tokens from merchant, the tokens from depot will be sent instead

cardstack/src/components/BalanceSection/BalanceSection.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export const BalanceSection = ({ tokens }: BalancesProps) => {
+  const { navigate } = useNavigation();
+
+  const onPress = useCallback(
+    (token: TokenType) => () => {
+      navigate(Routes.EXPANDED_ASSET_SHEET_DRILL, {
+        asset: token,
+        type: 'token',
```
danibonilha, line 20: Missing the safeAddress param

eshinkawa: I am passing as optional params, see what you think @danibonilha

danibonilha (approved):

cardstack/src/components/BalanceSection/BalanceSection.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const BalanceSection = ({ tokens, navProps = null }: BalancesProps) => {
+  const { navigate } = useNavigation();
+
+  const onPress = useCallback(
+    (token: TokenType) => () => {
+      navigate(Routes.EXPANDED_ASSET_SHEET_DRILL, {
+        asset: token,
+        type: 'token',
+        ...navProps,
```
danibonilha, line 21: It doesn't really need to be optional, ideally we can pass the safeAdress from depot too, we didn't do it before, because it used to default just depot transactions

eshinkawa: @danibonilha passing safeParams on DepotScreen now
https://github.com/cardstack/cardwallet/blob/679bafa627faa206b193ae59d1c3c72178c268d5/cardstack/src/screens/DepotScreen.tsx#L147

PR #507, opened 2021-11-16
[CS-2420] Update eoa deposit/withdraw transactions with account info

topwebtek7:
### Description

Updated EOA related transactions(both `Sent` and `Received` transactions, added To address for `Sent` transactions) view using account name, avatar color

- [x] Completes #CS-2420

### Checklist

### Screenshots

Screen Shot 2021-11-17 at 1 04 38 AM
Screen Shot 2021-11-17 at 1 15 44 AM

### DIFF
--- a/cardstack/src/components/Transactions/ERC20Transaction.tsx
+++ b/cardstack/src/components/Transactions/ERC20Transaction.tsx
@@ -1,6 +1,4 @@
 import React from 'react';
-
-import { NetworkBadge } from '../NetworkBadge';
 import { getDisplayDataByStatus } from './statusToDisplayData';
 import {
   TransactionBase,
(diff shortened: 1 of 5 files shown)


cardstack/src/components/Transactions/ERC20Transaction.tsx
```diff
@@ -63,5 +76,15 @@ export const ERC20Transaction = ({ item, ...props }: ERC20TransactionProps) => {
...
 
-  return <TransactionBase {...props} {...calculatedProps} />;
+  return (
+    <TransactionBase
+      {...props}
+      {...calculatedProps}
+      toAddress={
+        item.to !== accountAddress
+          ? `To: ${getAddressPreview(item.to)}`
+          : undefined
```
danibonilha: nit: we could have this on a meaningful variable, like has recipient or something, so its easier to understand that if we have a to parameter it means it's a sent transaction

cardstack/src/components/Transactions/ERC20Transaction.tsx
```diff
@@ -36,9 +39,14 @@ export const ERC20Transaction = ({ item, ...props }: ERC20TransactionProps) => {
           />
         ),
         Header: (
-          <Container paddingTop={4} paddingHorizontal={5}>
-            <NetworkBadge />
-          </Container>
+          <SafeHeader
+            address={accountName}
```
danibonilha: This is a little confusing, since we are not passing an address anymore, instead of a conditional `isFormattedAddress` we could have a new prop, like `title`, `from`, `account` or whatever, and then we can choose whether to show the new prop or the address.

danibonilha (approved):

PR #508, opened 2021-11-16
[CS-2504] [CS-2473] - Bump SDK v 0.27.26 -> Update available balances after claim all

eshinkawa:
### Description

SDK update solves balances update on Xdai / Sokol

### Checklist

### Screenshots
Nov-16-2021 14-23-01
Nov-16-2021 13-44-03


lukemelia (approved):

kdowlin:
We should probably bump the DID resolver to keep it inline with the sdk version as well.

PR #509, opened 2021-11-17
[CS-2552] Fix inputAmount not allowing decimals for SPD currency

topwebtek7:
### Description

Fixed inputAmount and currency util functions to don't allow decimals for SPD. note: there's still more work to do for improving inputAmount behavior for cs-2477, yet to be decided how gonna handle it.

- [x] Completes #CS-2552

### Checklist

### Screenshots

Nov-17-2021 17-46-52

### DIFF
--- a/cardstack/src/utils/currency-utils.ts
+++ b/cardstack/src/utils/currency-utils.ts
@@ -35,13 +35,10 @@ export function formatNative(value: string | undefined, currency = 'USD') {
     return '';
   }
 
-  const noDecimal = currency === NativeCurrency.SPD;
+  const decimalAllowed = currency !== NativeCurrency.SPD;
+  const isIncludeOneDot = (value.match(/\./g) || []).length === 1;
 
-  if (
-    !noDecimal &&
-    value.endsWith('.') &&
-    (value.match(/\./g) || []).length === 1
-  ) {
+  if (decimalAllowed && value.endsWith('.') && isIncludeOneDot) {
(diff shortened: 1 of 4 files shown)


cardstack/src/utils/currency-utils.ts
```diff
@@ -27,26 +35,35 @@ export function formatNative(value: string | undefined, currency = 'USD') {
     return '';
   }
 
-  if (value.endsWith('.') && (value.match(/\./g) || []).length === 1) {
+  const noDecimal = currency === NativeCurrency.SPD;
```
danibonilha: we are  "double" negating this twice, how about: 
```ts
const allowDecimals = currency !== NativeCurrency.SPD 
```
Then we don't need to negate `!`

cardstack/src/utils/currency-utils.ts
```diff
@@ -27,26 +35,35 @@ export function formatNative(value: string | undefined, currency = 'USD') {
...
+  ) {
     return `${formattedCurrencyToAbsNum(value).toLocaleString('en-US', {
       currency,
       maximumFractionDigits: 20,
       maximumSignificantDigits: 20,
     })}.`;
   }
 
-  if ((value.match(/\./g) || []).length === 1) {
+  if (!noDecimal && (value.match(/\./g) || []).length === 1) {
```
danibonilha: This check is happening more than once, maybe this could be a variable

danibonilha (approved):
Looks good, considering we will be handling locales next, just added small suggestions

PR #516, opened 2021-11-18
[CS-2364] - Can not hide a coin balance or prepaid card

eshinkawa:
### Description

- [X] Completes #2364

### Checklist

### Screenshots

Nov-17-2021 21-20-39-min

Nov-18-2021 15-31-22

### DIFF
--- a/cardstack/src/components/AssetList/AssetFooter.tsx
+++ b/cardstack/src/components/AssetList/AssetFooter.tsx
@@ -1,13 +1,9 @@
 import React from 'react';
 import { LayoutAnimation } from 'react-native';
-import { AssetListSectionItem, Button, Container } from '@cardstack/components';
+import { Button, Container } from '@cardstack/components';
 import { usePinnedAndHiddenItemOptions } from '@rainbow-me/hooks';
 
-interface AssetListProps {
-  sections: AssetListSectionItem<any>[];
-}
-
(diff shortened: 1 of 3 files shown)


cardstack/src/components/AssetList/AssetFooter.tsx
```diff
@@ -22,12 +18,7 @@ export const AssetFooter = ({ sections }: AssetListProps) => {
   const isInitialSelectionPinned = pinned.includes(selected[0]);
   const isInitialSelectionHidden = hidden.includes(selected[0]);
 
-  const sectionCount = sections.find(section =>
-    section.data.every(item => (item.address = selected[0]))
-  )?.header.count;
-
```
danibonilha, line 28: @kdowlin is it okay to hide all tokens ? Or should we keep at least one ?

danibonilha (commented):
Suggestion: We could use this opportunity to fix the broken layout while hiding the prepaid card, the overlay is out of place

cardstack/src/components/AssetList/AssetFooter.tsx
kdowlin: @danibonilha  we can hide all tokens as long as we keep the header.

danibonilha (approved):

PR #517, opened 2021-11-18
[CS-2519] - Add warning to users about not sending non .CPXD tokens to safe addresses

eshinkawa:
…o safe addresses

### Description

Add red warning label to address copy section for Depots/Merchants

Copy:

"I acknowledge that I can only send 

DAI.CPXD and CARD.CPXD to this address.

All other funds may be lost."

- [X] Completes #2519

### Checklist

### Screenshots

image
image

### DIFF
--- a/cardstack/src/screens/CopyAddressSheet.tsx
+++ b/cardstack/src/screens/CopyAddressSheet.tsx
@@ -1,5 +1,6 @@
-import React, { useState } from 'react';
+import React, { useCallback, useState } from 'react';
 import { useRoute } from '@react-navigation/core';
+import { TouchableOpacity } from 'react-native';
 import {
   ToastPositionContainerHeight,
   CopyToast,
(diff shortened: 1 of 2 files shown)


lukemelia (approved):

lukemelia:
Looks fine. My question is whether the tokens should come from SDK config. We have plans to support other .CPXD tokens, and this seems like the kind of thing we would be likely to forget to update.

cardstack/src/screens/CopyAddressSheet.tsx
```diff
@@ -47,6 +50,22 @@ const CopyAddressSheet = () => {
             alignItems="center"
             width="100%"
           >
+            <Container flexDirection="row" marginBottom={6}>
+              <Container flex={1} justifyContent="center">
+                <Checkbox
+                  isSelected={checked}
+                  onPress={() => setChecked(!checked)}
```
danibonilha: nit: This could be a function `toggleCheckBox` wrapped on useCallback since we are doing the same thing on both places

danibonilha:
@kdowlin  There's the case where we show the address of the prepaidCard, we already don't allow to copy, so should we have the text without the checkbox ? Or not even have the text ?

image

kdowlin:
For Prepaid Cards we should not show this for the time being.

PR #518, opened 2021-11-18
[Refactor]: Refactor safes

danibonilha:
### Description

This is the Draft of the RTK implementation, there are lots of things to be done yet, but since it's getting big, I decided to open this draft in case we need to discuss anything. In this current state  #(CS-2474) is fixed, but there are a couple of things broken, like claiming revenue update. So the plan is to first get all the current features working with global safes, once we get that and it's stable, we can move to handling each safe instance, then handling chartPrices, using normalized cache with entities etc. 

## Update
This is now ready to be reviewed since the broken things have been fixed, but just as a follow-up I'm adding the things we still need to do on upcoming PRs

- Extract apis sdk function to its own "service" files
- Split safes fetching by types
- Create wrapper to handle mutations/queries
- Migrate code that uses rainbow selector to retrieve safes data (txs, signing requests, etc)
- Remove all unused safes handling functions
- Migrate all sdk interactions to rtk queries/mutations (buying prepaidCard, sendtokens etc)
- Add right types to rtk stuff
- Create rtk query test structure
- Remove local storage after fetching safes once we migrate everything to rtk
- Fix network storage/redux race condition


cardstack/src/components/TransactionConfirmationSheet/displays/components/sections/PrepaidCardTransactionSection.tsx
```diff
@@ -20,10 +20,17 @@ export const PrepaidCardTransactionSection = ({
...
-    card => card.address === prepaidCardAddress
+  const { prepaidCard } = useGetSafesDataQuery(
+    { address: accountAddress, nativeCurrency },
+    {
+      selectFromResult: ({ data }) => ({
+        prepaidCard: data?.prepaidCards.find(
+          (card: PrepaidCardSafe) => card.address === prepaidCardAddress
+        ),
+      }),
+    }
```
lukemelia, line 33: It seems like we should have a hook for fetching a specific safe by address and type rather than force components through `useGetSafesDataQuery` when all they care about is one safe. This can be a future change.

cardstack/src/screens/MerchantScreen.tsx
```diff
@@ -395,3 +433,62 @@ const RecentActivitySection = ({
...
+    accountAddress,
+    claimRevenue,
+    merchantSafe.address,
+    merchantSafe.revenueBalances,
+    network,
+    selectedWallet,
+    showLoadingOverlay,
+  ]);
+
+  // isRefreshing may be false when isSucess is truthy on the first time
```
lukemelia: s/isSucess/isSuccess/

cardstack/src/screens/MerchantScreen.tsx
```diff
@@ -395,3 +433,62 @@ const RecentActivitySection = ({
...
+      dismissLoadingOverlay();
+    }
+  }, [dismissLoadingOverlay, isSuccess, hasUpdated]);
+
+  useEffect(() => {
+    if (isError) {
+      dismissLoadingOverlay();
+      logger.sentry('Error claiming revenue', error);
+      Alert.alert(
+        'Could not claim revenue, please try again. If this problem persists please reach out to support@cardstack.com'
```
lukemelia, line 488: Unrelated to this PR: we should introduce a constant for our support email address.

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -20,33 +12,21 @@ import {
...
 
 const CHART_HEIGHT = 600;
 
 export default function UnclaimedRevenueExpandedState({
   asset: merchantSafe,
+  customFunction: onClaimAll,
 }: {
   asset: MerchantSafeType;
+  customFunction: () => void;
 }) {
```
lukemelia: The return type syntax is getting confusing to read. Maybe extract a named interface to make it more readable.

cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
```diff
@@ -47,49 +34,64 @@ export const usePaymentMerchantUniversalLink = () => {
...
-
-      return;
     }
-  }, [isLoadingAssets, goBack, prepaidCards.length]);
+  }, [isLoadingCards, goBack, prepaidCards.length, isLoading]);
 
   useEffect(() => {
-    if (qrCodeNetwork && accountNetwork && qrCodeNetwork !== accountNetwork) {
+    if (
+      !isLoading &&
```
lukemelia: nit: I would prefer to read this as an affirmative standalone guard:

```js
if (isLoading) {
  return;
}
// continue ...
```

More verbose but easier to read IMO

lukemelia (approved):
A few minor comments. Otherwise looks good.

cardstack/src/components/TransactionConfirmationSheet/displays/components/sections/PrepaidCardTransactionSection.tsx
danibonilha: Agreed, this is what we are aiming for

cardstack/src/screens/MerchantScreen.tsx
danibonilha: Agreed, we need to map all places we use the support email address and fix it in one place

PR #520, opened 2021-11-21
Simplify fallbackExplorer by removing unnecessary distinction between xdaiAssets and mainnetAssets

lukemelia:
The fallbackExplorer reducer contained conditional logic that distinguished between xdai assets and mainnet assets. There was no actual difference between how these assets were fetched or exposed though.

### DIFF
--- a/src/redux/fallbackExplorer.js
+++ b/src/redux/fallbackExplorer.js
@@ -48,12 +48,8 @@ const getCurrentAddress = address => {
 };
 
 const findNewAssetsToWatch = () => async (dispatch, getState) => {
-  const { accountAddress, network } = getState().settings;
-  const {
-    xdaiChainAssets,
-    mainnetAssets,
-    latestTxBlockNumber,
-  } = getState().fallbackExplorer;
+  const { accountAddress } = getState().settings;
+  let { assets, latestTxBlockNumber } = getState().fallbackExplorer;
 
   const newAssets = await findAssetsToWatch(
     accountAddress,
@@ -63,24 +59,11 @@ const findNewAssetsToWatch = () => async (dispatch, getState) => {
   if (newAssets.length > 0) {
     logger.log('😬 Found new assets!', newAssets);
 
-    const assets =
-      network === networkTypes.xdai
-        ? {
-            xdaiChainAssets: uniqBy(
-              [...xdaiChainAssets, ...newAssets],
-              token => token.asset.asset_code
-            ),
-          }
-        : {
-            mainnetAssets: uniqBy(
-              [...mainnetAssets, ...newAssets],
-              token => token.asset.asset_code
-            ),
-          };
+    assets = uniqBy([...assets, ...newAssets], token => token.asset.asset_code);
 
     dispatch({
       payload: {
-        ...assets,
+        assets,
       },
       type: FALLBACK_EXPLORER_SET_ASSETS,
     });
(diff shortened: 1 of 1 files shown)


src/redux/fallbackExplorer.js
```diff
@@ -48,12 +48,8 @@ const getCurrentAddress = address => {
...
 
 const findNewAssetsToWatch = () => async (dispatch, getState) => {
-  const { accountAddress, network } = getState().settings;
-  const {
-    xdaiChainAssets,
-    mainnetAssets,
-    latestTxBlockNumber,
-  } = getState().fallbackExplorer;
+  const { accountAddress } = getState().settings;
+  let { assets, latestTxBlockNumber } = getState().fallbackExplorer;
```
danibonilha: nit:  Usually we don't use `let` while  destructuring, as it's dangerous to mistakenly reassign without noticing, we could have it as const, and inside the newAssets condition, create a `const updatedAssets = uniqBy...`  and pass it to the assets payload.

src/redux/fallbackExplorer.js
```diff
@@ -416,17 +399,12 @@ export const fetchAssetsBalancesAndPrices = async () => {
 
   const formattedNativeCurrency = toLower(nativeCurrency);
 
-  const {
-    xdaiChainAssets,
-    mainnetAssets,
+  let {
+    assets,
```
danibonilha: same here, we could rename this as `oldAssets`, `currentAssets`  i.e
```
 const {
    assets: currentAssets,
```
This way we still have the` assets` keyword if we want to use in the rest of the code

src/redux/fallbackExplorer.js
```diff
@@ -565,35 +543,24 @@ export const fetchAssetsBalancesAndPrices = async () => {
...
-          }
-        : {
-            mainnetAssets: mainnetAssets.concat(newMainnetAssets),
-          };
+    assets = assets.concat(newAssets);
 
     await dispatch({
       payload: {
-        ...assets,
+        assets,
```
danibonilha: And here we could have `assets: [...assets, ...newAssets]`, then there's no need to reassign `assets` and use `let`

src/redux/fallbackExplorer.js
```diff
@@ -623,22 +590,18 @@ const INITIAL_STATE = {
...
-      const { mainnetAssets, xdaiChainAssets } = action.payload;
+      const { assets } = action.payload;
 
       return {
         ...state,
-        mainnetAssets: mainnetAssets ? mainnetAssets : state.mainnetAssets,
-        xdaiChainAssets: xdaiChainAssets
-          ? xdaiChainAssets
-          : state.xdaiChainAssets,
+        assets: assets ? assets : state.assets,
```
danibonilha: nit: when checking if something exists and should be used first, we don't need a ternary, we could use `||` or `??`. 
`??` only checks `null/undefined` cases,  and `||` checks  all .js falsy values, including the ones from `??`, plus empty string, false, etc

so we could have 
`assets: assets || state.assets,`
if assets is truthy, use assets, if it's any falsy value, fallback to state.assets

danibonilha (approved):
Looks good, just some nit comments

src/redux/fallbackExplorer.js
lukemelia: ✅

src/redux/fallbackExplorer.js
lukemelia: ✅

src/redux/fallbackExplorer.js
lukemelia: ✅

src/redux/fallbackExplorer.js
lukemelia: ✅

PR #521, opened 2021-11-22
[CS-2425, CS-2586] Add merchant withdraw/deposit transaction strategies, fix receipient name with wallet/merchant name, refactoring

topwebtek7:
### Description

- Add new merchant withdrawn, deposited transaction strategies and new transaction type components
- Updated address with account/merchant name
- Update Transaction to Payment History in Merchant screen
- Update merchant available balance activities

- [x] Completes #CS-2425
- [x] Completes #CS-2586
- [x] Completes #CS-2390

### Checklist

### Screenshots

Screenshot 
Screenshot


kdowlin:
@topwebtek7 Confirm the token is rendering correctly. This example shows DAI which doesn't make sense and should be DAI.CPXD if it is in the Account Balances tab of a Business Account. (Probably nothing but wanted to acknowledge)

topwebtek7:
> @topwebtek7 Confirm the token is rendering correctly. This example shows DAI which doesn't make sense and should be DAI.CPXD if it is in the Account Balances tab of a Business Account. (Probably nothing but wanted to acknowledge)

basically it's Sokol merchant account balances tab and it's all DAI for now

kdowlin:
Ah yes - forgot Sokol is all dai with no cpxd

cardstack/src/hooks/merchant/useNameOrPreviewFromAddress.ts
```diff
@@ -0,0 +1,25 @@
+import { useRainbowSelector } from '@rainbow-me/redux/hooks';
+import { useAccountProfile } from '@rainbow-me/hooks';
+import { getAddressPreview } from '@cardstack/utils';
+
+export const useNameOrPreviewFromAddress = (address: string) => {
```
danibonilha, line 5: We could rename it to `useMerchantNameOrPreviewFromAddress` since it only works for merchant names

cardstack/src/services/merchant-service.ts
```diff
@@ -0,0 +1,20 @@
...
+    return {
+      ...merchantSafe,
+      merchantInfo,
+    };
+  } catch (e) {
+    logger.sentry('Error getting cardCustomization', e);
+  }
+
+  return merchantSafe;
+};
```
danibonilha: we have `updateMerchantSafeWithCustomization` on merchant utils, we can either keep this one and update on fetchSafes and move this file into the merchant folder, or we could remove this one and let `updateMerchantSafeWithCustomization` only, which I'll move to the right place following PRs

danibonilha (commented):
There's something odd happening, the Transactions modal doesn't scroll anymore, on this branch

https://user-images.githubusercontent.com/20520102/143054848-4daa02c7-1280-4927-8134-47292a4b45a8.mp4

cardstack/src/hooks/merchant/useNameOrPreviewFromAddress.ts
topwebtek7: > We could rename it to `useMerchantNameOrPreviewFromAddress` since it only works for merchant names

no it works for merchant name or account name, that's why I named like that

cardstack/src/services/merchant-service.ts
topwebtek7: cool. I will remove it so you can move it to right place in your following pr

topwebtek7:
> There's something odd happening, the Transactions modal doesn't scroll anymore, on this branch
> 
>  Simulator.Screen.Recording.-.iPhone.11.-.2021-11-23.at.12.36.08.mp4

it looks ok to me
Nov-23-2021 23-57-21

cardstack/src/hooks/merchant/useNameOrPreviewFromAddress.ts
topwebtek7: could be extended supporting prepaidcard names and more, I think

danibonilha: My bad, I thought it was merchant only, since it's inside the merchant folder

danibonilha (approved):

PR #523, opened 2021-11-23
[CS-2554] Fix getting symbol character from wallet address to be consistent

topwebtek7:
### Description

Updated getting symbol character from wallet address behavior to be consistent. All address starts with `0x` so used next character after that as symbol.

- [x] Completes #CS-2554

### Checklist

### Screenshots

Screen Shot 2021-11-23 at 8 47 25 PM

### DIFF
--- a/src/hooks/useAccountProfile.js
+++ b/src/hooks/useAccountProfile.js
@@ -1,5 +1,4 @@
 import { get, toUpper } from 'lodash';
-import { removeFirstEmojiFromString } from '../helpers/emojiHandler';
 import networkTypes from '../helpers/networkTypes';
 import useAccountSettings from './useAccountSettings';
 import useWallets from './useWallets';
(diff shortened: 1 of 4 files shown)


src/hooks/useAccountProfile.js
```diff
@@ -48,13 +50,13 @@ export default function useAccountProfile() {
...
-  const accountSymbol = new GraphemeSplitter().splitGraphemes(
+  const accountSymbol = getSymbolCharacterFromAddress(
     network === networkTypes.mainnet
-      ? labelOrAccountName || toUpper(accountENS) || `${index + 1}`
+      ? labelOrAccountName || toUpper(accountENS) || address
       : label === accountENS
       ? toUpper(accountName)
       : toUpper(label) || toUpper(accountName)
-  )[0];
+  );
```
danibonilha, line 55: I know it's "legacy code" but can we simplify or make this nested ternary more readable ? There are so many conditions there, it's hard to read, I'm not even sure if we need to handle all those cases, since generating an emoji is not something we care about, probably we just need to check the label, or the addressPreview

danibonilha (approved):

PR #524, opened 2021-11-23
[Fix]: Show loading while refetchn safes w/ new acc

danibonilha:
### Description

While switching account, or creating a new one, the isLoading flag from rtk query is not triggered, what is triggered is the isFetching flag, which is also triggered when we pull to refresh, so we need a condition to check if it's not manually triggered by the user and we isFetching, it means some args changed and the data will be different so we need to show the fullscreen loading.

### Screenshots

https://user-images.githubusercontent.com/20520102/143048879-7ed2855e-6fdd-40cf-8f98-aa1695f780cf.mp4


topwebtek7 (approved):
LGTM

PR #527, opened 2021-11-24
[CS-2555][CS-2635] - Hide / Unhide business account / Remove pin button

eshinkawa:
### Description

User should be able to hide and unhide a business account.  Also the pin button was removed

- [X] Completes #[CS-2555][CS-2635]

### Checklist

### Screenshots

Nov-24-2021 12-21-01-min

image

### DIFF
--- a/cardstack/src/components/MerchantSafe/MerchantSafe.tsx
+++ b/cardstack/src/components/MerchantSafe/MerchantSafe.tsx
@@ -1,6 +1,7 @@
-import React, { useCallback } from 'react';
+import React from 'react';
 import { ContactAvatar } from '@rainbow-me/components/contacts';
 import {
+  CenteredContainer,
   Container,
   Icon,
   SafeHeader,
@@ -10,6 +11,10 @@ import {
 import { MerchantInformation, MerchantSafeType } from '@cardstack/types';
 import { useNavigation } from '@rainbow-me/navigation';
 import Routes from '@rainbow-me/routes';
+import {
+  PinnedHiddenSectionOption,
+  useHiddenItemOptions,
+} from '@rainbow-me/hooks';
 
 interface MerchantSafeProps extends MerchantSafeType {
   networkName: string;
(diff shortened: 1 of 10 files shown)


cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -21,40 +27,141 @@ interface MerchantSafeProps extends MerchantSafeType {
...
+  const [isScrollable] = useState(false);
+  const Wrapper = isScrollable ? ScrollView : Container;
+
+  const {
+    editing,
+    selected,
+    hidden,
+    select,
+    deselect,
+  } = usePinnedAndHiddenItemOptions();
```
danibonilha: If we won't have the pin feature anymore we could rename all pinned related stuff to reflect that

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -21,40 +27,141 @@ interface MerchantSafeProps extends MerchantSafeType {
   infoDID: string;
 }
 
+const SELECT_ICON_WIDTH = '13%';
+
 export const MerchantSafe = ({ merchantInfo, ...props }: MerchantSafeProps) => {
   const { navigate } = useNavigation();
+  const [isScrollable] = useState(false);
+  const Wrapper = isScrollable ? ScrollView : Container;
```
danibonilha: i don't understand this piece, we never set isScrollable

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -21,40 +27,141 @@ interface MerchantSafeProps extends MerchantSafeType {
...
+      if (isSelected) {
+        deselect(props.address);
+      } else {
+        select(props.address);
+      }
+    } else {
+      const merchantData = { ...props, merchantInfo };
+      navigate(Routes.MERCHANT_SCREEN, { merchantSafe: merchantData });
+    }
+  };
```
danibonilha: isn't this the exact same behavior as prepaidCard ? Maybe we should create wrapper to handle this, since these features should work the same

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -21,40 +26,129 @@ interface MerchantSafeProps extends MerchantSafeType {
...
+  if (!isEditing && isHidden) {
+    return null;
+  }
+
+  const isSelected = selected.includes(props.address);
+  const showIcon = isHidden;
+  const iconName = 'eye-off';
+  const iconFamily = isHidden ? 'Feather' : 'MaterialCommunity';
+  const editingIconName = isSelected ? 'check-circle' : 'circle';
+
```
danibonilha, line 48: Maybe I didn't explained clearly, I meant: this whole hiding/editing block and components are the same as the one on prepaidcard section, the suggestion was to create something we could use across both, without duplicating  all those const and conditions, it doesn't make much sense to abstract just the onPress function.

src/hooks/useHiddenItemOptions.tsx
```diff
@@ -125,6 +131,18 @@ export const PinnedHiddenItemOptionProvider = ({
...
+    isSelected: boolean,
+    address: string
+  ) => {
+    if (isSelected) {
+      deselect(address);
+    } else {
+      select(address);
+    }
+  };
+
```
danibonilha, line 145: `isEditing` is never used why is it being passed ?

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -10,6 +11,10 @@ import {
 import { MerchantInformation, MerchantSafeType } from '@cardstack/types';
 import { useNavigation } from '@rainbow-me/navigation';
 import Routes from '@rainbow-me/routes';
+import {
+  PinnedHiddenSectionOption,
```
danibonilha, line 15: This still refers to the Pinned feature

danibonilha:
@eshinkawa I just read the ticket, and we shouldn't completely remove the Pin functionally only feature flag for now, which means we need to create a flag were we can set to true/false to enable or disable the Pin functionally.

PR #528, opened 2021-11-24
[CS-2621]: Only load safes once wallet is initialized with required info

danibonilha:
### Description

Since safes don't come from fallbackExplorer anymore we need to make sure all the information needed to build the data structure is ready before fetching the safes, props to @topwebtek7 on figuring we were missing conversionsRates on some cases. Also handles loading states that were flickering and bumps the sdk.

PS: I've tried to create a wrapper for the query bc we use the same params, but it's really complex to type properly, since we are moving to have each individual query next, it was not worth to spend too much time on that.


lukemelia (approved):
👍

PR #529, opened 2021-11-24
Improve generic transaction signing request experience

lukemelia:
- Fixes a bug where generic transactions on xdai chain using XDAI native token didn't work because the code was trying to use ETH
- Fixes an issue where transactions that don't specify gasPrice and/or gasLimit were showing these attributes as `NaN` in generic transaction signing requests
- delete some unused code 

image


danibonilha (approved):
LGTM, what happens if you tap on the black label?

lukemelia:
> LGTM, what happens if you tap on the black label?

Didn't get a chance to try it. Feel free to pick up or I will look on Monday.

topwebtek7 (approved):
LGTM! 🚢

lukemelia:
> LGTM, what happens if you tap on the black label?

Crash. It is the same as on develop branch, so I think this PR is good to merge. I'll look at that issue next.

PR #530, opened 2021-11-26
[Feat]: Add query to fetch safes by prepaidcard type

danibonilha:
### Description

This PR adds the query to fetch PrepaidCards safes only, it is still not in use, bc we need to create for each safe type bf migrating, but I wanted to merge because it also creates a `queryPromiseWrapper` helper so we can use across all ours `queryFn` that automatically handles success/error cases and we can start to migrate the mutations too.

### DIFF
--- a/cardstack/src/services/utils/index.ts
+++ b/cardstack/src/services/utils/index.ts
@@ -1,0 +1,44 @@
+import { captureException } from '@sentry/minimal';
+import logger from 'logger';
+
+type QueryError = {
+  error: {
+    status: number;
+    data: any;
+  };
+  data?: undefined;
+};
+
+type QuerySuccess<TResult> = {
+  data: TResult;
+  error?: undefined;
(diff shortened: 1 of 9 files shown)


cardstack/src/services/utils/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    captureException(error);
+
+    return {
+      error: {
+        status: options?.errorStatus || 418,
+        data: error,
+      },
+    };
+  }
+};
```
lukemelia, line 44: It would be nice to have some simple unit tests for this so there is an easy demonstration of usage.

lukemelia (approved):
Good to go with some unit tests

cardstack/src/services/utils/index.ts
danibonilha: image

PR #531, opened 2021-11-29
[CS-2585] Refactor transactions to use correct historyPrices in native currency conversion

topwebtek7:
### Description

`CARD` token is not supported correctly in cryptocompare api so used kucoin exchange to get CARD price in USDT(kucoin does not supports nativecurrencies, support USDT) and then convert it to native currency by calling cryptocompare again. Also found many transactions used sdk to get native currency balance but it's not history price so updated them to use history price properly at that transaction's timestamp.

- [x] Completes #CS-2585

### Checklist

### Screenshots

Screen Shot 2021-11-29 at 10 11 51 PM

### DIFF
--- a/cardstack/src/services/historical-pricing-service.ts
+++ b/cardstack/src/services/historical-pricing-service.ts
@@ -2,14 +2,6 @@ import { CRYPTOCOMPARE_API_KEY } from 'react-native-dotenv';
 import logger from 'logger';
 import { removeCPXDTokenSuffix } from '@cardstack/utils';
 
-const getRoundedTimestamp = (timestamp: string | number) => {
-  const date = new Date(Number(timestamp) * 1000);
-
-  date.setHours(0, 0, 0, 0);
-
-  return date.getTime();
-};
-
 export const fetchHistoricalPrice = async (
   symbol: string,
   timestamp: string | number,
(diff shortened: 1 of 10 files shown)


cardstack/src/services/historical-pricing-service.ts
```diff
@@ -1,28 +1,46 @@
...
 export const fetchHistoricalPrice = async (
   symbol: string,
   timestamp: string | number,
   nativeCurrency: string
 ) => {
   try {
+    if (!symbol) {
+      return 0;
+    }
+
```
danibonilha, line 17: Since we have this condition to check if symbol exists, we can make symbol an optional param, so we don't need to add a fallback to empty string on all the strategies

cardstack/src/services/historical-pricing-service.ts
```diff
@@ -1,28 +1,46 @@
...
+      const nativeCurrencyDataForUSDT = await (
+        await fetch(
+          `https://min-api.cryptocompare.com/data/pricehistorical?fsym=USDT&tsyms=${nativeCurrency}&ts=${timestamp}&api_key=${CRYPTOCOMPARE_API_KEY}`
+        )
+      ).json();
+
+      const nativeCurrencyPriceForUSDT =
+        nativeCurrencyDataForUSDT.USDT[nativeCurrency];
+
+      return usdtPrice * nativeCurrencyPriceForUSDT;
```
danibonilha: nit: I'd suggest to move this piece to its own function, just to make it easier to read, something like:
```ts
if (tokenSymbol === cryptoCurrencies.CARD.currency) { 
      const cardPrice = await fetchCardHistoricalPrice(tokenSymbol, timestamp, nativeCurrency)
      return cardPrice
}
```

cardstack/src/services/historical-pricing-service.ts
```diff
@@ -1,28 +1,46 @@
...
     // For cpxd tokens we chop off .CPXD in order to get mainnet's historical price
     const tokenSymbol = removeCPXDTokenSuffix(symbol);
 
-    const roundedTimestamp = getRoundedTimestamp(timestamp);
+    // cryptocompare does not support CARD price history correctly,
+    // so uses kucoin exchange to get CARD price in USDT(kucoin supports only USDT) and convert it to native currency
+    if (tokenSymbol === cryptoCurrencies.CARD.currency) {
+      const usdtPriceData = await (
+        await fetch(
+          `https://min-api.cryptocompare.com/data/pricehistorical?fsym=${tokenSymbol}&tsyms=USDT&ts=${timestamp}&e=kucoin&api_key=${CRYPTOCOMPARE_API_KEY}`
```
danibonilha: nit: we could have this part of this as a baseURL, so there's no need copy all three times

cardstack/src/transaction-mapping-strategies/transaction-mapping-strategy-types/bridge-to-layer1-strategy.ts
```diff
@@ -27,26 +27,25 @@ export class BridgeToLayer1EventStrategy extends BaseStrategy {
       return null;
     }
 
-    const nativeBalance = await getNativeBalanceFromOracle({
-      symbol: bridgeToLayer1Event.token.symbol,
-      balance: bridgeToLayer1Event.amount,
-      nativeCurrency: this.nativeCurrency,
-    });
+    const symbol = bridgeToLayer1Event.token.symbol || '';
```
danibonilha, line 30: this is the type of fallback I suggested we could get rid of

danibonilha (approved):
It's looking good, just a couple of suggestions

cardstack/src/services/historical-pricing-service.ts
topwebtek7: basically we couldn't make first param only optional as timestamp, nativeCurrency are required params and didn't want to change the order of params in fetchHistoricalPrice. also there are several places that require symbol as string(not optional), that's another reason I made it not optional in fetchHistoricalPrice and added like `const symbol = transaction.token.symbol || '';`. I think it will be answer to below comment as well.

PR #534, opened 2021-11-30
[CS-2655]: Keep local cache for safes in case of error

danibonilha:
### Description

In case of an error while fetching safes we fallback to local cache if it exists and store/show a timestamp so we can let the user know the data might be stale

- [x] Completes #(CS-2655)

### Screenshots

- First fetch 9:48AM 
- Pull to refresh at 9:49AM, forced an error, kept stale data from 9:48AM
- 9:49AM refreshed without error, updated data at 9:49AM

Nov-30-2021 09-58-05

### DIFF
--- a/cardstack/src/components/AssetList/AssetList.tsx
+++ b/cardstack/src/components/AssetList/AssetList.tsx
@@ -25,7 +25,7 @@ import {
   Icon,
   Text,
 } from '@cardstack/components';
-import { isLayer1, Device } from '@cardstack/utils';
+import { isLayer1, Device, dateFormatter } from '@cardstack/utils';
 import {
   PinnedHiddenSectionOption,
   useAccountProfile,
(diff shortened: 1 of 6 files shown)


topwebtek7 (approved):
It's good idea for caching safes and looks good to me! 🚢 
just a comment

cardstack/src/components/AssetList/AssetList.tsx
```diff
@@ -207,6 +218,22 @@ export const AssetList = (props: AssetListProps) => {
...
+          justifyContent="center"
+        >
+          <Text color="white" size="xs">
+            {strings.lastUpdatedAt(timestamp)}
+          </Text>
+        </Container>
+      ) : null,
+    []
+  );
+
```
topwebtek7, line 236: Is it just for temporary test purpose only or intentionally added? If it's not test, then we can create a util function for lastUpdatedAt rather than to have `strings.lastUpdatedAt(` in assetList

danibonilha: This is only used in this component, so at least for now it makes more sense to have it scoped within the component, rather than creating an util function for it.

PR #535, opened 2021-11-30
[CS-2589] Update merchant home screen, get rid of recent activity screen, graph

topwebtek7:
### Description

Removed recent activity block, graph on merchant home screen and merged into payment history, drilling into payment history goes to earning history and see graph.

- [x] Completes #CS-2589

### Checklist

### Screenshots

Screenshot


cardstack/src/screens/MerchantScreen.tsx
```diff
@@ -134,9 +119,8 @@ const MerchantScreen = () => {
       />
       <Container height="100%" justifyContent="flex-end" paddingBottom={4}>
         <ScrollView
-          flex={1}
           width="100%"
-          contentContainerStyle={{ alignItems: 'center', paddingBottom: 400 }}
+          contentContainerStyle={{ alignItems: 'center', paddingBottom: 200 }}
```
danibonilha: It probably don't need to be `400` but we do need more than `200`, maybe `260` idk, bc with 200 we are not able to scroll to the end and see the bottom border and the full +more keyword

image

cardstack/src/screens/MerchantScreen.tsx
```diff
@@ -399,46 +324,50 @@ const SectionWrapper = ({
...
-          {...recentActivityDataSectionData}
-          includeBorder
-          isFullWidth
-        />
-      ) : (
-        <SectionWrapper disabled={!firstActivityDataListItem}>
-          <Text variant="subText">No recent activity</Text>
-        </SectionWrapper>
-      )}
+      <SectionWrapper onPress={onPress} hasDetailsText={false}>
```
danibonilha: Maybe we should disable this section if there are no transactions ? Because this has no info at all 
image

src/components/expanded-state/LifetimeEarningsExpandedState.tsx
```diff
@@ -80,7 +80,7 @@ const ChartSection = ({ merchantSafe }: { merchantSafe: MerchantSafeType }) => {
   return (
```
danibonilha, line 80: nit: we are still referencing everything to LifetimeEarnings, but now it's actually payment history, we could think about renaming this

cardstack/src/screens/MerchantScreen.tsx
```diff
@@ -399,46 +324,50 @@ const SectionWrapper = ({
...
-  const firstActivityDataListItem = sections[0]?.data[0];
+  const numberOfTransactions = sections[0]?.data?.length || 0;
 
-  const recentActivityDataSectionData = {
-    item: firstActivityDataListItem,
-    onPressTransaction: onPress,
-  };
+  const first3ActivityLists = sections[0]?.data
+    ?.slice(0, 3)
+    .map(item => <TransactionItem item={item} isFullWidth disabled />);
```
danibonilha, line 352: This is more question than a suggestion, should we really disable the 3 transactions blockscout tap? If we have the `+more` section, should we enable the default tx item behavior to `view on blockscout` and when the user click +more we show the modal ? Or should we leave the way it is, making the whole section tappable straight to the modal ? cc: @kdowlin

kdowlin: I think it's ok to disable for now. I really don't think there's much value in the blockscout view. This will ultimately become the payment received share sheet. And then we will need to make sure we keep the transactions clickable.

danibonilha (approved):

PR #539, opened 2021-12-01
[CS-2682]: Fix not showing alert when payment link diff net

danibonilha:
### Description

When we are at a diff network from the one from the payment request we won't be able to retrieve the merchant info, so it doesn't make sense to wait for it to show the alert and handle the other steps since we can have separated loadings for merchantInfo only.
 

Video was added to the ticket, since it's too big for github


topwebtek7 (approved):
LGTM 🚢

PR #540, opened 2021-12-02
[CS-2477] Update currency utils to support number format based on device's country setting

topwebtek7:
### Description

- Added react-native-localize instead of react-native-languages as it is deprecated and also to support getting device's locale, number format, etc
- Added numbro library to use formatting input amounts
- Refactored currency util functions using numbro, refactor amount input to be consistent with rainbow's ExchangeInput

- [x] Completes #CS-2477

### Checklist

### Screenshots

Dec-06-2021 20-35-07


danibonilha (commented):
It's not working properly for me, it's adding more zeros than expected, and it's confusing having commas as thousands separator sometimes, and also as decimal separators on the underling text.
I'm wondering if we could apply the same behavior as we have in the send flow, instead of adding these libraries and risking having unexpected bugs. If we just replace the comma with dot and not split in thousands we have a safer approach and also handles the bug, I'd like to handle localization all together once we decided to internationalize the app. Because in this PR we are not adding the same behavior across the app. Let me know your thoughts

danibonilha:
> It's not working properly for me, it's adding more zeros than expected, and it's confusing having commas as thousands separator sometimes, and also as decimal separators on the underling text. I'm wondering if we could apply the same behavior as we have in the send flow, instead of adding these libraries and risking having unexpected bugs. If we just replace the comma with dot and not split in thousands we have a safer approach and also handles the bug, I'd like to handle localization all together once we decided to internationalize the app. Because in this PR we are not adding the same behavior across the app. Let me know your thoughts

My region is Brazil and the decimals separator is comma, but then here, besides having the wrong amount, the comma is used as a thousand separator and the dot as decimal, which is pretty confusing
The link: https://wallet.cardstack.com/pay/xdai/0x7e337CDb0794597e818F585e56873c2e0C27C076?amount=10.88&currency=USD

image

topwebtek7:
> 

@danibonilha could you check on your side again? Fixed it and looks stable to me...
I understand what you mean.. I think this will improve user experience of payment request behaviors

danibonilha:
It's also confusing that, if I open a link with `10.25 USD` and my region is the one that uses comma as separator for decimals, if I go to edit an amount I get:

Screenshot 

Which in my region would be "wrong" because it should be `10,25`, so it seems we are adding more complexity and confusion trying to use both

topwebtek7:
> 

ok, refactored again to be consistent with Rainbow's Input behavior. btw just commented out numbro usage as it's really useful library and could use this when implementing regional support

cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -64,7 +103,11 @@ export const InputAmount = memo(
...
     useEffect(() => {
       if (nativeCurrency === NativeCurrency.SPD) {
-        setInputValue(formatNative(inputValue, nativeCurrency));
+        const formattedValue = removeLeadingZeros(inputValue);
+
+        if (inputValue !== formattedValue) {
+          setInputValue(formattedValue);
+        }
       }
       // eslint-disable-next-line react-hooks/exhaustive-deps
```
danibonilha: Do we really want this to be disabled ? If so why ?

cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -43,17 +46,53 @@ export const InputAmount = memo(
...
+        const parts = inputValue.split('.');
+
+        if (parts[0].length > 1 && !Number(parts[0])) {
+          onChangeText(`0.${parts[1]}`);
+        }
+      }
+
+      setIsFocused(false);
+      setIsTouched(false);
+    }, [inputValue, onChangeText]);
```
danibonilha: it seems, most of these behaviors are being copied from `ExchangeInput`, wouldn't make more sense to just reuse `ExchangeInput` component here ? 
Or at least evaluate if we really need all this isFocused/touched, blur, focus behavior for this specific case.

index.js
```diff
@@ -20,7 +21,43 @@ lang.translations = Object.assign(
...
+      }
+    }
+
+    if (tagLanguageData) {
+      numbro.setLanguage(tagOrCountryCode);
+    }
+  }
+};
+*/
+
```
danibonilha: I believe that if we are not going to use for a while maybe this could be on a diff branch, related with a i18l ticket, to avoid adding "dirty" to the code

ios/Podfile.lock
```diff
@@ -501,8 +501,8 @@ PODS:
     - React
   - RNKeychain (6.2.0):
     - React
-  - RNLanguages (3.0.2):
-    - React
+  - RNLocalize (2.1.5):
+    - React-Core
```
danibonilha, line 505: Replacing RNLanguages with RNLocalize should probably be done on a different PR as handling the comma/dot from the input with the mask, has not to do with i18n packages

danibonilha (commented):
While testing,  the `Enter Amount` button never gets enable, nor does the spend value gets updated

Screenshot

To avoid some confusion and back-and-forth, I'd like to suggest that we keep this comma/dot change, as simples as possible.
So we could have a PR with just the input with the mask change, and this is the one related to the ticket, if possible, probably just reusing the `ExchangeInput`
Its my understanding that currency-utils is working almost as expected, which means we probably just need a small change to remove the thousand separation without needing to rely on `numbro` for the time being.
But if you believe adding RNLocalize and numbro is a must have right now we can  create a new PR for each of those changes to keep things simpler.

ios/Podfile.lock
topwebtek7: RNLanguages is deprecated and replaced with RNLocalize, that's why I updated it, I think we should do it to get away deprecations. And just wonder if really need to create another pr as we all understand what it is

cardstack/src/components/Input/InputAmount.tsx
topwebtek7: just didn't want it called when inputValue changes

cardstack/src/components/Input/InputAmount.tsx
topwebtek7: Right but can't use ExchangeInput as design's quite different and type mismatch, and these were needed to make InputMask work as expected.. otherwise works very weirdly

index.js
topwebtek7: np

topwebtek7:
@lukemelia @danibonilha can you review this PR again? I did a bunch of cleanups in this PR and got rid of currency-util.ts file as used sdk's utils instead

cardstack/src/components/Input/InputAmount.tsx
danibonilha: Can we have a comment at the top of this effect saying what is expected to do, I know it's to update the input if the currency is spd, but in which scenario does this need to happen ?

cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -43,17 +46,53 @@ export const InputAmount = memo(
...
+
```
danibonilha: Don't think we need this

cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -43,17 +46,53 @@ export const InputAmount = memo(
...
 
     const onChangeText = useCallback(
-      text => {
-        setInputValue(formatNative(text, nativeCurrency));
+      formatted => {
+        let text = formatted;
+
+        if (isTouched && !text.length && !inputValue) {
+          text = '0.';
+        }
```
danibonilha: We probably don't need this piece either

cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -43,17 +46,53 @@ export const InputAmount = memo(
...
+
+    const handleChange = useCallback(() => {
+      if (isFocused && !isTouched) {
+        InteractionManager.runAfterInteractions(() => setIsTouched(true));
+      }
+    }, [isFocused, isTouched]);
+
+    const handleFocus = useCallback(() => {
+      setIsFocused(true);
+    }, []);
```
danibonilha: Don't need this also

cardstack/src/components/Input/InputAmount.tsx
```diff
@@ -104,22 +144,25 @@ export const InputAmount = memo(
...
               color={isInvalid ? 'red' : 'black'}
               fontSize={30}
               fontWeight="bold"
-              keyboardType="numeric"
-              maxLength={(inputValue || '').length + 1} // just to avoid possible flicker issue
+              keyboardType="decimal-pad"
+              mask={INPUT_MASK}
+              onBlur={handleBlur}
+              onChange={handleChange}
+              onFocus={handleFocus}
```
danibonilha: I believe we can safely remove these, tested locally and it doesn't seem to have any effects, lmk if there's any scenario it's needed

src/App.js
```diff
@@ -154,7 +154,6 @@ class App extends Component {
 
   componentWillUnmount() {
     Logger.sentry('Unmount');
-    AppState.removeEventListener('change', this.handleAppStateChange);
     this.onTokenRefreshListener?.();
```
danibonilha: Is this change intended ?

danibonilha (commented):
It's looking good, just a couple of comments to wrap up

src/App.js
topwebtek7: oh no. good catch!

topwebtek7:
> It's looking good, just a couple of comments to wrap up

@danibonilha good catch and updated as your comments.

danibonilha (approved):
Let's ship it 🚢

PR #541, opened 2021-12-02
[CS-2577] Fix crash issue when press gas price

topwebtek7:
### Description

- [x] Completes #CS-2577

### Checklist

### Screenshots

Dec-03-2021 03-22-12

### DIFF
--- a/cardstack/src/theme/colors.ts
+++ b/cardstack/src/theme/colors.ts
@@ -31,6 +31,7 @@ export const palette = {
   whiteOpacity50: '#FFFFFF80',
   grayButtonBackground: '#D1D5DB',
   buttonDisabledBackground: '#2e2d38',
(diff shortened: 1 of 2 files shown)


danibonilha (approved):
LGTM

lukemelia:
This looks like it fixes the crash. My understanding is that we don't want to support changing gas price on xdai chain. @habdelra @kdowlin can you confirm or deny?

habdelra:
@lukemelia confirm

danibonilha:
@kdowlin we probably need a diff ticket to disable this on xDai, as the fix of the crash is not related with the gasPrice itself

kdowlin:
> @kdowlin we probably need a diff ticket to disable this on xDai, as the fix of the crash is not related with the gasPrice itself

will do

PR #542, opened 2021-12-03
[CS-2687]: Use SDK to get token balances instead of eo contract

danibonilha:
### Description

The balance contract was intermittently throwing  errors probably because of the amount of tokens, this PR fixes this issue by fetching the balances from the sdk also it:

- Migrates assets service to typescript
- Clean up unused code
- Clean up dup code
- Centralize assets data structure to be modified on fallbackHelper, instead of being "parsed" in the UI only, which was proving to be difficult to find why some UI had the right behavior and other didn't.

Ps: I'll be adding tests to assets services, just put the PR up so the reviews can start 

- [x] Completes #(CS-2687)

### Screenshots

Screenshot

### DIFF
--- a/cardstack/src/helpers/__mocks__/dataMocks.ts
+++ b/cardstack/src/helpers/__mocks__/dataMocks.ts
@@ -3,7 +3,7 @@ const assets = [
   {
     asset: {
       asset_code: 'spoa',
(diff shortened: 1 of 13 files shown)


topwebtek7 (approved):
looks good to me 🚀

PR #545, opened 2021-12-03
[CS-2700]: Fix importing wallet infinite loading

danibonilha:
### Description
While initializing a wallet, walletReady was never set to true if importing an account, which is already handle a few lines above this function, this `if(!importing)` was wrong so I removed it, because now we wait for walletReady to be true in order to fetch the safes. There's no .gif to preserve the account seed phrase safe but you can test locally, by totally reseting all the the app data and importing a new account from scratch.


topwebtek7 (approved):
🚢

PR #547, opened 2021-12-03
[CS-2669][Android]: Update bundle id and google services

danibonilha:
### Description
Removes rainbow references and update with cardstack bundle id.
Also created the android app on firebase console so we could update the credentials

image

### Screenshots

Screen Shot 2021-12-03 at 15 07 47


lukemelia (approved):
👍 ⚡ 👍 ⚡ 👍 ⚡

PR #550, opened 2021-12-05
Restore BalanceCoinRow, Button, Container, InputAmount, OptionItem, SystemNotification component tests

lukemelia:
- [x] Relies on https://github.com/cardstack/cardwallet/pull/549
- [x] Relies on https://github.com/cardstack/cardwallet/pull/551


jest.config.js
```diff
@@ -22,9 +22,6 @@ module.exports = {
   testMatch: [
     '<rootDir>/cardstack/**/*.test.ts',
     '<rootDir>/cardstack/**/*.test.js',
-    // Avoid testing components for now
-    '!<rootDir>/cardstack/**/*.test.jsx',
-    '!<rootDir>/cardstack/**/*.test.tsx',
   ],
```
danibonilha, line 27: We need these components extensions without the negation in order to execute the tests, the way it's right now we are collecting coverage from the components, but not actually testing them, it's ` 107` when it should be around `127` tests
```suggestion
    '<rootDir>/cardstack/**/*.test.jsx',
    '<rootDir>/cardstack/**/*.test.tsx',
  ],
```

danibonilha (commented):
Code wise it's looking good, but we are not actually executing the components tests

jest.config.js
lukemelia: Whoops good catch. I had all the component tests listed one at a time until the end and then... removed them. 🤦‍♂️

danibonilha (approved):
🚀

PR #551, opened 2021-12-06
Update @react-native-community/async-storage to @react-native-async-storage/async-storage

lukemelia:
@react-native-community/async-storage has been renamed to @react-native-async-storage/async-storage, and the new version includes a mock suitable for use with jest.

These dev env steps are required after this change:

* remove Pods directory (ios/Pods)
* clean build directory through Xcode (in menu: Product -> Clean Build Folder)
* install dependencies again (pod install)


lukemelia:
I verified this by deleting the app from my simulator, and using Flipper to see that there was no data in async storage at boot and that it was then written to.

topwebtek7 (approved):
🚢

danibonilha (approved):

PR #554, opened 2021-12-06
[CS-2708]: Use queryWrapper on claim revenue

danibonilha:
### Description

queryWrapper has a sentry breadcrumb and exception step on error cases, since we need to add a breadcrumb for this step, I took the opportunity to already refactor claimingRevenue to use queryWrapper and be a "service" which automatically adds the sentry step.

### DIFF
--- a/cardstack/src/services/merchant/merchant-service.ts
+++ b/cardstack/src/services/merchant/merchant-service.ts
@@ -1,0 +1,58 @@
+import { getSDK } from '@cardstack/cardpay-sdk';
+import Web3 from 'web3';
+import BigNumber from 'bignumber.js';
+import { ClaimRevenueQueryParams } from './merchant-types';
+import { TokenType } from '@cardstack/types';
+import Web3Instance from '@cardstack/models/web3-instance';
+import HDProvider from '@cardstack/models/hd-provider';
+
+// Mutations
+
+export const claimMerchantRevenue = async ({
+  selectedWallet,
+  network,
+  accountAddress,
+  revenueBalances,
+  merchantSafeAddress,
+}: ClaimRevenueQueryParams) => {
+  const web3 = await Web3Instance.get({
+    selectedWallet,
+    network,
+  });
+
+  const revenuePool = await getSDK('RevenuePool', web3);
+
+  const promises = revenueBalances.map(async (token: TokenType) => {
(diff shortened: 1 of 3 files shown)


topwebtek7 (approved):
lgtm

cardstack/src/services/merchant/merchant-service.ts
```diff
@@ -0,0 +1,58 @@
...
+      new BigNumber(token.balance.amount)
+        .div(new BigNumber('2'))
+        .toPrecision(8)
+        .toString()
+    );
+
+    const gasEstimate = await revenuePool.claimGasEstimate(
+      merchantSafeAddress,
+      token.tokenAddress,
+      // divide amount by 2 for estimate since we can't estimate the full amount
```
lukemelia: There is a disconnect between this comment location and where the division by 2 is happening.

cardstack/src/services/merchant/merchant-service.ts
```diff
@@ -0,0 +1,58 @@
...
+    network,
+  });
+
+  const revenuePool = await getSDK('RevenuePool', web3);
+
+  const promises = revenueBalances.map(async (token: TokenType) => {
+    const claimEstimateAmount = Web3.utils.toWei(
+      new BigNumber(token.balance.amount)
+        .div(new BigNumber('2'))
+        .toPrecision(8)
```
lukemelia: What is the significance of 8 here?

lukemelia (approved):
Good to go after reviewing my comments and addressing if you see fit.

cardstack/src/services/merchant/merchant-service.ts
danibonilha: Great question, I didn't write any piece of this, was just moving from one place to the other, so will need to figure what it means

danibonilha: @lukemelia great catch on this one, [`toPrecision`](https://mikemcl.github.io/bignumber.js/#toP) here may introduce some errors while dealing with huge numbers, since it rounds to the most significant digits having exponential notation which is not well accepted on `Web3.utils.toWei` throwing `[Error: while converting number to string, invalid number value '3.7500000e+11', should be a number matching (^-?[0-9.]+).]` for example.

PR #555, opened 2021-12-07
Break up shim.js

lukemelia:
This file was a mish-mash of different responsibilities and an exported const that was being imported by two screens was causing a lot of surprising behavior when debugging certain modules and tests.

In the future, we could consider moving this to App.js

### DIFF
--- a/shim.js
+++ b/shim.js
@@ -1,186 +1,20 @@
 import 'react-native-get-random-values';
 import '@ethersproject/shims';
-import AsyncStorage from '@react-native-community/async-storage';
 import { enableES5 } from 'immer';
-// eslint-disable-next-line import/default
-import ReactNative from 'react-native';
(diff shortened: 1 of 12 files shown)


topwebtek7:
As far as I know, shim.js is auto generated by running `yarn nodeify`..  anyway looks good if it's helpful for tests

lukemelia:
Ah, seems like it was originally generated with the content in https://github.com/tradle/rn-nodeify/blob/master/shim.js and collected all sorts of stuff after that.

danibonilha (commented):
LGTM, tested locally, everything seems to be working fine

danibonilha:
@lukemelia   I'm sorry, I had some internet issues and wrongfully clicked on `close pull request` 😓  once the page restarted

danibonilha (approved):

PR #556, opened 2021-12-07
[CS-2641] Use sdk's convertToSpend util func to be consistent with Dapp

topwebtek7:
### Description

Used sdk's convertToSpend util function to be consistent with dapp, and get rid of old functions as not using anymore.

- [x] Completes #CS-2641

### Checklist

### Screenshots

Screen Shot 2021-12-07 at 1 09 05 PM
Screen Shot 2021-12-07 at 1 08 45 PM


danibonilha (approved):
LGTM

PR #557, opened 2021-12-07
[CS-2722]: Disable FoR on Android, split AssetList components

danibonilha:
### Description

The task was simple, just to add a conditional based on the platform, but as I was going thorough the `AssetList's` code it was pretty confusing to add another ternary where we already had a couple, so I decided to split the code into simpler components to have better readability and reusability.
 - Created `PinnedHiddenSectionMenu` (The three dots and actions)
 - Created `AssetSectionHeader` to handle all the header's logic
 - Created, types, strings and components to `AssetList`
 - Renamed and improved `AssetFooter` to `PinHideOptionsFooter` which gives more context 
 - Cleaned up `AssetList`
 - Created layoutOpacityAnimation to share across components 
 - Enabled layoutAnimations on Android
 - Disabled fiat on ramp on Android
 - Improved sections typing

### Screenshots

Screenshot 

Screenshot


cardstack/src/components/AssetList/AssetList.tsx
```diff
@@ -254,76 +288,88 @@ export const AssetList = (props: AssetListProps) => {
...
+      </Container>
+      {isEmptyPrepaidCard && (
+        <Container marginHorizontal={4} alignItems="center">
+          <ListEmptyComponent
+            text={strings.emptyCardMessage}
+            width="100%"
+            hasRoundBox
+            textColor="blueText"
+          />
+          {Device.isIOS && (
```
lukemelia: It seems like it would be better to use a semantic flag calculated in one place like `supportsFiatOnRamp` rather than sprinkling platform checks around.

lukemelia (changes requested):

cardstack/src/components/AssetList/AssetList.tsx
danibonilha: Agreed, since the support is based on the platform I thought it makes sense to add to the Device object, lmk if that's ok

lukemelia (approved):
🚀  Looks good!

PR #558, opened 2021-12-08
Fix an issue handling transaction signing requests that use EIP712 with stringified structured data

lukemelia:
- Manifested itself in a CollabLand dapp transaction
- Also refactored useTransactionConfirmation to reduce the number of hooks in use
- Adjusted types to reflect real world experience


cardstack/src/hooks/transaction-confirmation/utils/parse-message-request-json.ts
```diff
@@ -0,0 +1,9 @@
+export const parseMessageRequestJson = (displayDetails: any) => {
```
danibonilha: We could type displayDetails since we know the return structure of `getTransactionDisplayDetails`

cardstack/src/hooks/transaction-confirmation/utils/extract-payload-params.ts
```diff
@@ -0,0 +1,31 @@
...
+  payload: EIP712TypedDataPayload
+): TypedData => {
+  const { params } = payload;
+
+  if (typeof params[1] === 'string') {
+    return JSON.parse(params[1]);
+  }
+
+  return params[1] || payloadParamsFallback;
+};
```
danibonilha: nit: the utils functions for txs are really small, maybe we could combine all in the same file

cardstack/src/hooks/transaction-confirmation/use-transaction-confirmation.ts
```diff
@@ -1,37 +1,91 @@
...
+          network,
+          nativeCurrency
+        ).getDecodedData();
+
+        setData(result);
+      } catch (error) {
+        logger.error(`Decoding data error - ${error}`);
+      }
+
+      setLoading(false);
```
danibonilha, line 68: FYI: we have an `useWorker` hook, that wraps any promise into a try/catch and handles the loading state, I think just the error logger is not implemented yet.

cardstack/src/hooks/transaction-confirmation/use-transaction-confirmation.ts
```diff
@@ -1,37 +1,91 @@
...
+    payload.params
+  );
+
+  const payloadRef = useRef<any>();
+  const typedDataRef = useRef<TypedData>();
+
+  if (payloadRef.current !== payload) {
+    payloadRef.current = payload;
+    typedDataRef.current = extractPayloadParams(payload);
+  }
```
danibonilha: For this specific case we could take advantage a diff hook:

```ts
const typedData = useMemo(() =>  extractPayloadParams(payload), [payload])
```
which is basically syntax sugar for what you tried to achieve, [this article](https://thoughtspile.github.io/2021/04/05/useref-usememo/) has a nice explanation

cardstack/src/hooks/transaction-confirmation/use-transaction-confirmation.ts
```diff
@@ -1,37 +1,91 @@
...
+    setDecodedData();
+  }, [message, domain, primaryType, network, nativeCurrency]);
 
   return {
     data,
     loading,
-    onConfirm,
-    onCancel,
+    onConfirm: useConfirmTransaction(),
+    onCancel: useCancelTransaction(),
```
danibonilha: super nit: I prefer to have most of the hooks at the top of the hook/component, but it's a totally personal opinion

danibonilha (approved):
Looks good, tested prepaidCard issuance, sushi swap and collab, just small comments

cardstack/src/hooks/transaction-confirmation/utils/parse-message-request-json.ts
lukemelia: 👍  converted the relevant bits to Typescript

cardstack/src/hooks/transaction-confirmation/utils/extract-payload-params.ts
lukemelia: ✅

cardstack/src/hooks/transaction-confirmation/use-transaction-confirmation.ts
lukemelia: ✅

cardstack/src/hooks/transaction-confirmation/use-transaction-confirmation.ts
lukemelia: Applied useMemo ✅

cardstack/src/hooks/transaction-confirmation/use-transaction-confirmation.ts
lukemelia: I took a look and will keep this in mind for the future.

danibonilha (approved):
Nice refactor! 🚀

PR #560, opened 2021-12-09
Update README to include how devs can run the app on Android

lukemelia:


danibonilha (approved):
Thanks! I forgot about the google_sync

PR #561, opened 2021-12-10
[CS-2611] Register with hub for push notifications

topwebtek7:
### Description

Refactored firebase model that stores fcmToken to asyncStorage by network type and check if already registered when account loads.
Also implemented hub authentication flow, and registered FCM token to hub api. I've tested and it works fine on Sokol, will confirm on xDai after hub deployed.
Btw the prompt message shows only when authenticate with passcode, don't show up when use biometric authenticate.

- [CS-2611] Completes #(Linear ticket)

### Checklist

- [CS-2611] All UI changes have been tested on a small device

### Screenshots

Dec-10-2021 22-31-06

Screen Shot 2021-12-09 at 9 46 44 PM


cardstack/src/models/firebase.ts
```diff
@@ -0,0 +1,198 @@
...
+};
+
+const getPermissionStatus = (): Promise<FirebaseMessagingTypes.AuthorizationStatus> =>
+  messaging().hasPermission();
+
+export const getFCMToken = async (): Promise<FCMTokenStorageType> => {
+  const {
+    data: fcmToken,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    storageVersion,
```
danibonilha: If we are not using, why is it being destructed and the rule is disabled ?

cardstack/src/models/firebase.ts
```diff
@@ -0,0 +1,198 @@
...
+    ...addressesByNetwork
+  } = (await getLocal(DEVICE_FCM_TOKEN_KEY)) as any;
+
+  if (!fcmToken) {
+    return { fcmToken: null };
+  }
+
+  return { fcmToken, addressesByNetwork };
+};
```
danibonilha, line 39: I find this data structure a bit odd, would it be possible to store everything we need within data ? Having `fcmToken`, `addressesByNetwork` in the same structure, so we won't need be remapping those values

cardstack/src/models/firebase.ts
```diff
@@ -0,0 +1,198 @@
...
+          {
+            onPress: async () => {
+              resolve(true);
+            },
+            style: 'cancel',
+            text: 'Dismiss',
+          },
+        ],
+        message: lang.t('wallet.push_notifications.please_enable_body'),
+        title: lang.t('wallet.push_notifications.please_enable_title'),
```
danibonilha, line 174: Should we replace this with local strings?

cardstack/src/models/firebase.ts
```diff
@@ -0,0 +1,198 @@
...
+          });
+        } else {
+          saveLocal(DEVICE_FCM_TOKEN_KEY, {
+            data: newFcmToken,
+            ...addressesByNetwork,
+            [network]: [
+              ...(addressesByNetwork?.[network] || []),
+              walletAddress,
+            ].filter(
+              (address, index, self) => self.indexOf(address) === index // remove duplicates
```
danibonilha: I don't get why we need to store the address right now and use [network] as a object key, could you explain to me ? If the main goal is just to store a token based on the network wouldn't make sense to create a `DEVICE_FCM_TOKEN_KEY+network` for example ?

cardstack/src/models/web3-provider.ts
```diff
@@ -10,8 +10,9 @@ let provider: WebsocketProvider | null = null;
 
 const Web3WsProvider = {
   get: async (network?: Network) => {
-    if (provider === null || network || !provider?.connected) {
-      const currentNetwork = await getNetwork();
+    if (provider === null || !provider?.connected) {
+      const currentNetwork = network || (await getNetwork());
+
```
danibonilha: Once we reset the network config, the `etherWeb3SetHttpProvider` is set with the new network, that's why we have this `|| network` here, in case we receive a network as parameter, it doesn't matter if we have a provider connected or not, it just sets a new provider with the new network, otherwise it doesn't even make sense to receive network as a parameter, if you can make sure that this does not breaks anything, the network param is unnecessary

cardstack/src/services/hub-service.ts
```diff
@@ -20,78 +34,79 @@ const axiosConfig = (authToken: string) => {
...
+  if (selectedWallet) {
+    try {
+      const hdProvider = await HDProvider.get({
+        walletId: selectedWallet.wallet.id,
+        network,
+        seedPhrase,
+        keychainAcessAskPrompt: HUBAUTH_PROMPT_MESSAGE,
+      });
+
+      const web3 = new Web3(hdProvider);
```
danibonilha: using 
```ts
const web3 = Web3Instance.get({
        walletId: selectedWallet.wallet.id,
        network,
        seedPhrase,
        keychainAcessAskPrompt: HUBAUTH_PROMPT_MESSAGE,
      })
```

   wraps this

cardstack/src/services/hub-service.ts
```diff
@@ -20,78 +34,79 @@ const axiosConfig = (authToken: string) => {
...
+      const authAPI = await getSDK('HubAuth', web3, hubURL);
+      // load wallet address when not provided as an argument(this keychain access does not require passcode/biometric auth)
+      const address = walletAddress || (await loadAddress()) || '';
+      const authToken = await authAPI.authenticate({ from: address });
+      return authToken;
```
danibonilha, line 60: Should we cache the hubAuth token somehow so we just make this call once ? Or is it necessary to call all the time ?

cardstack/src/services/hub-service.ts
```diff
@@ -20,78 +34,79 @@ const axiosConfig = (authToken: string) => {
...
+    } catch (e) {
+      logger.sentry('Hub authenticate failed', e);
+
+      return null;
+    }
+  }
+
+  return null;
+};
```
danibonilha, line 69: We are not reseting the HDProvider after the request, which might trigger the PollingBlocker error

src/hooks/useInitializeWallet.js
```diff
@@ -43,10 +44,10 @@ export default function useInitializeWallet() {
       checkedWallet = null,
     } = {}) => {
       try {
-        logger.sentry('Start wallet setup');
+        logger.log('Start wallet setup');
 
         await resetAccountState();
-        logger.sentry('resetAccountState ran ok');
+        logger.log('resetAccountState ran ok');
```
danibonilha: do we really want to remove all these breadcrumbs to be just logs right now ?

danibonilha (commented):
Doesn't work for me, I keep getting
 `error fcm token - cannot register fcm token! [TypeError: null is not an object (evaluating '_ref.data')]`

cardstack/src/models/firebase.ts
topwebtek7: just didn't want to have storageVersion key in `addressesByNetwork`, anyway updated that structure a bit

cardstack/src/models/firebase.ts
topwebtek7: This context looks ok, so didn't change yet

cardstack/src/models/firebase.ts
topwebtek7: yeah, the main goal is to check if token is registered to the hub through this value stored in asyncStorage. we will need to register token to the hub based on Account and Network but with same fcmToken string, so I stored account addresses that already registered to the hub on the specific network, and did reset when the token refreshes or changed. so I believe this structure is most efficient way to verify if already registered to hub.

cardstack/src/models/web3-provider.ts
topwebtek7: all works fine to me with this change.. but if there was that reason with this, I will revert this change

cardstack/src/services/hub-service.ts
topwebtek7: yes I'm doing that in next pr along with unregistering token on hub

cardstack/src/services/hub-service.ts
topwebtek7: I know, but didn't use Web3Instance intentionally as it caused issues when you try hub auth and loading assets almost simultaneously after  wallet initialized at app startup, so kind of created new web3 instance for just grabbing hub auth token.

src/hooks/useInitializeWallet.js
topwebtek7: basically I think we have too much logs in dev mode now and if it's ok looking at sentry log without these, gonna remove/cleanup them in next prs

cardstack/src/services/hub-service.ts
topwebtek7: hmm.. ok, basically I added it initially and then removed after that cause didn't see issues for a while without reset.. and just wanted do not have biometric authenticate again when need to use hdwallet-provider instance next time.
also if we reset after use once, do we really need this global `provider` variable and  `get`, `reset` behavior in hd-provider.ts file?

topwebtek7:
> Doesn't work for me, I keep getting `error fcm token - cannot register fcm token! [TypeError: null is not an object (evaluating '_ref.data')]`

pushed a couple fixes, and should work fine

cardstack/src/services/hub-service.ts
danibonilha: Makes sense, could you add a comment, something like, needs it's own web3 instance or so, just so we don't forget eventually and replace.

cardstack/src/services/hub-service.ts
danibonilha: We can observe the behavior without reseting this one, keeping as a single provider and reseting was the only way to decrease the amount of pollingblocker error we were having, especially coming from foreground/background. Before we had multiple diff providers, having a singleton made it more organized and less prone to errors

src/hooks/useInitializeWallet.js
danibonilha: I think if your motivation is just because there are too many logs on DEV mode, we can create a flag on the 'logger' to enable or disable those logs, and use `console.log` for specific debugging issues, instead of removing all the breadcrumbs that appears on sentry

danibonilha (commented):
I'm still getting `error fcm token - cannot register fcm token! [TypeError: undefined is not an object (evaluating '_ref$data.fcmToken')]` and also unable to use wallet connect: 

Screen Shot 2021-12-09 at 9 46 44 PM

topwebtek7:
> I'm still getting `error fcm token - cannot register fcm token! [TypeError: undefined is not an object (evaluating '_ref$data.fcmToken')]` and also unable to use wallet connect:


Resolved this bug and should be ok now

PR #562, opened 2021-12-10
[CS-2780]: Fix blank text input on android

danibonilha:
### Description

Fixes blank text input on android, on some android devices, the default text is not black, so we need to explicitly set the input text color 

### Screenshots

Screenshot


topwebtek7 (approved):
🚢

PR #563, opened 2021-12-10
[CS-2668]: Fix android crash, improve perf on scanner

danibonilha:
### Description

The useEffect was only enabled on Android, and it seems it was a workaround on a previous version of RNCamera, but since RNCamera was updated, we don't seem to need it anymore, as it was trying to call a method when the ref didn’t existed, did some cleanup to have the same behavior on both platforms and replace the default emulator image with a better one 

### Screenshots

https://user-images.githubusercontent.com/20520102/145598349-d95ba666-0c6f-4db9-93b3-3191789c634d.mp4


topwebtek7 (commented):
@danibonilha This pr looks good to me. I know it's not related to this ticket but I think react-native-camera version upgrade would help perf improvement, we're way behind of it and can see some android issue fixes - https://github.com/react-native-camera/react-native-camera/releases

danibonilha:
@topwebtek7 
Right, thanks for the suggestion, but RNCamera is about to be deprecated, it's looking for maintainers and they suggest to migrate to another package, I'm evaluating migrating to Expo-Camera, which we can add to the backlog and handle individually since the main functionally to read QRCode is working good 😊

topwebtek7 (approved):
:shipit:

PR #566, opened 2021-12-13
[CS-2794]: Fix iOS crash on loading

danibonilha:
### Description

We are using `useBlockBackButton` on LoadingOverlay to block the user from exiting  while loading, using the back button on Android, but it's a abstraction of a native module built only for Android, and it was crashing on iOS bc it doesn't exist, I migrated to typescript and typed correctly to avoid this mistake.
 
### Screenshots

https://user-images.githubusercontent.com/20520102/145821799-624c5621-de84-4b57-972b-0dc6dfecb89f.mp4


topwebtek7 (approved):
looks good to me. I had this crash issue and confirm this pr resolves it :shipit:

PR #567, opened 2021-12-13
[CS-2612] Unregister with hub from push notification, fix delete account behavior

topwebtek7:
### Description
When delete account, fetched hub auth token based on deleting account's address(cannot use hub token grabbed on current wallet if it's deleting different account) and also added overlay loading behavior when deleting

- [ ] Merge after #561 is merged

- [x] Completes #CS-2612

### Checklist

### Screenshots

Dec-13-2021 22-08-55


cardstack/src/hooks/prepaid-card/useAuthToken.ts
```diff
@@ -1,30 +1,30 @@
...
 
   const network = useRainbowSelector(
     state => state.settings.network
   ) as Network;
 
   const { callback: getAuthToken, error, isLoading } = useWorker(async () => {
-    const web3 = await Web3Instance.get({ selectedWallet, network });
-    const authAPI = await getSDK('HubAuth', web3, hubURL);
-    setAuthToken(await authAPI.authenticate({ from: accountAddress }));
+    const hubURLString = hubURL || getHubUrl(network);
```
danibonilha: it defaults to `getHubUrl(network)` so we don't need to received `hubUrl` as a param, since the only place `hubURL` is being passed on useBuyPrepaidCard, it does the same exact thing, calls `getHubUrl(network)`

cardstack/src/hooks/prepaid-card/useAuthToken.ts
```diff
@@ -1,30 +1,30 @@
...
 import { useRainbowSelector } from '@rainbow-me/redux/hooks';
 import { Network } from '@rainbow-me/networkTypes';
 import { useWorker } from '@cardstack/utils';
 import logger from 'logger';
 import { useAccountSettings } from '@rainbow-me/hooks';
-import HDProvider from '@cardstack/models/hd-provider';
+import { getHubAuthToken, getHubUrl } from '@cardstack/services/hub-service';
 
-export const useAuthToken = (hubURL: string) => {
+export const useAuthToken = (hubURL?: string, seedPhrase?: string) => {
```
danibonilha: We never pass `seedPhrase` for `useAuthToken`, since the hdProvider is loading the seedPhrase already, so these two params seems useless

cardstack/src/hooks/prepaid-card/useAuthToken.ts
```diff
@@ -1,30 +1,30 @@
...
   ) as Network;
 
   const { callback: getAuthToken, error, isLoading } = useWorker(async () => {
-    const web3 = await Web3Instance.get({ selectedWallet, network });
-    const authAPI = await getSDK('HubAuth', web3, hubURL);
-    setAuthToken(await authAPI.authenticate({ from: accountAddress }));
+    const hubURLString = hubURL || getHubUrl(network);
 
-    // resets signed provider and web3 instance to kill poller
-    await HDProvider.reset();
```
danibonilha, line 27: it seems we do need to reset the provider, testing this branch, once I went to background and returned, I got the pollingblocker error again:

cardstack/src/models/firebase.ts
```diff
@@ -0,0 +1,201 @@
...
+};
+
+// check if token's stored by confirming addresses includes wallet address
+export const isFCMTokenStored = async (
+  walletAddress: string
+): Promise<{
+  isTokenStored: boolean;
+  addressesByNetwork?: Record<Network, string | string[]>;
+  fcmToken: string | null;
+}> => {
```
danibonilha: nit: it could be it's own interface

cardstack/src/models/firebase.ts
```diff
@@ -0,0 +1,201 @@
...
+    } else {
+      logger.sentry('FCM token already registered for this account!');
+    }
+  } catch (error) {
+    logger.sentry('error fcm token - cannot register fcm token!', error);
+  }
+};
+
+export const requestPermission = () => {
+  return new Promise((resolve, reject) => {
```
danibonilha: nit: could omit return

cardstack/src/services/hub-service.ts
```diff
@@ -20,78 +34,121 @@ const axiosConfig = (authToken: string) => {
...
+    const authToken = await getHubAuthToken(
+      hubURL,
+      network,
+      walletAddress,
+      seedPhrase
+    );
+
+    if (!authToken) {
+      return { success: false };
+    }
```
danibonilha: This piece is the same for register and unregister, maybe we could have a wrapper to reuse this

cardstack/src/services/hub-service.ts
```diff
@@ -20,78 +34,121 @@ const axiosConfig = (authToken: string) => {
...
+
+    const results = await axios.delete(
+      `${hubURL}/api/push-notification-registrations/${fcmToken}`,
```
danibonilha, line 142: the path is the same, `/api/push-notification-registrations/` it could be a const, in case the api changes we just need to update in one place

src/components/change-wallet/WalletList.js
```diff
@@ -94,7 +94,7 @@ export default function WalletList({
     sortedKeys.forEach(key => {
       const wallet = allWallets[key];
       const filteredAccounts = wallet.addresses.filter(
-        account => account.visible
+        account => !!account.visible
       );
```
danibonilha, line 98: just for context, what does this fix ?

src/screens/ChangeWalletSheet.js
```diff
@@ -148,15 +151,27 @@ export default function ChangeWalletSheet() {
       const visibleAddresses = newWallets[walletId].addresses.filter(
         account => account.visible
       );
+      setIsWalletLoading(WalletLoadingStates.DELETING_WALLET);
```
danibonilha, line 154: 👏

src/screens/ChangeWalletSheet.js
```diff
@@ -148,15 +151,27 @@ export default function ChangeWalletSheet() {
       const visibleAddresses = newWallets[walletId].addresses.filter(
         account => account.visible
```
danibonilha, line 152: in case casting to boolean is needed, would it be needed here too right ?

src/screens/ChangeWalletSheet.js
```diff
@@ -148,15 +151,27 @@ export default function ChangeWalletSheet() {
...
       );
+      setIsWalletLoading(WalletLoadingStates.DELETING_WALLET);
+
+      const network = await getNetwork();
+      const { fcmToken, addressesByNetwork } = await getFCMToken();
+      if (
+        addressesByNetwork[network] &&
+        addressesByNetwork[network].includes(address)
+      ) {
+        const unreigsterResponse = await unregisterFcmToken(fcmToken, address);
```
danibonilha: typo, unregisterResponse

danibonilha (commented):
Besides the pollingBlocker error, I also got `Unable to getHdSignedProvider [Error: Could not create addresses from your mnemonic or private key(s). Please check that your inputs are correct.]`  but not sure how to repro though, (maybe bc it tried to face-id and it failed!?) and after this error I get stuck on the deleting account state.
I'm also concerned about how many times we are asking for permissions, we may have same rerender problem, bc it seems to ask 5 times, another weird behavior might be using face-id to unregister, wouldn't make sense once we fetch the hubAuthToken, we store it in the localStorage, so we don't need to face-id again ?

cardstack/src/services/hub-service.ts
topwebtek7: it's just 2 hub api calls yet and we can add wrapper if have more such hub apis calls

cardstack/src/services/hub-service.ts
topwebtek7: just wanted to be consistent with other hub api call functions in this file

cardstack/src/hooks/prepaid-card/useAuthToken.ts
topwebtek7: yea but I want to keep `seedPhrase` param as we have seedPhrase value in some places and if have that, don't need faceId pass

cardstack/src/hooks/prepaid-card/useAuthToken.ts
topwebtek7: yep just wanted to skip duplicated operations but ok `getHubUrl`'s simple operation so will remove hubUrl param

cardstack/src/hooks/prepaid-card/useAuthToken.ts
topwebtek7: hmm weird.. ok let's add resetting provider for now

topwebtek7:
> Besides the pollingBlocker error, I also got `Unable to getHdSignedProvider [Error: Could not create addresses from your mnemonic or private key(s). Please check that your inputs are correct.]` but not sure how to repro though, (maybe bc it tried to face-id and it failed!?) and after this error I get stuck on the deleting account state. I'm also concerned about how many times we are asking for permissions, we may have same rerender problem, bc it seems to ask 5 times, another weird behavior might be using face-id to unregister, wouldn't make sense once we fetch the hubAuthToken, we store it in the localStorage, so we don't need to face-id again ?

Yes if you fail face-id it logs that error but does not do any action and looks fine to me. not sure why you see faceId multiple times.. that should not be related with re-render problem. let me know where you see faceId multiple times if you can replicate it again.
also I think it's ok showing face-id to ask confirming deleting accounts. we can't store hubAuthToken for each accounts(hubAuthToken is different for each account, and when you try to delete the other account, need to get authToken with that account's seedPhrase)

src/components/change-wallet/WalletList.js
topwebtek7: just cleaning up, it's not a specific fix

topwebtek7:
@danibonilha added resetting provider as discussed. please let me know if you see any others

danibonilha (approved):
Approving to move forward, was not able to get a notification but it seems main issue is with the hub. 

I think you can close the registering one, and merge this at once

PR #568, opened 2021-12-13
[CS-2777] Fix currency bug when loading from deeplink

topwebtek7:
### Description

The issue was resolved setting nativeCurrency value with deeplink's currency value as default cause didn't set at initial loading app from camera app

- [x] Completes #CS-2777

### Checklist

### Screenshots

Dec-14-2021 00-12-05

### DIFF
--- a/cardstack/src/screens/PayMerchant/usePayMerchant.ts
+++ b/cardstack/src/screens/PayMerchant/usePayMerchant.ts
@@ -235,15 +235,14 @@ export const usePayMerchant = () => {
     }
   }, [initialAmount]);
 
-  const spendAmount = useMemo(
-    () =>
-      convertToSpend(
-        convertStringToNumber(inputValue || '0'),
-        nativeCurrency,
-        currencyConversionRates[nativeCurrency]
-      ),
-    [currencyConversionRates, inputValue, nativeCurrency]
-  );
+  const spendAmount = useMemo(() => {
+    const currency = nativeCurrency || initialCurrency || 'SPD';
+    return convertToSpend(
+      convertStringToNumber(inputValue || '0'),
+      currency,
+      currencyConversionRates[currency]
+    );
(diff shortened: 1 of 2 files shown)


cardstack/src/screens/PayMerchant/usePayMerchant.ts
```diff
@@ -177,7 +177,11 @@ export const usePayMerchant = () => {
     qrCodeNetwork,
   } = data;
 
-  const { paymentChangeCurrency, currency: nativeCurrency } = usePayment();
+  const {
+    paymentChangeCurrency,
+    currency: nativeCurrency = initialCurrency || NativeCurrency.SPD,
```
danibonilha: Maybe we could extract `initialCurrency || NativeCurrency.SPD` to a variable to make it easier to read, or even make the `initialCurrency` default to `NativeCurrency.SPD` first

danibonilha (approved):

cardstack/src/screens/PayMerchant/usePayMerchant.ts
```diff
@@ -177,7 +177,11 @@ export const usePayMerchant = () => {
...
+  } = usePayment();
+
```
lukemelia:
```suggestion
  const {
    paymentChangeCurrency,
    currency
  } = usePayment();
  const nativeCurrency = currency || initialCurrency || NativeCurrency.SPD;
```

Personally, I find this a lot easier to read than the destructuring assignment with new variable name plus default values. If you and @danibonilha prefer the original though, I can get used to it.

lukemelia (approved):
1 comment for consideration. Otherwise OK.

cardstack/src/screens/PayMerchant/usePayMerchant.ts
lukemelia: heh, didn't see this before I left my suggestion :D

PR #569, opened 2021-12-13
[CS-2796]: Fix sending EOA tokens on Android

danibonilha:
### Description

Giving the same behavior across devices to have the fee after the send button, as the screen adapts depending on device size, fixed keyboard sentry error as the flag was not even being used, fixed circular dependency issue on showAction sheet and fix button press we have time to show.

### Screenshots

Screenshot

Screenshot


lukemelia (approved):
Tried this out. Worked correctly but it refreshed my entire assets screen afterwards and left me scrolled to the top.

PR #570, opened 2021-12-13
[CS-2799]: Hide amounts for Android

danibonilha:
### Description

[CS-2799]: Hide amounts for Android

### Screenshots

Screenshot


lukemelia (approved):
IDK why this is different on iOS and Android, but assuming there is a good reason, this change looks fine.

PR #571, opened 2021-12-13
[CS-2792]: Add header info on tabs

danibonilha:
### Description

Ideally we should have one Header component that we could share over the tabs, but right now there's one wrapper and each button has a diff behavior, so for now I'm just adding a text component on each tab instead of rebuilding a shareable header, because we'll do it on the nav redesign.

### Screenshots

Screenshot

Screenshot


lukemelia (approved):
👍

PR #572, opened 2021-12-13
[CS-2790]: Fix android prepaidCardModal overflow

danibonilha:
### Description

The iOS bar is handled internally by the native Navigation style, and for Android we need to set the paddingTop based on the bar height, which is given by `StatusBar.currentHeight`  on Android and it's null on iOS

### Screenshots

Screenshot


lukemelia (approved):
Easy nuff

PR #573, opened 2021-12-13
[CS-2797]: Fix android send token address UI

danibonilha:
### Description

Fix android send token address UI 

### Screenshots

Screenshot


lukemelia (approved):
Removing conditionals FTW!

PR #576, opened 2021-12-14
[CS-2800] Fix slow issue with opening show qr request

topwebtek7:
### Description
On Android, navigating to QRCode screen was so slow cause of rendering qrcode canvas, so navigate to the screen first and started rendering qrcode with rendering indicator

- [x] Completes #CS-2800

### Checklist

### Screenshots

Dec-15-2021 03-40-16
Dec-15-2021 03-41-45

### DIFF
--- a/cardstack/src/screens/ShowQRCodeModal.tsx
+++ b/cardstack/src/screens/ShowQRCodeModal.tsx
@@ -1,6 +1,7 @@
 import { useRoute } from '@react-navigation/core';
-import { useNavigation } from '@react-navigation/native';
-import React, { useCallback } from 'react';
+import { useNavigation, useFocusEffect } from '@react-navigation/native';
+import React, { useCallback, useState } from 'react';
+import { ActivityIndicator } from 'react-native';
 import CardstackColorLogo from '../assets/cardstackColorLogo.png';
 import { ContactAvatar } from '@rainbow-me/components/contacts';
 import {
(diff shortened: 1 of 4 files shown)


cardstack/src/screens/ShowQRCodeModal.tsx
```diff
@@ -33,14 +35,25 @@ export const AmountQRCode = ({
...
+  const [isFocus, setFocus] = useState(false);
+  const [isQRCodeRendering, setQRCodeState] = useState(true);
+  const onQRCodeRendered = useCallback(() => setQRCodeState(false), []);
+  useFocusEffect(
+    useCallback(() => {
+      setFocus(true);
+
+      return () => setFocus(false);
+    }, [])
+  );
```
danibonilha, line 55: Why do we need to check focus ? If we are saying, once it's on focus, show `<QRCode` it seems we can just render the `<QRCode` directly without using any onFocus flag, the behavior should be the same, I did a test here, and it seems fine, lmk if there are any quirks.

danibonilha (commented):
It's looking good, just not sure about the onFocus stuff. There are still slowness on the button press, but I think it's a different global issue with the android side, which we can handle later.

cardstack/src/screens/ShowQRCodeModal.tsx
topwebtek7: @danibonilha  To me on Android, this `<QRCode` makes transition to qrcode screen so slow like user's able to press Show QR Code button several times meanwhile, but after added this rendering after focused, it's resolved and qrcode screen shows instantly for me, so user's not able to press button multiple times. I think slowness with button is related with rendering canvas on Android, started render after navigated seems resolve it.
and it's not much difference on iOS.

danibonilha (approved):

PR #577, opened 2021-12-14
[CS-2670]: Workaround for Sheets covering statusbar on android

danibonilha:
### Description

We have the `SlackSheet` and some native navigation behaviors that works great for iOS but not so much for Android, the goal is to have both platforms using the same component and navigation stack, which will require a good amount of work, right now to avoid breaking what works good on iOS but add some improvement to android I created this workaround which is not great, but it's useful, as we can gradually start to migrate every component that uses `SlackSheet` to the new `Sheet` component while having Android showing at least the status bar.
Also, as requested by @kdowlin it reverts the change to just ask for backup if user has assets 

### Screenshots

Screenshot 

Screenshot 

Screenshot

### DIFF
--- a/src/components/sheet/SlackSheet.js
+++ b/src/components/sheet/SlackSheet.js
@@ -3,6 +3,7 @@ import React, { Fragment, useEffect, useMemo, useRef } from 'react';
 import {
   KeyboardAvoidingView,
   Pressable,
+  StatusBar,
   StyleSheet,
   TouchableWithoutFeedback,
 } from 'react-native';
@@ -116,6 +117,13 @@ export default function SlackSheet({
       )
     : props => <Container backgroundColor={bg} flex={1} {...props} />;
 
+  const oldTopCondition =
+    deferredHeight || ios
+      ? 0
+      : contentHeight && additionalTopPadding
+      ? deviceHeight - contentHeight
+      : 0;
+
   return (
     <Fragment>
       {android ? (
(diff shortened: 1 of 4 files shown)


src/components/sheet/SlackSheet.js
```diff
@@ -116,25 +117,29 @@ export default function SlackSheet({
       )
     : props => <Container backgroundColor={bg} flex={1} {...props} />;
 
+  const oldTopCondition =
+    deferredHeight || ios
+      ? 0
+      : contentHeight && additionalTopPadding
+      ? deviceHeight - contentHeight
+      : 0;
+
```
danibonilha, line 126: Not worrying to much about this or this file at all, because once we migrate all the sheets we can remove this whole file, so just wanted to keep the previous behavior without spending too much time figuring why it exists

src/screens/BackupSheet.js
```diff
@@ -185,7 +186,29 @@ export default function BackupSheet() {
...
 }
+
+// Temp workaround until nav redesign
+const PlataformSheet = ({ children, step }) => {
+  const isFullScreenStep =
+    step === WalletBackupStepTypes.manual ||
+    step === WalletBackupStepTypes.cloud;
+
+  const bgColor =
+    step === WalletBackupStepTypes.cloud ? 'black' : 'transparent';
```
danibonilha, line 201: For cloudStep we need to set the color, bc the settings modal gets all messed up, I guess it's the keyboard faults, but with the redesign we will keep rid of that and can remove this

topwebtek7 (approved):
tested on android and looks good to me! just still experiencing keyboard flickering in payment request sheet when type amount, I think we can look at it in separate ticket

danibonilha:
> tested on android and looks good to me! just still experiencing keyboard flickering in payment request sheet when type amount, I think we can look at it in separate ticket

yeah the keyboard is a issue of it's own, we need to handle that eventually

src/screens/BackupSheet.js
```diff
@@ -185,7 +186,29 @@ export default function BackupSheet() {
...
     <Container flex={1} testID="backup-sheet">
       <StatusBar barStyle="light-content" />
-      <SlackSheet flex={1}>{renderStep()}</SlackSheet>
+      <PlataformSheet step={step}>{renderStep()}</PlataformSheet>
     </Container>
   );
 }
+
+// Temp workaround until nav redesign
+const PlataformSheet = ({ children, step }) => {
```
lukemelia, line 195: There's only one "a" in platform in english

danibonilha: Portuguese messing me up 😂  it's `Plataforma` here 😅

PR #578, opened 2021-12-15
[CS-2804] Fix claiming behavior on Android without faceid

topwebtek7:
### Description

On Android device does not support Biometry auth, used in-app pin code to crypt privateKey/seedPhrase(which was needed to access revenue pool), so used in-app pin code behavior 

### Checklist

### Screenshots

Untitled2

Untitled1

### DIFF
--- a/cardstack/src/screens/MerchantScreen.tsx
+++ b/cardstack/src/screens/MerchantScreen.tsx
@@ -427,9 +427,7 @@ const useClaimAllRevenue = ({
           if (canGoBack()) {
             goBack();
           } else {
-            navigate(Routes.WALLET_SCREEN, {
-              forceRefreshOnce: true,
-            });
+            navigate(Routes.WALLET_SCREEN);
           }
         });
       }
--- a/cardstack/src/screens/MerchantScreen.tsx
+++ b/cardstack/src/screens/MerchantScreen.tsx
@@ -1,6 +1,6 @@
 import { useRoute } from '@react-navigation/native';
 import React, { memo, useCallback, useEffect, useMemo } from 'react';
-import { Alert, StatusBar } from 'react-native';
+import { Alert, StatusBar, InteractionManager } from 'react-native';
 import { useMerchantTransactions } from '@cardstack/hooks';
 import { ContactAvatar } from '@rainbow-me/components/contacts';
 import {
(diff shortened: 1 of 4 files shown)


cardstack/src/screens/MerchantScreen.tsx
```diff
@@ -419,8 +421,27 @@ const useClaimAllRevenue = ({
...
+
+      if (Device.isAndroid) {
+        InteractionManager.runAfterInteractions(() => {
+          if (canGoBack()) {
+            goBack();
+          } else {
+            navigate(Routes.WALLET_SCREEN, {
+              forceRefreshOnce: true,
+            });
+          }
```
danibonilha: Why navigate to walletScreen?  the claimMerchant is a mutation that already refetches the safe, we shouldn't need to forceRefresh on wallet screen

src/model/wallet.ts
```diff
@@ -1183,24 +1204,6 @@ export const loadSeedPhraseAndMigrateIfNeeded = async (
...
-              // Dencrypt with the PIN
-              seedPhrase = await encryptor.decrypt(userPIN, seedPhrase);
-            } else {
-              return null;
-            }
-          } catch (e) {
-            return null;
-          }
-        }
-      }
```
danibonilha: Instead of moving this piece of code to `getSeedPhrase`,  why don't we scope `getSeedPhrase` to be accessible only within this file, and use `loadSeedPhraseAndMigrateIfNeeded` across the app when we need to load the seedPhrase ? This would mean replacing the `getSeedPhrase` at `HdProvider` to be `loadSeedPhraseAndMigrateIfNeeded` , this way we don't risk breaking any currently functionality.  We can even make the name short, like `loadSeedPhrase`

lukemelia:
This bug is a good example of why we should develop a pattern where we can use our standard transaction confirmation UI in a situation like this rather than having to handle these kinds of signing/confirmation details in various spots.

cardstack/src/screens/MerchantScreen.tsx
topwebtek7: oh yes, we don't need to forceRefresh on wallet screen, will fix it. we navigate to WALLET_SCREEN only if not able to goBack as experienced not able to goBack(to close out available revenue sheet after claimed all) after overlay screen is disappear.

src/model/wallet.ts
topwebtek7: good point

danibonilha (approved):
🚀

PR #579, opened 2021-12-16
[CS-2692]: Fix collectible sheet not closing sometimes

danibonilha:
### Description

The `ExpandedStateAssetSheet` has some issues when redux updates the state, also it's not used only for Assets which is even more confusing, anyways, it doesn't provide any meaningful abstraction for collectibles, or for some other flows that uses it, with the new `Sheet` component we can migrate all those expanded state routes, to their own routes, which will improve the code readability. Started with a simple one, the collectibles since we had a bug on it, migrated to the cardstack side, cleaned up a bit and improved the gesture handler for Android. Migrating the ones that have keyboard/input will be harder but we can eventually do it. 

### Screenshots

Dec-16-2021 18-02-22


cardstack/src/components/Sheet/Sheet.tsx
```diff
@@ -87,10 +87,41 @@ const Sheet = ({
...
+  const onScroll = useCallback(
+    ({ nativeEvent: { contentOffset } }) => {
+      if (Device.isIOS) {
+        return;
+      }
+
+      const hasScrollUp = contentOffset.y <= Device.scrollSheetOffset;
+
+      const fullVerticalDistance = Device.screenHeight;
+      const smallVerticalDistance = fullVerticalDistance * 0.3;
```
lukemelia, line 102: consider using an intention-revealing named constant in place of `0.3`

lukemelia (approved):
Nice cleanup!

cardstack/src/components/Sheet/Sheet.tsx
danibonilha: @lukemelia I thought about it, like `percentage ?`, but couldn’t find any good name, bc 0.3 is not actually a percentage, I want 30% of the screenHeight, maybe a comment in this case makes more sense, I'm open to suggestions

lukemelia: sure, just go with a comment that says `// 30% of screen height`. I think that reveals your intent clearly.

PR #580, opened 2021-12-16
[CS-2828] Update cardpay-sdk and did-resolver to 0.28.2

lukemelia:
'@cardstack/cardpay-sdk' is now the only importable path, so I updated several spots where we were specifying a more specific path unnecessarily.


cardstack/src/utils/cardpay-utils.ts
```diff
@@ -4,7 +4,7 @@ import {
   convertAmountToNativeDisplay,
   getConstantByNetwork,
 } from '@cardstack/cardpay-sdk';
-import { NativeCurrency } from '@cardstack/cardpay-sdk/sdk/currencies';
+import { NativeCurrency } from '@cardstack/cardpay-sdk';
```
danibonilha: I think we can combine this in the import above

lukemelia: ✅

danibonilha (approved):
Nooice 🚀

PR #582, opened 2021-12-17
[CS-2837]: Improve android keyboard dismissal on sheet

danibonilha:
### Description

Android's keyboard was only closing after the sheet being entirety dismissed, which caused the previous screen to have some layout breakage, to fix it i added a listener that dismisses the keyboard when the modal is closing, is not perfect yet, but it's better than before.

- I noticed there was a `isNativeStack` flag that has been set to false almost andyear, so I removed that, and all the conditions that were not used, and cleaned up iOS navigator.
- Migrated `IMPORT_SEED_SHEET` to typescript and added to the shared routes so we use the same on both platforms, the less platform conditions, the better.
- Split `IMPORT_SEED_SHEET` into components and hooks (wanted to add some tests, but won't have enough time before vacation  🏝).
- Removed some dup screens on Android and iOS navigators.
- Renamed collectibles path to same pattern as ImportSeed.

PS: on Android's emulator the behavior with keyboard is too laggy, I'd suggest testing on a Device.

### Screenshots

https://user-images.githubusercontent.com/20520102/146597269-4a3b2feb-5cab-4416-a4eb-fe10bfa4f9dc.mp4


topwebtek7 (approved):
Nice keyboard improvement and cleanup in this pr, looks good to me! 
I can see keyboard flickering issue on payment request sheet on Android, but we can take a look at in a separate ticket.

PR #583, opened 2021-12-20
[CS-2618]Add merchant payment received sheet

topwebtek7:
### Description

- [x] Completes #CS-2618

### Checklist

### Screenshots

Dec-20-2021 22-43-32

Screen Shot 2022-01-04 at 9 21 09 PM


cardstack/src/components/TransactionConfirmationSheet/displays/components/sections/MerchantSectionCard.tsx
```diff
@@ -26,7 +30,20 @@ const MerchantSectionCard = ({
       </Container>
     ) : (
       <Container alignItems="center">
-        {merchantInfoDID ? (
+        {isPaymentReceived ? (
+          <Container
+            borderRadius={70}
```
lukemelia: Is 70 a standard for us? Should it be captured as a named constant?

src/components/expanded-state/MerchantSpendReceivedExpandedState.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const network = useRainbowSelector(state => state.settings.network);
+
+  const earnedTxnData = {
+    ...transactionData,
+    subText: transactionData.netEarned,
+  };
+
+  const rowProps = {
+    ...props.asset,
+    CoinIcon: <CoinIcon size={30} symbol="DAI" />,
```
lukemelia: Should DAI be hard-coded here? I believe it would be possible for funds to be received as other tokens.

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -206,40 +150,54 @@ export default function MerchantTransactionExpandedState(
...
+      borderColor="borderGray"
+      borderRadius={10}
+      borderWidth={1}
+      marginBottom={8}
+      overflow="scroll"
+      paddingHorizontal={2}
+    >
+      {Object.values<string>(ClaimStatuses).includes(
+        props.asset.claimStatus
+      ) ? (
```
lukemelia: Extract this to a local variable to improve readability.

lukemelia (changes requested):
I added a few suggestions as review comments

lukemelia (approved):
LGTM

topwebtek7:
@kdowlin do you want to wait this pr for design change or it's ok to merge for now?

kdowlin:
Screen Shot 2021-12-28 at 10 53 54 AM
Remove the received spend transaction and it should be good for now.

topwebtek7:
ok. btw will wait for #588 to be merged

cardstack/src/components/TransactionConfirmationSheet/displays/components/sections/MerchantSectionCard.tsx
```diff
@@ -26,26 +32,41 @@ const MerchantSectionCard = ({
...
+          <Container paddingTop={3} marginBottom={4}>
+            <Text
+              weight="extraBold"
+              size="medium"
+              ellipsizeMode="tail"
+              numberOfLines={1}
+            >
+              {merchantInfoDID?.name || ''}
+            </Text>
+          </Container>
```
danibonilha: I have two concerns about this: 
 1. Mixing up the goal of this component which is to only display the MerchantInfo.
 2. Using multiple nested ternaries which is a bit confusing. 

For the first one I get the layout is almost the same, so it would make sense to adapt this one to reuse, but I'd suggest that, instead of having a boolean flag based on the payment, we add a prop called, `iconFallback`, `customIcon` something like that, then
 `<Icon name="user" size={USER_ICON_SIZE} />`  could be replaced with some helper function, `getFallbackIcon` which would render  the `user` icon, or the one received from props, bc we wouldn't have merchantInfo anyway, so the icon would be rendered. This way we can even use this for different scenarios if we need. Ideally if we add the checkmarkCircled icon svg with the right colors already on our side we wouldn't even need the helper function, it would be as simple as  `<Icon name={iconFallbackName || 'user'} size={USER_ICON_SIZE} />`

cardstack/src/components/TransactionConfirmationSheet/displays/components/sections/MerchantSectionCard.tsx
```diff
@@ -1,17 +1,23 @@
 import React, { memo, ReactNode } from 'react';
 import { ActivityIndicator } from 'react-native';
 import { MerchantInformation } from '@cardstack/types';
-import { Container, ContainerProps, Icon, Text } from '@cardstack/components';
+import { Container, ContainerProps, Text } from '@cardstack/components';
 import { ContactAvatar } from '@rainbow-me/components/contacts';
+import { Icon } from '@rainbow-me/components/icons';
```
danibonilha: We were using cardstack's Icon component, why change to rainbow's one ? We should actually add the icon svg on our side.

cardstack/src/components/TransactionConfirmationSheet/displays/components/sections/MerchantSectionCard.tsx
```diff
@@ -26,26 +32,41 @@ const MerchantSectionCard = ({
...
+            <Text
+              weight="extraBold"
+              size="medium"
+              ellipsizeMode="tail"
+              numberOfLines={1}
+            >
+              {merchantInfoDID?.name || ''}
+            </Text>
+          </Container>
         )}
```
danibonilha, line 54: it would be something like this:
```suggestion
import React, { memo, ReactNode } from 'react';
import { ActivityIndicator } from 'react-native';
import { MerchantInformation } from '@cardstack/types';
import { Container, ContainerProps, Icon, Text } from '@cardstack/components';
import { ContactAvatar } from '@rainbow-me/components/contacts';
import { IconName } from '@cardstack/components/Icon';

interface Props extends ContainerProps {
  merchantInfoDID?: MerchantInformation;
  isLoading?: boolean;
  children: ReactNode;
  iconFallback: IconName;
}

const MerchantSectionCard = ({
  merchantInfoDID,
  children,
  isLoading = false,
  iconFallback,
  ...props
}: Props) => (
  <Container
    backgroundColor="grayCardBackground"
    borderRadius={10}
    padding={8}
    {...props}
  >
    {isLoading ? (
      <Container flex={1} justifyContent="center" alignItems="center">
        <ActivityIndicator />
      </Container>
    ) : (
      <Container alignItems="center">
        {merchantInfoDID ? (
          <ContactAvatar
            color={merchantInfoDID?.color}
            size="xlarge"
            value={merchantInfoDID?.name}
            textColor={merchantInfoDID?.textColor}
          />
        ) : (
          <Icon name={iconFallback || 'user'} size={80} />
        )}
        {merchantInfoDID && (
          <Container paddingTop={3} marginBottom={4}>
            <Text
              weight="extraBold"
              size="medium"
              ellipsizeMode="tail"
              numberOfLines={1}
            >
              {merchantInfoDID?.name || ''}
            </Text>
          </Container>
        )}
      </Container>
    )}
    {children}
  </Container>
);

export default memo(MerchantSectionCard);
```

then just an example with a icon that already exists in the project:

```tsx
   <MerchantSectionCard iconFallback="check-circle">
```

Screenshot

cardstack/src/components/Transactions/Merchant/MerchantEarnedSpendAndRevenueTransaction.tsx
```diff
@@ -19,6 +21,16 @@ export const MerchantEarnedSpendAndRevenueTransaction = ({
...
+  const { navigate } = useNavigation();
+
+  const onPressTransaction = useCallback(
+    (assetProps: TransactionBaseProps) =>
+      navigate(Routes.EXPANDED_ASSET_SHEET, {
+        asset: { ...assetProps },
+        type: 'merchantSpendReceivedExpandedState',
+      }),
+    [navigate]
+  );
```
danibonilha, line 32: I think this is a great opportunity to use the new Sheet component instead of adding on more dependency on the ExpandedAssetState, since this is not an asset, you can check `CollectibleSheet`  for reference.

src/components/expanded-state/payment-item-details/PaymentItemDetails.tsx
```diff
@@ -0,0 +1,111 @@
...
+  isPrepaidCard = false,
+}: {
+  title: string;
+  color?: string;
+  textColor?: string;
+  info: any;
+  infoColor?: ResponsiveValue<keyof Theme['colors'], Theme>;
+  name?: string;
+  isTimestamp?: boolean;
+  isPrepaidCard?: boolean;
```
danibonilha: it's a bit crowded we could create it's own interface

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -166,35 +113,32 @@ const ClaimedTransaction = ({
...
-) {
-  const { setOptions } = useNavigation();
-
-  useEffect(() => {
-    setOptions({
-      longFormHeight: CHART_HEIGHT,
-    });
-  }, [setOptions]);
+export const MerchantTransactionExpandedStateBody = (
+  props: MerchantTransactionExpandedStateProps & { inlineContent?: boolean }
```
danibonilha: where is `inlineContent` being used ?

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -134,19 +79,21 @@ const EarnedTransaction = (data: EarnedTransactionProps) => {
...
+              <MerchantPaymentItemDetail
+                description={item.description}
+                key={index}
+                subValue={item.subValue}
+                symbol={item.symbol}
+                value={item.value}
+              />
+            );
+          }
+        )}
```
danibonilha: nit: earnedItems could be wrapped on useMemo and return the mapped items

danibonilha (commented):
There are some improvements to be done, also  @kdowlin agreed on using the `check-circle` icon so no need to add the blue one.

src/components/expanded-state/MerchantTransactionExpandedState.tsx
```diff
@@ -166,21 +37,64 @@ const ClaimedTransaction = ({
...
+      marginBottom={8}
+      overflow="scroll"
+      paddingHorizontal={2}
+    >
+      {isClaimedTransaction ? (
+        <ClaimedTransaction {...transactionData} txRowProps={props.asset} />
+      ) : (
+        <EarnedTransaction
+          {...transactionData}
+          subText={transactionData.netEarnedNativeDisplay}
```
lukemelia: Seems weird that we are passing `subText` here as well as in `txRowProps`. Looks like the top level one is not used and can be removed from the props type and from here.

lukemelia (changes requested):

src/components/expanded-state/MerchantTransactionExpandedState.tsx
topwebtek7: yes, good catch! 👍🏼

cardstack/src/components/TransactionConfirmationSheet/displays/components/sections/MerchantSectionCard.tsx
```diff
@@ -34,18 +40,20 @@ const MerchantSectionCard = ({
...
+          <Container paddingTop={3} marginBottom={4}>
+            <Text
+              weight="extraBold"
+              size="medium"
+              ellipsizeMode="tail"
+              numberOfLines={1}
+            >
+              {merchantInfoDID?.name || ''}
+            </Text>
+          </Container>
```
danibonilha: We can check just for merchantInfoDID, if `merchantInfoDID` exists we render the name, otherwise we don't, so no need for `isPaymentReceived` flag.

```suggestion
        {!!merchantInfoDID && (
          <Container paddingTop={3} marginBottom={4}>
            <Text
              weight="extraBold"
              size="medium"
              ellipsizeMode="tail"
              numberOfLines={1}
            >
              {merchantInfoDID?.name || ''}
            </Text>
          </Container>
```

cardstack/src/components/Transactions/Merchant/MerchantEarnedSpendAndRevenueTransaction.tsx
```diff
@@ -19,6 +21,16 @@ export const MerchantEarnedSpendAndRevenueTransaction = ({
   ...props
 }: MerchantEarnSpendAndRevenueTransactionProps) => {
   const { merchantInfoDID } = useMerchantInfoFromDID(item.infoDid);
+  const { navigate } = useNavigation();
+
+  const onPressTransaction = useCallback(
+    (asset: TransactionBaseProps) =>
+      navigate(Routes.PAYMENT_RECEIVED_SHEET, {
+        asset,
+        item,
```
danibonilha: why `asset`  and `item`?  Maybe we can merge them and pass it as a single param as `paymentTxDetails` or so.

cardstack/src/screens/sheets/PaymentReceived/PaymentReceivedSheet.tsx
```diff
@@ -0,0 +1,151 @@
...
+    params: { asset, item },
+  } = useRoute<RouteType>();
+
+  const { setOptions } = useNavigation();
+
+  useEffect(() => {
+    setOptions({
+      longFormHeight: screenHeight,
+    });
+  });
```
danibonilha: no need for this in the new sheet

cardstack/src/screens/sheets/PaymentReceived/PaymentReceivedSheet.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const rowProps = {
+    ...asset,
+    CoinIcon: (
+      <CoinIcon size={30} symbol={removeCPXDTokenSuffix(token.symbol || '')} />
+    ),
+    primaryText: `+ ${transactionData.netEarned.display}`,
+    subText: transactionData.netEarnedNativeDisplay,
+  };
+
+  return useMemo(
```
danibonilha: the `memo` on the export is enough, we don't need to wrap on useMemo

danibonilha (approved):
Looks good overall, just some minor improvements

cardstack/src/components/Transactions/Merchant/MerchantEarnedSpendAndRevenueTransaction.tsx
topwebtek7: you will see why have asset and item if look at PaymentReceivedSheet. it's also kind of consistent with other transactions

cardstack/src/components/TransactionConfirmationSheet/displays/components/sections/MerchantSectionCard.tsx
topwebtek7: nice catch! just missed it

src/components/expanded-state/MerchantTransactionExpandedState.tsx
topwebtek7: @lukemelia this is fixed. could you approve so can merge this pr?

cardstack/src/components/Transactions/Merchant/MerchantEarnedSpendAndRevenueTransaction.tsx
danibonilha: I saw the use of asset and item inside, `PaymentReceivedSheet `, still I don't think this name is meaningful since we are not passing a asset, we are actually passing the transaction details, this whole thing of using ExpandedAsset, or anything related to Asset, for other things that are not actual assets, doesn't make much sense, and it would nice to move away from it, that's why I questioned it. Hopefully we can migrate little by little.

topwebtek7: ok, updated param merged with asset, item. anyway I think we will need to revisit this when handing push notification for received payment.

lukemelia (approved):
👍  Good to go from my perspective

PR #584, opened 2021-12-21
[CS-2842] Add error handling for auth failures in Android backup

lukemelia:
The android Java code that manages this log-in (https://github.com/brunobar79/react-native-cloud-fs/blob/master/android/src/main/java/org/rncloudfs/RNCloudFsModule.java#L135) has no error handling or way to check whether the login was successful. I forked that repository and updated this branch to use the fork. Let's try it on a device demonstrating the issue (like Kiel's) and see if it is improved.


topwebtek7 (approved):
Looks good to me! 🚀

kdowlin:
@lukemelia I'm still getting a forever pending back up screen. I'm testing on version 26 which was the one you cut off this branch I believe.

danibonilha (commented):
I also get the backing up info forever, maybe are we missing to `setIsWalletLoading(null)` onError case ?

PR #590, opened 2022-01-03
[Chore]: Bump ios version

danibonilha:
### Description
Itunes' upload is failing bc we already have a version deployed to the store, for now I'm bumping it manually but we can add a task to bump using github actions.


topwebtek7 (approved):
🚀

PR #592, opened 2022-01-06
[CS-2672]: Fix button layout 

danibonilha:
### Description
The global `<Button />` uses a custom animated button wrapper `ButtonPressAnimation`, which has two different complicated and old implementation, one for iOS and on for Android,  just to achieve a bouncy/scaly button on press, which we can easily do nowadays with react native's Animated api, so I created the `<AnimatedPressable />`  and already migrated the `<Button />`, once we migrate all we can remove `ButtonPressAnimation`, the only thing missing right now, is the scale animation fixed on a starting point, like left/right, but I don't think will have this on the redesign, so we need to evaluate if it's worth. 
In summary, we had a diff behavior on Android bc the component was diff, now we are using the same to keep it consistent. 

I encountered a bug on mainnet, where android was navigating to an iOS screen, to buy funds with apple, so I fixed that, and with the button redesign we were able to enable the prefilled amount for android too. 
 
Also simplified the disabled state of the button, bf it had an diff component positioned on top of the main button, that caused a couple issues, now it applies a default disable variant, and in case we need to disable just the press, and use another variant, we can use the disablePress prop, added a couple of tests for these scenarios.

As <Button /> is global, i'd appreciate if you could go through the app too, I tested as much as I could but I might have missed something.

PS: As we know, animations on android might be slow on debug/emulator, it's best to test on a device.

- [x] Completes #(CS-2672)

### Screenshots

Android copy address:
Screen Shot 2022-01-07 at 11 36 14

Screen Shot 2022-01-06 at 20 35 05

PrefilledAmount: 
Screen Shot 2022-01-06 at 20 15 04

New button animation:
Simulator Screen Recording - iPhone 11 - 2022-01-07 at 11 27 38


cardstack/src/components/Button/Button.tsx
```diff
@@ -48,19 +49,10 @@ const VariantRestyleComponent = createVariant({
...
 );
 
-const ButtonContentWrapper = ({
-  children,
-  wrapper,
-}: {
-  children: ReactNode;
-  wrapper: ButtonWrappper;
-}) =>
-  wrapper === 'fragment' ? <>{children}</> : <Container>{children}</Container>;
```
topwebtek7, line 62: If we don't use wrapper prop anymore, we can remove this `wrapper` prop in ButtonProps

cardstack/src/utils/theme-utils.ts
```diff
@@ -43,19 +43,40 @@ export const useVariantValue = <T extends keyof SafeVariants<Theme>>(
...
+  const variantObject = theme[themeKey];
+  const defaultStyles = variantObject.defaults;
+  const variantStyles = variantObject[variant];
+
+  return {
+    variantStyles,
+    defaultStyles,
+    mergedStyles: { ...defaultStyles, ...variantStyles },
+  };
+};
```
topwebtek7, line 82: Nice util function 👏🏼

cardstack/src/theme/buttonVariants.ts
```diff
@@ -80,6 +80,11 @@ const invalid = {
   },
 };
 
+const disabledInvalid = {
+  ...invalid,
+  backgroundColor: 'invalidDark',
+};
+
```
topwebtek7, line 87: Looks good with this new variant, just for more context, when will need to use this button variant?

topwebtek7 (approved):
I did tests on both iOS, Android and overall looks good to me, left few comments.

oidoug (approved):
I really like your new implementation using Pressable and onPressIn and Out calls, a lot cleaner.

cardstack/src/components/Button/Button.tsx
danibonilha: good catch! thanks, I'll remove on #594 though, just to avoid the need of rebasing

cardstack/src/theme/buttonVariants.ts
danibonilha: This is being used on `HoldToAuthorizeButton`.

PR #593, opened 2022-01-07
Don't hide zero-balance prepaid cards

lukemelia:
### Description

Passes `viewAll: true` option to SDK when retrieving safes so that zero-balance prepaid cards are included in the results. Handling this without error required adjusting the balance formatting as well.
 
- [x] Completes #CS-1276

### Screenshots

Screenshot


oidoug (approved):
Tested on current production (TestFlight), it did not show a 0.01 SPEND card.
In this implementation, it shows my card as $0.00

topwebtek7 (approved):
looks good to me!

PR #594, opened 2022-01-07
[Refactor]: Cleanup unused code and use AnimatedPressable

danibonilha:
### Description

While searching for components that were using` <ButtonPressAnimation>` to replace with the one,  there were lots of components that are not being used anymore, so I removed those, and migrated some others to use `AnimatedPressable`, there's still lots to be migrate and some features in the AnimatedPressable might be missing still, so we can gradually improve this.

Depends on #592 , after 582 is merged, I'll base this on develop


cardstack/src/components/AnimatedPressable/AnimatedPressable.tsx
```diff
@@ -25,7 +30,17 @@ const styles = StyleSheet.create({
   },
 });
 
-const AnimatedPressable = ({ children, ...props }: PressableProps) => {
+interface AnimatedPressableProps extends PressableProps {
+  enableHapticFeedback?: boolean;
+  hepaticType?: HapticFeedbackTypes;
```
oidoug: I believe you have a typo on "hepticType".

src/navigation/effects.js
```diff
@@ -291,49 +255,6 @@ export const backgroundPreset = {
   gestureResponseDistance,
 };
 
-export const emojiPreset = {
```
oidoug, line 294: Just for me to understand, I see you removed setting avatar as emoji. Wasn't this feature working?

oidoug (changes requested):
Small change, there's a typo on heptic in AnimatedPressable.

src/navigation/effects.js
danibonilha: Rainbow had a feature to automatically create an emoji as an avatar if the user didn't add any name, but it seems it didn't work so great and it's been disabled from more than 8 months, so I think it's safe to delete it ,if we want to eventually have something like this again, which I doubt it, we can build from scratch and reference some previous code, if you want to add an emoji from the keyboard, it's still possible.

lukemelia (approved):
Hooray for deleted code!

oidoug (approved):

PR #595, opened 2022-01-10
[CS-2830] Update payment request expanded sheet, fix input flickering issue on Android

topwebtek7:
### Description

Moved requestPaymentRequest screen from ExpandedAssetSheet to new cardstack route with new sheet component, and also updated sheet component to have fixed Header, Footer. also did some UI tweaks for different device/screen sizes.

- [x] Completes #CS-2830

### Checklist

### Screenshots

Screen Shot 2022-01-11 at 3 23 22 AM
Screen Shot 2022-01-11 at 3 22 51 AM

Screen Shot 2022-01-11 at 3 22 51 AM

### DIFF
--- a/cardstack/src/components/Sheet/Sheet.tsx
+++ b/cardstack/src/components/Sheet/Sheet.tsx
@@ -38,6 +38,8 @@ export interface SheetProps {
   scrollEnabled?: boolean;
   shadowEnabled?: boolean;
   overlayColor?: string;
+  Header?: JSX.Element;
+  Footer?: JSX.Element;
 }
 
 const Sheet = ({
@@ -48,6 +50,8 @@ const Sheet = ({
   scrollEnabled = false,
   shadowEnabled = false,
   overlayColor = 'transparent',
+  Header,
+  Footer,
 }: SheetProps) => {
   const insets = useSafeArea();
   const { goBack, setOptions } = useNavigation();
@@ -141,6 +145,7 @@ const Sheet = ({
         <CenteredContainer paddingVertical={4}>
           {!hideHandle && <SheetHandle />}
         </CenteredContainer>
+        {Header}
         <ScrollView
           onMomentumScrollBegin={onMomentumScrollBegin}
           overScrollMode="always"
@@ -154,6 +159,7 @@ const Sheet = ({
         >
           {children}
         </ScrollView>
+        {Footer}
       </KeyboardAvoidingView>
     </Container>
   );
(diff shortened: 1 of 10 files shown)


cardstack/src/screens/PaymentRequest/RequestPaymentConfirmation.tsx
```diff
@@ -218,29 +218,33 @@ export const RequestPaymentConfirmation = ({
...
+    style={
+      Device.isIOS
+        ? shadow.buildAsObject(0, -1, 2, 'rgba(0, 0, 0, 0.25)', 1)
+        : {
+            borderTopWidth: 0.5,
+            borderTopColor: 'rgba(0, 0, 0, 0.25)',
+            shadowColor: 'rgba(0, 0, 0, 0.25)',
+            elevation: 3,
+          }
+    }
```
danibonilha: Can we extract this to an style const outside the component ?  If you get any typing error, you can just type the style as `StyleProp<ViewStyle>`

```tsx
const footerStyle: StyleProp<ViewStyle> = ... 
```

cardstack/src/components/Sheet/Sheet.tsx
```diff
@@ -141,6 +145,7 @@ const Sheet = ({
         <CenteredContainer paddingVertical={4}>
           {!hideHandle && <SheetHandle />}
         </CenteredContainer>
+        {Header}
         <ScrollView
```
danibonilha, line 149: nit: would make sense to use StickyHeader from ScrollView in this case ?

cardstack/src/screens/PaymentRequest/PaymentRequestExpandedSheet.tsx
```diff
@@ -88,20 +98,16 @@ const PaymentRequestExpandedState = (props: { asset: MerchantSafeType }) => {
...
-      bottomInset={editMode ? 50 : 110}
-      hasKeyboard={editMode}
-      height="100%"
-      renderFooter={() =>
-        editMode ? <EditFooter /> : <RequestPaymentConfirmationFooter />
-      }
-      renderHeader={() => (
+    <Sheet
+      isFullScreen
+      scrollEnabled
```
danibonilha: Maybe we should disable the scroll if we are at the editMode, bc on iOS we get bouncy behavior on the input.

danibonilha (approved):
Looking good, just small suggestions

cardstack/src/screens/PaymentRequest/PaymentRequestExpandedSheet.tsx
topwebtek7: good point! 👍🏼

cardstack/src/components/Sheet/Sheet.tsx
topwebtek7: hmm I've tried but didn't work with StickyHeader for some reason.. gonna merge this pr and will look at it more later

PR #596, opened 2022-01-11
[CS-2750] Service degradation notice

oidoug:
### Description

New `Notice` Component that will be (preferably) rendered inside the lower area of the navigation bar. `Notice` is reusable and will work similarly to a Toast component, but it's fixed and can be used as a button or link.

New `ServiceStatusNotice` uses `Notice` and will handle fetching the network status information to be displayed.

- [x] Completes #(CS-2750)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/148952344-2ad36648-0f34-455a-b681-cb620f29ae02.mp4


lukemelia:
Looking good so far!

cardstack/src/components/Notice/Notice.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export const Notice = ({
+  description,
+  onPress,
+  type = 'warning',
+  icon,
+}: NoticeProps) => {
+  return (
+    <AnimatedPressable onPress={onPress} testID="notice-pressable">
```
danibonilha: I think we can use a simple `Touchable` here, bc it's more of a `toast` than a button itself, or even a touchable without feedback ? 🤔

package.json
```diff
@@ -4,7 +4,8 @@
   "scripts": {
     "setup": "yarn install && yarn postinstall",
     "build:ios": "node node_modules/react-native/local-cli/cli.js bundle --entry-file='index.js' --bundle-output='./ios/main.jsbundle' --dev=false --platform='ios'",
-    "android": "adb uninstall com.cardstack.cardpay || true && react-native run-android",
+    "android": "react-native run-android",
+    "android:clean": "adb uninstall com.cardstack.cardpay || yarn android",
```
danibonilha: thanks 👏

cardstack/src/components/Notice/Notice.tsx
oidoug: Yes, I was considering the same. Maybe a Touchable will feel more in line with all the other elements in the wallet screen.

oidoug:
https://user-images.githubusercontent.com/129619/149335882-4888cc08-c07c-4f25-844e-136ee57a2c5f.mp4

We might need to fake some incidents for you to test this properly.
During development, I used a hard-coded fake json server for it.

cardstack/src/components/Notice/Notice.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { NoticeType } from '@cardstack/types';
+import { ColorTypes } from '@cardstack/theme';
+
+if (
+  Platform.OS === 'android' &&
+  UIManager.setLayoutAnimationEnabledExperimental
+) {
+  UIManager.setLayoutAnimationEnabledExperimental(true);
+}
+
```
danibonilha: we have this set globally on `App.js`

cardstack/src/components/Notice/ServiceStatusNotice.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const { data } = useGetServiceStatusQuery(null);
+
+  return (
+    <Notice
+      isVisible={!!data}
+      description={data?.name || 'Incident reported.'}
+      type="warning"
+      onPress={() => {
+        Linking.openURL(`https://status.cardstack.com`);
+      }}
```
danibonilha: we can wrap this on useCallback to avoid recreating on new render

cardstack/src/services/service-status-api.ts
```diff
@@ -0,0 +1,47 @@
...
+    })[0];
+  }
+
+  const major = incidents.filter((a: IncidentType) =>
+    a.impact?.includes('major')
+  );
+
+  return (
+    major.sort((a: IncidentType, b: IncidentType) => {
+      return +new Date(b.started_at) - +new Date(a.started_at);
```
danibonilha: nit: maybe this could be a helper function since we are using twice, we also have lodash in the project,  if you're familiar with it

cardstack/src/services/service-status-api.ts
```diff
@@ -0,0 +1,47 @@
...
+    })[0] || null
+  );
+};
+
+export const serviceStatusApi = createApi({
+  reducerPath: 'serviceStatusApi',
+  baseQuery: fetchBaseQuery({ baseUrl: STATUS_PAGE_URL }),
+  tagTypes: [...Object.values(ServiceStatusTags)],
+  endpoints: builder => ({
+    getServiceStatus: builder.query({
```
danibonilha: we can type the response right after `build.query<SomeResponseType, void>`  than `transformResponse` won't need to be any
Ps. this should probably remove the necessity to pass null as an argument for the query

cardstack/src/components/AssetList/AssetList.tsx
```diff
@@ -51,6 +52,7 @@ const onScrollToIndexFailed = () => {
 export const AssetList = (props: AssetListProps) => {
   const sectionListRef = createRef<SectionList>();
   const refresh = useRefreshAccountData();
+  const { refetch: refetchServiceStatus } = useGetServiceStatusQuery(null);
```
danibonilha: Why do we need to pass null? maybe we are missing to type something

danibonilha (commented):
Props on using RTK 👏 , it's really looking good, just a couple of suggestions, it would be nice to add on this PR the frequency to refetch the status api also, if there are any queries we should invalidate the status cache to be refetched, or if refetching every hour or so is enough, or just once the app first starts.

cardstack/src/components/Notice/Notice.tsx
```diff
@@ -0,0 +1,99 @@
...
+const textColorByType: {
+  [key in NoticeType]: ColorTypes;
+} = {
+  info: 'black',
+  warning: 'black',
+  error: 'white',
+};
+
+const iconColorByType: {
+  [key in NoticeType]: ColorTypes;
```
danibonilha: nit: this could be extracted to it's own type, so we can reuse on the others styles
``` ts
type NoticeVariant = { [key in NoticeType]: ColorTypes } 
```
or
``` ts
type NoticeVariant = Record<NoticeType, ColorTypes>
```
or we could even use one noticeConfig with all the variants in it

``` tsx
interface NoticeStyle {
  iconColor: ColorTypes;
  backgroundColor: ColorTypes;
  textColor: ColorTypes;
}


const noticeConfig: Record<NoticeType, NoticeStyle> = {
  info: {
    iconColor: 'appleBlue',
    backgroundColor: 'backgroundGray',
    textColor: 'black',
  },
  ...
};

<>
  backgroundColor={noticeConfig[type].backgroundColor}
</>
```

cardstack/src/components/Notice/Notice.tsx
oidoug: Ok, I also took the chance to improve the animation a bit.

cardstack/src/components/Notice/Notice.tsx
oidoug: Loved that last suggestion, so much cleaner :) Thanks for the quality review, Dani!

cardstack/src/services/service-status-api.ts
oidoug: I moved the incident filter to service/utils refactored it with lodash to work with any incident type (not only critical or major), and added some tests.

cardstack/src/components/Notice/Notice.tsx
```diff
@@ -56,12 +43,14 @@ export const Notice = ({
   type = 'warning',
   icon,
 }: NoticeProps) => {
-  useEffect(() => {
-    LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut);
-  }, []);
+  useLayoutEffect(() => {
+    LayoutAnimation.configureNext(
+      LayoutAnimation.create(200, 'easeInEaseOut', 'opacity')
+    );
```
danibonilha: we have this as a helper function => `layoutEasingAnimation`

danibonilha (approved):
LGTM, just one small "fix", congrats on your first feature 🚀

lukemelia:
Way to go @douglaslondrina! 👏 👏 👏  🥇

PR #597, opened 2022-01-11
[CS-2925] Fix Depot/EOA send screen UI breaks on Android

topwebtek7:
### Description

SendSheet had an issue with Android keyboard open state handling, so refactored using Sheet component and also updated UI to be keep same when keyboard state changes.

- [x] Completes #CS-2925

### Checklist

### Screenshots

Jan-12-2022 17-00-03

Screen Shot 2022-01-12 at 4 52 17 PM
Screen Shot 2022-01-12 at 4 51 49 PM

Screen Shot 2022-01-12 at 4 52 06 PM

### DIFF
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -28,7 +28,6 @@ import {
 import {
   bottomSheetPreset,
   expandedPreset,
-  sheetPreset,
 } from '@rainbow-me/navigation/effects';
 import { nativeStackModalConfig } from '@rainbow-me/navigation/config';
 import RainbowRoutes from '@rainbow-me/navigation/routesNames';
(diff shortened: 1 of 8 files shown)


danibonilha (commented):
Props for the refactor to use `<Sheet />`, but I noticed two things:
1. The items are all squished together, we should try to match as close as possible the prev design: 

Screen Shot 2022-01-12 at 18 58 18
     Maybe if we add a maxHeight based on the device screen it might work ? idk


2. Hitting the Max button doesn't out fill both inputs anymore.

topwebtek7:
> Props for the refactor to use `<Sheet />`, but I noticed two things:
> 
> 1. The items are all squished together, we should try to match as close as possible the prev design:
> Maybe if we add a maxHeight based on the device screen it might work ? idk
> 2. Hitting the Max button doesn't out fill both inputs anymore.

I've tried to match new design in this ticket(https://linear.app/cardstack/issue/CS-2413/update-transaction-screen-to-include-gas-coin-in-estimate) when refactor sendSheet
Also updated inputs UI and Max buttons works now as well.

Screen Shot 2022-01-13 at 1 17 16 PM

danibonilha (approved):
On bigger devices still a bit squashed, but it's good overall , we can investigate later, so I think it's good to go  🚀

PR #599, opened 2022-01-12
[CS-2929]: Fix dust amount and migrate unclaimedRevenue

danibonilha:
### Description
This PR fixes the dust amount and moves the unclaimedRevenue from expandedState, to the new Sheet component. Abstracts the `RouteType` type to be shared across all screens. I wanted to test more on Android, but unable to build right now since Jcenter is out, but the code is good be reviewed already.

- [x] Completes #(CS-2929)

### Screenshots

Simulator Screen Recording - iPhone 11 - 2022-01-12 at 17 27 05


cardstack/src/components/TransactionList/TransactionListLoading.tsx
```diff
@@ -15,8 +15,14 @@ const TransactionsSection = ({
...
 );
 
-export const TransactionListLoading = ({ light }: { light?: boolean }) => (
-  <Container padding={5} width="100%">
+export const TransactionListLoading = ({
+  light,
+  removePadding,
+}: {
+  removePadding?: boolean;
+  light?: boolean;
```
danibonilha, line 23: This is temporary, we can remove it when we migrate all the conditional sectionLists

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
```diff
@@ -51,11 +52,7 @@ export default function UnclaimedRevenueExpandedState({
               <TokenItem key={token.tokenAddress} token={token} />
             ))}
           </Container>
-          <Button
-            marginTop={8}
-            onPress={isDust ? () => {} : onClaimAll}
-            variant={isDust ? 'disabled' : undefined}
-          >
+          <Button disabled={!isDust} marginTop={8} onPress={onClaimAll}>
```
lukemelia: Something seems off. You should not be able to claim dust, but this logic says the button should be disabled if it is NOT dust.

cardstack/src/screens/sheets/UnclaimedRevenue/UnclaimedRevenueSheet.tsx
```diff
@@ -0,0 +1,159 @@
...
+    [revenueBalances]
+  );
+
+  return (
+    <Container paddingHorizontal={5} paddingVertical={3}>
+      <Text size="medium">Claim available revenue</Text>
+      <Container flexDirection="column" marginTop={5}>
+        {renderTokens}
+      </Container>
+      <Button disabled={!isDust} marginTop={8} onPress={onClaimAll}>
```
lukemelia: Same here. It should be disabled if there is nothing to claim or only dust to claim.

cardstack/src/screens/sheets/UnclaimedRevenue/UnclaimedRevenueSheet.tsx
```diff
@@ -18,40 +18,25 @@ import { ClaimStatuses } from '@cardstack/utils';
 import { RouteType } from '@cardstack/navigation/types';
 
 interface Params {
-  asset: MerchantSafeType;
-  customFunction: () => void;
+  merchantSafe: MerchantSafeType;
+  onClaimAllPress: () => void;
```
lukemelia, line 22: 👍  Much better name :)

lukemelia (changes requested):

src/components/expanded-state/UnclaimedRevenueExpandedState.tsx
danibonilha: my bad, I committed the wrong piece 🤦‍♀️

danibonilha: `isDust` should be `hasClaimableAmount`, bc on gnosis safe we already check for dustAmount, and if we have one, the native amount is set to zero, so in unclaimedRevenue we check if native amount is a valid value. 
`const hasClaimableAmount = !!nativeAmount`. 
The previous name was misleading as we are not checking for dust anymore.

lukemelia (approved):
🙌  LGTM!

PR #600, opened 2022-01-13
[CS-2922] Fix confirmation sheet cover over statusBar in Android

topwebtek7:
### Description

- [x] Completes #CS-2922

### Checklist

### Screenshots

before:
Screen Shot 2022-01-14 at 12 49 21 AM

after:
Screen Shot 2022-01-14 at 12 58 19 AM

### DIFF
--- a/cardstack/src/screens/TransactionConfirmation.tsx
+++ b/cardstack/src/screens/TransactionConfirmation.tsx
@@ -1,11 +1,16 @@
 import React from 'react';
+import { StatusBar } from 'react-native';
 import { GasSpeedButton } from '../../../src/components/gas';
 import { useTransactionConfirmation } from '@cardstack/hooks';
+import { Device } from '@cardstack/utils';
 import {
   Container,
   TransactionConfirmationSheet,
   SafeAreaView,
 } from '@cardstack/components';
+
+// converted Status bar height that restyle paddingTop prop can understand
+const StatusBarHeightInStyle = Math.ceil((StatusBar.currentHeight || 24) / 4);
 
 const TransactionConfirmation = () => {
   const {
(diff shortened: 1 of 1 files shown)


cardstack/src/screens/TransactionConfirmation.tsx
```diff
@@ -1,12 +1,17 @@
...
+import { Device } from '@cardstack/utils';
 import {
   Container,
   TransactionConfirmationSheet,
   SafeAreaView,
 } from '@cardstack/components';
 
+// converted Status bar height that restyle paddingTop prop can understand
+const StatusBarHeightInStyle = Math.ceil((StatusBar.currentHeight || 24) / 4);
+
```
danibonilha: The spacing is incremented by 4 , if by any chance `StatusBarHeightInStyle` is not a 4 multiplier, the app will probably crash, I think it's safer to use the style prop directly

cardstack/src/screens/TransactionConfirmation.tsx
```diff
@@ -21,7 +26,13 @@ const TransactionConfirmation = () => {
   } = useTransactionConfirmation();
 
   return (
-    <SafeAreaView backgroundColor="black" flex={1} width="100%">
+    <SafeAreaView
+      backgroundColor="black"
+      flex={1}
+      width="100%"
+      paddingTop={Device.isAndroid ? StatusBarHeightInStyle : 0}
+    >
```
danibonilha:
```suggestion
    <SafeAreaView  style={styles.safeview}
    >
```

cardstack/src/screens/TransactionConfirmation.tsx
danibonilha, line 19:
```suggestion
const styles = StyleSheet.create({
  safeArea: {
    backgroundColor: 'black',
    flex: 1,
    width: '100%',
    paddingTop: StatusBar.currentHeight || 0
  }
})
```

danibonilha: since  `StatusBar.currentHeight` only returns a value for Android we don't need to check the platform

danibonilha (approved):

PR #608, opened 2022-01-19
[Chore]: Bump iOS next version

danibonilha:
### Description
iOS needs to be bumped to the next version to prevent itunes uploading issues. This will hopefully be fixed in the next devops task


lukemelia (approved):
Bumpity bump bump

PR #609, opened 2022-01-19
Update "Earned" values on Available Revenue screen to reflect the actual net earned amount

lukemelia:
Show net earned amounts on "Earned" transactions on Available Revenue screen.

- [x] Completes CS-2030

### Screenshots

Before:

Before -->

After:

After -->


danibonilha (approved):
LGTM

PR #610, opened 2022-01-19
[CS-2945] Eliminate the "no backups found" error

lukemelia:


cardstack/src/screens/WelcomeScreen/hooks.tsx
```diff
@@ -17,7 +17,7 @@ export const useWelcomeScreen = () => {
     StackNavigationProp<ParamListBase>
   >();
 
-  const [userData, setUserData] = useState(null);
+  const [userData, setUserData] = useState(null as JSON | null);
```
danibonilha: Instead of casting, we can define it like this 
```ts
const [userData, setUserData] = useState<null | JSON>(null);
```
Maybe we can try to figure the backup structure and properly type it, instead of using a generic JSON type

lukemelia: I added some types. Please have a look and LMK what you think.

cardstack/src/screens/WelcomeScreen/hooks.tsx
```diff
@@ -11,13 +11,14 @@ import { Device } from '@cardstack/utils';
...
 import logger from 'logger';
+import { ICloudBackupData } from '@rainbow-me/model/backup';
 
 export const useWelcomeScreen = () => {
   const { navigate, replace } = useNavigation<
     StackNavigationProp<ParamListBase>
   >();
 
-  const [userData, setUserData] = useState(null);
+  const [userData, setUserData] = useState(null as ICloudBackupData | null);
```
danibonilha:
```suggestion
  const [userData, setUserData] = useState<ICloudBackupData | null>(null);
```

danibonilha (approved):

PR #611, opened 2022-01-20
[CS-2983]: Add initial tabBar layout as feature flag

danibonilha:
### Description

This PR adds the tabbar with a feature flag since we need to map internal flows and behaviors in order to replace the current navigation style, it's mostly design, don't expect to navigate correctly in inner flows.

### Screenshots

Screen Shot 2022-01-20 at 16 50 59

iOS:

Simulator Screen Recording - iPhone 11 - 2022-01-20 at 16 51 24

Android:
android-tabbar

### DIFF
--- a/cardstack/src/navigation/tabBarNavigator.tsx
+++ b/cardstack/src/navigation/tabBarNavigator.tsx
@@ -1,7 +1,6 @@
 import React, {
   createContext,
   Dispatch,
-  ReactNode,
   SetStateAction,
   useContext,
   useMemo,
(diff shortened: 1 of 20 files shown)


cardstack/src/navigation/tabBarNavigator.tsx
```diff
@@ -0,0 +1,125 @@
...
+}
+
+const TabBarFeatureContext = createContext<TabBarContextType>({
+  isTabBarEnabled: false,
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  setIsTabBarEnabled: () => {},
+});
+
+export const TabBarFeatureProvider: React.FC = ({ children }) => {
+  const [isTabBarEnabled, setIsTabBarEnabled] = useState(false);
```
danibonilha, line 110: Set this to true to enable the tab if you want to check it out

cardstack/src/navigation/tabBarNavigator.tsx
```diff
@@ -0,0 +1,125 @@
...
+      isTabBarEnabled,
+      setIsTabBarEnabled,
+    }),
+    [isTabBarEnabled]
+  );
+
+  return (
+    <TabBarFeatureContext.Provider value={contextValues}>
+      {isTabBarEnabled ? <StackNavigator /> : children}
+    </TabBarFeatureContext.Provider>
```
danibonilha, line 123: We haven't yet decided how to enable the feature in a release, or internal build, but I created as a context to make it easier to apply when we do decide

topwebtek7 (commented):
Looks good to me! @danibonilha what about we upgrade React-navigation to latest version(v6) when change to TabBar... I think it could fix QRCode scanner blackout things sometimes happens(cuz of `useFocusEffect`'s not stable) And if we can, get rid of react-native-cool-modals would be perfect, IMO.

lukemelia (approved):
I tried it out on iOS and the tab bar and interaction look good 👍

oidoug (approved):
Looks good on both ios and android.
I feel the wallet should be in the middle.

danibonilha:
> Looks good to me! @danibonilha what about we upgrade React-navigation to latest version(v6) when change to TabBar... I think it could fix QRCode scanner blackout things sometimes happens(cuz of `useFocusEffect`'s not stable) And if we can, get rid of react-native-cool-modals would be perfect, IMO.

I thought about migrating to v6 but as of today there are still lots of diff navigation configs, my thought is to simplify the nav first, then upgrade to the latest version, it's on the radar.

PR #612, opened 2022-01-24
[CS-2569] Notifications Preferences

oidoug:
### Description

Adds new Notifications Preferences:
- Notifications section to Settings screen;
- Service update on hub-service for getting and updating notif. options;
- Hook useUpdateNotificationPreferences for handling authToken and options state.

Note: Loading the auth token from using getHubAuthToken from hub-services is often failing with the error:
``` ERROR  Failed. {"errors": [{"detail": "Expired nonce", "status": "401", "title": "Invalid signature"}]}```

I'm investigating that now. The error does not seem to be related to this PR.

- [x] Completes #(CS-2569)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/151046073-24cf323f-de80-4584-a78d-f5116aa2d473.mp4

### DIFF
--- a/cardstack/src/hooks/notifications-preferences/useUpdateNotificationPreferences.ts
+++ b/cardstack/src/hooks/notifications-preferences/useUpdateNotificationPreferences.ts
@@ -33,7 +33,7 @@ export const useUpdateNotificationPreferences = () => {
   }, [authToken]);
 
   const onUpdateOptionStatus = useCallback(
-    async (value, item) => {
+    async (item, value) => {
       const updateList = options?.map(
         (option: NotificationsPreferenceDataType) => {
           if (
--- a/cardstack/src/hooks/notifications-preferences/useUpdateNotificationPreferences.ts
+++ b/cardstack/src/hooks/notifications-preferences/useUpdateNotificationPreferences.ts
@@ -1,0 +1,76 @@
+import { useEffect, useCallback, useState } from 'react';
+import { useAuthToken } from '@cardstack/hooks';
+
+import { getFCMToken } from '@cardstack/models/firebase';
+import {
+  getNotificationsPreferences,
+  setNotificationsPreferences,
+} from '@cardstack/services/hub-service';
+import { NotificationsPreferenceDataType } from '@cardstack/types';
+
+export enum NotificationsOptionsStrings {
+  'merchant_claim' = 'Merchant Claim',
+  'customer_payment' = 'New Payment Received',
+}
+
+export const useUpdateNotificationPreferences = () => {
+  const { authToken, isLoading, error } = useAuthToken();
+
+  const [options, setOptions] = useState<
+    NotificationsPreferenceDataType[] | undefined
+  >([]);
+
+  const fetchNotificationPreferences = useCallback(async () => {
+    const { fcmToken } = await getFCMToken();
+
+    if (fcmToken) {
+      const result = await getNotificationsPreferences(authToken, fcmToken);
+
+      if (result && result.length > 0) {
+        setOptions(result);
+      }
+    }
+  }, [authToken]);
+
+  const onUpdateOptionStatus = useCallback(
+    async (value, item) => {
+      const updateList = options?.map(
+        (option: NotificationsPreferenceDataType) => {
+          if (
+            option.attributes['notification-type'] ===
+            item.attributes['notification-type']
+          ) {
+            option.attributes.status = value ? 'enabled' : 'disabled';
+          }
+
+          return option;
+        }
+      );
+
+      setOptions(updateList);
+
+      const updateItem = updateList?.find(
+        o =>
+          o.attributes['notification-type'] ===
+          item.attributes['notification-type']
+      );
+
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Review by lukemelia — APPROVED** [review]
Looks good to me 👍

File: cardstack/src/hooks/notifications-preferences/useUpdateNotificationPreferences.ts [hunk 1]
```diff
@@ -0,0 +1,76 @@
...
+
+export const useUpdateNotificationPreferences = () => {
+  const { authToken, isLoading, error } = useAuthToken();
+
+  const [options, setOptions] = useState<
+    NotificationsPreferenceDataType[] | undefined
+  >([]);
+
+  const fetchNotificationPreferences = useCallback(async () => {
+    const { fcmToken } = await getFCMToken();
```
danibonilha: nit: we can't really call any of the hub notifications services without the fcmToken right ? Maybe we can retrieve this inside of the services functions

cardstack/src/hooks/notifications-preferences/useUpdateNotificationPreferences.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      const updateItem = updateList?.find(
+        o =>
+          o.attributes['notification-type'] ===
+          item.attributes['notification-type']
+      );
+
+      const { fcmToken } = await getFCMToken();
+
+      if (fcmToken && authToken && updateItem) {
```
danibonilha: ditto

src/components/settings-menu/NotificationsSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+    </Container>
+  );
+
+  const ListLoading = () => (
+    <Container flexDirection="column" paddingHorizontal={6} paddingVertical={2}>
+      <Skeleton height={40} light marginBottom={1} width="100%" />
+      <Skeleton height={40} light width="100%" />
+    </Container>
+  );
+
```
danibonilha: Those component functions are all "anonymous" so on every new render they will be recomputed, we can either extract those outside the `NotificationsSection` component, or wrap then with `useMemo/useCallback`

src/components/settings-menu/NotificationsSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const ListError = () => (
+    <Container alignItems="center" flex={1} justifyContent="center">
+      <Text>{error}</Text>
+    </Container>
+  );
+
+  const ListLoading = () => (
+    <Container flexDirection="column" paddingHorizontal={6} paddingVertical={2}>
+      <Skeleton height={40} light marginBottom={1} width="100%" />
+      <Skeleton height={40} light width="100%" />
```
danibonilha: nit: we could share these props, in case we need to change something it's easier to keep consistent

src/components/settings-menu/NotificationsSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+    </Container>
+  );
+
+  return (
+    <FlatList
+      ListEmptyComponent={error ? ListError : ListLoading}
+      contentContainerStyle={{ paddingTop: 16 }}
```
danibonilha: nit: we could extract this style object outside the component

src/components/settings-menu/NotificationsSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+      data={options}
+      keyExtractor={(item: NotificationsPreferenceDataType) =>
+        item.attributes['notification-type']
+      }
```
danibonilha: nit: this could be it's own function so it wouldn't be anonymous, we don't need to worry to much about it on this screen though, bc it's a light list, but it's nice to keep on mind the perf tips around flatlist

danibonilha (approved):
It's looking good, just a couple of suggestions, also I think that migrating the hub services to RTK could make easier to  reuse authToken and fcmToken, as our currently implementation is a bit confusing, but we can have the current state merged in and then we add the refactoring  to the backlog. Also can we get screenshots or .gifs for the current implementation  ?

cardstack/src/hooks/notifications-preferences/useUpdateNotificationPreferences.ts
oidoug: Good point, call moved as suggested.

src/components/settings-menu/NotificationsSection.tsx
oidoug: Forgot to do that, thanks for the pointer.

src/components/settings-menu/NotificationsSection.tsx
oidoug: Maybe we could have a more generalized loading list effect, but it's usually hard to know the number of items beforehand. For now, I kept these skeleton comps as a simple list.

oidoug:
@danibonilha , want to take a second look before I merge this?

danibonilha (approved):
🚢

PR #613, opened 2022-01-25
[CS-3004] Stores hub auth token

oidoug:
### Description

Saves Hub Auth Token
- In hub-service.getHubAuthToken, tries to load a token from local storage.
- If none is present, generate a new one with authAPI.authenticate and store it locally.
- If for some reason the saved token gets rejected by the hub API, the stored token pointing to stored key 'hubToken-[NETWORK]' is deleted.
- Some minor improvements were made to common storage calls `saveLocal` and `getLocal` to allow for using of `ids` alongside keys, so deleting entries with a shared key is a lot easier.
 
Refactored calls to HDProvider.reset to single place in getHubAuthToken.

- [x] Completes #(CS-3004)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/150995647-d29b93a7-f696-4a75-94d3-c27b11278fb2.mp4


oidoug:
Note: Just notice it's not yet handling if the saved token is invalidated. If a request errors on a token we should try to re-create it.

cardstack/src/services/hub-service.ts
```diff
@@ -355,3 +405,19 @@ export const getWyrePrice = async (
...
+
+axios?.interceptors?.response?.use?.(undefined, async (error: AxiosError) => {
+  if (error?.response?.status !== 401) {
+    return Promise.reject(error);
+  }
+  // Got API error 401, auth token should be refreshed.
+
+  // Remove token from local storage
+  const network: Network = await getNetwork();
+  await removeLocal(hubTokenStorageKey(network));
```
topwebtek7: Can we filter only Hub Api calls here? I think we're using Axios for other api calls as well like Wyre, OpenSea, etc..

package.json
```diff
@@ -194,7 +194,7 @@
     "react-native-dynamic": "^1.0.0",
     "react-native-extra-dimensions-android": "1.2.2",
     "react-native-fast-image": "^8.1.5",
-    "react-native-flipper": "^0.114.0",
+    "react-native-flipper": "^0.130.0",
```
topwebtek7: I think will need to add Pod lock file change as well if upgraded deps

cardstack/src/services/hub-service.ts
```diff
@@ -36,6 +43,38 @@ const axiosConfig = (authToken: string) => {
...
+) => {
+  // expires in a day.
+  await saveLocal(
+    tokenStorageKey,
+    { data: { authToken } },
+    1000 * 3600 * 24,
+    undefined,
+    walletAddress
+  );
+};
```
topwebtek7, line 76: hmm In my opinion, it's not a good idea use `walletAddress`  as `version` parameter of `saveLocal` function.. I think version parameter is meant to be used when need to change specific asyncStorage key's behavior/structure at some point of app updates.
Instead, we can generate keys with address, network value, like as `getKey` function does

```
export const getKey = (prefix, accountAddress, network) =>
  `${prefix}-${toLower(accountAddress)}-${toLower(network)}`;
```

cardstack/src/services/hub-service.ts
```diff
@@ -64,9 +116,11 @@ export const getHubAuthToken = async (
     // didn't use Web3Instance.get and created new web3 instance to avoid conflicts with asset loading, etc that uses web3 instance
     const web3 = new Web3(hdProvider);
     const authAPI = await getSDK('HubAuth', web3, hubURL);
-    // load wallet address when not provided as an argument(this keychain access does not require passcode/biometric auth)
-    const address = walletAddress || (await loadAddress()) || '';
     const authToken = await authAPI.authenticate({ from: address });
+    await HDProvider.reset();
```
topwebtek7, line 120: The idea behind of `getHubAuthToken` didn't have `await HDProvider.reset();` inside of it was - just to improve performance of getHubAuthToken a little bit, as we can reset HDProvider(resetting HDProvider's async action) even after authToken is returned, just FYI.. but seems it's really tiny delay so looks ok

topwebtek7 (commented):
Looks good so far - left some comments

cardstack/src/services/hub-service.ts
oidoug: Oh, good catch. I think this interceptor only acts on this file's context, but just to be sure I'll add a specific instance of axios for the hub calls.

package.json
oidoug: I don't think this upgrade is needed now, reverting.

cardstack/src/services/hub-service.ts
oidoug: I updated the `saveLocal` signature, so the `walletAddress` in this case is actually an ID that will be used alongside the key to store the token. Version parameter is set to `undefined` (so will use the default value).

I added this ID (supported by rn-storage) so it's easier to remove a group of entries by a shared key - that in this case uses the current network as key and wallletaddress as id.

cardstack/src/services/hub-service.ts
oidoug: Now getHubAuthToken pull the auth token from storage instead of requesting it from HDProvider.
That has drastically speedup the token pullup process, so leaving reset here isn't an issue anymore.

oidoug:
@topwebtek7 , I wasn't able to update the jest tests to work with axios's custom instance. Since we're probably going to refactor this to use RTK, I don't want to spend too much more time on this, so I just used the instance in the untested calls. Let me know if you are ok with this or if you would prefer it differently.

cardstack/src/services/hub-service.ts
```diff
@@ -36,6 +43,40 @@ const axiosConfig = (authToken: string) => {
   };
 };
 
+export const hubapi = axios.create(axiosConfig(''));
```
danibonilha: `hubApi`

yarn.lock
```diff
@@ -17647,10 +17647,10 @@ react-native-fast-image@^8.1.5:
...
   integrity sha512-AJ0b4BEswRwa0bh4SibYUtXszEiaO88Lf4CZ1ib+t5ZfkAgsMk9Liv3L0LYnDblMJmSeGTr1+2ViIM8F2vamjg==
 
-react-native-flipper@^0.114.0:
-  version "0.114.0"
-  resolved "https://registry.yarnpkg.com/react-native-flipper/-/react-native-flipper-0.114.0.tgz#e68064f6c29caf88b2adf751f4d889033c457f5c"
-  integrity sha512-/D87UlXE0k4elX0Cn3Ip73InBdn3IXx37efqVLzWCxL5cqWU5eUFY9RVvXQgmjS5o9YHzllfZjqvJqtiInKnkA==
+react-native-flipper@^0.130.0:
+  version "0.130.0"
+  resolved "https://registry.yarnpkg.com/react-native-flipper/-/react-native-flipper-0.130.0.tgz#91e907257000af514c4173c8f7b748943c08cf5c"
+  integrity sha512-F6UzGTRSy24tEFroQGxSfm6+7qL3LuV8RYG3O/91vGzUKX+IBZdRdEJWeuELk/6WBtFAUuWEE7mlIWe0ZwwrQQ==
```
danibonilha: How is updating flipper related to this task ? If we do need to update I think the package.json change is missing then.

danibonilha (approved):
LGTM, just some small nits, and regarding the tests we can skip for this instance since it’s temporary

yarn.lock
oidoug: It's not needed, I had issues and tried updating, but it wasn't related. Reverting...

PR #614, opened 2022-01-27
Fix merchant claim notification handling and show local notification on same network and EOA only

topwebtek7:
### Description

- Fixed merchant claim notification handling so it can open Merchant Screen properly from notification when app closed state
- Displayed local notification when notification's network and EOA's matched

Tried to switch account when different EOA notification banner pressed, but it has some difficulties, I think we can tackle this in another ticket

- [x] Completes #CS-2617

### Checklist

### Screenshots

Jan-28-2022 19-13-55

### DIFF
--- a/cardstack/src/notification-handler/index.ts
+++ b/cardstack/src/notification-handler/index.ts
@@ -61,20 +61,17 @@ const notificationConfig: Record<NotificationType, NotificationConfig> = {
 };
 
 export const notificationHandler = async ({ data }: NotificationInfoType) => {
-  logger.log('notifi====== ', data);
-
   if (data?.notificationType && data.network && data.ownerAddress) {
     const currentNetwork = await getNetwork();
     const address = await loadAddress();
-    logger.log('currentNetwork====== ', currentNetwork, address);
     const { notificationType, network, ownerAddress } = data;
 
-    // handle notification only in same network
+    // handle notification in same network only
     if (currentNetwork === network) {
       if (address === ownerAddress) {
         notificationConfig?.[notificationType]?.handler(data);
       } else {
-        // if different EOA switch account and then handle notification
+        // if different EOA with notification address, switch account to correct one and then handle notification
       }
     }
   }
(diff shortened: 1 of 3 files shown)


cardstack/src/notification-handler/index.ts
```diff
@@ -58,10 +60,20 @@ const notificationConfig: Record<NotificationType, NotificationConfig> = {
   },
 };
 
-export const notificationHandler = ({ data }: NotificationInfoType) => {
-  if (data?.notificationType) {
-    const { notificationType } = data;
-    notificationConfig?.[notificationType]?.handler(data);
+export const notificationHandler = async ({ data }: NotificationInfoType) => {
+  if (data?.notificationType && data.network && data.ownerAddress) {
```
danibonilha, line 64: Should we use optional chain for data.network and data.ownerAddress ? Or we are always sure those fields are returned ?

cardstack/src/notification-handler/merchantClaim.ts
```diff
@@ -12,16 +18,23 @@ export const merchantClaimHandler = async (
...
+
+      const extendedMerchantSafe = await updateSafeWithTokenPrices(
+        await updateMerchantSafeWithCustomization(merchantSafe),
+        nativeCurrency || NativeCurrency.USD
+      );
+
+      const revenueBalances = await getRevenuePoolBalances(
+        merchantSafe.address,
+        nativeCurrency
+      );
```
danibonilha: We should probably map this to be called only if the merchantData doesn't exist, bc if the app is on background, or open, there's no need to refetch all this info since we already have the data

danibonilha (commented):
It's looking good, my only concern is regarding the unnecessary merchant call

cardstack/src/notification-handler/index.ts
topwebtek7: we're always sure these fields are included in notification data, as these info are required to handle notifications

danibonilha (approved):
LGTM

PR #615, opened 2022-01-27
[CS-2995] Investigate feasibility of onramper

oidoug:
`Onramper` works both with a Widget and an API, so in theory, we could integrate it with our own UI, but only for the first part of the purchase, once users select the gateway we'll have to open a Webview anyways.

That said, the widget looks simpler to integrate and will lead to less maintenance on our end.

While opening the widget, we can set what wallets the funds will be deposited to.

Do we have a production key already? If not, we can request one here: https://docs.google.com/forms/d/e/1FAIpQLSdnmTskkGA5QJGjC1eVRcqXZouuGe_ojltlBFs5nFClrSl_gA/viewform

This PR is a simple POC for integrating Onramper by using the standard Linking API from RN. It's missing our apiKey, so I'm not sure if it will work properly or not. It can be accessed inside our Developer Settings, as in the video below.

- [x] Completes #(CS-2995)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/151373559-50c4d47d-d356-4a09-b78b-6cb794ebfde8.mp4

### DIFF
--- a/src/components/settings-menu/DeveloperSettings.js
+++ b/src/components/settings-menu/DeveloperSettings.js
@@ -38,7 +38,7 @@ const DeveloperSettings = () => {
    */
   const openOnramperWebview = useCallback(async () => {
     const params = {
-      apiKey: 'OUR_API_KEY',
+      // apiKey: 'OUR_API_KEY',
       color: '00EBE5', // button color
       wallets: `ETH:${selectedWallet?.addresses?.[0].address}`,
       isAddressEditable: 'false',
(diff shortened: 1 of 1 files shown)


danibonilha:
If this is this a POC should we merge it anyways ? @kdowlin what's the recommendation regarding this ?

oidoug:
Not merging this since it's just a POC, and we'll likely not move forward with `onramper` atm.

PR #616, opened 2022-01-27
[CS-2988] Update Service Degradation Banner to have same border radius as the "Prepaid Cards are denominated in Spend" box

oidoug:
### Description

Using 10pt for the radius of the notice too.

It would be better to have a generic Card component with these styles attributes.

- [x] Completes #(CS-2988)

### Checklist

### Screenshots

Screenshot_20220127-114453_Card Wallet

Screenshot_20220127-114444_Card Wallet


danibonilha (approved):
LGTM

topwebtek7 (approved):

PR #618, opened 2022-01-27
[CS-2693]: Automate Android Release 

danibonilha:
### Description

This PR sets up the android build and release to google play internal track, it works locally and it has the initial github action .yml, which I'm not able to test until we merge this PR, so, for this one we should only consider the local commands, build 30 was uploaded by running `android:publish:internal`

image


.github/workflows/release-android.yml
```diff
@@ -0,0 +1,71 @@
...
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+      - name: Yarn Install
+        run: yarn install --network-timeout 300000 --network-concurrency 1
+      - name: Yarn Setup
+        run: yarn postinstall
+      - name: Sync google_services
+        run: yarn contexts:sync:google_services
+      - name: Sync google_play_credentials
+        run: yarn contexts:sync:sync_google_play_credentials
```
oidoug: nit: I think this action is missing in package.json.

lukemelia (approved):
👍  Let's give it a try!

.github/workflows/release-android.yml
danibonilha: Good catch!, it should be actually bundle exec fastlane.

oidoug (approved):

PR #620, opened 2022-01-28
[CS-2693]: Setup github actions for android release

danibonilha:
### Description
This PR fixes the build script to build and release Android to Play Store, for some weird reason Cryptex only works on macOS, and we also need to skip some prompts if we are running in the CI. 

### Screenshots

Screenshot 

Screenshot


oidoug (approved):
LGTM

PR #621, opened 2022-01-31
[Chore]: Clean up package.json scripts

danibonilha:
### Description

Clean up unused package json commands and simplified some fastlane commands, we can just run `yarn fastlane {desired_lane}` instead of having all commnds in the package.json, if we eventually find ourselves using multiple manual commands we can re-add to the package.json, also removes all the detox stuff, once we setup it again we can add needed commands.


oidoug (approved):
Great!

PR #622, opened 2022-02-02
[CS-3024]: Improve dmg logs, createWallet  and remove migrations

danibonilha:
### Description

While investigating the damaged wallet error, I noticed there was a bunch of code we didn't need, from some rainbow older PRs it seems they had a diff way to create a wallet once they changed it, they needed to keep it backwards compatible, hence all the migrations the app had, in our case all the wallets were created using the new structure, so we don't seem to need those migrations.

This PR:

- Removes migrations
- Makes sure to pass a address to getHubAuthToken (in the refreshToken scenario no address was provided and we faced an error)
- Improves damaged wallet log, it used to say it the wallet was damaged even before having a wallet ready, now we first check if the app has entirely loaded before verifying the wallet integrity 
- Splits createWallet into smaller functions to make it more readable. 

I'd recommend we all test this internally, it messes with all the wallet creation, acc creation, backup, fresh account, imported account, I tested as many scenarios as I could but we need to validate if there isn't any race conditions 

PS: The `addAccountsWithTxHistory` doesn't seem to work on layer2, the behavior is the same before or after this PR, needs investigation.


src/model/wallet.ts
```diff
@@ -498,6 +498,119 @@ export const getWalletByAddress = ({
   return wallet;
 };
 
+interface SaveSeedAndPkeyParams {
+  walletSeed: string;
+  pkey: string;
```
lukemelia: consider renaming this to `privateKey`. `pkey` is not that descriptive, since both "private" and "public" start with "p"

lukemelia (approved):
Nice deletions! 🙌

src/model/wallet.ts
danibonilha: good call

danibonilha:
This is on hold, just until @gabiburger has time to validate some important flows

oidoug (approved):
Awesome work! Worked perfectly and apparently, it loaded faster! 👏

PR #623, opened 2022-02-03
[CS-3067]: Bump app versions 

danibonilha:
### Description
Bumps ios to next version to fix testflight builds, and bumps  android to current version to match iOS's.


lukemelia (approved):
🚀

PR #625, opened 2022-02-03
[CS-3057]: Fix restore multiple google drive backups

danibonilha:
### Description

On android we were adding `${REMOTE_BACKUP_WALLET_DIR}` before the file name, and this is only needed for the UserData.json, after a generated backup the filename already contains  the `REMOTE_BACKUP_WALLET_DIR`, so the path was duplicated and it was never being matched,  now with the filenameMatcher we check all possible matches to make sure the document is found. 

- [x] Completes #(CS-3057)

### Screenshots

android-backup


lukemelia (approved):
🙏  Thanks!

PR #626, opened 2022-02-03
[CS-2775]: Load wallet state after updating name

danibonilha:
### Description

This PR fixes the wallet name not updating the UI after editing the name .

### Screenshots

Simulator Screen Recording - iPhone 13 mini - 2022-02-03 at 16 46 19


lukemelia (approved):
One-liner...

PR #630, opened 2022-02-07
[CS-3075]: Fix crash on backup flow after failing pass fetch

danibonilha:
### Description

`BackupConfirmPasswordStep` was using a Cardstack Text component, but receiving invalid props, thus the crash, but this screen also had an outdated layout, so I refactored it and shared a couple of components.

### Checklist

### Screenshots

Before: 
Screenshot

After: 
Screenshot

PS: The minimum text is hidden bc there's a valid password, we just can't see it filled bc of the device's protected screenshot

### DIFF
--- a/src/components/backup/BackupCloudStep.tsx
+++ b/src/components/backup/BackupCloudStep.tsx
@@ -9,29 +9,19 @@ import React, {
   useState,
 } from 'react';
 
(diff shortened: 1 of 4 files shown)


oidoug (approved):
Tested and all seems fine, good job!
FYI, I added a minor commit changing one last "square" (􀍢) Unicode emoji left behind.

PR #631, opened 2022-02-07
[CS-2813] Android assets cards blend mode

oidoug:
### Description

Android default blend mode for opacity makes some views weird when using `TouchableOpacity`. This PR adds two props to favor compositing "offscreen", to all affected container views under a Touchable, in all components listed in the AssetList screen.

~~I tried creating a wrapper view for both the touchable and container, but I was getting more complex than necessary, especially to use it in the `PrepaidCard` component so I scrapped it.~~ Added wrap specifically for the container with the opacity props, so we don't forget why they are there.

- [x] Completes #(CS-2813)

### Checklist

### Screenshots

Now:

https://user-images.githubusercontent.com/129619/152837008-076aeeca-d28c-439d-8f97-00fbeae02bee.mp4

Before:

https://user-images.githubusercontent.com/129619/152837034-3877bea7-aa51-44b3-a48e-f0cb467012b8.mp4

### DIFF
--- a/cardstack/src/components/BalanceCoinRow/BalanceCoinRow.tsx
+++ b/cardstack/src/components/BalanceCoinRow/BalanceCoinRow.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 
 import { AssetWithNativeType } from '../../types';
 import {
+  BetterOpacityContainer,
   CenteredContainer,
   Container,
   Icon,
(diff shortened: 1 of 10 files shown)


topwebtek7 (approved):
checked on both iOS and Android and looks good to me! 👍🏼

danibonilha (commented):
My only concern is related to performance, bc the docs states that `renderToHardwareTextureAndroid ` should be set to false after interactions  => `The downside is that this can use up limited video memory, so this prop should be set back to false at the end of the interaction/animation.` and now it's set to true on multiple components in the first screen, maybe we should have a touchable for the prepaidcard only that we trigger the opacity and this flag onPressIn and  disable onPress out

oidoug:
@danibonilha, I didn't notice any difference, but you are right, I'll try your suggestion.

PR #632, opened 2022-02-07
[CS-2985]: Add UI loadingOverlay instead of using global state

danibonilha:
### Description

This PR fixes the deleting with PIN case, the problem was caused by using the global setIsWalletLoading state, which replaces the whole navigation with a `Portal` component. In order to prevent this issue on other flows I replaced all the possible global states with the loading overlay, unfortunately there's one piece missing, it's the restore/backup icloud flow within the Setting flow.  The settings flow is a complex nested navigator which turns out to be challenging to apply the loading overlay. Since the backup flow is going to be rewritten we can wait until then to remove this last occurrence of the Portal. I recorded just the deleting scenario, but I tested all the other ones but any new tests are welcome! 

### Screenshots

deleting-pin-acc

### DIFF
--- a/src/model/wallet.ts
+++ b/src/model/wallet.ts
@@ -39,7 +39,6 @@ import {
   isValidMnemonic,
 } from '../handlers/web3';
 import showWalletErrorAlert from '../helpers/support';
-import WalletLoadingStates from '../helpers/walletLoadingStates';
 import { EthereumWalletType } from '../helpers/walletTypes';
 import store from '../redux/store';
 import { setIsWalletLoading } from '../redux/wallets';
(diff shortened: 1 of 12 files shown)


src/model/wallet.ts
```diff
@@ -943,11 +928,7 @@ export const generateAccount = async (
       // Fallback to custom PIN
       if (!hasBiometricsEnabled) {
         try {
-          const { dispatch } = store;
-          // Hide the loading overlay while showing the pin auth screen
-          dispatch(setIsWalletLoading(null));
           userPIN = await authenticateWithPIN();
-          dispatch(setIsWalletLoading(WalletLoadingStates.CREATING_WALLET));
```
oidoug, line 950: Won't removing these store updates affect the `src/hooks/useWallets.js` hook? It's used a lot throughout the app and it depends on `isWalletLoading`.

danibonilha: The `isWalletLoading ` is only being used in the `Portal` and in the `BackupCloud`  components which are the two scenarios I've mentioned that we are not able to remove yet, also this `isWalletLoading` is really confusing because it doesn't mean that a wallet is loading, is more like, isLoadingOverlayBeingShown, the loading state of a wallet is tracked by the `walletReady` property.

oidoug (approved):
Tested and it looks like all is working fine.

PR #634, opened 2022-02-09
[CS-2818]: Fix "random" PIN screen on fresh account

danibonilha:
### Description

The issue was cause by not waiting all the setup before checking the notifications,  so the loading was dismissed and the user would get a random PIN screen that seemed related to the backup but it was not. it depends on #632, because the previous PR already handles the dismiss of the PIN screen. This PR: 

- Adds a custom message to the PIN screen so the user knows why it needs to authenticate.
- Waits all the setup to keep the loading and ask for the PIN after
- Fixes the sheet minHeight (the import sheet was to small), as its a global component, the current implementation has more priority than the WC sheet, since it will be redesigned 
- Adds different messages for each hub authentication step 
- Adds a loading while switching accounts 
- Removes the setTimeout from check notifications, because now we are waiting for the account to be initialized.
- Adds HD.reset on finally block to also reset in case of an error

### Screenshots

Part-1 :
android-creating-acc1
Part-2:
android-creating-acc2

### DIFF
--- a/cardstack/src/components/Sheet/Sheet.tsx
+++ b/cardstack/src/components/Sheet/Sheet.tsx
@@ -15,7 +15,7 @@ const styles = StyleSheet.create({
   wrapperBase: {
     justifyContent: 'flex-end',
     backgroundColor: 'white',
-    minHeight: '30%',
+    minHeight: '40%',
   },
 });
 
--- a/cardstack/src/services/hub-service.ts
+++ b/cardstack/src/services/hub-service.ts
@@ -27,7 +27,15 @@ import HDProvider from '@cardstack/models/hd-provider';
 import { getFCMToken } from '@cardstack/models/firebase';
 
 const HUBAUTH_PROMPT_MESSAGE =
-  'To enable notifications, please authenticate your ownership of this account with the Cardstack Hub server';
+  'Please authenticate your ownership of this account with the Cardstack Hub server';
+
+const hubAuthPromptMessages = {
+  notifications: {
+    register: `${HUBAUTH_PROMPT_MESSAGE} to enable notifications`,
+    unregister: `${HUBAUTH_PROMPT_MESSAGE} to unregister`,
+  },
+  default: HUBAUTH_PROMPT_MESSAGE,
+};
 
 const HUBTOKEN_KEY = 'hubToken';
 
@@ -83,7 +91,8 @@ export const getHubAuthToken = async (
   hubURL: string,
   network: Network,
   walletAddress?: string,
-  seedPhrase?: string
+  seedPhrase?: string,
+  keychainAcessAskPrompt: string = hubAuthPromptMessages.default
 ): Promise<string | null> => {
   // load wallet address when not provided as an argument(this keychain access does not require passcode/biometric auth)
   const address = walletAddress || (await loadAddress()) || '';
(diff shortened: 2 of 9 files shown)


src/hooks/useInitializeWallet.js
```diff
@@ -110,17 +110,13 @@ export default function useInitializeWallet() {
 
         hideSplashScreen();
         logger.sentry('Hide splash screen');
-        initializeAccountData();
+        await initializeAccountData();
```
oidoug, line 113: Nice catch!

cardstack/src/services/hub-service.ts
```diff
@@ -125,6 +133,8 @@ export const getHubAuthToken = async (
     logger.sentry('Hub authenticate failed', e);
 
     return null;
+  } finally {
+    await HDProvider.reset();
```
oidoug, line 137: Nice!

oidoug (approved):
Creating an account flow looks better now, with fewer screen changes. 👍

cardstack/src/components/Sheet/Sheet.tsx
```diff
@@ -15,7 +15,7 @@ const styles = StyleSheet.create({
   wrapperBase: {
     justifyContent: 'flex-end',
     backgroundColor: 'white',
-    minHeight: '30%',
+    minHeight: '40%',
```
topwebtek7, line 18: Any specific reason for this change? basically I changed to 30% before as bottom WC notification sheet had too much space with just small text..

src/hooks/useInitializeWallet.js
```diff
@@ -110,17 +110,13 @@ export default function useInitializeWallet() {
...
 
         dispatch(appStateUpdate({ walletReady: true }));
 
-        logger.log('💰 Wallet initialized');
-        setTimeout(
-          () => {
-            checkPushPermissionAndRegisterToken(walletAddress, seedPhrase);
-          },
-          isNew ? 1000 : 0 // hub auth fails if we try too soon after wallet created
-        );
```
topwebtek7, line 123: just curious if hub auth works immediately as soon as wallet is created with seedPhrase.. I'm testing more on my side now but we need to make sure if this works properly, I think.

topwebtek7 (commented):
Looks good, left a couple comments

cardstack/src/components/Sheet/Sheet.tsx
danibonilha: Yes, the smallest bottom modal before the wc connect was the import one, and the expected behavior is to be bigger, because it might have another section in the bottom modal, if you check the Add an existing account modal on develop, you'll see the layout has changed, I changed back because WC notification will be removed, if we want to have diff behavior for same sheets we should actually add a custom prop, which I don't think it's necessary right now as wc notificaiton will be removed

danibonilha: We can see how 30% it's a bit crowded 
image

topwebtek7: OK, got it. let's leave it for now.
btw WC notification will be removed? We will update the UI but won't remove the WC notification as far as I know, right? :)

src/hooks/useInitializeWallet.js
danibonilha: from my tests, being newly created, or imported it's been registering correctly in the hub, the imported one since it already has the decrypted key it doesn't prompt to authenticate, there's one issue with android with deriving accounts that is not related to this PR

cardstack/src/components/Sheet/Sheet.tsx
danibonilha: It will probably not be a sheet anymore, but will need to wait the designs

PR #635, opened 2022-02-09
[CS-3094]:Decrypt phrase to get right derived path

danibonilha:
### Description

The decryptedSeedPhrase was not being passed forward, thus the account derivation was wrong bc the phrase was never decrypted. 

### Screenshots

Screen Shot 2022-02-09 at 14 00 29

### DIFF
--- a/src/model/wallet.ts
+++ b/src/model/wallet.ts
@@ -929,15 +929,17 @@ export const generateAccount = async (
     }
 
     const seedData = await getSeedPhrase(id);
-    const seedphrase = seedData?.seedphrase;
+    let seedphrase = seedData?.seedphrase;
 
     if (userPIN) {
       try {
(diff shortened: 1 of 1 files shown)


src/model/wallet.ts
```diff
@@ -929,15 +929,17 @@ export const generateAccount = async (
...
       try {
-        const decryptedSeedphrase = await encryptor.decrypt(
+        const decryptedSeedPhrase = await encryptor.decrypt(
           userPIN,
           seedphrase
         );
-        if (!decryptedSeedphrase) {
+        if (decryptedSeedPhrase) {
+          seedphrase = decryptedSeedPhrase;
+        } else {
```
topwebtek7, line 942: Nice catch!

topwebtek7 (approved):
LGTM! 🚀

PR #637, opened 2022-02-10
[CS-3102]: Create new HomeScreen for TabBar

danibonilha:
### Description

This PR adds the new HomeScreen with the tabBar design, the address behavior was changed by @kdowlin requests, and the navigate icons doesn't work yet, because there's  other route in the tabNavigator, which will be mapped later. Also removed some unused code and add reusable components for the next screens.

### Screenshots

Screen Shot 2022-02-10 at 19 29 31

Screen Shot 2022-02-10 at 19 29 51

Screen Shot 2022-02-10 at 19 30 40

### DIFF
--- a/cardstack/src/components/DiscordPromoBanner/DiscordPromoBanner.tsx
+++ b/cardstack/src/components/DiscordPromoBanner/DiscordPromoBanner.tsx
@@ -5,7 +5,7 @@ import { Image, Touchable } from '@cardstack/components';
 import { screenWidth } from '@cardstack/utils';
 
 const layouts = {
-  maxWidth: screenWidth * 0.92, // magic number but it works good on big and smaller devices
+  maxWidth: screenWidth * 0.91, // magic number but it works good on big and smaller devices
 };
 
 export const DiscordPromoBanner = ({ onPress }: { onPress: () => void }) => (
--- a/cardstack/src/components/TransactionList/TransactionList.tsx
+++ b/cardstack/src/components/TransactionList/TransactionList.tsx
@@ -1,49 +1,23 @@
-import React, { memo, useCallback } from 'react';
+import React, { memo, useCallback, useEffect, useMemo, useRef } from 'react';
 import { useFocusEffect } from '@react-navigation/native';
-import {
-  RefreshControl,
-  SectionList,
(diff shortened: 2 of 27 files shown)


cardstack/src/components/DiscordPromoBanner/DiscordPromoBanner.tsx
```diff
@@ -5,7 +5,7 @@ import { Image, Touchable } from '@cardstack/components';
 import { screenWidth } from '@cardstack/utils';
 
 const layouts = {
-  maxWidth: screenWidth * 0.92, // magic number but it works good on big and smaller devices
+  maxWidth: screenWidth * 0.91, // magic number but it works good on big and smaller devices
 };
```
danibonilha, line 9: Decreased this one bc I realized it wasn't properly aligned

cardstack/src/components/TransactionList/TransactionList.tsx
```diff
@@ -55,39 +29,79 @@ export const TransactionList = memo(({ Header }: TransactionListProps) => {
...
     <SectionList
       ListEmptyComponent={
-        <ListEmptyComponent
-          text={`You don't have any\ntransactions yet`}
-          textColor="blueText"
-          hasRoundBox
-        />
+        // Use fallback to avoid flickering empty component,
+        // when fetching hasn't started yet
+        isLoadingTransactions || isLoadingFallback ? (
```
oidoug, line 89: nit: I got your work around here, but couldn't isLoadingTransactions be started as true in useFullTransactionList hook? (if it doesn't break other use cases.. :)

oidoug (approved):

cardstack/src/components/TransactionList/TransactionList.tsx
danibonilha: It could, but the way the transactions are implemented right now is a bit messy, and it breaks other flows, but hopefully this workaround is only temporarily

topwebtek7 (commented):
LGTM! 
@danibonilha can we create a new branch for navigation redesign and create prs to that branch until stable enough to merge into develop branch? that way would be easy to review prs as well with not much changes.. what do you think?

PR #639, opened 2022-02-11
[CS-2413] Update send transaction screen to include gas estimates

topwebtek7:
### Description

Updated transaction gas estimate to have native gas price as well.
I've noticed gas calc is not correct and also does not properly change when amount changes.

- [x] Completes #(Linear ticket)

### Checklist

### Screenshots

Screen Shot 2022-02-11 at 10 05 22 PM
Screen Shot 2022-02-11 at 9 21 52 PM


danibonilha (approved):
LGTM

PR #640, opened 2022-02-11
[CS-3109]: Redesign settings screen and share on both navigator 

danibonilha:
### Description

This PR enables the scrollview in the settings and restyle it to be a screen from left to right to avoid gestures conflicts with the scroll, also adds the possible shareable routes into the tabBar navigator, which will be refactored soon to specific navigators based on each flow.  This change is applied on the current app and in the tabBar

### Screenshots

Simulator Screen Recording - iPhone 13 mini - 2022-02-11 at 17 05 01


topwebtek7 (approved):
LGTM 🚀

PR #642, opened 2022-02-14
[CS-3101] CardPressable 

oidoug:
### Description

Adds a new CardPressable component that can be used to safely utilize Android's hardware blending mode only when animating the component.

Also, this PR refactors all uses of BetterOpacityContainer and removes it, in favor of CardPressable.

- [x] Completes #(CS-3101)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/153925904-7e09dbae-8514-49fd-9093-aaa0214247f6.mp4


cardstack/src/components/CardPressable/CardPressable.tsx
```diff
@@ -0,0 +1,116 @@
...
+}: CardPressableProps & RestyleProps) => {
+  const [animating, setAnimating] = useState(false);
+  const animatedValue = useRef(new Animated.Value(Animate.out)).current;
+
+  const onPressOpacity = useCallback(
+    (toValue: Animate) => () => {
+      setAnimating(true);
+      Animated.timing(animatedValue, {
+        toValue,
+        duration: toValue === Animate.in ? 50 : 120,
```
danibonilha: nit: we could use interpolate to avoid this type of ternary

cardstack/src/components/CardPressable/CardPressable.tsx
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export const CardPressable = ({
+  children,
+  enableHapticFeedback = false,
+  hapticType = 'selection',
+  onPress,
+  ...rest
+}: CardPressableProps & RestyleProps) => {
+  const [animating, setAnimating] = useState(false);
```
danibonilha: nit: I'm wondering, bc Idk for sure so, to achieve the expected behavior does this flag needs to be a state, or a useRef could be enough ?

danibonilha (approved):
Nicely done! 👏 , just two small nits

cardstack/src/components/CardPressable/CardPressable.tsx
oidoug: These values are durations, and I think they can't be interpolated. Or am I missing something? 🤔

danibonilha: oh my bad, I misread. 
So to make it clear maybe  we could add:
 `duration: toValue === Animate.in ? duration.fast : duration.slow`

cardstack/src/components/CardPressable/CardPressable.tsx
oidoug: I had the impression the component needed to re-render for the flag to update, but I just tested with a ref and it worked, so thanks for the tip :)

PR #646, opened 2022-02-15
[Bug]: Fix QRCrosshair icon misplace

danibonilha:
### Description

This PR fixes the misplaced qrCode icon, ~~now that icons are pressables, the absolute positioning maybe me challenging, so we changed to aligned the text as absolute and we can evaluate the need to change the` <Icon />` to support this eventually~~  by replacing two icon wrappers (TouchableOpacity/Container) with one simple` <Touchable>`  component which has all the styles options.

### Screenshots

image

### DIFF
--- a/cardstack/src/components/Icon/Icon.tsx
+++ b/cardstack/src/components/Icon/Icon.tsx
@@ -3,12 +3,11 @@ import React from 'react';
 import Feather from 'react-native-vector-icons/Feather';
 import MaterialCommunity from 'react-native-vector-icons/MaterialCommunityIcons';
 
-import { TouchableOpacity } from 'react-native';
+import { Touchable, TouchableProps } from '../Touchable';
 import { CustomIconNames, customIcons } from './custom-icons';
 import { FeatherIconNames } from './feather-icon-names';
 import { MaterialCommunityIconNames } from './material-community-icon-names';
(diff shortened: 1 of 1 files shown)


topwebtek7:
@danibonilha In my opinion, updating Icon component would be better rather than fixing just this issue.. basically most icons  we are using now does not need Pressable and will be wrapped with TouchableOpacity unnecessarily.. hmm I think we can simply update Icon component so it supports Pressable only when has onPress prop.

danibonilha:
> @danibonilha In my opinion, updating Icon component would be better rather than fixing just this issue.. basically most icons we are using now does not need Pressable and will be wrapped with TouchableOpacity unnecessarily.. hmm I think we can simply update Icon component so it supports Pressable only when has onPress prop.

Alright, it seems the main issue was to have a TouchableOpacity wrapping the restyled container, which didn't allow custom styles to be passed forward, the TouchabeOpacity is the same as view when disabled, so what do you think about, instead of adding a ternary condition to choose what to wrap, we just use the `Touchable` as the default container, since it's a restyle component, it will have all the container features, plus the option to be pressed, I updated the PR to reflect this, lmk your thoughts @topwebtek7

oidoug (approved):
LGTM

topwebtek7 (approved):
🚀

PR #647, opened 2022-02-16
[CS-3177]: Fix infinite activities loading on empty state

danibonilha:
### Description
Silly mistake, we were comparing the object as the flag instead of the value itself 

- [x] Completes #(CS-3177)

### Screenshots

image


topwebtek7 (approved):
LGTM

PR #648, opened 2022-02-16
[CS-3105]: Update android store paths on force update screen

danibonilha:
### Description

This PR adds the right links for android, removes unnecessary ios links, and bumps the android version 

PS: The links don't work on iOS simulator, only on devices

### Screenshots

upda-android-link

### DIFF
--- a/cardstack/src/components/MinimumVersion/MinimumVersion.tsx
+++ b/cardstack/src/components/MinimumVersion/MinimumVersion.tsx
@@ -18,6 +18,11 @@ const strings = {
   button: 'Update',
 };
 
+const screenWidthPercentage = {
+  '100': screenWidth,
+  '110': screenWidth * 1.1,
+};
+
 // Not quite sure about these magical numbers but it works
 const Constants = {
   iconSize: screenWidth * 0.25,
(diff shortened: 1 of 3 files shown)


cardstack/src/components/MinimumVersion/MinimumVersion.tsx
```diff
@@ -10,23 +10,17 @@ import {
...
-    uri: `itms-apps://apps.apple.com/app/1549183378?mt=8`,
-    url: `https://apps.apple.com/app/1549183378?mt=8`,
-  },
-  testFlight: {
-    uri: `itms-beta://beta.itunes.apple.com/v1/app/1549183378?mt=8`,
-    url: `https://beta.itunes.apple.com/v1/app/1549183378?mt=8`,
-  },
+const screenWidthPercentage = {
+  '100': screenWidth,
+  '110': screenWidth * 1.1,
```
topwebtek7, line 23: Just for more context, any reason why screen width is diff on iOS and Android

topwebtek7 (approved):
LGTM, just a comment out of curiosity

cardstack/src/components/MinimumVersion/MinimumVersion.tsx
danibonilha: I believe is the way the platform manage images differently but not quite sure, doesn't make much sense, but on all android devices, the image was cropped  🤷🏻‍♀️

PR #653, opened 2022-02-18
[CS-3201] Using shared Sheet component for show QR code screen

oidoug:
### Description

Refactors screen ShowQRCodeModal to use navigation options `sheetPreset` and also `Sheet` component for better display on Android, since Sheet already considers StatusBar.currentHeight.

Also, added `cardBackgroundColor` prop to `Sheet` so we can match previous style for ShowQRCodeModal.

- [x] Completes #(CS-3201)

### Checklist

### Screenshots

Screenshot

### DIFF
--- a/cardstack/src/screens/ShowQRCodeModal.tsx
+++ b/cardstack/src/screens/ShowQRCodeModal.tsx
@@ -60,7 +60,6 @@ export const AmountQRCode = ({
         justifyContent="space-between"
       >
         <Container alignItems="center">
-          {/* <SheetHandle /> */}
           <Text size="body" color="black" fontWeight="bold" marginTop={4}>
             Scan to Pay
           </Text>
--- a/cardstack/src/screens/ShowQRCodeModal.tsx
+++ b/cardstack/src/screens/ShowQRCodeModal.tsx
@@ -8,8 +8,7 @@ import {
   Touchable,
   Container,
   QRCode,
-  SafeAreaView,
-  SheetHandle,
+  Sheet,
   Text,
 } from '@cardstack/components';
 import { MerchantInformation } from '@cardstack/types';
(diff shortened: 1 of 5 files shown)


cardstack/src/screens/ShowQRCodeModal.tsx
```diff
@@ -53,23 +52,15 @@ export const AmountQRCode = ({
...
       <Container
         flex={1}
         flexDirection="column"
         alignItems="center"
         justifyContent="space-between"
-        paddingTop={4}
       >
         <Container alignItems="center">
-          <SheetHandle />
+          {/* <SheetHandle /> */}
```
topwebtek7: As we converted to Sheet component, we don't need SheetHandle and I think safely can remove this part

topwebtek7: Feb-18-2022 23-12-52
On iOS device, opening animation of the ShowQRCodeModal looks a bit weird to me(especially ShowQRCodeModal downside corners).. maybe can we keep RequestPayment screen size, not zooming out when goes behind the ShowQRCodeModal? @oidoug 

@kdowlin @danibonilha thoughts?

oidoug: Thanks for checking that. Pretty odd, on my smaller iPhone and on my android it did fill up the whole screen 🤔 
I used the nav options `sheetPreset` we use on other screens, strange it behaves differently. I'll try to improve it!

danibonilha: Now that is a sheet, maybe we can consider to move `SHOW_QRCODE_MODAL` from globalRoutes to mainRoutes, because basically this division is to make the animation diff on iOS, but since we want to apply the same behavior on both platforms we might be able to move it now

danibonilha (approved):
It's looking good, just a small suggestion

PR #655, opened 2022-02-21
[CS-3088] Font weight work-around

oidoug:
### Description

This PR adds supported font-weight variations for font OpenSans-Regular on android.

Differently than on iOS, on Android, the only font used after Theme translation by `restyle` is `OpenSans-Regular`. This can be easily tested by trying to use the wrong font-weight on any Text, it will show an error with the available options for the font.

- [x] Completes #(CS-3088)

### Checklist

### Screenshots

Screenshot


danibonilha (commented):
The change looks good, but I noticed there are places we are using `fontWeight` and other `weight`,  in order to to apply this across the app we would need to replace the `fontWeight` with `weight` in all used Text reestyle components and not allow `Text` to receive `fontWeight` anymore to avoid making a android mistake, an easy way, is to apply this on the `Text.ts` line 13: 

```ts
export type TextProps = Omit<ShopifyTextProps<Theme>, 'fontWeight'> &
```

src/components/walletconnect-list/WalletConnectExplainer.js
```diff
@@ -25,10 +25,10 @@ const IconQRScanImg = styled(ImgixImage).attrs({
 const CustomText = props => (
   <Text
     color="blueText"
-    fontWeight={props.bold ? '500' : '400'}
     letterSpacing={0.32}
     size="xs"
     textAlign="center"
+    weight={props.bold ? 'bold' : 'regular'}
     {...props}
```
danibonilha: nit: This one already passes the props forward, we could remove this condition, and on `   <CustomText bold>` replace with `   <CustomText weight='bold'>`

danibonilha (commented):
small suggestion, and missing some eslint fixes, once it's green, it's good to go.

src/components/walletconnect-list/WalletConnectExplainer.js
oidoug: Good catch 👍

oidoug:
@danibonilha , I refactored all Texts to use weight and added the Omit as you suggested.

There are a few places that are still using `fontWeight` but they are either for Inputs or are not using the Theme (prob. older Rainbow implementation). If you don't mind I'll leave these cases behind for now.

danibonilha (approved):
🚀

PR #657, opened 2022-02-21
[CS-3149] Enable direct WC deeplink handling

topwebtek7:
### Description

Enabled direct WC deeplink handling on both ios&android

- [x] Completes #CS-3149

### Checklist

### Screenshots

- Android
Feb-21-2022 22-31-25
- iOS
Feb-21-2022 22-30-34

### DIFF
--- a/cardstack/src/utils/url-utils.ts
+++ b/cardstack/src/utils/url-utils.ts
@@ -1,0 +1,13 @@
+export const isEncodedUri = (uri: string): boolean => {
+  uri = uri || '';
+
+  return uri !== decodeURIComponent(uri);
+};
+
+export const fullyDecodeURI = (uri: string): string => {
+  while (isEncodedUri(uri)) {
+    uri = decodeURIComponent(uri);
+  }
+
+  return uri;
+};
--- a/cardstack/src/utils/url-utils.ts
+++ b/cardstack/src/utils/url-utils.ts
@@ -1,10 +1,7 @@
-export const isEncodedUri = (uri: string): boolean => {
-  uri = uri || '';
+export const isEncodedUri = (uri = ''): boolean =>
+  uri !== decodeURIComponent(uri);
 
-  return uri !== decodeURIComponent(uri);
-};
-
-export const fullyDecodeURI = (uri: string): string => {
+export const decodeNestedURI = (uri: string): string => {
   while (isEncodedUri(uri)) {
     uri = decodeURIComponent(uri);
   }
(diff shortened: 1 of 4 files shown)


cardstack/src/utils/url-utils.ts
```diff
@@ -0,0 +1,13 @@
...
+  return uri !== decodeURIComponent(uri);
+};
+
+export const fullyDecodeURI = (uri: string): string => {
+  while (isEncodedUri(uri)) {
+    uri = decodeURIComponent(uri);
+  }
+
+  return uri;
+};
```
danibonilha, line 10: I'm confused by this loop it seems the loop will only run once,  isn't  using `decodeURIComponent` on `handleDeepLink` enough to achieve the expected result ? Not sure we need this util, unless if it's for a very specific case, also, how can we generate the wc: request in order to test it ?

topwebtek7: WC link might be encoded multiple times cause it has url inside WC uri, i.e. `cardwallet%3A%2F%2Fwc%3Furi%3Dwc%3A4c1d2407-b74d-4d7f-bbff-574426859bbe%401%3Fbridge%3Dhttps%253A%252F%252Fsafe-walletconnect.gnosis.io%252F%26key%3Dd1a9705af10a94a885a83a3ff44b4f410e51f758b9ce32da5381a2fde9c311f2` and it will not be decoded properly when decode only once.

danibonilha: Got it, makes sense, I'll add a couple of suggestions then

cardstack/src/utils/url-utils.ts
```diff
@@ -0,0 +1,13 @@
+export const isEncodedUri = (uri: string): boolean => {
+  uri = uri || '';
+
+  return uri !== decodeURIComponent(uri);
+};
```
danibonilha:
```suggestion
export const isEncodedUri = (uri: string = ''): boolean => uri !== decodeURIComponent(uri);
```

cardstack/src/utils/url-utils.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const fullyDecodeURI = (uri: string): string => {
```
danibonilha: To be more straightforward on why we need this, we could rename this to be `decodeNestedURI` or add a comment

danibonilha (approved):
It's looking good, just a couple of suggestions

oidoug (approved):
👍

PR #659, opened 2022-02-22
[CS-3209]: Add initial TransferCard UI 

danibonilha:
### Description
This PR adds the TransferCard Screen and it handles the navigation to it

PS:We don't have designs for this, so this is a template just to get the functionality usable 
### Screenshots

Simulator Screen Recording - iPhone 13 mini - 2022-02-22 at 10 39 58

### DIFF
--- a/cardstack/src/components/PrepaidCard/PrepaidCard.tsx
+++ b/cardstack/src/components/PrepaidCard/PrepaidCard.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useCallback, useState } from 'react';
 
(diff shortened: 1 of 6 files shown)


oidoug (approved):
I know we don't have mocks for this feature, so it's more of a nit, but to me, this could look and work closer to what we have for other token transfers that are using `SendSheet.js`.

danibonilha:
> I know we don't have mocks for this feature, so it's more of a nit, but to me, this could look and work closer to what we have for other token transfers that are using `SendSheet.js`.

@oidoug  We are actually trying to move away from using Sheets with Inputs, this is kind of an experiment towards that, but I appreciate the suggestion.

PR #662, opened 2022-02-23
[CS-3127] Update PaymentRequest link share info

topwebtek7:
### Description

Updated PaymentRequest link share info.

- [x] Completes #CS-3127

### Checklist

### Screenshots

Screen Shot 2022-02-23 at 9 56 56 PM
Screen Shot 2022-02-23 at 9 52 20 PM

### DIFF
--- a/cardstack/src/utils/merchant-utils.ts
+++ b/cardstack/src/utils/merchant-utils.ts
@@ -1,13 +1,12 @@
 import { Resolver } from 'did-resolver';
 import { getResolver } from '@cardstack/did-resolver';
-import { Share } from 'react-native';
+import { Share, Platform } from 'react-native';
 import {
   convertAmountToNativeDisplay,
   convertRawAmountToBalance,
   MerchantSafe,
   subtract,
 } from '@cardstack/cardpay-sdk';
-import { Device } from './device';
 import {
   MerchantClaimFragment,
   MerchantRevenueEventFragment,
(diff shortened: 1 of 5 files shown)


oidoug (approved):
LGTM!

cardstack/src/utils/merchant-utils.ts
```diff
@@ -114,23 +114,30 @@ export const shareRequestPaymentLink = (
...
+      };
 
-  return Share.share(
-    {
-      message,
-      title,
-    },
-    options
-  );
+  return Share.share(content, options);
```
danibonilha: nit: a way to minimize checking the Device, is to use Platform.select, so we could replace this piece with: 
```ts


  const shareConfig = {
    options: Platform.select({
      android: {
        dialogTitle: title,
      },
      ios: {
        excludedActivityTypes,
        subject: title,
      },
    }),
    content: Platform.select({
      android: {
        title,
        message,
      },
      ios: {
        title,
        message,
        url: paymentRequestLink,
      },
      default: { message },
    }),
  };

  return Share.share(shareConfig.content, shareConfig.options);
```

danibonilha: Then on the tests we need to mock the platform select: 

```ts
  describe('shareRequestPaymentLink', () => {
    beforeAll(() => {
      jest
        .spyOn(Platform, 'select')
        .mockImplementation(obj => obj[Platform.OS]);
    });
```
And  replace  
- `Device.isIOS = true` with `Platform.OS = 'ios' `
- `Device.isIOS = false` with `Platform.OS = 'android' `

danibonilha (commented):
It's looking good, just a small suggestion to "shrinking" it, but it's nit

PR #663, opened 2022-02-23
[CS-3012] Notification icon on android

oidoug:
### Description

Added new assets for notification icon on android.

- [x] Completes #(CS-3012)

### Checklist

### Screenshots

Screenshot

Screenshot


danibonilha (commented):
I think we already have this asset as `ic_launcher_foreground`, you can check this [PR](https://github.com/cardstack/cardwallet/pull/553/files) to make sure, at the time I just added `ic_launcher` on `com.google.firebase.messaging.default_notification_icon` bc notifications were not implemented yet, I find odd that it didn't work, maybe it doesn't accept a `xml`, but looking at[ firebase's docs](https://github.com/firebase/quickstart-android/tree/master/messaging#custom-default-icon) it seems if we remove this property it should take the icon launcher as default, maybe we can try that and see how it goes.

oidoug:
@danibonilha, you're right, if we don't set any resource it builds it using ic_launcher, with foreground and background.

I'm pushing an update soon.

One thing we need to keep an open eye, notifications on android are taking a lot longer to arrive.

topwebtek7 (approved):
LGTM! 👍🏼

danibonilha (approved):
Thanks for looking into it! 🚢

PR #664, opened 2022-02-24
[CS-3227] Bump sdk to 0.28.19

topwebtek7:
### Description

Bump sdk to 0.28.19
(It includes sdk currency conversion bug fix)

- [x] Completes #CS-3227


oidoug (approved):
👍

oidoug:
I have a question related to the SDK. Although it works fine, I'm always seeing a warning related to cardpay-sdk that is a bit odd:

```warn Package @cardstack/cardpay-sdk has been ignored because it contains invalid configuration. Reason: Package subpath './package.json' is not defined by "exports" in /Users/douglas/Cardstack/code/cardwallet/node_modules/@cardstack/cardpay-sdk/package.json```

It shows right after I start metro. Do you know what that's about?

**Comment by topwebtek7** [comment]
I also have that warning and not sure exactly but I think we can add like this in sdk side to get rid of this warning
```
"exports": {
    "./package.json": "./package.json",
 ```

PR #665, opened 2022-02-24
[CS-3210]: Adds ability to transfer a prepaid card 

danibonilha:
### Description
This PR:

- Adds the mutation to transfer a prepaid card 
- Adds a new hook `useMutationEffects`, which can be reused across useMutation hooks to handle loading, success and error status 
- Resolves a TODO, to move payMerchant from api creation to prepaidcard services 
- Adds the ability to disable HDProvider on the finally block of `queryWrapper`
- Adds UI and logic test for transfer screen 
- Disables Scan QR Code until feature is built 

As I mentioned on the meeting, we can use this screen as a model for new screens where the structure/pattern is like this:
image
~~We could move even further if we want to extract the strings into it's own file too, the gain would be already having the strings out for i18n and for testing, we don't need to rely on the text itself, because in need to change one copy we are able 
 to just update the strings file without needing to mess with the tests and the screen file, but I'll leave it open to discussion.~~ (Moved to their own file, after team's discussion)

### Screenshots

The flow [GIF](https://linear.app/cardstack/issue/CS-3210#comment-fd95cd30) is on Linear bc it was too big for gtb.

Here I'll leave the test coverage: 

image

it's missing just the onScan function, which hasn’t been implemented yet.

### DIFF
--- a/cardstack/src/screens/TransferCardScreen/strings.ts
+++ b/cardstack/src/screens/TransferCardScreen/strings.ts
@@ -5,4 +5,16 @@ export const strings = {
   inputPlaceholder: 'Enter address (0x...)',
   scanQrBtn: 'Scan QR Code',
   transferBtn: 'Transfer',
+  loadingTitle: 'Transferring Prepaid Card...',
+  alert: {
+    btnLabel: 'Okay',
+    success: {
+      title: 'Success',
+      message: 'Your Prepaid Card has been transferred!',
+    },
+    error: {
+      title: 'Oops!',
+      message: 'Something went wrong',
+    },
+  },
 };
--- a/cardstack/src/screens/TransferCardScreen/strings.ts
+++ b/cardstack/src/screens/TransferCardScreen/strings.ts
@@ -1,0 +1,8 @@
+export const strings = {
+  title: 'Transfer Card',
+  subtitle:
+    'You can enter an EOA address or Scan it to transfer this PrepaidCard',
+  inputPlaceholder: 'Enter address (0x...)',
+  scanQrBtn: 'Scan QR Code',
+  transferBtn: 'Transfer',
+};
--- a/cardstack/src/screens/TransferCardScreen/useTransferCardScreen.ts
+++ b/cardstack/src/screens/TransferCardScreen/useTransferCardScreen.ts
@@ -1,19 +1,97 @@
-import { useCallback, useState } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 
+import { useNavigation, useRoute } from '@react-navigation/core';
 import { useSendAddressValidation } from '@rainbow-me/components/send/SendSheet';
+import { useAccountSettings, useWallets } from '@rainbow-me/hooks';
+import { RouteType } from '@cardstack/navigation/types';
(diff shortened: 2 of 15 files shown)


cardstack/src/screens/TransferCardScreen/useTransferCardScreen.ts
```diff
@@ -1,19 +1,97 @@
...
+            message: 'Something went wrong',
+          }),
+        },
+      }),
+      [isError, isSuccess, onTransferFinishedAlert]
+    )
+  );
+
+  const onTransferPress = useCallback(async () => {
+    showLoadingOverlay({ title: 'Transferring Prepaid Card...' });
```
oidoug: nit: Since you have your local strings file already, maybe you could move the text in this hook up there too.

cardstack/src/screens/TransferCardScreen/strings.ts
```diff
@@ -0,0 +1,8 @@
+export const strings = {
+  title: 'Transfer Card',
+  subtitle:
+    'You can enter an EOA address or Scan it to transfer this PrepaidCard',
+  inputPlaceholder: 'Enter address (0x...)',
+  scanQrBtn: 'Scan QR Code',
+  transferBtn: 'Transfer',
+};
```
oidoug: Question: Are you planning on using react-native-localization (https://github.com/stefalda/ReactNativeLocalization) for i18n?

Asking because how you are putting up text on a strings file looks more closely related to that lib.

There's another lib that I used before with a different approach, having a function `t("string")` that is used to translate and keeping all text in json files in a well organized i18n folder structure:
https://react.i18next.com/

cardstack/src/services/prepaid-cards/prepaid-card-api.ts
```diff
@@ -25,40 +29,33 @@ const prepaidCardApi = safesApi.injectEndpoints({
...
+      PrepaidCardTransferQueryParams
+    >({
+      async queryFn(params) {
+        return queryPromiseWrapper<
+          TransactionReceipt,
+          PrepaidCardTransferQueryParams
+        >(transferPrepaidCard, params, {
+          errorLogMessage: 'Error while transferring prepaid card',
+          resetHdProvider: true,
+        });
```
oidoug, line 58: All of this is a significant improvement! 👏  nice!
I'll have to study it more to fully comprehend, but looks like a better way forward!

oidoug (approved):
I couldn't test it fully because all my prepaid cards are not transferrable.
Regardless, excellent work as usual 🥇

cardstack/src/screens/TransferCardScreen/strings.ts
danibonilha: We haven't discussed anything related to i18n libs, we are totally open to evaluate libraries, this approach is just how I have used on other projects and regardless of the libraries we can still reference the strings, even with the second lib suggestion, like `t(strings.title)`

cardstack/src/screens/TransferCardScreen/useTransferCardScreen.ts
danibonilha: oh thanks, I forgot this one!

oidoug (approved):
🚀

PR #668, opened 2022-02-25
[CS-3221]: Fix wrong address and migrate expanded sheet to new sheet

danibonilha:
### Description

This PR fixes the payment confirmation screen which was showing the owner-address instead of the safe address, since we needed to update this screen I decided to refactor it to use the new sheet component ,while migrating it, though I noticed a bug was introduced on the MerchantTxSheet so I also migrated it to use the new sheet, it easier to get the context following the commit history, by refactoring this piece we were able to remove workaround we had previously built 💯 . Also we were passing all the section information to each txItem with no need, there are still some quirkiness in it but it's definitely better than before. Also it minimizes the long tokens name being cropped across the screen.

- [x] Completes #(CS-2401)
- [x] Completes #(CS-3221)

### Screenshots

The safe address: 
image

The `TO` matching the merchant safe address above: 
image

Tx screen on big and small devices: 
image

Sheet Behavior: 

Simulator Screen Recording - iPhone 13 mini - 2022-02-25 at 17 27 45


oidoug (approved):
Nice refactoring!

topwebtek7 (approved):
LGTM! 🚀

PR #669, opened 2022-03-01
[CS-3217] Setting a merchant as primary

oidoug:
### Description

Adds MerchantSafe primary selection feature:
- New setting's option "Payment Account" (not sure about the naming, just a suggestion)
- On new screen, you can select a primary Merchant Safe
- Redux slice for primary + `usePrimaryMerchant` hook for ease of use
- AssetList's `useMerchantSafeSection` hook updated to use primary Merchant Safe if set

[UPDATED]

Scope of PR chanted a bit, it now focused on providing a primary merchant safe. If none is set, `usePrimaryMerchant` will set the first safe created as primary.

ProfileScreen shows the primary safe as a placeholder for now but it will be replaced soon by the safe's information in a new PR.

UI for selecting primary was removed from settings and will be integrated into safes in a new PR.

- [x] Completes #(CS-3217)

### Checklist

### Screenshots

[UPDATED]

https://user-images.githubusercontent.com/129619/156761133-aea22804-c041-41db-8f77-bb51db49c188.mp4


oidoug:
I changed this PR to be the base work for the new profile. 

@topwebtek7 , once this is merged, you can use`usePrimaryMerchant` for your needs.

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -96,7 +96,8 @@ export const MerchantInfo = ({
         >
           {name || ''}
         </Text>
-        <Text variant="subText">Business Account</Text>
+        {/* Note: Will be later used for a different term. */}
+        {/* <Text variant="subText">Profile</Text> */}
```
danibonilha: Let's remove this, it's just a text anyway, if we want to re-add it's a easy one

cardstack/src/redux/hooks/usePrimaryMerchant.ts
```diff
@@ -0,0 +1,46 @@
...
+    data: { merchantSafes },
+    error,
+  } = useGetSafesDataQuery({
+    address: accountAddress,
+    nativeCurrency,
+  });
+
+  const changePrimaryMerchant = useCallback(
+    (merchant: MerchantSafeType) =>
+      dispatch(slideChangePrimaryMerchant({ primary: merchant })),
```
danibonilha: What does slide stands for ?

cardstack/src/redux/hooks/usePrimaryMerchant.ts
```diff
@@ -0,0 +1,46 @@
...
+    [dispatch]
+  );
+
+  // Ensures primary will always be valid if theres at least one merchant safe.
+  useEffect(() => {
+    if (merchantSafes?.length > 0 && !primaryMerchant) {
+      changePrimaryMerchant(merchantSafes[merchantSafes.length - 1]);
```
danibonilha: Isn't index 0 the latest merchant created ? Is it intentional to use the oldest one (merchantSafes.length - 1)?

cardstack/src/redux/hooks/usePrimaryMerchant.ts
danibonilha: If it's just a way to rename this, I'd suggest we keep as `changePrimaryMerchant` and this local function we name something like,  `setPrimaryMerchant`  or even `setPrimarySafeAccount` (since the merchant product context will not exist in the future)

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -1,11 +1,23 @@
...
+import usePrimaryMerchant from '@cardstack/redux/hooks/usePrimaryMerchant';
 
 const ProfileScreen = () => {
+  const { primaryMerchant } = usePrimaryMerchant();
+
   return (
     <Container backgroundColor="backgroundDarkPurple" flex={1}>
       <MainHeader title="PROFILE" />
+      <CenteredContainer flex={1}>
+        <MerchantSafe {...(primaryMerchant as MerchantSafeProps)} />
```
danibonilha: Why is this being casted ? If there are props not being used, they should be removed, or if they are needed for one piece but not the other we can make it optional

src/redux/store.ts
```diff
@@ -1,19 +1,38 @@
-import { configureStore } from '@reduxjs/toolkit';
+import AsyncStorage from '@react-native-async-storage/async-storage';
+import { combineReducers, configureStore } from '@reduxjs/toolkit';
 import { setupListeners } from '@reduxjs/toolkit/dist/query';
+import { TypedUseSelectorHook, useSelector } from 'react-redux';
+import { persistReducer, persistStore } from 'redux-persist';
 import reducers from './reducers';
+import { name as merchantPath } from '@cardstack/redux/merchantSlice';
```
danibonilha: Should we give a more meaningful name while exporting, can't it be `merchantPath` already ? Or `merchantSliceName`

cardstack/src/redux/hooks/usePrimaryMerchant.ts
```diff
@@ -0,0 +1,46 @@
...
+  const { accountAddress } = useAccountSettings();
+  const primaryMerchant = useTypedSelector(selectPrimaryMerchantSafe);
+
+  const {
+    data: { merchantSafes },
+    error,
+  } = useGetSafesDataQuery({
+    address: accountAddress,
+    nativeCurrency,
+  });
```
topwebtek7: just out of curiosity, why don't we get merchantSafes simply from redux selector?

cardstack/src/redux/merchantSlice.ts
```diff
@@ -0,0 +1,34 @@
...
+  name: 'merchant',
+  initialState,
+  reducers: {
+    changePrimaryMerchant(
+      state,
+      { payload }: PayloadAction<{ primary: MerchantSafeType }>
+    ) {
+      state.primary = payload.primary;
+    },
+  },
```
danibonilha: We should probably store this by network, because let's say I'm on sokol, it will set a sokol safe, as the primary merchant and save on the storage, once I switch to xdai even though the app is restarted the sokol safe will persist, so we either need to store by network or clean-up the primary safe if the network changes.

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
topwebtek7: just wanna confirm, is it also requested - getting rid of `Business Account` below of merchant name in Assets list

src/redux/store.ts
```diff
@@ -1,19 +1,38 @@
...
-    [safesApi.reducerPath]: safesApi.reducer,
-    [serviceStatusApi.reducerPath]: serviceStatusApi.reducer,
-  },
+  reducer: persistedReducer,
   middleware: getDefaultMiddleware => {
     const middlewares = getDefaultMiddleware({
+      // Note on redux-persist:
+      // If we eventually start to use serialization for redux, we need to ignore some actions
+      // from redux-persist `ignoredActions: [FLUSH, REHYDRATE, PAUSE, PERSIST, PURGE, REGISTER]`,
+      // reference: https://redux-toolkit.js.org/usage/usage-guide#use-with-redux-persist
```
danibonilha, line 34: 👍

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -1,11 +1,23 @@
...
+      </CenteredContainer>
```
danibonilha: Is the expect UI this one ? : 
image

If we want to match what is on the Wallet tab we can replace this `CenteredContainer` with a `Container` with `justifyContent="center"`

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
danibonilha: maybe on <MerchantSafe> , `MerchantSafeType` is enough as props, also we should probably check if `primaryMerchant` exists before trying to render it as it can be undefined

src/redux/reducers.js
```diff
@@ -54,5 +55,6 @@ export default {
   uniswapLiquidity,
   walletconnect,
   wallets,
+  merchant,
```
topwebtek7: Can we use same name for reducer name and reducer file name to avoid possible confusion?

danibonilha (commented):
I'm glad we simplified this a lot, there's still some things to validate, but nice one

topwebtek7 (commented):
@oidoug I don't think `usePrimaryMerchant` is working as expected cuz tested and it returns before account's primary Merchant after switched account/network as persisted before state.. I think we need to confirm how to use redux-persist. IMO, we need to purge redux-persist when switch account/network

cardstack/src/redux/hooks/usePrimaryMerchant.ts
oidoug: Ok, it makes sense. Do you think it's better to rename it all in this new code from Merchant to SafeAccount?

cardstack/src/redux/hooks/usePrimaryMerchant.ts
oidoug: Per discussion with Kiel, we want to display by default the first safe created.

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
oidoug: Good point, the casting isn't necessary if some of the unused props are optional and if `primaryMerchant` exists.

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
oidoug: Ok! This will be replaced soon, just wanted to have something there.

cardstack/src/redux/hooks/usePrimaryMerchant.ts
oidoug: MerchantSafes are now being pulled and stored by rtk-query, so we need to ask rtk-query for that resource and it will pull cached data from the store if available, if not it will query the server.

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
oidoug: I think we already did remove it in the current merged develop. But I'll double check with Kiel if not.

src/redux/reducers.js
oidoug: I didn't want to have the store pathname for this to be `merchantSlice`, makes better sense to just be `merchant`. And the exported name here is used for type-checking the path, so I prefer to leave it as `merchant` if you are ok if it.

I'm also considering renaming merchant to "safeAccount". Wdyt?

cardstack/src/redux/merchantSlice.ts
oidoug: You're right, I didn't consider that. I think the best way to do that is to just have the network as an optional MerchantSafeType value, and validate it in the usePrimaryMerchant hook before returning it. Wdyt?

src/redux/reducers.js
topwebtek7: just merchant sounds good to me but I think we can discuss that naming in the meeting?

kdowlin: I think we should likely leave all references to merchant for the time being - until we implement the generic safes at the protocol level- at that time we will know where we can rename.

oidoug:
> IMO, we need to purge redux-persist when switch account/network

Yes, you're right, that didn't cross my mind. I'll check for the network before setting the primary.

Redux-persist only job is to save whatever we have as the current global state as it changes, and as soon as the app launches, to put that back in the global state. So clearing redux-persist is not a good strategy, since when the app launches it would have already populated the state. 

My thinking is that we have to consider the current state and handle it as we see fit. In this case, it means checking the network and updating the state if not correct.

cardstack/src/redux/merchantSlice.ts
oidoug: Never mind my question, we actually need to save a preferred primary for each network, or we'll have to "reset" to first created at each network change, my idea above wouldn't permit that.

cardstack/src/redux/__tests__/primarySafeSlice.test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+// interface AccountType {
+//   [accountAddress: string]: {
+//     primary?: MerchantSafeType;
+//   };
+// }
+
+// type SliceType = Record<Network, AccountType>;
+
+// const initialState: SliceType = Object.keys(Network)
```
danibonilha: leftover ?

cardstack/src/redux/primarySafeSlice.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type SliceType = Record<Network, AccountType>;
+
+const initialState: SliceType = {
+  kovan: {},
+  mainnet: {},
+  sokol: {},
+  xdai: {},
+};
```
danibonilha: We have an enum and Network defined as types, maybe we should use here as keys, in case we add or remove networks it's easy to keep it consistent

cardstack/src/redux/hooks/usePrimarySafe.ts
```diff
@@ -0,0 +1,51 @@
...
+const safesInitialState = {
+  merchantSafes: [],
+};
+
+export default function usePrimarySafe() {
+  const dispatch = useDispatch();
+  const [nativeCurrency] = useNativeCurrencyAndConversionRates();
+  const { accountAddress, network } = useAccountSettings();
+
+  const primarySafe = useSelector(selectPrimarySafe(network, accountAddress));
```
danibonilha, line 21: What happened with the typedSelector ? PrimarySafe has `any` type, we should maybe at least type the return of selectPrimarySafe

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -1,11 +1,17 @@
...
+import usePrimarySafe from '@cardstack/redux/hooks/usePrimarySafe';
 
 const ProfileScreen = () => {
+  const { primarySafe } = usePrimarySafe();
+
   return (
     <Container backgroundColor="backgroundDarkPurple" flex={1}>
       <MainHeader title="PROFILE" />
+      <Container justifyContent="center" flex={1}>
+        {primarySafe && <MerchantSafe {...primarySafe} />}
```
danibonilha: It's safer to either cast this to boolean `!!primarySafe` in case it's undefined, or check `!!primarySafe?.address`, bc if primarySafe is an empty object  `!!{}` returns true.

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -11,10 +11,10 @@ import { MerchantInformation, MerchantSafeType } from '@cardstack/types';
...
 import Routes from '@rainbow-me/routes';
 
-interface MerchantSafeProps extends MerchantSafeType {
-  networkName: string;
-  nativeCurrency: string;
-  currencyConversionRates: {
+export interface MerchantSafeProps extends MerchantSafeType {
+  networkName?: string;
+  nativeCurrency?: string;
+  currencyConversionRates?: {
```
danibonilha, line 17: I don't think we are using these props in this component, nor in the children, can you double check ? and if so remove it

cardstack/src/redux/__tests__/primarySafeSlice.test.ts
oidoug: Yes, my bad, already removed.

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
danibonilha: I'm not sure we are using these props in this component, nor in the children, can you double check ? and if so remove it

danibonilha (approved):
It's looking good, just small suggestions

cardstack/src/redux/primarySafeSlice.ts
oidoug: I tried using the enum but was having issues fulfilling the type requirements so I went with something simpler. I'll try again.

cardstack/src/redux/hooks/usePrimarySafe.ts
oidoug: I removed the typedselector because the return is already magically typed by rtk, prob. by keeping the type of the saved object. So it's an unnecessary complication, `primarySafe` in this case is returned as `const primarySafe: MerchantSafeType | undefined`.

cardstack/src/redux/primarySafeSlice.ts
oidoug: Ok, kinda did hehe, a bit overkill but makes sense.

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
oidoug: Nice catch, I added the boolean casting. It can't return an empty Obj because the selector needs to go through the whole chain. So it's either a safe or undefined.

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
oidoug: Apparently not, I removed them as well as infoDID that is already defined in MerchantSafeType.

PR #670, opened 2022-03-01
[CS-3247]: Use ppCard address for receipt header

danibonilha:
### Description

The `receipt.from` refers to the relay server, when we actually want to show the selected PrepaidCard address that is from, this PR fixes this issue.


oidoug (approved):
👍

PR #673, opened 2022-03-02
[CS-3262]: Share url only on payment request

danibonilha:
### Description

Since we have SSR now we can share only the url link

### Screenshots

image

image

image


oidoug (approved):
LGTM

PR #675, opened 2022-03-02
[Chore]: Match android versionName with iOS

danibonilha:
### Description

We need to match the versions in other to force-update, I'm working on the version tagging so we don't need to keep track of this manually, hopefully it might start working this week, but for now, this is what we got.


oidoug (approved):
👍

oidoug:
Are we going to align the build number on both too?

danibonilha:
> Are we going to align the build number on both too?

I'm still dwelling on this,  I'm a little concern about the failing builds, which might not occur on both platforms at the same time, so probably its best to match the versionName only, which is allowed to be pushed as the same as long as the build number increases

PR #676, opened 2022-03-03
[CS-3212] Same section header padding on home and wallet

oidoug:
### Description

- [x] Completes #(CS-3212)

### Checklist

### Screenshots

iOS:

https://user-images.githubusercontent.com/129619/156558086-9d02e845-af7a-484e-a686-622fcaaf8f42.MP4

Android:

https://user-images.githubusercontent.com/129619/156558265-060c4fce-bd57-41e4-bc48-8a0c5414d010.mp4

### DIFF
--- a/cardstack/src/components/TransactionList/TransactionList.tsx
+++ b/cardstack/src/components/TransactionList/TransactionList.tsx
@@ -55,8 +55,7 @@ export const TransactionList = memo(({ Header }: TransactionListProps) => {
   const renderSectionHeader = useCallback(
     ({ section: { title } }: { section: { title: string } }) => (
       <Container
-        paddingVertical={2}
-        paddingHorizontal={5}
+        padding={4}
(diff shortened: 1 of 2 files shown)


danibonilha (approved):
looks good padding-wise, but as mentioned in the ticket there's a behavioral difference, should we consider setting `stickySectionHeadersEnabled ` true ? To apply the same thing on Android, we would need to check the performance though.

topwebtek7 (approved):

lukemelia (approved):

oidoug:
> should we consider setting `stickySectionHeadersEnabled ` true

I just tested a bit with it on both the transaction list and asset list. It works pretty well except when a big list is building, at the first few renders it glitches a bit.
So, I'll merge this without the sticky header on android and maybe have a separate PR with it on for the team to test.

PR #677, opened 2022-03-03
Refactor scan screens to support two different views

topwebtek7:
### Description

- Added SwitchSelector custom component
- Added supporting 2 different views in Scan screen
- Moved all scan js components to TS and refactored, did some cleanup
- Improved camera handling to fix bugs with camera blank out
- Linked to request merchant payment request and sharing paymentLink
- UI/UX update with new designs

ToDo: 
- Add opacity surrounds scanning square and scan through that square only(maybe need to upgrade react-native-camera or use another library)
- Add more tests

- [x] Completes #3213
- [x] Completes #3215
- [x] Completes #3216
- [x] Completes #3187
- [x] Completes #3104

### Checklist

### Screenshots

iOS
Screenshot

Android
Mar-03-2022 20-31-54

iPhone SE
Screenshot

Simulator
Screenshot

iphone12
Screen Shot 2022-03-03 at 9 11 22 PM

Screen Shot 2022-03-03 at 9 11 36 PM


kdowlin:
@topwebtek7 I think the one thing I see is we want to introduce an opacity to the scan sheet that surrounds the scan square. I.e. the entire scan should not be rendering the camera view only a portion of the screen in the square.

topwebtek7:
> @topwebtek7 I think the one thing I see is we want to introduce an opacity to the scan sheet that surrounds the scan square. I.e. the entire scan should not be rendering the camera view only a portion of the screen in the square.

@kdowlin ok, understood.
basically tried but that does not work well with current `react-native-camera` library.. (react-native-camera is deprecated now, fyi - https://github.com/react-native-camera/react-native-camera) so I'm giving a try with other libraries like `react-native-camera-kit` locally as it supports better for QRCode scanning..

cardstack/src/screens/QRScannerScreen/components/CameraDimmer.tsx
```diff
@@ -0,0 +1,52 @@
...
+        ),
+      ]),
+    }),
+    []
+  );
+
+  return (
+    <Animated.View style={[style, styles.dimStyle]}>{children}</Animated.View>
+  );
+}
```
danibonilha: it doesn't seem this component is being used

danibonilha (approved):
Although there are some improvements to be made like the switch slider gesture is a bit laggy, unused components, tests to be added etc, I do need this screen to continue my transfer task, so I'll be approving in favor of it and I can update the missing pieces alongside with adding the overlay of the QR code. Also this should be a new screen and we could have kept the previous behavior, but I think it's okay that was migrated, we had a miscommunication issue and we can keep as it is.

cardstack/src/screens/QRScannerScreen/components/CameraDimmer.tsx
topwebtek7: good catch. removed this usage at the end of changes as thought it's not needed.. will get rid of it soon

PR #680, opened 2022-03-04
[CS-3237] Add WalletConnect Sessions in Settings

topwebtek7:
### Description

Added WC sessions handling screen in Settings and updated disconnect WC sessions behavior. while implementing disconnect, found an issue with dapp has no dappName and dappUrl only in peerMeta and handled that case as well.

note: need to update WC icon after got provided svg

- [x] Completes #CS-3237
- [x] Completes #CS-3238

### Checklist

### Screenshots

Mar-04-2022 21-39-25


src/components/settings-menu/WalletConnectSessionsSection.tsx
```diff
@@ -0,0 +1,103 @@
...
+            flex={1}
+            flexDirection="row"
+            justifyContent="flex-end"
+          >
+            <Button
+              borderColor="buttonSecondaryBorder"
+              height={50}
+              onPress={() =>
+                handleDisconnectSession(item.dappName || item.dappUrl)
+              }
```
danibonilha: nit: `handleDisconnectSession` could have a second arrow function in it then we could avoid the anonymous on on render

src/components/settings-menu/WalletConnectSessionsSection.tsx
```diff
@@ -0,0 +1,103 @@
...
+      );
+    },
+    [handleDisconnectSession]
+  );
+
+  return (
+    <Container paddingTop={4}>
+      <FlatList
+        data={walletConnectorsByDappName}
+        keyExtractor={(item, index) => `${item.dappUrl}_${index}`}
```
danibonilha: nit: This could be extract into a `useCallback`

src/components/settings-menu/WalletConnectSessionsSection.tsx
```diff
@@ -0,0 +1,103 @@
...
+        ]);
+      }
+    },
+    [walletConnectDisconnectAllByDappNameOrUrl]
+  );
+
+  const renderItem = useCallback(
+    ({ item }) => {
+      const title =
+        item.dappName || (item.dappUrl || '').replace('https://', '');
```
danibonilha, line 40: nit: instead of doing `item.dappName || item.dappUrl` twice  we could assign to a `const`, `sessionName` or something

danibonilha (approved):
Looks good, just some suggestions.

src/components/settings-menu/WalletConnectSessionsSection.tsx
topwebtek7: basically it's just for handling rare cases that has empty `dappName`.
`item.dappName || item.dappUrl` and `const title =
        item.dappName || (item.dappUrl || '').replace('https://', '');` are different. `title` is for adding dapp name like `sushi.com` rather than `https://sushi.com` when empty dappName.
I think it does make sense and don't need to add new variable for that rare case (just wanna less variable definitions inside renderItem if makes sense)

PR #682, opened 2022-03-04
[CS-3290]: Fix tabHeader size and make it customizable 

danibonilha:
### Description

Make mainHeader more dynamic to match or at least get pretty close to the `Settings` header, also it makes the StatusBar.currentHeight as default on SafeArea's paddingTop (Idk why it doesn't work out of box for android anymore, or if it's something on this project). MainHeader now is fully customizable, it may not make much sense right now, but it will definitely be useful on my next task.

### Checklist

### Screenshots

Simulator Screen Recording - iPhone 12 - 2022-03-04 at 16 25 39

image


topwebtek7 (approved):
LGTM! 👍🏼

PR #683, opened 2022-03-07
[CS-3280] Update profile screen to have CTA to create a profile

topwebtek7:
### Description

Update profile screen to have a CTA to create a profile if has no safe in their wallet. (Design is still pending and it's temporary workaround)

- [x] Completes #CS-3280

### Checklist

### Screenshots

Screen Shot 2022-03-07 at 8 47 49 PM

### DIFF
--- a/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
+++ b/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
@@ -1,11 +1,36 @@
 import React from 'react';
 
-import { Container, MainHeader } from '@cardstack/components';
+import { Button, Container, MainHeader, Text } from '@cardstack/components';
+import { useRainbowSelector } from '@rainbow-me/redux/hooks';
 
 const ProfileScreen = () => {
+  // TODO: replace this with usePrimaryMerchant hook
+  const merchantSafes = useRainbowSelector(state => state.data.merchantSafes);
+  const hasSafes = (merchantSafes || []).length > 0;
+
+  // TODO: This is a temporary workaround as designs are pending yet
+  const NoSafesPlaceholder = () => (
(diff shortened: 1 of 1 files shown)


cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -1,11 +1,36 @@
 import React from 'react';
 
-import { Container, MainHeader } from '@cardstack/components';
+import { Button, Container, MainHeader, Text } from '@cardstack/components';
+import { useRainbowSelector } from '@rainbow-me/redux/hooks';
 
 const ProfileScreen = () => {
+  // TODO: replace this with usePrimaryMerchant hook
+  const merchantSafes = useRainbowSelector(state => state.data.merchantSafes);
```
oidoug, line 8: My understanding is that our new implementation for pulling updated safes from redux is using RTK-Query and the hook you should use is `useGetSafesDataQuery()`, as in your case, like so:

```
const {
    data: { merchantSafes },
    error,
  } = useGetSafesDataQuery({
    address: accountAddress,
    nativeCurrency,
  });
```

This `data.merchantSafes` is updated on calling `fetchSafes`, but with rtk-query we are pretty much duplicating this state.

topwebtek7: Yes that might be better option.. btw it's just temporary workaround yet and will be replaced later with hook function that you will be implemented in your pr.

oidoug: You're right, the new hook encapsulates that.

kdowlin:
Sorry - see you just closed it! My mistake! @topwebtek7

PR #685, opened 2022-03-07
[CS-3277]: Add QrCode overlay on ScannerScreen

danibonilha:
### Description

Creates and adds the QRCode overlay in the ScannerScreen, I've build a custom svg scalable based on the devices dimensions, there are of course some magical numbers, which were chosen based on try and errors to match the design, but I  tested on multiple devices and it looks good. 

### Checklist

### Screenshots

image

Iphone 12 mini:
image

QROverlay


oidoug (approved):
Congratulations, well done! New SVG skills unlocked!

topwebtek7 (commented):
@danibonilha Really good job! 👍🏼 
basically the reason I tried other libraries like react-native-vision-camera, react-native-camera-kit rather than react-native-camera when working on it, was to support such frame using library and able to scan qrcode inside that frame only, that was the initial scope when discussing with Kiel(it's able to scan outside of frame as well now). somehow it's looking good and we can keep it as it is, and try to update RNCamera to other library later as RNCamera is deprecated now.

PR #686, opened 2022-03-07
[CS-3294]: Simplify and restyle SwitchSelector

danibonilha:
### Description

The selector gesture was laggy and it was defined that it shouldn't really be a slider, so the design was updated to reflect that, more as a On/Off toggle, than a slider.

PS: Right now the borderRadius logic supports 2 options only (as specified on design), if we come across more options we can update it to reflect that.

### Checklist

### Screenshots

Simulator Screen Recording - iPhone SE (2nd generation) - 2022-03-07 at 18 34 29

### DIFF
--- a/cardstack/src/components/SwitchSelector/SwitchSelector.tsx
+++ b/cardstack/src/components/SwitchSelector/SwitchSelector.tsx
@@ -1,160 +1,65 @@
-import React, {
-  useState,
-  useEffect,
-  useMemo,
-  useCallback,
-  useRef,
-} from 'react';
-import {
-  Animated,
-  Easing,
-  PanResponder,
-  GestureResponderEvent,
-  PanResponderGestureState,
-} from 'react-native';
+import React, { useState, useMemo, useCallback } from 'react';
+
 import {
   Container,
   ContainerProps,
   Text,
   Touchable,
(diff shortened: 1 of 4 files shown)


cardstack/src/components/SwitchSelector/SwitchSelector.tsx
```diff
@@ -1,226 +1,92 @@
...
   return (
     <Container
+      height={height}
+      borderWidth={1.5}
       flexDirection="row"
-      accessibilityLabel={accessibilityLabel}
-      testID={testID}
+      borderColor="whiteLightOpacity"
       backgroundColor="darkGrayOpacity"
+      alignItems="center"
```
topwebtek7, line 90: Screen Shot 2022-03-08 at 7 10 03 PM
Screen Shot 2022-03-08 at 7 09 52 PM

As you can see screenshots, item Touchable overlaps container border, so suggest this change.

```suggestion
      alignItems="center"
      style={{
        paddingHorizontal: 1,
      }}
```

topwebtek7 (approved):
Nice simplify on this pr! just a small suggestion

oidoug (approved):
No more than what Brian noted.

PR #688, opened 2022-03-08
[CS-3179] Add CreateProfile step one

topwebtek7:
### Description

Created first step of profile component and adding steps indicator in another pr

- [CS-3179] Completes #(Linear ticket)

### Checklist

### Screenshots

Screen Shot 2022-03-10 at 7 11 02 PM


cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -1,16 +1,74 @@
...
+    const networkName = getConstantByNetwork('name', network);
+
+    return {
+      nativeCurrency,
+      currencyConversionRates,
+      networkName,
+    };
+  }, [currencyConversionRates, nativeCurrency, network]);
+
+  const CreateProfileStepOne = () => (
```
oidoug: It would be better to have this component in a separate file. We'll have a few steps like this in the end.

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -13,31 +13,39 @@ import Routes from '@rainbow-me/routes';
...
+  ...props
+}: MerchantSafeProps) => {
   const { navigate } = useNavigation();
 
   const onPress = useCallback(() => {
     const merchantData = { ...props, merchantInfo };
     navigate(Routes.MERCHANT_SCREEN, { merchantSafe: merchantData });
   }, [merchantInfo, navigate, props]);
 
+  const PressableContainer = notPressable ? Container : CardPressable;
```
oidoug: I think you can use CardPressable disabled prop instead of switching for a container here.
If the style doesn't match up, maybe adapt CardPressable to work as you expect by redefining onPress.

notPressable could also be just disabled, although disabled sometimes also implies a different style, so I'm not sure.

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -13,31 +13,39 @@ import Routes from '@rainbow-me/routes';
...
+
```
danibonilha: Instead of conditionally  choose the container, couldn't we simply use the disabled prop on `<CardPressable>` ? We could receive `disabled` as prop and pass down instead of using `notPressable` then we don't have to check anything in other to pass `onPress`

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -13,31 +13,39 @@ import Routes from '@rainbow-me/routes';
...
         testID="inventory-card"
-        onPress={onPress}
+        onPress={notPressable ? undefined : onPress}
       >
         <SafeHeader
           {...props}
           onPress={onPress}
           backgroundColor={merchantInfo?.color}
           textColor={merchantInfo?.textColor}
+          rightText={notPressable ? 'Profile' : undefined}
```
danibonilha: nit: Here we could receive the `rightText` as a prop, `HeaderRightText` which could default to 'Profile'

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -1,16 +1,74 @@
...
+
+  const exmapleMerchantExtraProps = useMemo(() => {
+    const networkName = getConstantByNetwork('name', network);
+
+    return {
+      nativeCurrency,
+      currencyConversionRates,
+      networkName,
+    };
+  }, [currencyConversionRates, nativeCurrency, network]);
```
danibonilha: There's a type on this function name, but it also seems we don't need this, what is the intention ?

oidoug (commented):
Looks good, but better refactor CreateProfileStepOne out of ProfileScreen before merging.

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -1,16 +1,74 @@
...
+          <MerchantSafe
+            {...exmapleMerchantExtraProps}
+            {...exampleMerchantData}
+            notPressable
+          />
+        </Container>
+        <Button marginTop={5}>{strings.continueButton}</Button>
+      </Container>
+    </Container>
+  );
```
danibonilha: This screen uses fixed padding etc, instead of flex,  it doesn't seem to work right on smaller devices 
image
We could think about adding a scrollView, but I think scaling the components based on the phone available space to flex might be better

Also we should extract this to their own file, we could have a `Form` context

cardstack/src/screens/ProfileScreen/fixture.ts
```diff
@@ -0,0 +1,26 @@
...
+    name: 'Mandello',
+    ownerAddress: '',
+    slug: 'mandello',
+    textColor: '#fff',
+  },
+  owners: [''],
+  revenueBalances: [],
+  tokens: [],
+  type: 'merchant',
+};
```
danibonilha: fixture is usually more related to testing suite, let's keep these type of files, either as strings, os as utils

danibonilha (commented):
It seems the design of the "merchantSafe" has changed, it's worth checking if the updated layout is expected in this task @kdowlin

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
topwebtek7: nice catch. these are required in MerchantSafe component, so updated with empty data except networkName which is  correct current network name.

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
```diff
@@ -13,9 +13,16 @@ import Routes from '@rainbow-me/routes';
 
 export interface MerchantSafeProps extends MerchantSafeType {
   merchantInfo?: MerchantInformation;
+  notPressable?: boolean;
```
oidoug: nit: isn't disabled a better term, since we're using disabled on cardpressable too?

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -1,16 +1,24 @@
...
+
   return (
     <Container backgroundColor="backgroundDarkPurple" flex={1}>
-      <MainHeader title="PROFILE" />
-      <Container justifyContent="center" flex={1}>
-        {!!primarySafe && <MerchantSafe {...primarySafe} />}
+      <MainHeader title={strings.title} />
+      <Container justifyContent="center" flexGrow={1}>
+        {primarySafe ? <MerchantSafe {...primarySafe} /> : <CreateProfile />}
```
oidoug, line 21: ~~nit: primarySafe could be double-negated (`!!primarySafe`), but doest mather much, I'll need to change this a bit later to merge PR #691 with it.~~ nevermind, I didn't see it's a ternary op.

oidoug (approved):
👍

cardstack/src/components/MerchantSafe/MerchantSafe.tsx
danibonilha: Agreed on rename it to disabled

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
danibonilha: We just need to cast to boolean if the assertion is boolean e.g `primarySafe && <Merchant>`, bc if primarySafe is a `positive` value, like a string for example, it will try to render as a "component", another way to prevent something like that would be to use `primarySafe ? <Merchant>: null`
in case of the ternary it's okay not to cast, since the default case will be to render the `<CreateProfile>`

cardstack/src/screens/ProfileScreen/components/CreateProfileStepOne.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  const exampleMerchantExtraProps = useMemo(() => {
+    const networkName = getConstantByNetwork('name', network);
+
+    return {
+      nativeCurrency: '',
+      currencyConversionRates: [],
+      networkName,
+    };
+  }, [network]);
```
danibonilha: As I've mentioned before, it doesn't seem we need this, if your branch is up-to-date with develop, you'll noticed those props were removed

cardstack/src/screens/ProfileScreen/components/CreateProfileStepOne.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <Container
+      flexGrow={1}
+      flexDirection="column"
+      justifyContent="space-between"
+      paddingVertical={8}
+      backgroundColor="black"
```
danibonilha: Why is the background black ? I think it's supposed to be the same as the other tabs, darkPurple
image

danibonilha (changes requested):
Nice adaptation to support smaller devices 👏 , but there are still some things to address:
- wrong backgroundColor
- removal of exampleMerchantExtraProps 
- using disabled as the pattern

cardstack/src/screens/ProfileScreen/components/CreateProfileStepOne.tsx
topwebtek7: oh i've added it for UI check purpose and forgot to remove, good catch!

danibonilha (approved):
LGTM

PR #691, opened 2022-03-09
[CS-3281] Marking a safe as primary

oidoug:
### Description

This PR adds a new feature to select a primary safe from any opened safe from the wallet.

It also concludes the refactor on MerchantScreen to separate its components and hooks to be more reusable.

Minor tweak: Sets the new tabbar navigation as default if on dev environment.

- [x] Completes #(CS-3281)
- [x] Completes #(CS-3282) 

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/157460129-25d3f7a2-28e7-4150-8cb8-9bd0a38dcb9d.mp4


cardstack/src/screens/MerchantScreen/MerchantScreen.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Container, MerchantContent } from '@cardstack/components';
+
+const MerchantScreen = () => {
+  const {
+    isRefreshingBalances,
+    merchantSafe,
+    safesCount,
+    isPrimarySafe,
+    changeToPrimarySafe,
+  } = useMerchantScreen();
```
danibonilha, line 14: Nice 👏  how about we add some tests for this one ?

oidoug: will do!

cardstack/src/components/MerchantContent/MerchantContent.tsx
```diff
@@ -0,0 +1,118 @@
...
+          onClaimAllPress,
+        }),
+      [onClaimAllPress, merchantSafe, navigate]
+    );
+
+    return (
+      <Container height="100%" justifyContent="flex-end" paddingBottom={4}>
+        <ScrollView
+          width="100%"
+          contentContainerStyle={{ alignItems: 'center', paddingBottom: 260 }}
```
danibonilha: nit: would be nice to extract this to its own style const outside the component

cardstack/src/components/MerchantContent/MerchantContent.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }
+);
+
+const strings = {
+  requestPayment: 'Request Payment',
+  readyToClaim: 'Ready to Claim',
+  noPendingPayment: 'No pending payments',
+  availableBalance: 'Your Available Balance',
+  noAvailableBalance: 'No balance available',
+};
```
danibonilha: nit: it would be nice if we could choose a pattern to wether place the strings, I have mixed feelings about this, I like to keep in the same file because you 'scope' the strings, but at the same time, having as a file might help on i18n case, anyways, nothing to worry right now, just keep on the radar

cardstack/src/components/MerchantContent/components/MerchantHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { Pressable, StyleSheet } from 'react-native';
+import { Container, Icon, Text } from '@cardstack/components';
+import { MerchantInformation } from '@cardstack/types';
+import { ContactAvatar } from '@rainbow-me/components/contacts';
+
+interface MerchantHeaderProps {
+  merchantInfo?: MerchantInformation;
+  isPrimarySafe: boolean;
+  changeToPrimarySafe?: () => void;
+  showSafePrimarySelection?: boolean;
```
danibonilha: nit: we could have this as its own type and extend on the places we need

cardstack/src/components/MerchantContent/components/MerchantHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export const MerchantHeader = ({
+  merchantInfo,
+  isPrimarySafe,
+  changeToPrimarySafe,
+  showSafePrimarySelection = false,
+}: MerchantHeaderProps) => {
+  return (
```
danibonilha: nit: could omit the return

cardstack/src/components/MerchantContent/components/MerchantHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+    <Container
+      width="100%"
+      flexDirection="row"
+      alignItems="center"
+      justifyContent="center"
```
danibonilha: it feels the alignment should be on the left and a bit more vertical padding

cardstack/src/components/MerchantContent/components/MerchantHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+};
+
+const strings = {
+  isPrimarySafeProfileLinkText: 'Primary Profile',
+  setToPrimaryProfileLinkText: 'Set as Primary Profile',
+};
+
+const styles = StyleSheet.create({
+  primaryText: {
+    width: 140,
```
danibonilha, line 76: would a percentage work ?

cardstack/src/components/MerchantContent/components/MerchantHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+      flexDirection="row"
+      alignItems="center"
+      justifyContent="center"
+      padding={10}
+    >
+      {merchantInfo ? (
+        <Container marginRight={2}>
+          <ContactAvatar
+            color={merchantInfo?.color}
+            size="large"
```
danibonilha: and this might be `x-large`

cardstack/src/components/MerchantContent/components/MerchantTokensList.tsx
```diff
@@ -0,0 +1,66 @@
...
+} from '@cardstack/components';
+import { TokenType } from '@cardstack/types';
+import { sortedByTokenBalanceAmount } from '@cardstack/utils';
+
+const isLastItem = (items: TokenType[], index: number): boolean =>
+  items.length - 1 === index;
+
+type onPressProps = {
+  onPress: () => void;
+};
```
danibonilha: nit: I believe when I first created this, we were using in more than one place, now that it's used just on one we can just add `onPress: () => void;` to `MerchantTokensListProps`

danibonilha (approved):
Looking really good, props on splitting  the components into smaller ones! 💯  just small nits

cardstack/src/components/MerchantContent/MerchantContent.tsx
oidoug: Yes, I agree, I had already moved this out for making tests easier, but that's wip.
I think we should try i18n something soon so we get a better idea for this pattern.

cardstack/src/components/MerchantContent/components/MerchantHeader.tsx
oidoug: I tried putting it to the left but it didn't look good, tbh. Kiel did the mock in the ticket just for reference, I think we'll need a better design soon.

cardstack/src/components/MerchantContent/components/MerchantHeader.tsx
oidoug: No, 140 is the minimum size for the longer text to not break the line. I need this fixed size so when switching the text the UI doesn't jump.

oidoug:
@danibonilha I added a simple test for MerchantScreen, mocking it was a pain since it's a whole screen.
I'll add more significant tests in a new PR, I don't want to hold this one too long, ok?

PR #696, opened 2022-03-10
[CS-3180, CS-3181] Add Create profile flow

topwebtek7:
### Description

Update create profile flow with 3 steps using custom ProgressSteps component and covered several screen devices, added checking business Id uniqueness, custom input form etc, added tests as well.

ToDo: working on next steps like prepaid card selection and pay $1 for creating profile flow. will do small UI padding tweaks as well.

- [x] Completes #CS-3180
- [x] Completes #CS-3181

### Checklist

### Screenshots

iPhone12
Mar-14-2022 21-35-45

iPhoneSE
Mar-14-2022 23-00-49


cardstack/src/screens/ProfileScreen/components/ProfileFormContainer.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }),
+    [businessColor, businessId, businessName, isUniqueId]
+  );
+
+  return (
+    <ProfileFormContext.Provider value={contextValues}>
+      {children}
+    </ProfileFormContext.Provider>
+  );
+};
```
danibonilha, line 78: Why do we need a context for this case? if everything is being updated in the same screen, couldn't we have the state logic in the parent and pass it down if needed ?

cardstack/src/screens/ProfileScreen/components/fixture.ts
```diff
@@ -5,16 +5,30 @@ export const strings = {
   createProfileDesc: 'All users need a profile to send and receive payments.',
   continueButton: 'Continue',
   example: 'Example:',
+  nameAndIdForProfile: 'Name and ID for Profile',
+  profileSubText:
+    "Subtext content here and here's the \nsecond line for good measure.",
+  businessName: 'Name (or Business Name)',
+  uniqueId: 'Unique ID',
+  unqiueIdDescription:
```
danibonilha: typo

cardstack/src/screens/ProfileScreen/components/fixture.ts
```diff
@@ -5,16 +5,30 @@ export const strings = {
   createProfileDesc: 'All users need a profile to send and receive payments.',
   continueButton: 'Continue',
   example: 'Example:',
+  nameAndIdForProfile: 'Name and ID for Profile',
+  profileSubText:
+    "Subtext content here and here's the \nsecond line for good measure.",
```
danibonilha: Since it's a draft I'm assuming this is temporary, let's add a TODO here to not forget to update

cardstack/src/screens/ProfileScreen/components/fixture.ts
```diff
@@ -5,16 +5,30 @@ export const strings = {
...
+  unqiueIdDescription:
+    'This is a unique ID that will be used to identify your payment profile. Please note this ID cannot be changed once the profile is created and can be used as a contact address.',
+  review: 'Review',
+  reviewDescription:
+    'Next you will confirm the creation of your profile. This will require a payment of $1.00USD.',
+  yourProfile: 'Your Profile:',
+  create: 'Create',
+  businessNameRequired: 'Business Name is required',
+  businessIdShouldBeUnique:
+    'This Business ID is already taken. Please choose another one',
```
danibonilha: we could contextualized those strings with object keys, e.g 
```
review: { 
    title: '...',
    description: '....'
business: { 
    nameRequired: '....',
    uniqueIdInfo: ' ....'
}
```



same goes to uniqueID etc,

cardstack/src/screens/ProfileScreen/components/ProgressSteps.tsx
```diff
@@ -0,0 +1,147 @@
...
+      </Container>
+      {children[currentStep]
+        ? React.cloneElement(children[currentStep], {
+            setActiveStep,
+            currentStep,
+          })
+        : null}
+    </Container>
+  );
+};
```
danibonilha: can you add some comments or wrap some thing into functions to be easier to understand ?  Also extract all the calculations to consts, so it easier to read the component

danibonilha: can you add some comments or wrap some thing into functions to be easier to understand ?  Also extract all the calculations to consts, so it easier to read the component

cardstack/src/screens/ProfileScreen/components/ProgressSteps.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  const setActiveStep = useCallback(
+    (step: number) => {
+      if (step >= stepCount - 1) {
+        setCurrentStep(stepCount - 1);
+      }
+
+      if (step > -1 && step < stepCount - 1) {
+        setCurrentStep(step);
+      }
```
danibonilha: what does `stepCount -1` means ?  Is not last step ?  What are the cases we want to cover here ?

danibonilha: what does `stepCount -1` means ?  Is not last step ?  What are the cases we want to cover here ?

cardstack/src/screens/ProfileScreen/components/ProgressSteps.tsx
```diff
@@ -0,0 +1,147 @@
...
+export const ProgressSteps = ({
+  activeStep = 0,
+  children,
+}: StepIndicatorProps) => {
+  const [currentStep, setCurrentStep] = useState<number>(activeStep);
+  const stepCount = useMemo(() => React.Children.count(children), [children]);
+
+  useEffect(() => {
+    setCurrentStep(activeStep);
+  }, [activeStep]);
```
danibonilha: What is this useEffect for ?

danibonilha: What is this useEffect for ?

cardstack/src/screens/ProfileScreen/components/ProgressSteps.tsx
```diff
@@ -0,0 +1,147 @@
...
+      .fill(null)
+      .map((_, index: number) => {
+        let stepStatus = StepStatus.DEFAULT;
+        const isLastStep = index === stepCount - 1;
+
+        if (index < currentStep) {
+          stepStatus = StepStatus.COMPLETED;
+        } else if (index === currentStep) {
+          stepStatus = StepStatus.ACTIVE;
+        }
```
danibonilha: suggestion: since we need the step status anyway, wouldn't mak sense to track the steps based on the status, instead of the index's ?

danibonilha: suggestion: since we need the step status anyway, wouldn't make more sense to track the steps based on the status, instead of the index's ?

cardstack/src/screens/ProfileScreen/components/ProgressSteps.tsx
```diff
@@ -0,0 +1,147 @@
...
+                    4) *
+                    index +
+                  ProgressStepSizes.circleWidth +
+                  2
+                }
+                borderTopWidth={1}
+                borderTopColor="borderLightColor"
+                width={ProgressStepSizes.circleDistance}
+              />
+            ) : null}
```
danibonilha: Is this the dash ?

danibonilha: Is this the dash ?

cardstack/src/screens/ProfileScreen/components/StepOne.tsx
```diff
@@ -1,8 +1,12 @@
-import React from 'react';
-import { exampleMerchantData, strings } from './';
+import React, { useCallback } from 'react';
+import { exampleMerchantData, StepActionType, strings } from '.';
 import { Button, Container, Text, MerchantSafe } from '@cardstack/components';
 
-export const CreateProfileStepOne = () => {
+export const StepOne = ({ setActiveStep, currentStep }: StepActionType) => {
+  const onPressContinue = useCallback(() => {
+    setActiveStep?.((currentStep || 0) + 1);
```
danibonilha: couldn't we have a setNextStep, then we don't need to add a plus one everytime

danibonilha: couldn't we have a setNextStep, then we don't need to add a plus one everytime

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
```diff
@@ -0,0 +1,160 @@
...
+                  name: 'check',
+                  color: 'greenColor',
+                  margin: 0,
+                  top: 14,
+                }
+              : undefined
+          }
+          spellCheck={false}
+          autoCorrect={false}
+        />
```
danibonilha, line 104: We could reuse parts of the input to make the component less verbose

danibonilha, line 104: We could reuse parts of the input to make the component less verbose

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
```diff
@@ -0,0 +1,160 @@
...
+        <Input
+          autoCapitalize="none"
+          placeholder="Enter Unique ID"
+          textContentType="username"
+          borderWidth={1}
+          borderColor={
+            (errors?.businessId || !isUniqueId) && isSubmitPressed
+              ? 'error'
+              : 'buttonSecondaryBorder'
+          }
```
danibonilha: let's extract this to a helper function

danibonilha: let's extract this to a helper function

danibonilha (commented):
I understand it's a draft , so I added some suggestion that we can talk about, also there are some blockers: 

- The layout is broken on smaller devices 
- We are unable to close the keyboard by clicking outside of it 

and one doubt, isn't the mandelo example supposed to be red? Why is it pink ?

cardstack/src/screens/ProfileScreen/components/CreateProfile.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export const CreateProfile = () => (
+  <ProfileFormContainer>
+    <ProgressSteps>
+      <StepOne />
+      <StepTwo />
+      <StepThree />
+    </ProgressSteps>
+  </ProfileFormContainer>
+);
```
oidoug, line 13: Will this CreateProfile flow only be started from ProfileScreen? Maybe it's better kept on global components folder.

cardstack/src/screens/ProfileScreen/components/ProfileFormContainer.tsx
oidoug: I also think using context API here isn't a good idea, it re-renders too much stuff when changes happen.
Having a hook with some local state is probably enough, but if you'll share a lot of state and call changes between the steps and maybe you want to persist the steps later, a RTK slice might be a good idea, it has persistence built-in now. Check `primarySafeSlice` for reference.

cardstack/src/screens/ProfileScreen/components/ProgressSteps.tsx
oidoug: > Is this the dash ?

Yes, it renders from the last circle forward, but not in the last one.

cardstack/src/screens/ProfileScreen/components/ProgressSteps.tsx
oidoug: It would be good to extract the dots with their "checked | selected | none" states to a component. They look real nice, maybe we can reuse them somewhere else later.

cardstack/src/screens/ProfileScreen/components/fixture.ts
```diff
@@ -5,16 +5,30 @@ export const strings = {
...
 };
```
oidoug: Better have strings always in a separate strings.ts file, so we can find them later for translation.

cardstack/src/screens/ProfileScreen/helper.ts
```diff
@@ -0,0 +1,66 @@
+import { useContext, useCallback, useMemo } from 'react';
+import { validateMerchantId } from '@cardstack/cardpay-sdk';
+import { ProfileFormContext, strings } from './components';
+import { useAuthToken } from '@cardstack/hooks';
+import { checkBusinessIdUniqueness } from '@cardstack/services/hub-service';
+
+export const useProfileForm = () => {
```
oidoug: Isn't better to call this helper file `useProfileForm.ts`, since it only has this hook?

oidoug (commented):
There's a lot to cover in these new screens 😅 

I'm missing a few things, sorry if you're aware or wip, just a heads up:
- Keyboard avoidance for inputs
- Jump between inputs, with a next tag in keyboard for name field
- Elements are squished in my screen, might need a scrollview (see screenshot)


image

cardstack/src/screens/ProfileScreen/components/ProfileFormContainer.tsx
topwebtek7: I was convinced context's best option here as we need to save form state. After filled form, need to navigate to prepaidcard selection screen and get back to it and the form state should be restored with selected prepaid id.
I'm not sure if RTK or redux persistent is needed here, IMO, that can make more complicated, which is simply can wrap only inside this screen/folder.

cardstack/src/screens/ProfileScreen/components/ProgressSteps.tsx
topwebtek7: yep it's dash

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
topwebtek7: ok

cardstack/src/screens/ProfileScreen/components/StepOne.tsx
topwebtek7: you can get back to completed step by pressing completed icon on top(for test purpose only yet but need to discuss this feature with Kiel), I'm using setActiveStep there as well

cardstack/src/screens/ProfileScreen/components/ProgressSteps.tsx
topwebtek7: it's for applying different styles based on Step status. please refer 
```
const StepIconStyles = {
  [StepStatus.DEFAULT]: {
    backgroundColor: 'underlineGray' as ColorTypes,
  },
  [StepStatus.COMPLETED]: {
    backgroundColor: 'teal' as ColorTypes,
  },
  [StepStatus.ACTIVE]: {
    backgroundColor: 'transparent' as ColorTypes,
    borderColor: 'teal' as ColorTypes,
    borderWidth: ProgressStepSizes.circleBorderWidth,
  },
};
```

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
topwebtek7: hmm not sure if we need to extract this to helper as well.. cause this makes sense enough and also isSubmitPressed is StepTwo component specific state, can't be used inside helper

cardstack/src/screens/ProfileScreen/components/CreateProfile.tsx
topwebtek7: It's being used inside profile screen only, will move to global when used in others

cardstack/src/screens/ProfileScreen/helper.ts
topwebtek7: ok, will keep as helper for now cause will have other helper functions as next steps like create did hub api call, etc

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
danibonilha: I didn't mean like a external function, it's just a function that returns the right borderColor of the input, since this check is done on two places, just with slight different conditions,

image

image

something as simple as 
```tsx

const getBorderColor =  useCallback((isValid) => () => isValid ? 'buttonSecondaryBorder' : 'error', [])

```

or if you were to create a input component for this screen to be reusable (which i'd recommend), you could track the error state with a flag within it

topwebtek7:
@danibonilha @oidoug Hey guys, added more refactoring and updates in this pr, please review again.

cardstack/src/components/ProgressSteps/CustomScrollView.tsx
```diff
@@ -0,0 +1,45 @@
...
+      keyboardVerticalOffset={KeyboardOffset}
+    >
+      <ScrollView contentContainerStyle={styles.scrollViewContentStyle}>
+        {childElements}
+      </ScrollView>
+    </KeyboardAvoidingView>
+  ) : (
+    <ScrollView flex={1} contentContainerStyle={styles.scrollViewContentStyle}>
+      {childElements}
+    </ScrollView>
```
oidoug: Why not have this component always be for KeyboardAvoidingView + Scrollview, and name it accordingly?

cardstack/src/components/ProgressSteps/ProgressSteps.tsx
```diff
@@ -0,0 +1,161 @@
...
+              {isCompletedStep ? (
+                <Icon
+                  color="black"
+                  iconSize="small"
+                  name="check"
+                  strokeWidth={2}
+                  height={14}
+                />
+              ) : null}
+            </Touchable>
```
oidoug: It would be nice to have at least the Step Icons be extracted as components.

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
```diff
@@ -0,0 +1,150 @@
...
+        <Text color="white" size="xs" fontWeight="600" marginBottom={2}>
+          {strings.businessName}
+        </Text>
+        <Input
+          placeholder="Enter Business Name"
+          textContentType="name"
+          borderColor={errors?.businessName ? 'error' : 'buttonSecondaryBorder'}
+          value={businessName}
+          onChange={onChangeBusinessName}
+          onSubmitEditing={onBusinessNameSubmitEditing}
```
oidoug, line 80: you should add the prop `returnKeyType="next"` here so it shows as next in the soft keyboard.

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
```diff
@@ -0,0 +1,150 @@
...
+          marginTop={4}
+          color="white"
+          size="xs"
+          fontWeight="600"
+          marginBottom={2}
+        >
+          {strings.uniqueId}
+        </Text>
+        <Input
+          autoCapitalize="none"
```
oidoug, line 94: and maybe as `returnKeyType="done"` in this one.

cardstack/src/screens/ProfileScreen/components/fixture.ts
```diff
@@ -5,16 +5,30 @@ export const strings = {
...
```
oidoug: Can you rename this one to strings.ts?

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -0,0 +1,120 @@
...
+        ? strings.businessIdShouldBeUnique
+        : validateMerchantId(businessId),
+    };
+  }, [isSubmitPressed, businessName, isUniqueId, businessId]);
+
+  const onSubmitForm = useCallback(() => {
+    setIsSubmitPressed(true);
+
+    if (isUniqueId && businessName.trim() && !validateMerchantId(businessId)) {
+      onUpdateProfileForm({
```
oidoug, line 86: Is this doing something? When I hit "Create" nothing changes.

oidoug (commented):
Looking at the design files at https://app.zeplin.io/project/62265495c4b7951732dd18ed/screen/6227691959f6f0b86231275c
I see the padding between elements are different in your implementation. And the safe card color is also mismatching.

Also, I'm not sure what should happen on pressing "Create" at the end now, but just as a heads up, for me it's not doing anything.

cardstack/src/screens/ProfileScreen/useProfileForm.ts
topwebtek7: yes, I'm working on it as next pr, it will open prepaid card selection sheet.

topwebtek7:
@oidoug After Step3, there will be prepaid cards selection sheet that I'm working on right now. paddings are a bit different to make it look ok on small devices as well but not much diff with designs, looks quite ok to me..

cardstack/src/components/ProgressSteps/CustomScrollView.tsx
topwebtek7: We need KeyboardAvoidingView only in the screen which has Input like StepTwo.  also we're using this inside ProgressSteps only as a wrapper of step components. so named it as CustomScrollViewWrapper

cardstack/src/screens/ProfileScreen/components/strings.ts
```diff
@@ -0,0 +1,41 @@
+export const strings = {
+  title: 'Profile',
+  headerRightText: 'Profile',
+  createProfile: 'Create a Profile',
+  createProfileDesc: 'All users need a profile to send and receive payments.',
+  continueButton: 'Continue',
+  example: 'Example:',
+  nameAndIdForProfile: 'Name and ID for Profile',
+  profileSubText:
+    "Subtext content here and here's the \nsecond line for good measure.", // ToDo: update with proper text
```
kdowlin: Remove this text for now. @topwebtek7

cardstack/src/components/ProgressSteps/CustomScrollViewWrapper.tsx
```diff
@@ -0,0 +1,45 @@
...
+    >
+      <ScrollView contentContainerStyle={styles.scrollViewContentStyle}>
+        {childElements}
+      </ScrollView>
+    </KeyboardAvoidingView>
+  ) : (
+    <ScrollView flex={1} contentContainerStyle={styles.scrollViewContentStyle}>
+      {childElements}
+    </ScrollView>
+  );
```
danibonilha, line 45: `<KeyboardAvoidingView>` is just a view with the keyboardPadding, does it make any difference when we render the screen without the keyboard for this case?

cardstack/src/screens/ProfileScreen/__tests__/useProfileForm.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      accountName: 'foo',
+    }));
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should useProfileForm return empty businessName, avatarName with default,isUniqueId at initial render', async () => {
+    const { result } = renderHook(() => useProfileForm());
```
danibonilha: I believe since the tests are about `useProfileForm` we can remove from the it declaration, `it('should return empty...`  reads better than `it('should useProfileForm return empty...`  or just remove the should

danibonilha: Ditto for all following declarations

cardstack/src/screens/ProfileScreen/components/ProfileFormContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+});
+
+export const ProfileFormContainer = ({
+  children,
+}: {
+  children: React.ReactNode;
+}) => {
+  const { accountColor } = useAccountProfile();
+  const [businessName, setBusinessName] = useState<string>('');
+  const [businessId, setBusinessId] = useState<string>('');
```
danibonilha: nit: this could be a single combined state since they're all updated together

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
```diff
@@ -0,0 +1,149 @@
...
+          placeholder="Enter Unique ID"
```
danibonilha: Do we want a placeholder here ? or the label is enough ? If we do want the placeholder the current color is too dark, so we are not able to see it

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  const checkIdUniqueness = useCallback(
+    async (id: string) => {
+      if (!isLoading && authToken && id) {
+        const uniquenessResult = await checkBusinessIdUniqueness(id, authToken);
+
+        if (uniquenessResult?.slugAvailable) {
+          setIdUniqueness(true);
+        }
+      }
```
danibonilha, line 39: We should probably check if it's a validMerchant before calling the service every time the user types something.

danibonilha (approved):
I had hopes the important suggestions/discussions would be taking into consideration, as they weren't I need to note that I'm approving it just in favor of finishing the flow and not because I think it's satisfactory.
Also the tabBar is being pushed on android 

image

cardstack/src/screens/ProfileScreen/useProfileForm.ts
topwebtek7: We still need to discuss and update design how to handle validations on this screen, so will update this in next pr when update it.

cardstack/src/components/ProgressSteps/CustomScrollViewWrapper.tsx
topwebtek7: will be updated in next pr along with tabbar on android

PR #697, opened 2022-03-10
[Fix]: Go back two screens after claiming

danibonilha:
### Description

With the addition of a second modal to claim the revenue, we now need to go back two screens, otherwise we are not able to see the value updated to zero, after successfully claiming the revenue


oidoug (approved):
👍

PR #698, opened 2022-03-11
[CS-3328] Modals not opening

oidoug:
### Description

- Screen options on main Stack Navigator to be modal;
- Adds `ModalScreen` to Modal Navigator;
- Fixes removing last global screen added to stack on Android.

Fixes modals on Android with less-risky changes, by just setting the main Stack Navigator mode to 'modal' and setting navigation `screenOptions` `gesturesEnabled` to be true on Android too (it defaults to true on ios and false on android).

https://reactnavigation.org/docs/5.x/stack-navigator/#mode
https://reactnavigation.org/docs/5.x/stack-navigator/#gestureenabled

- Adds ModalScreen to Modal Navigator;
- Fixes removing last global screen added to stack on Android.

- [x] Completes #(CS-3328)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/157885978-61c65aba-ed86-4a5b-bde9-63b7f9961f5e.mp4

### DIFF
--- a/cardstack/src/navigation/tabBarNavigator.tsx
+++ b/cardstack/src/navigation/tabBarNavigator.tsx
@@ -26,6 +26,7 @@ import { TabBarIcon } from '@cardstack/components';
 import { colors } from '@cardstack/theme';
 import { Device, screenHeight } from '@cardstack/utils';
 import ExpandedAssetSheet from '@rainbow-me/screens/ExpandedAssetSheet';
+import ModalScreen from '@rainbow-me/screens/ModalScreen';
 import { expandedPreset, sheetPreset } from '@rainbow-me/navigation/effects';
 
 const Tab = createBottomTabNavigator();
@@ -103,11 +104,16 @@ const StackNavigator = () => {
 
   // TODO: Create a navigator for each flow and split auth/non-auth
 
-  // Remove last item aka LoadingOverlay, to avoid dupe
-  cardstackGlobalScreens.pop();
+  // Remove last item aka LoadingOverlay, to avoid dupe (on iOS)
+  Device.isIOS && cardstackGlobalScreens.pop();
 
   return (
-    <Stack.Navigator headerMode="none" initialRouteName={initialRoute}>
+    <Stack.Navigator
+      headerMode="none"
+      mode="modal"
+      screenOptions={{ gestureEnabled: true }}
+      initialRouteName={initialRoute}
+    >
       <Stack.Screen
         component={TabNavigator}
         name={RainbowRoutes.SWIPE_LAYOUT}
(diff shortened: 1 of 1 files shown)


oidoug:
@danibonilha , this is a band-aid for our modals issue in android, because all our sheet screens are still in the main stack, and can't be swiped out in android (well, at least in Samsung's OneUI). I think we need to discuss what the best approach is.

For all I tested, changing the Root stack didn't affect anything on iOS, only drawback looks to be that sheets in Android are deallocating the previous screen and there's a bit of a flash on the visible translucent background - I think this will be fixed if we move all sheets to the correct stacking mode, "modal".

danibonilha:
> @danibonilha , this is a band-aid for our modals issue in android, because all our sheet screens are still in the main stack, and can't be swiped out in android (well, at least in Samsung's OneUI). I think we need to discuss what the best approach is.
> 
> For all I tested, changing the Root stack didn't affect anything on iOS, only drawback looks to be that sheets in Android are deallocating the previous screen and there's a bit of a flash on the visible translucent background - I think this will be fixed if we move all sheets to the correct stacking mode, "modal".

What if we move the sheets to be inside the main stack ? Let's pair on this to figure it out

cardstack/src/navigation/tabBarNavigator.tsx
```diff
@@ -103,11 +104,16 @@ const StackNavigator = () => {
...
-  cardstackGlobalScreens.pop();
+  // Remove last item aka LoadingOverlay, to avoid dupe (on iOS)
+  Device.isIOS && cardstackGlobalScreens.pop();
 
   return (
-    <Stack.Navigator headerMode="none" initialRouteName={initialRoute}>
+    <Stack.Navigator
+      headerMode="none"
+      mode="modal"
+      screenOptions={{ gestureEnabled: true }}
```
oidoug, line 114: defaults to false on android

cardstack/src/navigation/tabBarNavigator.tsx
```diff
@@ -103,11 +104,16 @@ const StackNavigator = () => {
...
-  // Remove last item aka LoadingOverlay, to avoid dupe
-  cardstackGlobalScreens.pop();
+  // Remove last item aka LoadingOverlay, to avoid dupe (on iOS)
+  Device.isIOS && cardstackGlobalScreens.pop();
 
   return (
-    <Stack.Navigator headerMode="none" initialRouteName={initialRoute}>
+    <Stack.Navigator
+      headerMode="none"
+      mode="modal"
```
oidoug, line 113: Default is 'card', having 'modal' on iOS doesn't seem to affect any screen.

When set to 'modal', `headerMode` becomes 'screen' so we need to set it to none to keep the same behavior.

oidoug:
@danibonilha it's ready for review.

danibonilha (approved):
🚀

PR #700, opened 2022-03-14
[CS-3229]: Add walletAddress screen and refactor copyToast

danibonilha:
### Description

This PR adds the wallet address QR code screen and the ability to copy it, also refactors the CopyToast to handle the component and the copying logic as this piece was duplicated on multiple components, although I'm not a fan of returning components within hooks, for this Toast case I believe it makes sense. I wish I had migrated more components, but as we are in a tight schedule I'll leave for next time. For the QR Address we don't want to show the logo anymore.

### Checklist

### Screenshots

Simulator Screen Recording - iPhone 12 - 2022-03-14 at 13 33 44

image


cardstack/src/screens/CopyAddressSheet.tsx
```diff
@@ -33,16 +30,12 @@ const CopyAddressSheet = () => {
...
 
-  const { setClipboard } = useClipboard();
-  const [copiedText, setCopiedText] = useState(undefined);
-  const [copyCount, setCopyCount] = useState(0);
   const [checked, setChecked] = useState(false);
 
-  const handleCopiedText = useCallback(() => {
-    setClipboard(address);
-    setCopiedText(abbreviations.formatAddressForDisplay(address));
-    setCopyCount(count => count + 1);
```
oidoug, line 44: awesome! I was wondering if that counter was still necessary 👏

oidoug (approved):
LGTM

PR #703, opened 2022-03-14
[CS-3211]: Transfer prepaid card using QR code 

danibonilha:
### Description

Adds the ability to read a QRCode within the transfer flow to auto fill the address. It may be mixing the contexts of Pay/Connect/Transfer, but since it's experimental I guess it's okay.

- [x] Completes CS-3211
- [x] Completes CS-3196

### Checklist

### Screenshots

scan-transfer

image


oidoug (approved):
🚀

PR #706, opened 2022-03-15
[CS-3340]: Hide toast on scan page

danibonilha:
### Description

There's probably some smoother way to hide is maybe within the `NetworkToast` but I don't know if it's because of rainbow navigation,  but sometimes it was not identifying the screen had changed, so the safest bet was to place in the SwipeNavigator itself, and we are going to remove this soon anyways, we cannot memoize the isOnScanPage, because if we do so, it will only store the first value.

### Screenshots

Simulator Screen Recording - iPhone 12 - 2022-03-15 at 16 42 30


oidoug (approved):
Ok! This badge won't be in the new tab bar structure, right? So I think it's fine.

topwebtek7 (approved):
LGTM

PR #707, opened 2022-03-16
[CS-3364] Fixes primary reassignment issue

oidoug:
Refactors usePrimarySafe hook to app function standard.

### Description

On changing EOA, the primary safe was reset to null but `merchantSafes` array was still dirty from the last EOA. Using flag `isFetching` also as an update dependency resolves the issue. 

- [x] Completes #(CS-3364)

### Checklist

### DIFF
--- a/cardstack/src/components/MerchantSafe/MerchantSafe.tsx
+++ b/cardstack/src/components/MerchantSafe/MerchantSafe.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback } from 'react';
 import { strings } from './strings';
-import usePrimarySafe from '@cardstack/redux/hooks/usePrimarySafe';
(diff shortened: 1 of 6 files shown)


danibonilha (commented):
Nice, how about we add some test for this hook to prevent this kind of issue ? I won't block for it, but it's a nice addition, will approve once the broken tests run

oidoug:
> Nice, how about we add some test for this hook to prevent this kind of issue ? I won't block for it, but it's a nice addition, will approve once the broken tests run

Yes, it would be nice. I started writing some but I need to study redux slice testing a bit more and I didn't want to hold off the fix.

topwebtek7 (approved):
LGTM

PR #709, opened 2022-03-16
[CS-3337] Add choose prepaidcard card screen

topwebtek7:
### Description

Refactored ChoosePrepaidCard component inside cardstack/components and added ChoosePrepaidCard sheet as well so can be used in other places. Also updated creating profile flow UI and ProgressSteps/ScrollableStepWrapper.

ToDo: handling navigate directly to confirmation screen if only one prepaid card and need to navigate to confirmation screen after selected prepaidcard.

- [x] Completes #CS-3337

### Checklist

### Screenshots

Mar-16-2022 20-58-52


cardstack/src/components/ChoosePrepaidCard/ChoosePrepaidCard.tsx
```diff
@@ -0,0 +1,144 @@
...
+    borderBottomLeftRadius: 20,
+    borderBottomRightRadius: 20,
+    position: 'absolute',
+    backgroundColor: 'white',
+    alignItems: 'center',
+  },
+  listContainer: { paddingBottom: 75 },
+});
+
+export const ChoosePrepaidCard = memo(
```
oidoug, line 49: I like the new pattern of having the screen and actions separated in UI and logic hook. I think this new screen is a good candidate for having a useChoosePrepaidCard hook.

cardstack/src/components/ChoosePrepaidCard/ChoosePrepaidCard.tsx
```diff
@@ -0,0 +1,144 @@
...
+    );
+
+    const onSelect = useCallback(
+      (item: PrepaidCardType) => {
+        onSelectPrepaidCard(item);
+      },
+      [onSelectPrepaidCard]
+    );
+
+    const renderItem = useCallback(
```
oidoug, line 85: Isn't better to use useMemo in this case?

cardstack/src/components/ChoosePrepaidCard/ChoosePrepaidCard.tsx
```diff
@@ -0,0 +1,144 @@
...
+        <Header
+          nativeBalanceDisplay={nativeBalanceDisplay}
+          onPressEditAmount={onPressEditAmount}
+        />
+        <FlatList
+          data={prepaidCards}
+          removeClippedSubviews
+          horizontal={false}
+          renderItem={renderItem}
+          scrollEventThrottle={16}
```
oidoug: It looks like you're not doing any scroll action that would need a throttle scroll event.
This needs to be used carefully since it can lead to performance issues.

cardstack/src/components/ChoosePrepaidCard/Header.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }) => (
+    <Container
+      alignItems="center"
+      backgroundColor="white"
+      width="100%"
+      borderTopLeftRadius={20}
+      borderTopRightRadius={20}
+    >
+      <Text marginTop={4} weight="bold" size="body">
+        Choose a Prepaid Card
```
oidoug: Could you move these to a strings.ts file? thanks.

cardstack/src/components/ChoosePrepaidCard/Header.tsx
```diff
@@ -0,0 +1,52 @@
...
+      </Text>
+      <Text variant="subText" weight="bold" marginTop={3} marginBottom={1}>
+        To Pay This Amount
```
oidoug: same as above.

cardstack/src/components/ChoosePrepaidCard/PrepaidCardItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+  currencyConversionRates: {
+    [key: string]: number;
+  };
+  spendAmount: number;
+  isLastItem: boolean;
+}
+
+// Android has a shadow issue,
+// the default opacity adds a weird shadow on press
+const activeOpacity = (Device || {}).isAndroid ? 0.85 : undefined; // added ( || {}) because Device is undefined in test only, works fine other than tests
```
oidoug: I think this is one of the problems CardPressable solves. You could try using it instead of your custom Touchable implementation bellow.

cardstack/src/components/ChoosePrepaidCard/PrepaidCardItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+                borderWidth={1}
+              />
+            )}
+          </Container>
+          <Container flexGrow={1} paddingRight={16}>
+            <Container flexDirection="row">
+              <NetworkBadge text={`ON ${networkName.toUpperCase()}`} />
+              {isInsufficientFund && (
+                <Text color="red" fontSize={10} weight="bold" paddingLeft={4}>
+                  INSUFFICIENT FUNDS
```
oidoug: Move to strings.ts maybe?

cardstack/src/components/ProgressSteps/ScrollableStepWrapper.tsx
```diff
@@ -2,44 +2,55 @@ import React from 'react';
...
     flex: 1,
     flexGrow: 1,
   },
-  scrollViewContentStyle: {
+  smallScreenScrollViewContentStyle: {
     flexGrow: 1,
     justifyContent: 'space-between',
     paddingVertical: 20,
   },
+  bigScreenScrollViewContentStyle: {
```
oidoug, line 17: since flex and justify are the same, you could add below style props additionally to the above.

cardstack/src/components/ProgressSteps/ScrollableStepWrapper.tsx
```diff
@@ -2,44 +2,55 @@ import React from 'react';
...
+
+  return (
     <KeyboardAvoidingView
       behavior={Device.keyboardBehavior}
       style={styles.keyboardAvoidView}
       keyboardVerticalOffset={KeyboardOffset}
+      enabled={keyboardEnabled}
     >
-      <ScrollView contentContainerStyle={styles.scrollViewContentStyle}>
+      <ScrollView contentContainerStyle={contentContainerStyle}>
```
oidoug: What about:

```
<ScrollView contentContainerStyle={[styles.smallScreenScrollViewContentStyle, isSmallPhone && styles.bigScreenScrollViewContentStyle]}>
```
and have better names for the styles.

cardstack/src/components/ProgressSteps/ScrollableStepWrapper.tsx
```diff
@@ -2,44 +2,55 @@ import React from 'react';
...
         {childElements}
       </ScrollView>
     </KeyboardAvoidingView>
-  ) : (
-    <ScrollView flex={1} contentContainerStyle={styles.scrollViewContentStyle}>
-      {childElements}
-    </ScrollView>
```
oidoug, line 44: Good removal, I didn't like this duplication 👏

cardstack/src/navigation/tabBarNavigator.tsx
```diff
@@ -49,6 +49,7 @@ const tabBarOptions = {
     elevation: 3,
   },
   showLabel: false,
+  keyboardHidesTabBar: true,
```
oidoug, line 52: isn't this available only on react-navigation v6?

cardstack/src/screens/ProfileScreen/components/StepThree.tsx
```diff
@@ -24,6 +30,21 @@ export const StepThree = ({ goToNextStep }: ProgressStepProps) => {
...
 
+  const onPressCreate = useCallback(() => {
+    if (selectedPrepaidCard) {
+      console.log({ selectedPrepaidCard });
+      showLoadingOverlay({ title: 'Creating profile...' });
+    } else {
+      navigate(MainRoutes.CHOOSE_PREPAIDCARD_SHEET, {
+        spendAmount: 100,
+        onConfirmChoosePrepaidCard: (prepaidCard: PrepaidCardType) => {
+          console.log({ prepaidCard });
```
oidoug: Plz, remove this debug log

cardstack/src/navigation/tabBarNavigator.tsx
danibonilha: it does exist on [v5](https://reactnavigation.org/docs/5.x/bottom-tab-navigator/#keyboardhidestabbar)

cardstack/src/components/ChoosePrepaidCard/ChoosePrepaidCard.tsx
topwebtek7: it's callback function which returns PrepaidCardItem component, so I think should be `useCallback`..
We have several very similar places that use useCallback in project, fyi

cardstack/src/components/ChoosePrepaidCard/ChoosePrepaidCard.tsx
topwebtek7: scroll action will be used when have multiple PrepaidCards and yes agreed, should be careful with scrollEventThrottle but it does not much affect here.. so will get rid of it

cardstack/src/navigation/tabBarNavigator.tsx
topwebtek7: yes it exists from v4, i think..

oidoug: sorry my bad, I did a quick search and didn't see it at the time.

cardstack/src/components/ChoosePrepaidCard/ChoosePrepaidCard.tsx
oidoug: You are right for using useCallback, useMemo ignores parameters and returns only a value, not a function.

cardstack/src/screens/sheets/ChoosePrepaidCardSheet/strings.ts
```diff
@@ -0,0 +1,4 @@
+export const strings = {
+  noPrepaidCardErrorMessage: `You don't own a Prepaid card!\nYou can create one at app.cardstack.com`,
+  noPrepaidCardErrorTitle: 'Ooops!',
```
danibonilha: There are too many o's  😅  `Oops`

cardstack/src/screens/sheets/ChoosePrepaidCardSheet/useChoosePrepaidCard.ts
```diff
@@ -0,0 +1,78 @@
...
+  params: {
+    spendAmount: number;
+    onConfirmChoosePrepaidCard: (selectedPrepaidCard: PrepaidCardType) => void;
+  };
+};
+
+export const useChoosePrepaidCard = () => {
+  const {
+    params: { spendAmount = 0, onConfirmChoosePrepaidCard },
+  } = useRoute() as ChoosePrepaidCardSheetRouteParams;
```
danibonilha: casting is not ideal, we have helpers that we can use to enforce better types.

```ts
type RouteParams = {
    spendAmount: number;
    onConfirmChoosePrepaidCard: (selectedPrepaidCard: PrepaidCardType) => void;
};

... } = useRoute<RouteType<RouteParams>>()
```

cardstack/src/screens/ProfileScreen/components/StepThree.tsx
```diff
@@ -24,6 +30,19 @@ export const StepThree = ({ goToNextStep }: ProgressStepProps) => {
     [businessColor, businessId, businessName]
   );
 
+  const onPressCreate = useCallback(() => {
+    if (selectedPrepaidCard) {
+      showLoadingOverlay({ title: 'Creating profile...' });
+    } else {
+      navigate(MainRoutes.CHOOSE_PREPAIDCARD_SHEET, {
+        spendAmount: 100,
```
danibonilha: we should extract this to a const, like, profileFee or something, in case we need to change, it's easier

cardstack/src/screens/ProfileScreen/components/StepThree.tsx
```diff
@@ -24,6 +30,19 @@ export const StepThree = ({ goToNextStep }: ProgressStepProps) => {
     [businessColor, businessId, businessName]
   );
 
+  const onPressCreate = useCallback(() => {
+    if (selectedPrepaidCard) {
+      showLoadingOverlay({ title: 'Creating profile...' });
```
danibonilha: Don't think we should add this right now as we have no way to dismiss

cardstack/src/screens/ProfileScreen/components/StepThree.tsx
```diff
@@ -24,6 +30,19 @@ export const StepThree = ({ goToNextStep }: ProgressStepProps) => {
...
+    } else {
+      navigate(MainRoutes.CHOOSE_PREPAIDCARD_SHEET, {
+        spendAmount: 100,
+        onConfirmChoosePrepaidCard: (prepaidCard: PrepaidCardType) => {
+          setPrepaidCard(prepaidCard);
+        },
```
danibonilha:
```suggestion
        onConfirmChoosePrepaidCard: setPrepaidCard,
```

cardstack/src/screens/sheets/ChoosePrepaidCardSheet/useChoosePrepaidCard.ts
```diff
@@ -0,0 +1,78 @@
...
+    []
+  );
+
+  const onConfirmSelectedCard = useCallback(() => {
+    if (selectedPrepaidCard) {
+      onConfirmChoosePrepaidCard(selectedPrepaidCard);
+      goBack();
+    }
+  }, [selectedPrepaidCard, onConfirmChoosePrepaidCard, goBack]);
+
```
danibonilha, line 65: Ok, I noticed there are lots of duplicated logic between this and payMerchant, I understand one is a route and the other more like a component, but the ideal scenario would be to make it as shareable as possible. The only advantage with this approach is that we can make sure PayMerchant still works, so for the sake of stability it's okay to be like this for now, but would be nice to revisit asap.

cardstack/src/screens/sheets/ChoosePrepaidCardSheet/useChoosePrepaidCard.ts
```diff
@@ -0,0 +1,78 @@
...
+export const useChoosePrepaidCard = () => {
+  const {
+    params: { spendAmount = 0, onConfirmChoosePrepaidCard },
+  } = useRoute() as ChoosePrepaidCardSheetRouteParams;
+
+  const [selectedPrepaidCard, selectPrepaidCard] = useState<PrepaidCardType>();
+  const { accountAddress, network, nativeCurrency } = useAccountSettings();
+  const { goBack } = useNavigation();
+
+  const walletReady = useRainbowSelector(state => state.appState.walletReady);
```
danibonilha: why is walletReady being used here ?

cardstack/src/components/ChoosePrepaidCard/PrepaidCardItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+  networkName: string;
+  nativeCurrency: string;
+  currencyConversionRates: {
+    [key: string]: number;
+  };
+  spendAmount: number;
+  isLastItem: boolean;
+}
+
+export const PrepaidCardItem = memo(
```
danibonilha: nit: For debugging purposes it's better to export default and wrap in the memo, something like `export default memo(PrepaidCardItem);`  bc react is smart to give a name to the component, otherwise, it will be anonymous

cardstack/src/screens/sheets/ChoosePrepaidCardSheet/useChoosePrepaidCard.ts
topwebtek7: It's for handling wait till wallet's ready as soon as wallet is initialized.. but right, I think it's not needed here as wallet will be ready when user access choosing prepaid card sheet, so getting rid of it.

cardstack/src/screens/sheets/ChoosePrepaidCardSheet/useChoosePrepaidCard.ts
topwebtek7: yes good catch!

cardstack/src/components/ChoosePrepaidCard/PrepaidCardItem.tsx
topwebtek7: ok.. I think `export const PrepaidCardItem = `'s fine as different with export default. anyway will update as you suggested.

cardstack/src/screens/sheets/ChoosePrepaidCardSheet/useChoosePrepaidCard.ts
topwebtek7: yes, there's some duplicated part between ChoosePrepaidCard and PayMerchant..  PayMerchant uses all EditAmount, ChoosePrepaidCard, Confirmation in single screen so couldn't use ChoosePrepaidCard sheet. somehow shares ChoosePrepaidCard UI component so not that bad, I think we can revisit it later refactoring PayMerchant flow.

oidoug (approved):
Few things to consider:
- Choosing a prepaid card gets stuck on loading if on mainnet;
- My comment about separating the UI and actions hook got closed without a change, I think it's a good idea doing that moving forward.

topwebtek7:
> 

Good catch. I will add exception handling if no prepaidcard on mainnet. And for 2nd comment, we have useChoosePrepaidCard hook already and thought you missed that.. anyway will need to revisit ChoosePrepaidCard later as it has some duplicated part with PayMerchant, for now agreed leave it for now to keep stability.

PR #711, opened 2022-03-16
[CS-3308]: Show buy PrepaidCard btn on section header

danibonilha:
### Description

This PR adds the buy prepaid card button on the section header of prepaidcards, also removes the `New card` pill on empty state, since we already the buy prepaid as default we don't need it anymore.

### Screenshots

Empty State:
image

The section wasn't reordered yet when I recorded the video, but the behavior is the same
Simulator Screen Recording - iPhone 12 - 2022-03-16 at 11 49 50


oidoug (approved):
🚀

PR #712, opened 2022-03-16
[CS-3283] Fixes name on MerchantSafe, mistakenly using slug.

oidoug:
### Description

I was testing different values and forgot to revert to name.

I'm having issues mocking restyle, so tests will be pushed in a new PR.

- [x] Completes #(CS-3283)

### Checklist


topwebtek7 (approved):
👍🏼

PR #713, opened 2022-03-17
[Dept] Adds better tests for Merchant Screen and Safe.

oidoug:
### Description

This PR adds tests to MerchantSafe and more for Merchant Screen, in response to a bug with wrong values rendered in MerchantSafe.

### DIFF
--- a/cardstack/src/components/MerchantSafe/__tests__/MerchantSafe.test.tsx
+++ b/cardstack/src/components/MerchantSafe/__tests__/MerchantSafe.test.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { render } from '../../../test-utils';
 import { MerchantSafe } from '../MerchantSafe';
 import { strings } from '../strings';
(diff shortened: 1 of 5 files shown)


cardstack/src/components/MerchantSafe/__tests__/MerchantSafe.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    textColor: 'black',
+    ownerAddress: '0xOwner',
+  },
+};
+
+// Mock navigation
+const mock_navigate = jest.fn();
+jest.mock('@rainbow-me/navigation', () => ({
+  useNavigation: () => ({
+    navigate: mock_navigate,
```
danibonilha: we just need to add an external mock function, if we are using it to assert something, so if we don't want to check if navigate was called we can this add `navigate: jest.fn()`

danibonilha (approved):
👏

PR #714, opened 2022-03-17
[CS-3411]: Fix tab profile/wallet switching 

danibonilha:
### Description

One of the toast positioning was rendering on the tab bar between the two tabs profile/wallet, added a negative bottom to make sure it's positioned outside the screen.

Simulator Screen Recording - iPhone 12 - 2022-03-17 at 10 58 56


topwebtek7 (approved):
LGTM

oidoug (approved):
Great that you found it! Thanks 👏

PR #718, opened 2022-03-18
[CS-3373] Create Reward Center Screen

oidoug:
### Description

This PR improves navigation for RewardsCenterScreen;
Adds new Mainheader/NavigationStackheader more appropriate to stack navigation;
Adds header banner to screen';
Creates RewardContent and RewardRow components.

- [x] Completes #(CS-3373)
- [x] Completes #(CS-3374)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/159005815-b2577ccc-2d3d-4926-8c9d-e34f15c30245.mp4


cardstack/src/components/MainHeader/NavigationStackHeader.tsx
```diff
@@ -0,0 +1,52 @@
...
+            <Text color="white" size="body" weight="bold">
+              {title}
+            </Text>
+          )}
+        </Container>
+      ) : (
+        children
+      )}
+      {/* Right end spacing so title gets properly centered. */}
+      <Container width={SIDES_SIZE} />
```
danibonilha: we could use flex, instead of fixed values, for the items at the sides we could have flex : 1 and for the text container flex: 4

cardstack/src/components/MainHeader/NavigationStackHeader.tsx
```diff
@@ -0,0 +1,52 @@
...
+    </MainHeaderWrapper>
+  );
+};
```
danibonilha, line 48: The header height doesn't quite match the others, probably bc it doesn't have the network, we can check if we want to show the network in this case, and if not, we need to create a fixed height header with a scaled value, but in this case we can have a new task to deal with that.

danibonilha, line 48: The header height doesn't quite match the others, probably bc it doesn't have the network, we can check if we want to show the network in this case, and if not, we need to create a fixed height header with a scaled value, but in this case we can have a new task to deal with that.

cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
```diff
@@ -1,19 +1,20 @@
...
-      alignContent="center"
-      justifyContent="center"
-      flex={1}
-    >
-      {rewardsSafe}
-      <Button onPress={onRegisterPress}>{strings.buttonRegisterText}</Button>
+    <Container backgroundColor="white" flex={1}>
+      <NavigationStackHeader title={strings.navigation.title} />
+      <Container backgroundColor="white" flex={1}>
+        <Image source={rewardBanner} />
```
danibonilha: image

The image is not filling the whole space on iPhone 12

danibonilha: image

The image is not filling the whole space on iPhone 12

cardstack/src/screens/RewardsCenterScreen/components/RegisterContent.tsx
```diff
@@ -0,0 +1,43 @@
...
+  subText: string;
+  onRegisterPress: () => void;
+}
+
+export const RegisterContent = ({
+  onRegisterPress,
+  primaryText = '3,200 CARD.CPXD',
+  subText = '$45.00 USD',
+}: RegisterContentProps) => (
+  <Container padding={5} alignItems="center">
```
danibonilha: I'd rather add the horizontal padding in the parent, so we don't need to add it to all the children, and in case we need to reuse any of this we don't need to worry about having extra padding

danibonilha: I'd rather add the horizontal padding in the parent, so we don't need to add it to all the children, and in case we need to reuse any of this we don't need to worry about having extra padding

danibonilha (commented):
It's looking good, but there are some tweaks to be made

cardstack/src/components/MainHeader/NavigationStackHeader.tsx
oidoug: To keep UI scalable and consistent with design mocks, I prefer to have edge elements and height sizes discrete, in pts.
It's the same problem we are having with the navbar.

I'll use flex now, but we might want to revise this approach at some point - and have a single navbar component to handle all cases.

cardstack/src/components/MainHeader/NavigationStackHeader.tsx
oidoug: In this case, the height is not matching because the title font size is smaller in this screen than the title in the Settings screen. I'm following the design for the font size, maybe we should do the same in Settings, or have a fixed height for the header content.

kdowlin: Let me know the path you choose and if a new ticket is needed

danibonilha (approved):
LGTM

PR #720, opened 2022-03-21
[CS-3425]: Delay longPress action on prepaidCard

danibonilha:
### Description
Decreases the long press delay to transfer card, also it extracts to a const so we can keep it consistent.


oidoug (approved):
👍

topwebtek7 (approved):
LGTM!

PR #722, opened 2022-03-21
[CS-3366, CS-3435] Add Create profile confirmation, update creating profile flow integrating with sdk, RTK

topwebtek7:
### Description

- Added Create profile confirmation screen
- Created TransactionConfirmationSheet so existing TransactionConfirmation screens can be used without WC notification
- Implemented create profile flow using SDK and RTK mutation
- Improved user experience with create profile flow

note: used existing overlay screen and success notification temporarily and will be updated with provided design.
Trying to add recorded video shortly.

ToDo: update screens with updated designs and add validation messages, success message

- [x] Completes #CS-3366
- [x] Completes #CS-3435

### Checklist

### Screenshots

Create profile success -

https://user-images.githubusercontent.com/16714648/159729082-e1709a50-d95d-4c5c-99d9-71d7f5b082da.mp4

Create profile failed handling -

https://user-images.githubusercontent.com/16714648/159729106-3e53d98f-a4bc-4973-aa01-32da0569656c.mp4

### DIFF
--- a/cardstack/src/navigation/hooks.tsx
+++ b/cardstack/src/navigation/hooks.tsx
@@ -1,4 +1,5 @@
 import React, { useCallback } from 'react';
+import { InteractionManager } from 'react-native';
 import { useNavigation } from '@react-navigation/core';
 import { MainRoutes, GlobalRoutes } from './routes';
 import { MainScreens, GlobalScreens, ScreenNavigation } from './screens';
@@ -33,6 +34,12 @@ export const useDismissCurrentRoute = (routeName: string) => {
   const checkAndDismissCurrentRoute = useCallback(() => {
     if (Navigation.getActiveRouteName() === routeName) {
       goBack();
+      InteractionManager.runAfterInteractions(() => {
+        // this fixes a bug where the overlay dismiss didn't work on create profile, will investigate more
+        if (Navigation.getActiveRouteName() === routeName) {
+          goBack();
+        }
+      });
     }
   }, [goBack, routeName]);
 
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -33,6 +33,7 @@ import {
   MerchantTransactionSheet,
   ChoosePrepaidCardSheet,
   RewardsCenterScreen,
+  TransactionConfirmationSheet,
 } from '@cardstack/screens';
 import {
   bottomSheetPreset,
@@ -176,6 +177,10 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
   REWARDS_CENTER_SCREEN: {
     component: RewardsCenterScreen,
     options: horizontalInterpolator,
+  },
+  TRANSACTION_CONFIRM: {
+    component: TransactionConfirmationSheet,
+    options: expandedPreset as StackNavigationOptions,
   },
 };
 
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -181,6 +181,7 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
   TRANSACTION_CONFIRMATION_SHEET: {
     component: TransactionConfirmationSheet,
     options: expandedPreset as StackNavigationOptions,
+    listeners: dismissAndroidKeyboardOnClose,
   },
 };
 
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -178,7 +178,7 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
     component: RewardsCenterScreen,
     options: horizontalInterpolator,
   },
-  TRANSACTION_CONFIRM: {
+  TRANSACTION_CONFIRMATION_SHEET: {
     component: TransactionConfirmationSheet,
     options: expandedPreset as StackNavigationOptions,
   },
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -181,7 +181,6 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
   TRANSACTION_CONFIRMATION_SHEET: {
     component: TransactionConfirmationSheet,
     options: expandedPreset as StackNavigationOptions,
-    listeners: dismissAndroidKeyboardOnClose,
   },
 };
 
--- a/cardstack/src/notification-handler/index.ts
+++ b/cardstack/src/notification-handler/index.ts
@@ -29,7 +29,8 @@ interface LocalNotificationDataType {
 }
 interface LocalNotificationType {
   notification: { title?: string; body: string };
-  data: LocalNotificationDataType;
+  data?: LocalNotificationDataType;
+  isManualNotification?: boolean;
 }
 
 interface NotificationConfig {
@@ -84,13 +85,14 @@ export const displayLocalNotification = async (
     const {
       notification: { title, body },
       data,
+      isManualNotification,
     } = notificationData;
 
-    const { notificationType, network } = data;
+    const { notificationType, network } = data || {};
     const currentNetwork = await getNetwork();
 
     // show local notification in same network only
-    if (network === currentNetwork) {
+    if (network === currentNetwork || isManualNotification) {
       // update notification title when no title from hub side(mostly had no title)
       const notificationTitle =
         title ||
@@ -99,7 +101,7 @@ export const displayLocalNotification = async (
       // A channel is required on Android for displaying local notifications.
       // The default channel can't have custom 'importance'.
       const channelId = await notifee.createChannel({
-        id: notificationType,
+        id: isManualNotification ? 'ManualLocalNotification' : notificationType,
         name: notificationTitle,
         importance: AndroidImportance.HIGH,
       });
--- a/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
+++ b/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
@@ -1,10 +1,26 @@
-import React from 'react';
+import React, { useMemo } from 'react';
+import { ActivityIndicator } from 'react-native';
 import { CreateProfile, strings } from './components';
 import { Container, MainHeader, MerchantContent } from '@cardstack/components';
 import { usePrimarySafe } from '@cardstack/redux/hooks/usePrimarySafe';
 
 const ProfileScreen = () => {
   const { primarySafe, isFetching, safesCount } = usePrimarySafe();
+
+  const ProfileBody = useMemo(
+    () =>
+      primarySafe ? (
+        <MerchantContent
+          showSafePrimarySelection={safesCount > 1}
+          isPrimarySafe
+          merchantSafe={primarySafe}
+          isRefreshingBalances={isFetching}
+        />
+      ) : (
+        <CreateProfile />
+      ),
+    [primarySafe, isFetching, safesCount]
+  );
 
   return (
     <Container
@@ -13,15 +29,10 @@ const ProfileScreen = () => {
     >
       <MainHeader title={strings.title} />
       <Container justifyContent="center" flexGrow={1}>
-        {primarySafe ? (
-          <MerchantContent
-            showSafePrimarySelection={safesCount > 1}
-            isPrimarySafe
-            merchantSafe={primarySafe}
-            isRefreshingBalances={isFetching}
-          />
+        {!primarySafe && isFetching ? (
+          <ActivityIndicator size="large" color="white" />
         ) : (
-          <CreateProfile />
+          ProfileBody
         )}
       </Container>
     </Container>
--- a/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
+++ b/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
@@ -27,7 +27,7 @@ const ProfileScreen = () => {
       backgroundColor={primarySafe ? 'white' : 'backgroundDarkPurple'}
       flex={1}
     >
-      <MainHeader title={strings.title} />
+      <MainHeader title={strings.header.profile} />
       <Container justifyContent="center" flexGrow={1}>
         {!primarySafe && isFetching ? (
           <ActivityIndicator size="large" color="white" />
--- a/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
+++ b/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
@@ -16,7 +16,7 @@ const ProfileScreen = () => {
         {primarySafe ? (
           <MerchantContent
             showSafePrimarySelection={safesCount > 1}
-            isPrimarySafe={true}
+            isPrimarySafe
             merchantSafe={primarySafe}
             isRefreshingBalances={isFetching}
           />
--- a/cardstack/src/screens/ProfileScreen/components/strings.ts
+++ b/cardstack/src/screens/ProfileScreen/components/strings.ts
@@ -19,6 +19,10 @@ export const strings = {
   businessIdShouldBeUnique:
     'This Business ID is already taken. Please choose another one',
   choosePrepaidCard: 'Choose a Prepaid Card',
+  createProfileErrorMessage:
+    'Could not create profile, please try again. If this problem persists please reach out to support@cardstack.com',
+  profileCreated: 'Profile Created',
+  profileCreatedMessage: 'Your profile has been created successfully!',
 };
 
 export const exampleMerchantData = {
--- a/cardstack/src/screens/ProfileScreen/components/strings.ts
+++ b/cardstack/src/screens/ProfileScreen/components/strings.ts
@@ -23,6 +23,7 @@ export const strings = {
     'Could not create profile, please try again. If this problem persists please reach out to support@cardstack.com',
   profileCreated: 'Profile Created',
   profileCreatedMessage: 'Your profile has been created successfully!',
+  creatingProfile: 'Creating Profile',
 };
 
 export const exampleMerchantData = {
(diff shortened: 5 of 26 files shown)


cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -113,6 +143,101 @@ export const useProfileForm = (params?: useProfileFormParams) => {
     [accountSymbol, businessName]
   );
 
+  const createProfile = useCallback(async () => {
```
oidoug: Suggestion: This function seems out of place considering `useProfileForm` deals with UI updates, so by separation of concerns I would break the service part into a service file or similar.

cardstack/src/screens/sheets/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
```diff
@@ -0,0 +1,30 @@
...
+} from '@cardstack/components';
+
+const TransactionConfirmation = () => {
+  const {
+    data,
+    loading,
+    onCancel,
+    onConfirm,
+    onConfirmLoading,
+  } = useTransactionConfirmationSheet();
```
oidoug, line 15: 👏

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -113,6 +143,101 @@ export const useProfileForm = (params?: useProfileFormParams) => {
...
+      if (!did) {
+        return;
+      }
+
+      const address = (await loadAddress()) || '';
+
+      const allWallets = await getAllWallets();
+
+      const walletId =
+        getWalletByAddress({ walletAddress: address, allWallets })?.id || '';
```
danibonilha: we have a hook with this info, useWallets, returns the selected wallet

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -113,6 +143,101 @@ export const useProfileForm = (params?: useProfileFormParams) => {
...
+        network,
+      });
+
+      const revenuePool = await getSDK('RevenuePool', web3);
+
+      try {
+        const newMerchant = await revenuePool.registerMerchant(
+          selectedPrepaidCard?.address,
+          did
+        );
```
danibonilha: this should be part of the merchant-apis, by creating this as a mutation you can just invalidate the tag, so it will refetch the safes automatically for example

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -113,6 +143,101 @@ export const useProfileForm = (params?: useProfileFormParams) => {
...
+
+        if (newMerchant?.merchantSafe?.address) {
+          await refetchSafes();
+          navigate(RainbowRoutes.PROFILE_SCREEN);
+        }
+      } catch (e) {
+        logger.sentry('Hub authenticate failed', e);
```
danibonilha: This log doesn't seem right

cardstack/src/screens/sheets/TransactionConfirmationSheet/useTransactionConfirmationSheet.ts
```diff
@@ -0,0 +1,35 @@
...
+  const {
+    params: { data, onConfirm },
+  } = useRoute<RouteType<Params>>();
+
+  const { goBack } = useNavigation();
+
+  const {
+    callback: onConfirmCallback,
+    isLoading: onConfirmLoading,
+  } = useWorker(onConfirm, []);
```
danibonilha: not sure all onConfirm actions will be a promise

cardstack/src/screens/sheets/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
danibonilha, line 15: I like the attempt to use this pattern, but for this case, I'm not sure how it makes it better, since all this info might came already from the navigation param.

cardstack/src/types/transaction-confirmation-types.ts
```diff
@@ -45,8 +46,9 @@ export interface IssuePrepaidCardDecodedData {
 
 export interface RegisterMerchantDecodedData {
   spendAmount: number;
-  infoDID: string;
-  prepaidCard: string;
+  infoDID?: string;
+  merchantInfo?: MerchantInformation;
+  prepaidCard?: string;
```
danibonilha: I'm not sure PrepaidCard can be optional here, if we need one to pay for the merchant account

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -113,6 +143,101 @@ export const useProfileForm = (params?: useProfileFormParams) => {
...
+        'text-color': '#fff',
+      };
+
+      const did = await createBusinessInfoDID(merchantInfoData, authToken);
+
+      if (!did) {
+        return;
+      }
+
+      const address = (await loadAddress()) || '';
```
danibonilha: don't we already have the address on accountAddress ?

cardstack/src/screens/sheets/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
topwebtek7: I also tried to use transaction-confirmation-strategies but that makes things more complicated cause strategies has decode behaviors and we already have all info needed in this kind cases so made just this TransactionConfirmationSheet screen to display confirmation sheet using existing confirmation templates

cardstack/src/screens/sheets/ChoosePrepaidCardSheet/useChoosePrepaidCard.ts
```diff
@@ -46,8 +46,26 @@ export const useChoosePrepaidCard = () => {
...
+      selectPrepaidCard(prepaidCards[0]);
+      onConfirmChoosePrepaidCard(prepaidCards[0]);
+    }
+  }, [
+    goBack,
+    isLoading,
+    onConfirmChoosePrepaidCard,
+    prepaidCards,
+    spendAmount,
+  ]);
```
danibonilha: not sure about this behavior, since it triggers the screen anyway, it looks like a bug when the screen flashes, it might be better to let the user continue the flow even with a single card, I mean, we can pre-select the card, but not dismiss the screen

cardstack/src/screens/sheets/TransactionConfirmationSheet/useTransactionConfirmationSheet.ts
```diff
@@ -0,0 +1,38 @@
...
+  } = useRoute<RouteType<Params>>();
+
+  const { goBack } = useNavigation();
+
+  const {
+    callback: onConfirmCallback,
+    isLoading: onConfirmLoading,
+  } = useWorker(async () => {
+    await onConfirm();
+  }, []);
```
danibonilha, line 25: not sure everything will be a promise here, we probably don't need this

cardstack/src/screens/sheets/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
danibonilha: I meant, was that we may not need the `useTransactionConfirmationSheet` , bc `useRoute` already returns all the info for the screen.

```ts
  const { params: {
    data,
    loading,
    onCancel,
    onConfirm,
  } } = useRoute()
```

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -113,6 +175,87 @@ export const useProfileForm = (params?: useProfileFormParams) => {
...
+        }
+
+        await createProfile({
+          selectedWallet,
+          network,
+          selectedPrepaidCardAddress: selectedPrepaidCard.address,
+          profileDID,
+        });
+
+        refetchSafes();
```
danibonilha: If we are invalidating the Safes tag, it already refetches automatically, no need for calling refetchSafes

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -1,18 +1,85 @@
...
+        },
+        data: {
+          notificationType: 'CreateProfile', // not important value, just for notification channelId
+          network,
+        },
+      });
+
+      navigate(RainbowRoutes.PROFILE_SCREEN);
+    }
+  }, [dismissLoadingOverlay, isSuccess, isFetching, navigate, network]);
```
danibonilha: we have an useMutationEffects that might abstract this a little

cardstack/src/screens/sheets/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
topwebtek7: ok, just wanted to split out feature(onConfirm, onCancel, onConfirmLoading behaviors) and UI like others

cardstack/src/screens/sheets/ChoosePrepaidCardSheet/useChoosePrepaidCard.ts
topwebtek7: already got rid of this in last commit 👍🏼

cardstack/src/screens/sheets/TransactionConfirmationSheet/useTransactionConfirmationSheet.ts
topwebtek7: yes right. somehow I think we can handle both cases(promise, not promise) with this code

cardstack/src/screens/ProfileScreen/useProfileForm.ts
topwebtek7: oh ok, good to know. then will clean it up

cardstack/src/notification-handler/index.ts
```diff
@@ -29,7 +29,8 @@ interface LocalNotificationDataType {
 }
 interface LocalNotificationType {
   notification: { title?: string; body: string };
-  data: LocalNotificationDataType;
+  data?: LocalNotificationDataType;
+  isManualNotification?: boolean;
```
oidoug, line 33: Can't `isManualNotification` attribute be part of `LocalNotificationDataType`?

cardstack/src/notification-handler/index.ts
```diff
@@ -99,7 +101,7 @@ export const displayLocalNotification = async (
       // A channel is required on Android for displaying local notifications.
       // The default channel can't have custom 'importance'.
       const channelId = await notifee.createChannel({
-        id: notificationType,
+        id: isManualNotification ? 'ManualLocalNotification' : notificationType,
```
oidoug, line 104: You should add this `ManualLocalNotification` also in `NotificationType` enum.

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -1,27 +1,38 @@
...
+        />
+      ) : (
+        <CreateProfile />
+      ),
+    [primarySafe, isFetching, safesCount]
+  );
+
   return (
     <Container
       backgroundColor={primarySafe ? 'white' : 'backgroundDarkPurple'}
```
oidoug, line 27: just occurred to me that while our safes are being fetched, we have a dark background, and after that it jumps to be white... We might need to rethink this, or even better, have a dark layout for MerchantContent.

cardstack/src/navigation/screens.ts
```diff
@@ -177,6 +178,11 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
     component: RewardsCenterScreen,
     options: horizontalInterpolator,
   },
+  TRANSACTION_CONFIRMATION_SHEET: {
+    component: TransactionConfirmationSheet,
+    options: expandedPreset as StackNavigationOptions,
+    listeners: dismissAndroidKeyboardOnClose,
```
danibonilha: Why do we need this listener?

cardstack/src/services/merchant/merchant-service.ts
```diff
@@ -49,3 +52,24 @@ export const claimMerchantRevenue = async ({
...
+    walletId: selectedWallet.id,
+    network,
+  });
+
+  const revenuePool = await getSDK('RevenuePool', web3);
+
+  const newProfile = await revenuePool.registerMerchant(
+    selectedPrepaidCardAddress,
+    profileDID
+  );
```
danibonilha, line 75: we should add the from param, contract options here, otherwise we are unable to create a profile on a derived account

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -113,6 +170,84 @@ export const useProfileForm = (params?: useProfileFormParams) => {
...
+        const profileDID = await createBusinessInfoDID(
+          merchantInfoData,
+          authToken
+        );
+
+        if (!profileDID) {
+          return;
+        }
+
+        await createProfile({
```
danibonilha: we don't need this await

danibonilha:
Not sure why, but it seems the error case is not dismissing the overlay, idk if it's bc it's a context, never experienced this issue before.

cardstack/src/navigation/screens.ts
topwebtek7: yes right, we don't need this listener on this screen, will remove

cardstack/src/screens/ProfileScreen/useProfileForm.ts
topwebtek7: I think we need await here cause we wait for `onConfirmCreateProfile` to be completed and uses onConfirmLoading, catching error inside `TransactionConfirmationSheet` screen using useWorker

cardstack/src/services/merchant/merchant-service.ts
topwebtek7: I've tested several times and was able to create profile on derived accounts properly without from param.. maybe it's related with which account this prepaid card is included.. do you still think we need from param here to make sure?

cardstack/src/screens/ProfileScreen/components/strings.ts
```diff
@@ -12,13 +12,17 @@ export const strings = {
...
   yourProfile: 'Your Profile:',
   create: 'Create',
   businessNameRequired: 'Business Name is required',
   businessIdShouldBeUnique:
     'This Business ID is already taken. Please choose another one',
   choosePrepaidCard: 'Choose a Prepaid Card',
+  createProfileErrorMessage:
+    'Could not create profile, please try again. If this problem persists please reach out to support@cardstack.com',
+  profileCreated: 'Profile Created',
+  profileCreatedMessage: 'Your profile has been created successfully!',
```
oidoug: I think it was commented before, but just as a heads up, it would be better to have the strings separated objects contextualized by where in the screen they are, like:
```
{
 alert: {
   profile...
 },
 header: {
   title...
 }, etc...
}
```

cardstack/src/notification-handler/index.ts
topwebtek7: I added `isManualNotification` param to do not use `LocalNotificationDataType` for this kind local notifications. anyway it's temporary workaround for now

cardstack/src/notification-handler/index.ts
topwebtek7: basically ManualLocalNotification's not notification type.. and it's just channel id only for manual notifications. If add ManualLocalNotification in NotificationType, it makes types error in `notificationConfig`..

cardstack/src/screens/ProfileScreen/components/strings.ts
topwebtek7: ok, then will separate by screen.. basically I didn't separate cause there's some duplicates, and others are for error messages etc.

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
topwebtek7: basically that's why I added ActivityIndicator while fetching. There's loading indicator while fetching so looks fine to me.. anyway we can discuss in meeting if you think need better designs

cardstack/src/services/merchant/merchant-service.ts
danibonilha: That's odd, I've tested here and it didn't work, I got the signer is not the owner of the. account etc, better add it to be safe

topwebtek7: ok, sounds good.

topwebtek7:
> Not sure why, but it seems the error case is not dismissing the overlay, idk if it's bc it's a context, never experienced this issue before.

@danibonilha Fixed the dismissing overlay issue and updated recorded videos for success on derived account, and error handling, fyi.

cardstack/src/navigation/hooks.tsx
```diff
@@ -33,6 +34,12 @@ export const useDismissCurrentRoute = (routeName: string) => {
   const checkAndDismissCurrentRoute = useCallback(() => {
     if (Navigation.getActiveRouteName() === routeName) {
       goBack();
+      InteractionManager.runAfterInteractions(() => {
+        // this fixes a bug where the overlay dismiss didn't work on create profile, will investigate more
+        if (Navigation.getActiveRouteName() === routeName) {
+          goBack();
+        }
+      });
```
danibonilha, line 42: did you check other flows that are using the overlay ? just to make sure this doesn't break anything else ?

topwebtek7: Yes checked other places using overlay and it didn't break any other things

cardstack/src/notification-handler/index.ts
danibonilha: What if we add a `profileCreation` into the notificationType enum, with the default title, and make handler optional on `NotificationConfig` if we don't need it.

topwebtek7: Basically `NotificationType` is notification types enum that comes from Hub server side(we can move this enum to sdk side if there's more notifications to share with Hub side), and if add new type here might be confused with real pushnotification types that we're using in `notificationHandler` as well. That's why I added a new parameter in `LocalNotificationType` rather than add new type in `NotificationType`.
@oidoug @danibonilha If it does not make sense, please let me know better suggestions.

danibonilha (approved):

PR #724, opened 2022-03-21
[CS-3446]: Add fixed header height

danibonilha:
### Description
Updated the header height 

### Checklist

### Screenshots

Simulator Screen Recording - iPhone 12 - 2022-03-21 at 17 30 58

Not sure why the scan got messed up on the .gif, but it's working properly as you can see on the next screenshoot
Simulator Screen Recording - iPhone SE (2nd generation) - 2022-03-21 at 17 31 20

image


oidoug (approved):
👏

PR #725, opened 2022-03-22
[CS-3453]: Add reward register mutation 

danibonilha:
### Description
Adds the registration flow with mocked data and alerts, to get the flow working, it's not possible yet to retrieve the gas estimation from the sdk, so there's a mocked value, I'm using the alert to show the info until we build the UI and while adding the logic to claim the rewards, we might want to split the useRewardsCenter hook into useRegistration and useClaim or something, cause it's starting to get big. Also it orders the prepaidCard list by the spendable balance.
We also need to rethink about the choosePrepaidCard screen and maybe hide the `To pay this amount`, but it can be added to the backlog

### Screenshots

Simulator Screen Recording - iPhone 12 - 2022-03-22 at 13 31 04

### DIFF
--- a/cardstack/src/screens/RewardsCenterScreen/strings.ts
+++ b/cardstack/src/screens/RewardsCenterScreen/strings.ts
@@ -14,6 +14,7 @@ export const strings = {
       message:
         'Before you can claim, you need to create and register a reward account.  This is an on-chain event and will cost you a small gas fee - you can use any existing prepaid card to pay the transaction fee.',
     },
+    loading: 'Registering Account',
   },
   claim: {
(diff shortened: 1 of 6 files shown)


oidoug (approved):
Nice!

oidoug:
Question: How will we update the UI after a successful register? Will `mainPoolTokenInfo` update automatically within the hook?

danibonilha:
> Question: How will we update the UI after a successful register? Will `mainPoolTokenInfo` update automatically within the hook?

By invalidating the rewardsSafes tag, the refetch is trigger automatically

PR #726, opened 2022-03-23
[CS-3442] Claim sections and tabs UI

oidoug:
### Description

This PR adds an MVP for the ClaimRewards content, adds TabHeader component with a hook for keeping track of the selected tab, and extracts RewardsTitle to its own component.

The structure of this screen makes it complex to implement, especially since we would need a header that is a dynamic list and tabs that also have dynamic section lists. I'm studying a better approach but didn't want to hold off on the MVP so the development of the complete picture could continue.

- [x] Completes #(CS-3442)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/159706735-53ef6d7d-fe57-4516-9583-93bbd82f0432.mp4

### DIFF
--- a/cardstack/src/components/Button/Button.tsx
+++ b/cardstack/src/components/Button/Button.tsx
@@ -69,8 +69,6 @@ export const Button = ({
     props.variant
   );
 
-  console.log('::: textStyle bt', children, textStyle);
-
   const disabledTextStyle = useVariantValue(
     'buttonVariants',
     'disabledTextStyle',
--- a/cardstack/src/components/Button/Button.tsx
+++ b/cardstack/src/components/Button/Button.tsx
@@ -69,6 +69,8 @@ export const Button = ({
     props.variant
   );
 
+  console.log('::: textStyle bt', children, textStyle);
+
   const disabledTextStyle = useVariantValue(
     'buttonVariants',
     'disabledTextStyle',
@@ -115,7 +117,13 @@ export const Button = ({
               {...iconProps}
             />
           )}
-          <Text {...textStyle} {...disabledTextProps} allowFontScaling={false}>
+          <Text
+            {...textStyle}
+            {...disabledTextProps}
+            // Fixes text issue in buttons with set height prop
+            style={{ height: '100%' }}
+            allowFontScaling={false}
+          >
             {children}
           </Text>
         </Container>
--- a/cardstack/src/components/Button/Button.tsx
+++ b/cardstack/src/components/Button/Button.tsx
@@ -10,7 +10,7 @@ import {
   BorderProps,
   backgroundColor,
 } from '@shopify/restyle';
-import { ActivityIndicator } from 'react-native';
+import { ActivityIndicator, StyleSheet } from 'react-native';
 import React, { ReactNode, useMemo } from 'react';
 
 import { Text } from '../Text';
@@ -49,6 +49,13 @@ const AnimatedButton = createRestyleComponent<ButtonProps, Theme>(
   [layout, spacing, border, VariantRestyleComponent, backgroundColor],
   AnimatedPressable
 );
+
+const styles = StyleSheet.create({
+  // Ensures button's title size integrity if button's height gets to be manually set.
+  title: {
+    height: '100%',
+  },
+});
 
 /**
  * A button with a simple press animation
@@ -118,8 +125,7 @@ export const Button = ({
           <Text
             {...textStyle}
             {...disabledTextProps}
-            // Fixes text issue in buttons with set height prop
-            style={{ height: '100%' }}
+            style={styles.title}
             allowFontScaling={false}
           >
             {children}
--- a/cardstack/src/components/TabHeader/TabHeaderButton.tsx
+++ b/cardstack/src/components/TabHeader/TabHeaderButton.tsx
@@ -18,7 +18,7 @@ export const TabHeaderButton = ({
       justifyContent="center"
       onPress={onPress}
       paddingTop={5}
-      width="50%"
+      flex={1}
     >
       <Text
         color="black"
--- a/cardstack/src/components/TabHeader/TabHeaderButton.tsx
+++ b/cardstack/src/components/TabHeader/TabHeaderButton.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import { Container, Touchable, Text } from '@cardstack/components';
+
+export interface TabHeaderProps {
+  title: string;
+  isSelected?: boolean;
+  onPress: () => void;
+}
+
+export const TabHeaderButton = ({
+  title,
+  isSelected = false,
+  onPress,
+}: TabHeaderProps) => {
+  return (
+    <Touchable
+      alignItems="center"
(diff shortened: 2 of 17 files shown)


cardstack/src/components/Button/Button.tsx
```diff
@@ -115,7 +115,13 @@ export const Button = ({
               {...iconProps}
             />
           )}
-          <Text {...textStyle} {...disabledTextProps} allowFontScaling={false}>
+          <Text
+            {...textStyle}
+            {...disabledTextProps}
+            // Fixes text issue in buttons with set height prop
+            style={{ height: '100%' }}
```
danibonilha: let's extract to a style const

cardstack/src/components/TabHeader/TabHeaderButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <Text
+        color="black"
+        marginBottom={3}
+        opacity={isSelected ? 1 : 0.5}
+        weight={isSelected ? 'bold' : 'regular'}
+      >
+        {title}
+      </Text>
+      <Container
+        backgroundColor={isSelected ? 'buttonPrimaryBorder' : 'white'}
```
danibonilha, line 32: it saddens me that Restyle make it so hard to type that it's difficult  to extract these isSelected conditions to a variable.

cardstack/src/components/TabHeader/TabHeaderButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+  isSelected = false,
+  onPress,
+}: TabHeaderProps) => {
+  return (
+    <Touchable
+      alignItems="center"
+      justifyContent="center"
+      onPress={onPress}
+      paddingTop={5}
+      width="50%"
```
danibonilha: Instead of `width='50%` we should set to `flex={1}`, if so we add a third tab for example all of them will take the same available space, with width, even mapping out, just 2 tabs will be shown

cardstack/src/components/TabHeader/useTabHeader.tsx
```diff
@@ -0,0 +1,42 @@
...
+export interface TabProps {
+  title: string;
+}
+
+export interface TabsProps {
+  tabs: Array<TabProps>;
+}
+
+export const useTabHeader = ({ tabs }: TabsProps) => {
+  const [currentTab, setCurrentTab] = useState(0);
```
danibonilha: nit: I'd rather have a enum with a string or use the title as key or something instead of relying on indexes, specially because we will use the key to render the tabs, so in the case of someone reordering the enum we would be safe, but it's just a personal opinion.

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
```diff
@@ -0,0 +1,126 @@
...
+  {
+    coinSymbol: 'CARD',
+    primaryText: '100 CARD.CPXD',
+  },
+  {
+    coinSymbol: 'CARD',
+    primaryText: '50 CARD.CPXD',
+  },
+];
+
```
danibonilha, line 19: I think this was nice to see the UI but we can remove it, and show the No reward balance and history state, for the balance we have the rewardSafes, you can just take the first item of the array for now, since we are consideration just one safe

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
```diff
@@ -0,0 +1,126 @@
...
+enum Tabs {
+  BALANCE,
+  HISTORY,
+}
+
+export const ClaimContent = () => {
+  const { TabHeader, currentTab } = useTabHeader({
+    tabs: [{ title: strings.balance.title }, { title: strings.history.title }],
+  });
```
danibonilha: following up my nit comment, it would be something like: 
Outside the component: 
```ts
enum Tabs {
  BALANCE = 'BALANCE',
  HISTORY = 'HISTORY',
}

const tabs = [{ title: strings.balance.title, key:  Tabs.BALANCE }, { title: strings.history.title, key:  Tabs.HISTORY }],
```

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
```diff
@@ -0,0 +1,126 @@
...
+        />
+        {renderClaimList()}
+      </Container>
+      <TabHeader />
+      <Container padding={5}>
+        {currentTab === Tabs.BALANCE
+          ? renderBalanceList()
+          : renderHistoryList()}
+      </Container>
+    </ScrollView>
```
danibonilha, line 71: not sure the scrollview is the best approach here, for Balances I think it's okay to map over, but History for example, we might have a huge amount of data, which is better leverage by using a `FlatList, SectionList`, also it's not scrolling until the end

cardstack/src/components/TabHeader/TabHeaderButton.tsx
oidoug: True, but tbh I don't mind these small ternaries being around, in the end, we are doing a lot of in-place styling with restyle anyway.

cardstack/src/components/TabHeader/useTabHeader.tsx
oidoug: Agreed, I overlooked this.

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
oidoug: Yes, I agree, but I was hoping I could get the signature of the data we're showing before implementing the SectionList. I started it and went back to mapping so I could get what I had out sooner.

Do you have by chance mock data for the history list?

danibonilha: So the history I think we'll probably need to add a transaction mapping similar to the other transaction of the app, I think this can be a separate ticket, you can create the history with the data array empty and a listEmpty component, then I can show you the way we mapped other transactions

oidoug:
@danibonilha I don't have any reward in Sokol either, so I appreciate it if you can test this for me.

danibonilha (approved):
maybe we can improve the onClaimPress params, but I can do it in my PR, overall it's looking good 
image

cardstack/src/screens/RewardsCenterScreen/components/RegisterContent.tsx
```diff
@@ -16,21 +16,16 @@ export const RegisterContent = ({
...
+  <Container alignItems="center" padding={5}>
+    <RewardsTitle title={strings.register.title} paddingBottom={5} />
     <RewardRow
       coinSymbol={coinSymbol}
       primaryText={primaryText}
       subText={subText}
       paddingBottom={5}
+      onClaimPress={() => {
+        //pass
+      }}
```
danibonilha: we can remove this, because this state, means the user is unable to claim, and if we pass this function, the claim button appears

PR #727, opened 2022-03-23
[CS-3458]: Add ability to claim rewards

danibonilha:
### Description
This adds the ability to claim the rewards, ~~right now there's an issue where we are not able to claim multiple proofs at once, so I structured the code as we would use for claim multiple, but it has a temp slice to get just one item~~. Nvm just added to be sequential, it seems to work, although sometimes the balances are not up-to-date, but we can investigate on a diff ticket. The UI is just a placeholder to be able to claim and following improvements includes splitting the screen hook into smaller ones and handling fetching states. Also we are missing the confirmation to claim step.

### Screenshots

Each proof is worth around 10 CARD, since we are claming just one proof at a time the balance is updated to 10 CARD less
after the successful claim:
Simulator Screen Recording - iPhone 12 - 2022-03-23 at 15 28 25


oidoug (approved):
After claiming, I got the success alert, but on accepting it, I got redirected out of the rewards screen, not sure it's intended.

Since I redeemed I can't see any rewards anymore.

danibonilha:
> After claiming, I got the success alert, but on accepting it, I got redirected out of the rewards screen, not sure it's intended.
> 
> Since I redeemed I can't see any rewards anymore.

It's odd that it went back from rewards, since we don't have an extra go back 🤔 , but the flow is missing parts of the UI which we can use to double check

PR #728, opened 2022-03-24
[CS-3430] Update profile create flow, add validations

topwebtek7:
### Description

- Update create profile screens UI/UX
- Add validataions
- Add color input edit, fix bugs

- [x] Completes #CS-3430

### Checklist

### Screenshots

Mar-24-2022 21-40-37


cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -86,15 +86,18 @@ export const useProfileForm = (params?: useProfileFormParams) => {
...
-
-  // ToDo: update with custom color picker
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [businessColor, setBusinessColor] = useState<string>(businessColorData);
   const [isSubmitPressed, setIsSubmitPressed] = useState(false);
 
   const checkIdUniqueness = useCallback(
     async (id: string) => {
       if (!isLoading && authToken && id) {
+        const validateErrorMessage = validateMerchantId(id);
```
danibonilha, line 95: nice

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -107,7 +110,7 @@ export const useProfileForm = (params?: useProfileFormParams) => {
 
   useEffect(() => {
     if (businessId && authToken && !isLoading) {
-      // check unqiuness when get back to the screen again
+      // check unqiuness when returned to create profile screen
```
danibonilha: typo

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
```diff
@@ -33,6 +42,44 @@ const InputCommonProps = {
   blurOnSubmit: false,
 };
 
+const TextErrorProps = {
```
oidoug, line 45: Isn't it better to have these styled texts in individual components? Along with the Input, all separate and reusable. That way you wouldn't need this prop switching you are doing here.

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
```diff
@@ -33,6 +42,44 @@ const InputCommonProps = {
...
+  color: 'errorLight' as ColorTypes,
+  fontSize: 11,
+  fontWeight: '600' as any,
+  letterSpacing: 0.15,
+  marginTop: 1,
+};
```
danibonilha, line 51: nit: we could type this using the reestyle types, TextProps I guess 🤔

cardstack/src/screens/ProfileScreen/__tests__/useProfileForm.test.ts
```diff
@@ -78,12 +79,10 @@ describe('useProfileForm', () => {
...
 
     expect(result.current.errors?.businessName).toBe(
-      strings.validation.businessNameRequired
+      strings.validation.thisFieldIsRequied
     );
 
-    expect(result.current.errors?.businessId).toBe(
-      strings.validation.businessIdShouldBeUnique
-    );
+    expect(result.current.errors?.businessId).toBe(validateMerchantId(''));
```
danibonilha: we can extract `validaMerchantId('')` to => `const invalidMerchantIdAssertion = validaMerchantId('')` then use this value to assert, bc it's confusing to read we expect error to be validmerchantId

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -136,10 +147,10 @@ export const useProfileForm = (params?: useProfileFormParams) => {
...
       businessName: businessName.trim()
         ? undefined
-        : strings.validation.businessNameRequired,
-      businessId: !isUniqueId
-        ? strings.validation.businessIdShouldBeUnique
-        : validateMerchantId(businessId),
+        : strings.validation.thisFieldIsRequied,
+      businessId:
+        validateMerchantId(businessId) ||
+        (isUniqueId ? undefined : strings.validation.businessIdShouldBeUnique),
```
danibonilha: nit: the ternary could be a const just to give the context that is validMerchant or invalidUniqueId

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
```diff
@@ -41,37 +88,90 @@ export const StepTwo = ({ goToNextStep }: ProgressStepProps) => {
...
+            justifyContent="center"
+            alignItems="center"
+            position="absolute"
+            left={10}
+            top={10}
+          >
+            <Container
+              width={20}
+              height={20}
+              style={{ backgroundColor: validBusinessColor }}
```
danibonilha: nit: this could be memoized

danibonilha (approved):
just some nits

oidoug (approved):

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
topwebtek7: basically I've tried to create variants like Button does but that make Text more complicated.. so I think maybe we can add that when we need to use similar Text in other places.

cardstack/src/screens/ProfileScreen/components/StepTwo.tsx
topwebtek7: Good suggestion.. It's being used in this place only yet and I think we can convert to separated, reusable component if there's other places that uses similar feature.

PR #729, opened 2022-03-24
[CS-3460]: Display rewards claimed history

danibonilha:
### Description
Get reward claimed data from subgraph and display on rewards center, improvements to be made, is to make the row touchable to redirect to blockscout, and I'm thinking about extracting the history logic to the HistoryList itself, as there are no shared data, but we can decide that later when I split the rewards hooks.

### Screenshots

Simulator Screen Recording - iPhone 12 - 2022-03-24 at 18 09 24


oidoug (approved):
👏

PR #731, opened 2022-03-24
[Fix]: Update sheets to work on mobile redesign

danibonilha:
### Description

This PR fixes sheets opening over screen and also the chart/send sheet, I tested, but if you could, it would be really helpful to test across both navigation and devices. (of course the best fix would be to migrate the ones that are using the expanded state, to be it's own sheet, but we can do one by one next).

### Screenshots

image

image

image


oidoug (approved):
Nice!

Shouldn't `LifetimeEarningsExpandedState` and `LiquidityPoolExpandedState` also be converted to Sheet?

danibonilha:
> Nice!
> 
> Shouldn't `LifetimeEarningsExpandedState` and `LiquidityPoolExpandedState` also be converted to Sheet?

Yes, but right now they are hidden, so low priority

PR #732, opened 2022-03-25
[CS-3371]: Add loading state to rewards center

danibonilha:
### Description

I tweak the designs a little to match the components we have on the screen after the redesign, but it's basically the same thing. Also we use `isUninitialized` to set the first loading as default in case we don't have an accountAddress yet and the query is skipped, so we can prevent a registered user to see the unregistered state.

### Checklist

### Screenshots

Simulator Screen Recording - iPhone 12 - 2022-03-25 at 09 59 01


oidoug (approved):
LGTM

PR #735, opened 2022-03-25
[CS-3376] Register Account Transaction Confirmation Screen

oidoug:
### Description

This PR adds a new TransactionConfirmation sheet, `RewardsRegisterSheet`, all its navigation structure, and a new SectionIconTitle component for displaying icon + title in a transaction sheet.

- [x] Completes #(CS-3376)

### Checklist

### Screenshots

Simulator_Screen_Recording_-_iPhone_12_-_2022-03-25_at_11 51 06

### DIFF
--- a/cardstack/src/components/TransactionConfirmationSheet/displays/RewardsRegisterDisplay.tsx
+++ b/cardstack/src/components/TransactionConfirmationSheet/displays/RewardsRegisterDisplay.tsx
@@ -6,6 +6,7 @@ import { PayThisAmountSection } from './components/sections/PayThisAmountSection
 import {
   HorizontalDivider,
   TransactionConfirmationDisplayProps,
+  IconProps,
 } from '@cardstack/components';
 import { RewardsRegisterData } from '@cardstack/types';
 
@@ -14,6 +15,12 @@ interface RewardsRegisterDisplayProps
   data: RewardsRegisterData;
 }
 
+const rewardsIconProps = {
+  name: 'rewards',
+  size: 22,
+  color: 'teal',
+} as IconProps;
+
 export const RewardsRegisterDisplay = ({
   data: { programName, prepaidCard, spendAmount },
 }: RewardsRegisterDisplayProps) => (
@@ -21,7 +28,7 @@ export const RewardsRegisterDisplay = ({
     <SectionIconTitle
       sectionHeaderText={strings.rewards.program.title}
       title={programName}
-      iconProps={{ name: 'rewards', size: 22, color: 'teal' }}
+      iconProps={rewardsIconProps}
     />
     <HorizontalDivider />
     <PrepaidCardTransactionSection
--- a/cardstack/src/components/TransactionConfirmationSheet/displays/RewardsRegisterDisplay.tsx
+++ b/cardstack/src/components/TransactionConfirmationSheet/displays/RewardsRegisterDisplay.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import { strings } from '../strings';
+import { SectionIconTitle } from './components/SectionIconTitle';
+import { PrepaidCardTransactionSection } from './components/sections/PrepaidCardTransactionSection';
+import { PayThisAmountSection } from './components/sections/PayThisAmountSection';
+import {
+  Container,
+  HorizontalDivider,
+  Icon,
+  Text,
+  TransactionConfirmationDisplayProps,
+} from '@cardstack/components';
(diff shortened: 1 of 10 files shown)


cardstack/src/screens/RewardsCenterScreen/useRewardsCenterScreen.ts
```diff
@@ -157,23 +171,23 @@ export const useRewardsCenterScreen = () => {
...
-          },
-        ],
-        title: 'Register Account',
-        message: `Cardstack Rewards\n(${rewardDefaultProgramId[network]})\nPrepaid Card:\n${prepaidCard.address}`,
+      const data: RewardsRegisterData = {
+        type: TransactionConfirmationType.REWARDS_REGISTER,
+        spendAmount: convertToSpend(0.01, 'USD', 1),
+        prepaidCard: prepaidCard.address,
+        programName: 'Cardstack Rewards',
+        network: rewardDefaultProgramId[network],
```
danibonilha: Is network required? If so we need to pass the network prop and not the rewardProgramID, but it seems we don't need network for this case

cardstack/src/components/TransactionConfirmationSheet/displays/RewardsRegisterDisplay.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export const RewardsRegisterDisplay = ({
+  data: { programName, prepaidCard, spendAmount },
+}: RewardsRegisterDisplayProps) => (
+  <>
+    <SectionIconTitle
+      sectionHeaderText={strings.rewards.program.title}
+      title={programName}
+      iconProps={{ name: 'rewards', size: 22, color: 'teal' }}
```
danibonilha: we could extract this to an object outside, just add  ` const iconProps: IconProps = { ...}` to avoid a typing issue

cardstack/src/screens/RewardsCenterScreen/useRewardsCenterScreen.ts
oidoug: ah, forgot it there

cardstack/src/components/TransactionConfirmationSheet/displays/components/SectionIconTitle.tsx
```diff
@@ -0,0 +1,39 @@
...
+  Icon,
+  IconProps,
+  Text,
+} from '@cardstack/components';
+
+interface SectionIconTitleProps {
+  title?: string;
+  iconProps?: IconProps;
+  sectionHeaderText?: string;
+}
```
danibonilha, line 15: nit: does it make sense to all be optional ? if we don't have any of these props why use the component? to fill some space? Just something we can circle back on

cardstack/src/components/TransactionConfirmationSheet/displays/RewardsRegisterDisplay.tsx
```diff
@@ -0,0 +1,43 @@
...
+interface RewardsRegisterDisplayProps
+  extends TransactionConfirmationDisplayProps {
+  data: RewardsRegisterData;
+}
+
+const rewardsIconProps = {
+  name: 'rewards',
+  size: 22,
+  color: 'teal',
+} as IconProps;
```
danibonilha, line 22:
```suggestion
const rewardsIconProps: IconProps = {
  name: 'rewards',
  size: 22,
  color: 'teal',
};
```
When we use `as` we are casting the value, so even if we add a `bleh` variable ts won't complain, when we 'assign'  like my suggestion, if we try to add `bleh` it will throw an error since `bleh` is not on `IconProps`

danibonilha (approved):
Approving to make the release, the new comments can be address on a different PR

cardstack/src/components/TransactionConfirmationSheet/displays/components/SectionIconTitle.tsx
oidoug: Well, I didn't think this thought, but you can't have a title without an icon or the layout gets messed up. I don't think we'll have any case where they aren't all there, so better have them all required.

cardstack/src/components/TransactionConfirmationSheet/displays/RewardsRegisterDisplay.tsx
oidoug: Ah, yeah... thanks for the pointer.

PR #739, opened 2022-03-28
[CS-3441] Fix pay merchant's default currency with account's currency, typo

topwebtek7:
### Description

- Fix pay merchant's default currency to be account's setting currency when not specified(was SPD before)
- Fixed typo
- Handled currencies that sdk does not support
- Updated sdk version

- [x] Completes #CS-3441
- [x] Completes #CS-3462

### Checklist


danibonilha (approved):

oidoug (approved):
LGTM

PR #740, opened 2022-03-28
[CS-3480]: Trigger refetch when reward data updates

danibonilha:
### Description

Manually refetch claim history when data is updated, as claim history is from Apollo and the data is from RTK  we can't automatically invalidate and refetch, also added `refetchOnMountOrArgChange` to try and make the data be up-to-date

### Screenshots

Unfortunately the reward balance is not updated yet, but it's a db issue we will look into 

https://user-images.githubusercontent.com/20520102/160439472-c0c74229-0f92-472c-9d6f-21eb3765ac20.mp4


oidoug (approved):
LGTM!

PR #742, opened 2022-03-28
[CS-3386]: Add default timeout to queryWrapper

danibonilha:
### Description

This PR adds a default timeout of 2minutes for all promises wrapped with queryWrapper, it also adds an option to customize the amount of time

Completes CS-3386
Completes CS-3384

### Screenshots

In order to show the behavior I decrease the timeout to 10ms and displayed the error in the alert: 

Simulator Screen Recording - iPhone 12 - 2022-03-28 at 18 04 21


topwebtek7 (approved):
LGTM

PR #743, opened 2022-03-29
[CS-3336] Update networksection ui, move mainnet into testnet

topwebtek7:
### Description

- Updated "Show testnets" to "Show all networks" and removed Advanced Users
- Moved Ethereum Mainnet to testnet so can be controlled by toggle (`isTestnet` is being used by NetworkToast only and it does not show up in new Navigations, so used `isTestnet` existing flag for toggling, let me know if better thought)

- [x] Completes #CS-3336

### Checklist

### Screenshots

Mar-29-2022 21-21-18

### DIFF
--- a/src/components/settings-menu/NetworkSection.tsx
+++ b/src/components/settings-menu/NetworkSection.tsx
@@ -17,20 +17,20 @@ import {
 } from '@cardstack/components';
 
 const networks = values(networkInfo);
+const defaultNetwork = INITIAL_STATE.network;
 
 const NetworkSection = () => {
   const { network, showTestnets } = useAccountSettings();
   const [selectedNetwork, setSelectedNetwork] = useState(network);
   const dispatch = useDispatch();
   const networkSelected = useMemo(() => networkInfo[network], [network]);
-  const defaultNetwork = INITIAL_STATE.network;
 
   // transform data for sectionList
   const sectionListItems = useMemo(() => {
     const filteredNetworks = networks.filter(item =>
       showTestnets ? true : !item.isTestnet
     );
-    // merge networks by layer and then sort by title
+    // merge networks by layer and then sort by layer title
     const sectionLists = filteredNetworks
       .reduce((result: RadioItemProps[], curr, currentIndex) => {
         result[curr.layer] =
@@ -58,7 +58,7 @@ const NetworkSection = () => {
       );
 
     return sectionLists;
-  }, [defaultNetwork, selectedNetwork, showTestnets]);
+  }, [selectedNetwork, showTestnets]);
 
   useEffect(() => {
     if (networkSelected.isTestnet !== showTestnets) {
--- a/src/components/settings-menu/NetworkSection.tsx
+++ b/src/components/settings-menu/NetworkSection.tsx
@@ -1,5 +1,5 @@
 import { toLower, values } from 'lodash';
-import React, { useCallback, useState } from 'react';
+import React, { memo, useCallback, useEffect, useMemo, useState } from 'react';
 import { useDispatch } from 'react-redux';
 import networkInfo from '../../helpers/networkInfo';
 import { useAccountSettings } from '../../hooks';
@@ -8,13 +8,12 @@ import {
   settingsUpdateNetwork,
   toggleShowTestnets,
 } from '../../redux/settings';
-
 import {
   Button,
   Checkbox,
   Container,
+  RadioItemProps,
   RadioList,
-  Text,
 } from '@cardstack/components';
 
 const networks = values(networkInfo);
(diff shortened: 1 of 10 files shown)


src/helpers/networkInfo.ts
```diff
@@ -29,7 +29,7 @@ export const networkInfo = {
     shortName: 'Mainnet',
     layer: 1,
     value: networkTypes.mainnet,
-    isTestnet: false,
+    isTestnet: true,
```
oidoug: nit: And also rename this flag to something like `hidden`

src/components/settings-menu/NetworkSection.tsx
```diff
@@ -0,0 +1,116 @@
...
+  Container,
+  RadioItemProps,
+  RadioList,
+} from '@cardstack/components';
+
+const networks = values(networkInfo);
+const defaultNetwork = INITIAL_STATE.network;
+
+const NetworkSection = () => {
+  const { network, showTestnets } = useAccountSettings();
```
oidoug: nit: Better reflect the new label, maybe rename it to `showAllNetworks`

src/components/settings-menu/NetworkSection.tsx
```diff
@@ -0,0 +1,116 @@
...
+    [setSelectedNetwork]
+  );
+
+  const updateNetwork = useCallback(() => {
+    dispatch(settingsUpdateNetwork(selectedNetwork));
+  }, [dispatch, selectedNetwork]);
+
+  const onShowTestsPress = useCallback(() => dispatch(toggleShowTestnets()), [
+    dispatch,
+  ]);
```
oidoug: nit: Since you're moving this from js to tsx, maybe it would be nice to have the logic in a custom hook.

oidoug (commented):
Some minor nits.

src/components/settings-menu/NetworkSection.tsx
```diff
@@ -0,0 +1,116 @@
...
+        `${a.title}` < `${b.title}` ? 1 : -1
+      );
+
+    return sectionLists;
+  }, [selectedNetwork, showTestnets]);
+
+  useEffect(() => {
+    if (networkSelected.isTestnet !== showTestnets) {
+      dispatch(toggleShowTestnets());
+    }
```
danibonilha: I feel we can make cleanup this file a little, and it seems this `showTestnets` doesn't really need to live on Redux, it's used just inside this screen, we could track locally, and if a "testnet" is selected we activate the toggle, otherwise its active by the user

danibonilha: I feel we can make cleanup this file a little, and it seems this `showTestnets` doesn't really need to live on Redux, it's used just inside this screen, we could track locally, and if a "testnet" is selected we activate the toggle, otherwise its active by the user

src/components/settings-menu/NetworkSection.tsx
```diff
@@ -0,0 +1,116 @@
...
+                value: curr.value,
+                selected: toLower(selectedNetwork) === toLower(curr.value),
+                default: curr.value === defaultNetwork,
+              },
+            ],
+          } || {};
+
+        return result;
+      }, [])
+      .reduce((acc: RadioItemProps[], val) => acc.concat(val), [])
```
danibonilha: does this reduce flattens the array ?  If you want to use `.flat`, you can just add `es2019`  [lib on tsconfig ](https://www.typescriptlang.org/tsconfig#lib)

danibonilha, line 55: does this reduce flattens the array ?

src/helpers/networkInfo.ts
topwebtek7: basically `isTestnet` is still being used in NetworkToast.. so I think we can update it when agreed getting rid of NetworkToast later.

danibonilha: Actually, since the network toast appears for all networks, it's safe to remove the usage of isTestNet from the `NetworkToast` since it's only purpose is to default the visible state, which in the end doesn't matter, because `setVisible` is called with `isConnected` on mount

src/components/settings-menu/NetworkSection.tsx
topwebtek7: @danibonilha hmm I think you're checking old codes, I already got rid of `showTestnets` before

danibonilha: Github messed the comments, this is an old one, sorry

src/components/settings-menu/NetworkSection/useNetworkSection.ts
```diff
@@ -0,0 +1,98 @@
...
+const networks = values(networkInfo);
+const defaultNetwork = INITIAL_STATE.network;
+
+export const useNetworkSection = () => {
+  const { network } = useAccountSettings();
+  const dispatch = useDispatch();
+  const [selectedNetwork, setSelectedNetwork] = useState(network);
+  const [isShowAllNetworks, setShowAllNetworks] = useState<boolean>(false);
+  const selectedNetworkInfo = useMemo(() => networkInfo[network], [network]);
+
```
danibonilha: instead of network we should use `selectedNetwork` as the key since this is selectedNetworkInfo

src/components/settings-menu/NetworkSection/useNetworkSection.ts
```diff
@@ -0,0 +1,98 @@
...
+      setShowAllNetworks(selectedNetworkInfo.hidden);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedNetworkInfo.hidden]);
+
+  useEffect(() => {
+    if (network) {
+      setSelectedNetwork(network);
+    }
+  }, [network, setSelectedNetwork]);
```
danibonilha: Giving that we are setting the initial state on line 15, I'm not sure if we need this piece here

src/components/settings-menu/NetworkSection/useNetworkSection.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    return sectionLists;
+  }, [selectedNetwork, isShowAllNetworks]);
+
+  useEffect(() => {
+    if (selectedNetworkInfo.hidden !== isShowAllNetworks) {
+      setShowAllNetworks(selectedNetworkInfo.hidden);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedNetworkInfo.hidden]);
```
danibonilha: we could simply this and remove the linter rule, by doing 

```suggestion
  useEffect(() => {
    setShowAllNetworks(selectedNetworkInfo.hidden);
  }, [selectedNetworkInfo.isTestnet]);
```

src/components/settings-menu/NetworkSection/useNetworkSection.ts
```diff
@@ -0,0 +1,98 @@
...
+import { RadioItemProps } from '@cardstack/components';
+
+const networks = values(networkInfo);
+const defaultNetwork = INITIAL_STATE.network;
+
+export const useNetworkSection = () => {
+  const { network } = useAccountSettings();
+  const dispatch = useDispatch();
+  const [selectedNetwork, setSelectedNetwork] = useState(network);
+  const [isShowAllNetworks, setShowAllNetworks] = useState<boolean>(false);
```
danibonilha: nit: suggestion `isShowAllNetworks` => `showAllNetworks`

danibonilha: Also if we default this value to be `selectedNetworkInfo.hidden` we may not need that useEffect to setShowAllNetworks at all

src/components/settings-menu/NetworkSection/useNetworkSection.ts
```diff
@@ -0,0 +1,98 @@
...
+  const selectedNetworkInfo = useMemo(() => networkInfo[network], [network]);
+
+  // transform data for sectionList
+  const sectionListItems = useMemo(() => {
+    const filteredNetworks = networks.filter(item =>
+      isShowAllNetworks ? true : !item.hidden
+    );
```
danibonilha:
```suggestion
    const filteredNetworks = networks.filter(item =>
      isShowAllNetworks  ||  !item.hidden
    );
```

oidoug (approved):
LGTM

PR #744, opened 2022-03-29
[CS-3383]: Add error message on rewards register/claim failure

danibonilha:
### Description

Adds the default error message on rewards register/claim failure

- [x] Completes CS-3383
- [x] Completes CS-3385

### Screenshots

image


topwebtek7 (approved):
LGTM

PR #745, opened 2022-03-29
[FIX] Workaround bug for exception in SDK function handleSignificantDecimals.

oidoug:
### Description

`handleSignificantDecimals` function in SDK excepts with error for parameters `value = 0, decimals = any, buffer = 0`, because it tries to slice an empty value, which return undefined, then undefined is feed into `search` regex causing it to return -1 and then causing an exception in `new BigNumber5`.

A proper fix requires handleSignificantDecimals to handle 0 values and additionally it could be refactored, with the aid of unit tests, to be stripped into smaller more maintainable functions.

### DIFF
--- a/cardstack/src/services/gnosis-service.ts
+++ b/cardstack/src/services/gnosis-service.ts
@@ -269,8 +269,10 @@ export const addNativePriceToToken = async (
 
   const isAmountDust = nativeBalance < 0.01;
 
-  //decimal places formatting for residual crypto values
-  const bufferValue = isAmountDust ? 0 : undefined;
+  // Cecimal places formatting for residual crypto values.
+  // Using "2" is a workaround to a exception happening in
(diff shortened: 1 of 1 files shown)


cardstack/src/services/gnosis-service.ts
```diff
@@ -269,8 +269,10 @@ export const addNativePriceToToken = async (
 
   const isAmountDust = nativeBalance < 0.01;
 
-  //decimal places formatting for residual crypto values
-  const bufferValue = isAmountDust ? 0 : undefined;
+  // Cecimal places formatting for residual crypto values.
```
danibonilha: typo, Decimal

danibonilha (approved):
After fixing the type it's good to go, just a side note that it might show dust amount on claim merchant revenue, but at least the claim button is not enable.

PR #746, opened 2022-03-29
[CS-3489]: Add register gas estimated from sdk 

danibonilha:
### Description

This PR adds the gas estimation for registering, some workarounds were done, since the design expected the gas before the prepaidcard selection, which currently is not possible given that the sdk needs the card address, so we hide the amount on chooseCard and there's a loading now in between the screens to fetch the gas estimate, if there's an error we continue anyway and use the default value to display, if there's a gas estimation we show it. Also wrapped some instances in their own functions to shrink a little.

### Screenshots

Simulator Screen Recording - iPhone 12 - 2022-03-29 at 19 00 29


oidoug (approved):
LGTM

PR #747, opened 2022-03-30
[CS-3495]: Get and display est gas fee for claiming rewards

danibonilha:
### Description

This PR gets and displays the estimate gas fee on claiming rewards.
- Adds loading prop to claim button and fixes uncentered loading
- Caches validProofs to make the claim request quicker and cleans after claim
- Updates ClaimTxConfirmationDisplay to show formatted estimatedValue

PS: For this case we are not using the query from RTK to fetch the gas estimate, given the difficulty to store the parameters based on the each token,I'll be refactoring this piece anyways in order to improve readability given useRewards hook is huge.

### Screenshots

https://user-images.githubusercontent.com/20520102/160880407-6b179741-9448-4454-a238-bd7ce9c7a8c3.mp4

### DIFF
--- a/cardstack/src/services/rewards-center/rewards-center-service.ts
+++ b/cardstack/src/services/rewards-center/rewards-center-service.ts
@@ -1,4 +1,11 @@
-import { getSDK, Safe, TokenInfo } from '@cardstack/cardpay-sdk';
+import {
+  fromWei,
+  getSDK,
+  Proof,
+  Safe,
+  TokenInfo,
+  WithSymbol,
+} from '@cardstack/cardpay-sdk';
 import BN from 'bn.js';
 import {
   addNativePriceToToken,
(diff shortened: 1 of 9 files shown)


cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -118,6 +154,47 @@ export const getRegisterGasEstimate = async ({
...
+  );
+
+  const totalEstimatedGas = estimatedAmounts.reduce(
+    (total, amount) => total.add(amount),
+    new BN(0)
+  );
+
+  const totalEstimatedGasInEth = fromWei(totalEstimatedGas.toString());
+
+  const formattedTotalEstimatedGas = parseFloat(totalEstimatedGasInEth).toFixed(
```
oidoug, line 191: nit: I would prefer to have the display layer deal with clipping the values, but it's more of a preference thing.

oidoug (approved):
🚀

cardstack/src/services/rewards-center/rewards-center-service.ts
danibonilha: yeah, it was a 'hard' decision, for now it's simpler bc we don't need to deal with empty values on the screen, but we can circle back on this while refactoring

PR #748, opened 2022-03-30
[CS-3502]: Disable rewards safe flow

danibonilha:
### Description

This PR disables the send flow and adds refetchOnMount: true, i wanted to add a pollingInterval but it didn't seem to work great, so the safest bet right now is this.

### Screenshots

Simulator Screen Recording - iPhone 12 - 2022-03-30 at 14 10 55


oidoug (approved):
I can't test it right now, but looks fine. 👍

PR #750, opened 2022-03-31
[CS-3491] Update amount to 0 when invalid currency

topwebtek7:
### Description

- [x] Completes #CS-3491

### Checklist

### DIFF
--- a/cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
+++ b/cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
@@ -39,10 +39,13 @@ export const usePaymentMerchantUniversalLink = () => {
 
   const { goBack } = useNavigation();
 
-  const currencyName =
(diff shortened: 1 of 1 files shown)


cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
```diff
@@ -42,6 +42,8 @@ export const usePaymentMerchantUniversalLink = () => {
   const currencyName =
     currency && isSupportedCurrency(currency) ? currency : accountCurrency;
 
+  const validAmount = currency && isSupportedCurrency(currency) ? amount : '0';
+
```
danibonilha: nit: we could make the `currency && isSupportedCurrency(currency)` check just once on an useMemo and use for both cases

danibonilha (approved):

oidoug (approved):
👍

PR #752, opened 2022-03-31
[CS-3528] Update payment received UI

topwebtek7:
### Description

- [x] Completes #CS-3528

### Checklist

### Screenshots

Screen Shot 2022-03-31 at 21 46 50

### DIFF
--- a/cardstack/src/components/PaymentDetails/EarnedTransaction.tsx
+++ b/cardstack/src/components/PaymentDetails/EarnedTransaction.tsx
@@ -29,9 +29,9 @@ export const EarnedTransaction = (data: EarnedTransactionProps) => {
   const MerchantPaymentItemDetails = useMemo(
     () =>
       [
-        { description: strings.payment.toUpperCase(), value: revenueCollected },
+        { description: strings.payment, value: revenueCollected },
         {
-          description: `${strings.protocolFee.toUpperCase()} \n(0.5%)`,
+          description: strings.protocolFee,
           value: protocolFee,
         },
       ].map((item: MerchantPaymentItemDetailProps, index: number) => (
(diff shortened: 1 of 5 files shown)


danibonilha (commented):
Hm Idk about using `.toUpperCase()` on all strings, if we are not using the lowerCase state, what's the benefit of doing that? We could store the strings the way they are shown, so it's easier to search for if we need to

cardstack/src/components/PaymentDetails/EarnedTransaction.tsx
```diff
@@ -28,9 +29,9 @@ export const EarnedTransaction = (data: EarnedTransactionProps) => {
   const MerchantPaymentItemDetails = useMemo(
     () =>
       [
-        { description: 'REVENUE \nCOLLECTED', value: revenueCollected },
+        { description: strings.payment.toUpperCase(), value: revenueCollected },
         {
-          description: 'PROTOCOL FEE \n(0.5%)',
+          description: `${strings.protocolFee.toUpperCase()} \n(0.5%)`,
```
oidoug: Personal preference, but I would drop the `toUpperCase` above and here too, have the text correctly formatted in strings and add the `\n(0.5%)` to strings too.

oidoug (approved):

cardstack/src/components/PaymentDetails/EarnedTransaction.tsx
oidoug: I agree with @danibonilha, especially since all these strings are very context-dependent, we don't reuse them at all. Also, some languages might need lowercase representations to convey information.

topwebtek7:
> Hm Idk about using `.toUpperCase()` on all strings, if we are not using the lowerCase state, what's the benefit of doing that? We could store the strings the way they are shown, so it's easier to search for if we need to

In before projects, I used automatic translation library to translate to other languages so we followed this way to make more accurate translations.. but if it's not what we need to consider at the moment, will follow what you guys suggested

PR #754, opened 2022-03-31
[CS-3520] Adds gas fee InfoBanner to ClaimContent.

oidoug:
### Description

PR adds Claim Fee Notice to Rewards screen. 

The message in the banner is part in bold, so I updated InfoBanner to allow custom Text components to be added and had the message split into 3 parts to handle the bold part. Please let me know if there's a more elegant way of handling this scenario.

- [x] Completes #(CS-3520)

### Checklist

### Screenshots

Screenshot

### DIFF
--- a/cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
+++ b/cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
@@ -14,7 +14,6 @@ import {
   Container,
   useTabHeader,
   InfoBanner,
-  Text,
 } from '@cardstack/components';
 
 interface ClaimContentProps {
--- a/cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
+++ b/cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
@@ -9,7 +9,13 @@ import {
   RewardsHistoryList,
   RewardsHistoryListProps,
 } from '.';
-import { ScrollView, Container, useTabHeader } from '@cardstack/components';
+import {
+  ScrollView,
+  Container,
+  useTabHeader,
+  InfoBanner,
+  Text,
+} from '@cardstack/components';
 
 interface ClaimContentProps {
   claimList?: Array<RewardRowProps>;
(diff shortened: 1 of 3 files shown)


cardstack/src/screens/RewardsCenterScreen/strings.ts
```diff
@@ -17,6 +17,15 @@ export const strings = {
     loading: 'Registering Account',
     payCostDescription: 'To pay transaction cost',
     gasLoading: 'Getting estimated gas fee',
+    gasInfoBanner: {
+      title: 'Claiming Gas Fee',
+      message: {
+        part1:
+          'Claiming is an on chain transaction which required a small blockchain gas fee: ',
```
danibonilha: Typo: `required` -> `requires`

cardstack/src/screens/RewardsCenterScreen/strings.ts
```diff
@@ -17,6 +17,15 @@ export const strings = {
...
+        part2: '~0.25 CARD.CPXD. ',
+        part3: 'This is paid from your claimable reward.',
```
danibonilha: nit: I'd prefer this to be contextual rather than part1, 2,3

cardstack/src/screens/RewardsCenterScreen/strings.ts
```diff
@@ -17,6 +17,15 @@ export const strings = {
     loading: 'Registering Account',
     payCostDescription: 'To pay transaction cost',
     gasLoading: 'Getting estimated gas fee',
+    gasInfoBanner: {
+      title: 'Claiming Gas Fee',
+      message: {
+        part1:
+          'Claiming is an on chain transaction which required a small blockchain gas fee: ',
+        part2: '~0.25 CARD.CPXD. ',
```
danibonilha: @kdowlin  this is like for each proof, every gas estimate I did for more than on proof I got around 0.6 CARD, maybe we should reflect that somehow

oidoug: True... It can get complex if we try to show that info contextually, bc there can be multiple rewards for multiple programs. Maybe, for now, we can just add "~0.25 CARD.CPXD per claim" to the copy. Wdyt?

kdowlin: Let's just remove the cost for now. And just let them know there is a fee paid from their reward. So remove ': 0.25 CARD.CPXD per claim'

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
```diff
@@ -61,6 +67,19 @@ export const ClaimContent = ({
...
         <RewardsTitle title={title} width="100%" paddingBottom={5} />
+        <InfoBanner
+          paddingBottom={5}
+          title={strings.register.gasInfoBanner.title}
+          message={
+            <Text size="xs">
+              {strings.register.gasInfoBanner.message.intro}
+              <Text size="xs" weight="bold">
+                {strings.register.gasInfoBanner.message.gasEstimative}
+              </Text>
```
danibonilha: i believe we should remove this piece, other than that looks good

danibonilha (approved):

oidoug:
End result: 

Screenshot

kdowlin:
Per claim seems good add to me.

PR #755, opened 2022-03-31
[CS-3498]: Add pin and backup missing screens to new nav 

danibonilha:
### Description

This PR adds the missing PIN screen alongside with the backup one, which was adapted to be use by both platforms and both navigations, also it updates some missing send_flow routes and remove unused routes and components. I'm not adding gifs to this one, because it's important that you test it if you could, it will also pass through QA, bc after approved I'll generate a version with the tab as default .


cardstack/src/navigation/tabBarNavigator.tsx
```diff
@@ -154,6 +155,11 @@ const StackNavigator = () => {
         name={RainbowRoutes.RESTORE_SHEET}
         options={expandedPreset as StackNavigationOptions}
       />
+      <Stack.Screen
+        component={PinAuthenticationScreen}
+        name={RainbowRoutes.PIN_AUTHENTICATION_SCREEN}
+        options={{ ...sheetPreset, gestureEnabled: false }}
```
oidoug: `...sheetPreset` does nothing here since it's a function and will not unwrap itself, but you don't need a preset for this screen anyway, since it pops on top of everything, just the `gestureEnabled: false`.

oidoug: I guess the other use of sheetPreset on `ExpandedAssetSheet` may also be doing nothing.

src/components/backup/BackupManualStep.tsx
```diff
@@ -48,6 +41,13 @@ export default function BackupManualStep() {
     : 'These words are the keys to your account! Write them down or save them in your password manager.';
```
oidoug, line 41: nit: Since you refactoring this, maybe you could extract this file's strings too.

src/navigation/routesNames.js
```diff
@@ -67,10 +53,6 @@ export const NATIVE_ROUTES = [
 const RoutesWithNativeStackAvailability = {
   ...Routes,
   ADD_CASH_FLOW: Routes.ADD_CASH_SHEET,
-  SEND_FLOW: Device.isAndroid ? Routes.SEND_SHEET_NAVIGATOR : Routes.SEND_SHEET,
-  SPEND_FLOW: Device.isAndroid
-    ? Routes.SPEND_SHEET_NAVIGATOR
-    : Routes.SPEND_SHEET,
 };
```
oidoug, line 55: Question: for this PR, do we want to keep the old navigation working?

oidoug (commented):
Found a few issues related to PIN mode. I understand they might not be related to these PR changes, but wanted to report anyway, I can create new tickets if you prefer so.

There are also some comments.

1. On a device without secure-auth, loading a wallet (from google bkp) didn't finish restoring, so loading was never completed and Pin could not be entered. The same occurs when restoring a backup from settings. On refresh, it worked.

Screenshot

2. I couldn't make a payment using PIN, it errors out with the alert to contact support, in logs I see:
```
 ERROR  Unable to getHdSignedProvider [Error: Could not create addresses from your mnemonic or private key(s). Please check that your inputs are correct.]
 LOG  Interaction finished automatically
 LOG  Error while paying merchant [Error: Provider not set or invalid]
 LOG  Pay Merchant failed! {"data": [Error: Provider not set or invalid], "status": 418}
```

3. On "Secret Recovery Phrase", after tapping "Show Secret Recovery Phrase" button and entering the PIN, nothing changes, button is still visible.

src/navigation/routesNames.js
danibonilha: Ideally yes, but not a requirement, those changes shouldn't affect the old nav

src/components/backup/BackupManualStep.tsx
danibonilha: The goal for this PR is to get a version as soon as possible, so I decided to skip some improvements in favor of time

cardstack/src/navigation/tabBarNavigator.tsx
danibonilha: good catch, I just extracted this from `Routes.android` and didn't notice it

danibonilha: when it's not destructed though it works, bc it receives the navigation props, but you're right about not needing in the PIN screen

danibonilha: But we definitely can circle back on this to check which screen needs each preset eventually

danibonilha:
> Found a few issues related to PIN mode. I understand they might not be related to these PR changes, but wanted to report anyway, I can create new tickets if you prefer so.
> 
> There are also some comments.
> 
> 1. On a device without secure-auth, loading a wallet (from google bkp) didn't finish restoring, so loading was never completed and Pin could not be entered. The same occurs when restoring a backup from settings. On refresh, it worked.
> 
> 
> 2. I couldn't make a payment using PIN, it errors out with the alert to contact support, in logs I see:
> 
> ```
> 3. On "Secret Recovery Phrase", after tapping "Show Secret Recovery Phrase" button and entering the PIN, nothing changes, button is still visible.

1. This is a known issue `CS-3332` you can find more info in the ticket, and it's not related to this PR.
2. I made payments successfully, giving the logs, it was probably some punctual error, if the phone was left idle etc 
3. I'm able to see the recovery phrase. 

So I'd suggest to test on a diff device or we need to find more repro steps

oidoug (approved):
I'll report found issues in new tickets. This PR is good to go.

PR #756, opened 2022-03-31
[CS-3533]: Add change account on my wallet address

danibonilha:
### Description

Adds ability to change accounts on new navigation by clicking the preview address, we don't show feedback nor make explicit, bc it's a feature we want to discontinue.

### Screenshots

Simulator Screen Recording - iPhone SE (2nd generation) - 2022-03-31 at 17 44 33


oidoug (approved):
I was missing this, thanks!

PR #758, opened 2022-04-01
[Bug]: Fix expanded asset not opening on iOS

danibonilha:
### Description

By fixing the sheets on the new nav we broke stuff on the current one, so this is a 'hotfix'

### Screenshots

Simulator Screen Recording - iPhone SE (2nd generation) - 2022-04-01 at 11 20 38


oidoug (approved):
LGTM

PR #759, opened 2022-04-01
[CS-3526] Add profile id length validation check

topwebtek7:
### Description

Add profileId length validation check.

- [CS-3526] Completes #(Linear ticket)

### Checklist

- [CS-3526] All UI changes have been tested on a small device

### Screenshots

Screen Shot 2022-03-31 at 21 12 37

### DIFF
--- a/cardstack/src/screens/ProfileScreen/useProfileForm.ts
+++ b/cardstack/src/screens/ProfileScreen/useProfileForm.ts
@@ -93,7 +93,7 @@ export const useProfileForm = (params?: useProfileFormParams) => {
   const [businessColor, setBusinessColor] = useState<string>(businessColorData);
   const [isSubmitPressed, setIsSubmitPressed] = useState(false);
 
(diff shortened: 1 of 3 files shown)


cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -93,10 +93,20 @@ export const useProfileForm = (params?: useProfileFormParams) => {
   const [businessColor, setBusinessColor] = useState<string>(businessColorData);
   const [isSubmitPressed, setIsSubmitPressed] = useState(false);
 
+  const validateProfileId = useCallback((id: string) => {
```
oidoug: nit: For clarity, this function could be renamed to be clear it's the local part of the merchant id validation, since the SDK part is called validateMerchatId, it sounds like different things but it isn't.

oidoug (approved):
Minor nit, but looks good.

PR #760, opened 2022-04-01
[CS-3534]: Scaffolds  rewards withdraw flow and enables it 

danibonilha:
### Description

This PR adds the rewards withdraw flow structure to enable UI and logic to be build in parallel, also adds a developer page, called design system where we can see all the button variants. For  the placeholders I didn't extract anything into strings files, cause it's going to be removed, any UI stuff is available to change so didn't worry to much about that. For this new flow I created a new folder structure called flows, inside the rewards center, we may thing about moving the claiming and registering pieces there too eventually.

- Completes CS-3534
- Completes CS-3543
- Completes CS-3544

### Screenshots

Simulator Screen Recording - iPhone 12 - 2022-04-01 at 14 26 36

Not pretty but it helps 😅 
image

### DIFF
--- a/cardstack/src/screens/Dev/DesignSystemScreen.tsx
+++ b/cardstack/src/screens/Dev/DesignSystemScreen.tsx
@@ -1,0 +1,46 @@
+import React, { memo, useCallback } from 'react';
+import { SectionList } from 'react-native';
+import {
+  Button,
+  CenteredContainer,
+  Container,
+  Text,
+} from '@cardstack/components';
+import { buttonVariants } from '@cardstack/theme';
+
+// Not worrying about perfomance and typing as this is a developer feature
+const DesignSystemScreen = () => {
+  const sections = [
+    {
+      title: 'Buttons',
+      data: Object.keys(buttonVariants),
+    },
+  ];
+
+  const renderItem = useCallback(({ item, section: { title } }) => {
+    switch (title) {
+      case 'Buttons':
+        return (
+          <CenteredContainer padding={2} backgroundColor="overlayGray">
+            <Button variant={item}>{item}</Button>
+          </CenteredContainer>
+        );
+    }
+  }, []);
+
+  return (
+    <SectionList
+      renderItem={renderItem as any}
+      sections={sections}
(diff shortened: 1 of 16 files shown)


cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/useRewardWithdrawToScreen.ts
```diff
@@ -0,0 +1,47 @@
...
+        ].filter(Boolean),
+      }),
+    }
+  );
+
+  const onSafePress = useCallback(
+    safe => () => {
+      navigate(MainRoutes.REWARD_WITHDRAW_CONFIRMATION, {
+        tokenInfo,
+        fromRewardSafe: tokenInfo.safeAddress,
```
danibonilha, line 36: I'm passing tokenInfo and then fromRewardSafe which is within TokenInfo, for readability purposes, it's easier to understand that it's a from a reward safe, just to avoid mixing the multiple types of safeAddresses

cardstack/src/screens/Dev/DesignSystemScreen.tsx
```diff
@@ -0,0 +1,46 @@
...
+          <Text color="white" fontSize={20}>
+            {item.section.title}
+          </Text>
+        </Container>
+      )}
+    />
+  );
+};
+
+export default memo(DesignSystemScreen);
```
oidoug, line 46: I don't see us adding all of our base components to this screen, so maybe it's useful for specific tests, but not for a complete design system visualization.

oidoug (approved):
Looks good, thanks for the scaffolding!

cardstack/src/screens/Dev/DesignSystemScreen.tsx
danibonilha: It's also some kind of `scaffolding` we could also add icons, into inner sections or screens, I always find difficult to get the right variant or icon without visual

PR #761, opened 2022-04-04
[CS-3538] Adds UI for Safe Selection Item and List.

oidoug:
### Description

This PR adds a List and Item components to handle displaying and selecting Safes. Part 1 / 2 for withdrawing UI.

- [x] Completes #(CS-3538)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/161533754-473cdd0d-55a6-4eda-a3c9-d63398ccd7e8.mp4

### DIFF
--- a/cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionItem.tsx
+++ b/cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionItem.tsx
@@ -29,7 +29,7 @@ export const SafeSelectionItem = memo(
     );
 
     return (
-      <CardPressable onPress={onSafePress(safe)} paddingVertical={8}>
+      <CardPressable onPress={onSafePress(safe)}>
         <Container flexDirection="row">
           <Container justifyContent="center" alignContent="center">
             <ContactAvatar {...avatarProps} />
@@ -38,10 +38,10 @@ export const SafeSelectionItem = memo(
             <Text variant="body" weight="bold">
               {safe.merchantInfo?.name || safe.type}
             </Text>
-            <Text size="xs" color="darkGrayText">
+            <Text size="xs" color="secondaryText">
               {typeText}
             </Text>
-            <Text size="xs" color="darkGrayText">
+            <Text size="xs" color="tertiaryText">
               {getAddressPreview(safe.address)}
             </Text>
           </Container>
--- a/cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionItem.tsx
+++ b/cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionItem.tsx
@@ -38,7 +38,7 @@ export const SafeSelectionItem = memo(
             <Text variant="body" weight="bold">
               {safe.merchantInfo?.name || safe.type}
             </Text>
-            <Text size="xs" color="secondaryText">
+            <Text size="xs" weight="semibold" color="secondaryText">
               {typeText}
             </Text>
             <Text size="xs" color="tertiaryText">
--- a/cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionItem.tsx
+++ b/cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionItem.tsx
@@ -1,0 +1,62 @@
+import React, { memo, useMemo } from 'react';
+import { StyleSheet } from 'react-native';
+import { strings } from './strings';
+import { getAddressPreview } from '@cardstack/utils';
+import { CardPressable, Container, Text, Icon } from '@cardstack/components';
+import { ContactAvatar } from '@rainbow-me/components/contacts';
+import { MerchantSafeType } from '@cardstack/types';
+import { palette } from '@cardstack/theme';
+
+interface SafeSelectionItemProps {
+  safe: MerchantSafeType;
+  primary: boolean;
+  onSafePress: (safe: MerchantSafeType) => () => void;
+}
+
+export const SafeSelectionItem = memo(
+  ({ safe, primary, onSafePress }: SafeSelectionItemProps) => {
+    const typeText = primary ? strings.primary : strings[safe.type];
+
+    const avatarProps = useMemo(
+      () => ({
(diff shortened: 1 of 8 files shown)


cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionList.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const renderDivisor = useCallback(
+    () => <Container backgroundColor="borderGray" height={1} />,
+    []
+  );
+
+  const renderFooter = useCallback(
+    () => (
+      <Container>
+        {renderDivisor()}
```
danibonilha: nit: since for this case we always want a item separator, instead of this piece we could add the separator on `renderItem`  and not use `ItemSeparatorComponent`

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionList.tsx
```diff
@@ -0,0 +1,66 @@
...
+      </Container>
+    ),
+    [renderDivisor]
+  );
+
+  return (
+    <FlatList
+      contentContainerStyle={styles.content}
+      data={safes}
+      keyExtractor={item => item.address}
```
danibonilha, line 51: nit: this hardly will be a huge list  but wrapping in a useCallback would improve rendering

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/strings.ts
```diff
@@ -0,0 +1,10 @@
+export const strings: { [key: string]: any } = {
+  merchant: 'Profile',
+  depot: 'Depot',
+  primary: 'Primary profile',
+  note: {
+    title: 'Note',
+    message:
+      "You can only withdraw to a profile on your device. If you don't have one, please go to Profile screen and create yours.",
```
danibonilha, line 8: nit: maybe profile tab would make more sense ? @kdowlin

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionItem.tsx
```diff
@@ -0,0 +1,62 @@
...
+        value: safe.merchantInfo?.name || safe.type,
+        textColor: safe.merchantInfo?.textColor || 'white',
+        size: 'large',
+        style: styles.avatar,
+      }),
+      [safe]
+    );
+
+    return (
+      <CardPressable onPress={onSafePress(safe)} paddingVertical={8}>
```
danibonilha: onPress, we probably need to pass the safe and `avatarProps` so we can render it on next screen, but it's ok to handle on this on the next PR

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionItem.tsx
```diff
@@ -0,0 +1,62 @@
...
+          <Container flex={1} paddingHorizontal={3}>
+            <Text variant="body" weight="bold">
+              {safe.merchantInfo?.name || safe.type}
+            </Text>
+            <Text size="xs" color="darkGrayText">
+              {typeText}
+            </Text>
+            <Text size="xs" color="darkGrayText">
+              {getAddressPreview(safe.address)}
+            </Text>
```
danibonilha: These colors don't seem right,  I think it's `blueText` on safeAddress, and probably a darker color on type

danibonilha (approved):
It's looking good, most of the comments are nit besides the type/safe info text colors

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionItem.tsx
oidoug: The colors from the design aren't matching well app-defined ones. Another argument for an improved design system to be implemented. Anyway, I'll add the new colors.

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionList.tsx
oidoug: Ok, I was wondering if there will be cases where we don't want to show the info banner and the last divider wouldn't be needed. But I'm fine either way, I'll add the line to the item for now.

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionList.tsx
oidoug: Simple value checks are usually cheaper than the added `useCallback` overhead. But I don't know when that line is crossed 🤷

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/components/SafeSelectionItem.tsx
oidoug: SafeSelectionItem is rendered straight from a `MerchantSafeType`, and it's the same UI on both screens, so if we re-use it only the safe will be needed.

danibonilha: oh, you're right I didn't notice that.

cardstack/src/theme/colors.ts
```diff
@@ -24,6 +24,8 @@ export const palette = {
   grayBackgroundLight: '#DEDEDE',
   borderGray: '#E8E8E8',
   darkGray: '#423e4f',
+  slateGray: '#5a586a',
+  battleshipGrey: '#6b6a80',
```
danibonilha: there names are just randomly created by zeplin  I believe and we can totally  give another meaning to it, `battleshipGrey: #6b6a80` already exists under the `blueText` name

oidoug: Ah, didn't know that was auto-generated and I miss it was there already. `slateGray` is new though, so I'm keeping it.

PR #762, opened 2022-04-04
[CS-3410] Fix IssuePrepaidCardDisplay

topwebtek7:
### Description

Fixed IssuePrepaidCardDisplay so it supports issuing prepaidcard from merchant and shows merchant balance.

- [x] Completes #CS-3410

### Checklist

### Screenshots

https://user-images.githubusercontent.com/16714648/161556672-ef19c010-93d8-420c-9f53-d0431b2fc46b.mp4

Got WC push notification when safes are not fully loaded -

https://user-images.githubusercontent.com/16714648/161556686-46c87112-8552-47f2-b552-aae72492a7ac.mp4

### DIFF
--- a/cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
+++ b/cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
@@ -78,7 +78,7 @@ const FromSection = ({
   );
 
   const safeTypeText =
-    fromSafe?.type === 'depot' ? 'DEPOT' : fromSafe?.merchantInfo?.name;
+    fromSafe?.merchantInfo?.name || fromSafe?.type.toUpperCase();
 
   const token = getSafeTokenByAddress(fromSafe, tokenAddress);
   const tokenBalance = token ? token.balance.display : 'Insufficient Funds';
--- a/cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
+++ b/cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
@@ -1,5 +1,6 @@
 import { convertRawAmountToBalance } from '@cardstack/cardpay-sdk';
 import React from 'react';
+import { ActivityIndicator } from 'react-native';
 import { SectionHeaderText } from '../components/SectionHeaderText';
 import { ContactAvatar } from '@rainbow-me/components/contacts';
 import {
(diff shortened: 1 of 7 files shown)


cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
```diff
@@ -66,21 +97,25 @@ const FromSection = ({ tokenAddress }: { tokenAddress: string }) => {
           <Container marginLeft={4}>
             <Text weight="extraBold">{accountName}</Text>
             <NetworkBadge marginTop={2} />
-            {depot && (
+            {fromSafe ? (
```
danibonilha: Shouldn't we use the loading property here? If by any change there's an error while loading the safes, we really want to keep on a infinite loading and not enable the user to issue the card without the safe info ?

cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
```diff
@@ -43,14 +48,40 @@ export const IssuePrepaidCardDisplay = (
...
 
-  const token = getDepotTokenByAddress(depot, tokenAddress);
+  const { fromSafes } = useGetSafesDataQuery(
+    { address: accountAddress, nativeCurrency },
+    {
+      selectFromResult: ({ data }) => ({
+        fromSafes: [
+          ...(data?.merchantSafes || []),
+          ...(data?.depots || []),
+        ].filter(safe => safe && safe.address === safeAddress),
```
danibonilha: nit: instead of filter, we could use `find` which would return the value of the safe, so we don't need to create `const fromSafe = fromSafes?.[0];`.

cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
```diff
@@ -43,14 +48,40 @@ export const IssuePrepaidCardDisplay = (
...
+      }),
+    }
+  );
 
+  const fromSafe = fromSafes?.[0];
+
+  const safeTypeText =
+    fromSafe?.type === 'depot' ? 'DEPOT' : fromSafe?.merchantInfo?.name;
+
+  const token = getDepotTokenByAddress(fromSafe, tokenAddress);
```
danibonilha: since this is not used just by depots anymore we could rename it to getSafeTokenByAddress, and also replace the depot references inside the function, to safe

danibonilha (commented):
My concern is just about the loading of the safes, we should probably agree on a fallback solution to show just the from safe address in case we are not able to load the safes and use the isLoading flag instead of the safesData to show the loading

cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
```diff
@@ -66,21 +97,25 @@ const FromSection = ({ tokenAddress }: { tokenAddress: string }) => {
...
                   </Container>
                   <Text fontSize={15} weight="extraBold">
                     {tokenBalance}
                   </Text>
                 </Container>
               </Container>
+            ) : (
+              <Container marginTop={2}>
+                <ActivityIndicator size="small" />
+              </Container>
```
oidoug: nit: I think this loader would look better if left-aligned along with the text.

cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
```diff
@@ -43,14 +48,39 @@ export const IssuePrepaidCardDisplay = (
...
+          ...(data?.merchantSafes || []),
+          ...(data?.depots || []),
+        ].find(safe => safe && safe.address === safeAddress),
+        isLoadingSafe: isLoading || isUninitialized,
+      }),
+    }
+  );
+
+  const safeTypeText =
+    fromSafe?.type === 'depot' ? 'DEPOT' : fromSafe?.merchantInfo?.name;
```
danibonilha: nit: 
```suggestion
  const safeTypeText = fromSafe?.merchantInfo?.name || fromSafe?.type.toUpperCase();
```

danibonilha (approved):
looking good, minor nit

PR #764, opened 2022-04-04
[CS-3545]: add ability to withdraw from reward safe

danibonilha:
### Description

This PR adds the 'logic' ability to withdraw all funds from a reward safe.
- Only show withdraw button on balances greater than 0 
- Adds types on SelectionList, so the parent uses the derived types from the child, in case we need to change
- Fixes dust amount, we don't actually need buffer if we default to zero when the amount is dust
- Remove old unused functions of gnosis services.

-----------------------------------------------------
Update

- Add skeleton loading to amountSection 
- Fix minus gas display on claim amount 
- Add icon to confirm button and disable it if there's no gas amount 
- Fix padding on from section
- handle safeType to be based on the strings and not the type itself so it won't show `merchant` for example instead of profile

Regarding the footer buttons, I think we should extract to it's own component to reuse across the screens, but it's okay for now to get this shipped.

### Screenshots

Success: 
Simulator Screen Recording - iPhone 12 - 2022-04-04 at 20 34 11

Error: 
Simulator Screen Recording - iPhone 12 - 2022-04-04 at 16 36 34


cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawConfimation/useRewardWithdrawConfimationScreen.ts
```diff
@@ -14,20 +25,115 @@ interface NavParams {
...
+    () => ({
+      from: fromRewardSafe,
+      to: withdrawTo.address,
+      tokenAddress,
+      amount: balance?.wei || '0',
+    }),
+    [balance, fromRewardSafe, tokenAddress, withdrawTo]
+  );
+
+  // TODO: retry state for gas estimate failure
```
oidoug, line 49: nit: Isn't better to open a ticket?

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/components/SafeSelectionItem.tsx
```diff
@@ -13,10 +14,17 @@ import { ContactAvatar } from '@rainbow-me/components/contacts';
...
 import { palette } from '@cardstack/theme';
 
-interface SafeSelectionItemProps {
-  safe: MerchantSafeType;
+type OptionalUnion<T1, T2> = {
+  [P in keyof Omit<T1 & T2, keyof (T1 | T2)>]?: (T1 & T2)[P];
+} &
+  (T1 | T2);
+
+type MerchantOrDepotSafe = OptionalUnion<MerchantSafeType, DepotSafe>;
```
oidoug, line 22: Wont a simple `type MerchantOrDepotSafe = MerchantSafeType & DepotSafe;` work for this case? I tried and it doesn't seem we get any typing warning.

oidoug (approved):
Looks good, just some questions.

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/components/SafeSelectionItem.tsx
danibonilha: not really, this type is created to make all the non common properties of those types Optional.

With an intersection we would just merge all the fields, for example `accumulatedSpendValue` which only exists on merchantSafe would be a required field while passing a depotSafe

`type MerchantOrDepotSafe = MerchantSafeType & DepotSafe`

image

With OptionalUnion we made all different fields optional so we can handle properly

`type MerchantOrDepotSafe = OptionalUnion<MerchantSafeType, DepotSafe>`
image

PR #765, opened 2022-04-05
[CS-3552] Add refetch ability in merchant content, update usePrimarySafe

topwebtek7:
### Description

- Added refetch ability in merchant screen with pull down action
- Updated usePrimarySafe so it updates primarySafe reducer with updated primary safe

- [x] Completes #CS-3552

### Checklist

### Screenshots

https://user-images.githubusercontent.com/16714648/161764729-2f35109c-0598-458d-87a5-8479f2ff6db0.mp4

### DIFF
--- a/cardstack/src/redux/hooks/usePrimarySafe.ts
+++ b/cardstack/src/redux/hooks/usePrimarySafe.ts
@@ -1,6 +1,5 @@
 import { useCallback, useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
-import { useFocusEffect } from '@react-navigation/native';
 import {
   changePrimarySafe as setPrimarySafeAccount,
   selectPrimarySafe,
@@ -63,12 +62,6 @@ export const usePrimarySafe = () => {
       }
     }
   }, [changePrimarySafe, primarySafe, merchantSafes, isFetching]);
-
-  const onRefresh = useCallback(() => {
-    refetch && refetch();
-  }, [refetch]);
-
-  useFocusEffect(onRefresh);
 
   return {
     error,
--- a/cardstack/src/redux/hooks/usePrimarySafe.ts
+++ b/cardstack/src/redux/hooks/usePrimarySafe.ts
@@ -20,7 +20,12 @@ export const usePrimarySafe = () => {
 
   const primarySafe = useSelector(selectPrimarySafe(network, accountAddress));
 
-  const { data = safesInitialState, error, isFetching } = useGetSafesDataQuery({
+  const {
+    data = safesInitialState,
+    error,
+    isFetching,
+    refetch,
+  } = useGetSafesDataQuery({
     address: accountAddress,
     nativeCurrency,
   });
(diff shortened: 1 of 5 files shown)


cardstack/src/screens/MerchantScreen/MerchantScreen.tsx
```diff
@@ -7,6 +7,7 @@ import { Container, MerchantContent } from '@cardstack/components';
 const MerchantScreen = () => {
```
oidoug, line 7: HomeScreen updates when on focus, I think we should do the same here for similar reasons:
```
const onRefresh = useCallback(() => {
  refetch?.();
}, [refetch]);

useFocusEffect(onRefresh);
```

oidoug (commented):
I suggest we refresh the safe also on Merchant screen focus.

cardstack/src/screens/MerchantScreen/MerchantScreen.tsx
topwebtek7: @oidoug hm I think we don't need to refetch in MerchantScreen as we refetched already in Wallet screen before open Merchant Screen. 
I just wonder if need to refetch when focused on profile screen.. or just pulling down to refetch would be enough for now?
I think we can test with current approach and convinced refetch's needed when focused on Profile, then can add this on ProfileScreen as it's easy fix.

oidoug: You're right, autorefreshing the merchant screen might be too much, I meant that for the Profile.

cardstack/src/redux/hooks/usePrimarySafe.ts
```diff
@@ -20,7 +20,12 @@ export const usePrimarySafe = () => {
...
-  const { data = safesInitialState, error, isFetching } = useGetSafesDataQuery({
+  const {
+    data = safesInitialState,
+    error,
+    isFetching,
+    refetch,
+  } = useGetSafesDataQuery({
     address: accountAddress,
     nativeCurrency,
   });
```
danibonilha: It's not related to this PR, but I think we should probably add the skip options on this, since this is used on multiple  tabs and we might not have those info yet, this will probably fix the sentry error, `Don't know about the constant 'subgraphURL' for network mainnet`
`    {
      skip: isLayer1(network) || !accountAddress || !walletReady,
    }`

topwebtek7: ah ok, makes sense. will add skip here and in useMerchantScreen as well

cardstack/src/redux/hooks/usePrimarySafe.ts
```diff
@@ -37,14 +49,31 @@ export const usePrimarySafe = () => {
...
+      }
     }
   }, [changePrimarySafe, primarySafe, merchantSafes, isFetching]);
 
+  const onRefresh = useCallback(() => {
+    refetch && refetch();
+  }, [refetch]);
+
+  useFocusEffect(onRefresh);
+
```
danibonilha: Not sure we need this here 🤔,  we use this hooks on multiple places and it might trigger lots of unnecessary requests, either way the right way to update the data would be to use the refecthOnArgChange prop probably, so  let's circle back on the auto refresh of later, leave this PR with only the manual PullToRefresh

topwebtek7: oh sorry it was meant to be inside ProfileScreen, just update it

oidoug (approved):
LGTM

oidoug:
I noticed that sometimes when switching it doesn't show the refresh indicator, but I realized it's actually a nice side effect of using RTQ Query for the safes, bc it uses cached info if it's still valid 👏

danibonilha:
We can merge as it is, but we should rethink the design of auto refreshing, I feel it's a bit odd to have the same component as the pull to refresh one, maybe some skeletons would be better, but tbd

PR #766, opened 2022-04-05
[Fix]: Invalidate safes on reward withdraw to refresh

danibonilha:
### Description
Since we send to a safe we need to invalidate the tag to have the safes values up-to-date


oidoug (commented):
👍

oidoug (approved):

PR #767, opened 2022-04-05
[FIX] Restart call on DeveloperSettings.

oidoug:
### Description

`RNRestart` Restart function wasn't being properly called.

### DIFF
--- a/src/components/settings-menu/DeveloperSettings.js
+++ b/src/components/settings-menu/DeveloperSettings.js
@@ -1,7 +1,7 @@
 import AsyncStorage from '@react-native-async-storage/async-storage';
 import React, { useCallback, useContext } from 'react';
 import { Alert, ScrollView } from 'react-native';
-import { Restart } from 'react-native-restart';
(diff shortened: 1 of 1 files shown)


src/components/settings-menu/DeveloperSettings.js
```diff
@@ -63,7 +63,7 @@ const DeveloperSettings = () => {
         onPress={wipeKeychain}
         testID="reset-keychain-section"
       />
-      <ListItem label="🔄 Restart app" onPress={Restart} />
+      <ListItem label="🔄 Restart app" onPress={() => RNRestart.Restart()} />
```
danibonilha, line 66: superNit: `onPress={RNRestart.Restart}`

danibonilha (approved):

src/components/settings-menu/DeveloperSettings.js
oidoug: I tried, but Restart can't accept any params, and onPress passes the event on.

PR #768, opened 2022-04-05
[CS-3332]: Fix android restore PIN and cleanups

danibonilha:
### Description
This PR replaces the Portal loading on restore step with loadingOverlay, this way the PIN screen is shown above the loading, also removes some unused functions and workarounds for the SlackSheet, since we replaced with Sheets already. It prevents the PIN key to be saved on cloud backup, as this might be pretty confusing. The only place now that uses the Portal loading is the backup flow which hopefully we can remove from there too.
Also it moves the tokenRefreshListener of firebase, to be called just when there's a wallet since it needs to authenticate to the hub

### Screenshots

The .gif got cut off in the end, but you can see that the PIN screen is opening above the loading
restore-PIN

### DIFF
--- a/cardstack/src/navigation/presetOptions.ts
+++ b/cardstack/src/navigation/presetOptions.ts
@@ -49,8 +49,16 @@ export const overlayPreset: StackNavigationOptions = {
   cardStyleInterpolator: forFade,
 };
 
+/**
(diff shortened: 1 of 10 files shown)


oidoug (approved):
Tested and working on my pin only device 👍

PR #770, opened 2022-04-06
[CS-3583][CS-3582] Removes Discord Banner, Adds Rewards to WalletScreen

oidoug:
### Description

This PR only removes the visualization of Discord Banners on old and new screens, but it does not remove the banner component. 

Kiel mentioned it can be repurposed soon for another type of announcement.

- [x] Completes #(CS-3583)
- [x] Completes #(CS-3582)

### Checklist

### Screenshots

Screenshot 

Screenshot


oidoug:
@danibonilha , do you prefer we remove all Discord banner's code and assets?

danibonilha (approved):
In regards of removing the discordBanner component, if the chance of reusing is really small we should delete it, it's a hardcoded png, so we will not be able to repurpose it anyways, also let's not forget to add a ticket to update the rewardsBanner design once new nav is launched

oidoug:
Agreed, will remove it, we can re-add it if we find it necessary again.

About rewards, I think there will be some more substantial changes once we have multiple programs. What change do we need after the tab bar is released?

PR #772, opened 2022-04-06
[CS-2562]: Remove PrepaidCard non/reloadable references

danibonilha:
### Description

Removes text and props about reloadability of prepaidCards, I made it optional on  `PrepaidCardType` , bc the sdk still delivers this flag and we might need to use, so I just removed the UI related stuff.

### Screenshots

image


oidoug (approved):
LGTM

PR #773, opened 2022-04-06
[CS-3549]: Redirect on wc qr code success and handle other cases

danibonilha:
### Description
This PR redirects to homeScreen on wallet connect connect success, but for other cases as reject, we were having the same issue of infinite loading, so I added a flag to decide if we read or not, and also a timeout to delay the scan a bit when one code has been scanned already, also it decreases the invalid code timeout, as 30 seconds was too much time and wc response is usually really fast.
A bit unrelated but I added the missing spendAmount field to pay merchant decode data in a attempt to fix the issue we saw this morning.

### Screenshots

scan-wc-code

### DIFF
--- a/cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/useScanner.ts
+++ b/cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/useScanner.ts
@@ -49,10 +49,16 @@ export const useScanner = ({
 
   const walletConnectOnSessionRequestCallback = useCallback(
     (type: WCRedirectTypes) => {
-      if (type === WCRedirectTypes.qrcodeInvalid) {
-        navigate(Routes.WALLET_CONNECT_REDIRECT_SHEET, {
-          type,
-        });
+      switch (type) {
+        case WCRedirectTypes.qrcodeInvalid:
+          navigate(Routes.WALLET_CONNECT_REDIRECT_SHEET, {
+            type,
+          });
+
+          break;
+        case WCRedirectTypes.connect:
+          navigate(Routes.HOME_SCREEN);
+          break;
       }
     },
     [navigate]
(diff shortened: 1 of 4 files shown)


cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/useScanner.ts
```diff
@@ -49,13 +53,26 @@ export const useScanner = ({
...
+          break;
+        case WCRedirectTypes.connect:
+          navigate(Routes.HOME_SCREEN);
+          break;
       }
+
+      // WC QrCode becames invalid once its used,
+      // so this waits a bit b4 enabling the scan again
+      startTimeout(() => {
+        isWcScanEnabled.current = true;
```
oidoug, line 71: nit: So, once you scan a WC code, you want to wait to scan another WC code again, but you don't other types of codes? Is that it?

If you can wait for WC_SCAN_TIMEOUT for any type of scan/code, couldn't you use disableScanning and enableScanning?

oidoug (approved):
Just have a question about the disabled state, but regardless, it looks good.

cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/useScanner.ts
danibonilha: good question, this way I can give more context, the truth is that Scanning is always enabled, but we choose when to take an action, based on the flag, we use the `isScanningEnabled` flag to show the loading, so while scanning is not enabled but you are at the scan page and has scanned something, it means we are processing the scanned data (loading), even though most of the action are async and there's no way to await, the actions usually navigates to some other flow, which disables the scan, and o focus is enabled again. The problem with the Wallet connect one, is that is not a flow, is just a sheet and as soon as you hit connect, chances are you are still pointing to the same QR code you just read, and once it's read, it's invalid, which doesn't happen with the other types of QR codes. if you read a pay merchant one, close the sheet and read it again it will work right way, so there's no need to delay those, just the wc one. That being said having this separated flag prevents us from delaying other codes and it's not considering `loading` as the `isScanningEnabled` is

PR #774, opened 2022-04-06
[CS-3603]: Reset nav state to remove non-auth routes

danibonilha:
### Description
This PR rewrites the navigation state on a fresh imported wallet to prevent the user from returning to the welcome screen when it already has a wallet

### DIFF
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -286,3 +286,21 @@ export const tabLinking = {
     },
   },
 };
+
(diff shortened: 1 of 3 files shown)


oidoug:
I understand now why on wallet creation the bug didn't happen, it's because we don't need to push another sheet up for that. Using nav replace in a sheet with the wallet screen causes it to lose the safe area context we need.

I'm adding your new reset nav call to also to backup flow, it had the same issue as the import one.

oidoug (approved):
LGTM!

PR #776, opened 2022-04-07
[CS-3554] Improved Safe Area View top padding handling.

oidoug:
### Description

PR improves Safe Area View padding Top handling by using component and hook from safe-area-context lib. React-Native's one seems to not update values correctly.

Also, we need to use the top's navigator `mode="modal"`, so on Android, the navigator knows it needs to keep the instance of the underlying screen alive, with card mode the background "flashes".

--

Updates:

- Upgraded react-native-safe-area-view package to the latest. Now it works as we would expect, defining safe areas even for bottom handlers and proper sizes for the status bar. But that means we either use SafeAreaView or set insets, if both are set we see the inset doubled. The new version also works properly on Android, so there is no need to grab the `StatusBar.currentHeight` anymore.
- Refactored all uses of hook `useSafeArea` to `useSafeAreaInsets`. Works the same, just a better naming.

- [x] Completes #(CS-3554)

### Checklist

- [x] Tested on Android, iPhone SE, and iPhone MAX, iOS 12 up to 15.

### Screenshots

https://user-images.githubusercontent.com/129619/162227851-9ee53e9c-64c0-4fff-80aa-05a6a053fcac.MP4

https://user-images.githubusercontent.com/129619/162273560-86337dd7-bf46-4665-b99f-ad1a9e6538db.mov

### DIFF
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -163,7 +163,7 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
   ...LoadingOverlayComponent,
   SETTINGS_MODAL: {
     component: SettingsModal,
-    options: slideLeftToRightPreset,
+    options: { ...slideLeftToRightPreset, gestureEnabled: false },
   },
   TRANSFER_CARD: {
     component: TransferCardScreen,
--- a/cardstack/src/components/MainHeader/components/MainHeaderWrapper.tsx
+++ b/cardstack/src/components/MainHeader/components/MainHeaderWrapper.tsx
@@ -1,6 +1,4 @@
-import React, { memo, useMemo } from 'react';
-import { StatusBar } from 'react-native';
-import { useSafeArea } from 'react-native-safe-area-context';
+import React, { memo } from 'react';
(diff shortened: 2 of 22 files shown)


oidoug:
While checking a broken jest test, I realized the package `react-native-safe-area-context` was outdated, and that we could update it to the last version without much refactoring. So I'm taking the opportunity to update it.

The update is important since we are using its inset hook everywhere and the new version opens some more opportunities for diminishing our reliance on the inset hook by just using the top SafeAreaView, plus it has some other new utils.

cardstack/src/navigation/screens.ts
```diff
@@ -163,7 +163,7 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
   ...LoadingOverlayComponent,
   SETTINGS_MODAL: {
     component: SettingsModal,
-    options: slideLeftToRightPreset,
+    options: { ...slideLeftToRightPreset, gestureEnabled: false },
```
danibonilha, line 166: thanks 😅

danibonilha (approved):
LGTM, nice job on bumping the safe-area 👏

PR #777, opened 2022-04-07
[CS-3604]: Bump sentry and update release scripts

danibonilha:
### Description

Bump sentry and update release scripts, tested locally was able to generate the builds, but will trigger a build from this branch to make sure everything is good, also bumps the app next version. 
Please run `yarn fastlane sync_sentry_properties` to get the sentry file. This file is used to upload the sourcemaps to the release

### Screenshots

image

image


src/App.js
```diff
@@ -73,24 +73,10 @@ if (__DEV__) {
...
-
-  if (android) {
-    const dist = VersionNumber.buildVersion;
-    // In order for sourcemaps to work on android,
-    // the release needs to be named with the following format
-    // me.rainbow@1.0+4
-    const releaseName = `me.rainbow@${VersionNumber.appVersion}+${dist}`;
-    sentryOptions.release = releaseName;
-    // and we also need to manually set the dist to the versionCode value
-    sentryOptions.dist = dist.toString();
```
danibonilha, line 91: no need for this anymore on newer version, I noticed that actually sourceMaps were never uploaded, so we should probably take a look into that sometime

scripts/postinstall.sh
```diff
@@ -41,16 +41,3 @@ else
...
-
-if ! command -v sentry-cli &> /dev/null
-then
-  echo "Installing sentry-cli..."
-  brew install getsentry/tools/sentry-cli
-else
-  which sentry-cli
-  sentry-cli --version
-  echo "sentry-cli is already installed. Skipping" 
-fi
```
danibonilha, line 56: lavamoat-allowscript, sentry: true, replaces this

oidoug (approved):
I'm unable to properly test this in a timely manner, so I'm approving but I'll check in detail later.

danibonilha:
The android build was successfully deployed, but the iOS one had an error uploading the debug symbols, will check it before merging it.

PR #780, opened 2022-04-08
[CS-3623] Seed Phrase backup decryption

oidoug:
### Description

We identified some problems returning a valid seed phrase when a backup to Google Cloud of the wallet was made in:
- A: a Touch/Face ID enabled device, errors when trying to recover in a PIN-based device.
- B: the reverse.

So this workaround first checks if the seed phrase is indeed encrypted before trying to unpack it, or just returns it.

There are some side effects, such as not asking for PINs in PIN-based devices because the BKP doesn't need one, and asking for PIN on Touch/Face ID enabled devices since it's needed to decrypt.

A better approach will be to create a PIN regardless, encrypt it regardless, and use Touch/Face ID only to unlock the PIN.

## Update

Included also the same workaround for `loadPrivateKey` function.

-[x] Completes #(CS-3623)

### DIFF
--- a/src/model/wallet.ts
+++ b/src/model/wallet.ts
@@ -37,7 +37,7 @@ import {
   isValidMnemonic,
 } from '../handlers/web3';
 import showWalletErrorAlert from '../helpers/support';
-import { isValidSeedPhrase } from '../helpers/validators';
+import { isValidSeed } from '../helpers/validators';
 import { EthereumWalletType } from '../helpers/walletTypes';
 import { getRandomColor } from '../styles/colors';
 import { ethereumUtils } from '../utils';
@@ -360,19 +360,30 @@ const loadPrivateKey = async (): Promise<
     }
     const privateKey = get(privateKeyData, 'privateKey', null);
 
+    const noPIN = !(await getExistingPIN());
+
+    if (privateKey && isValidSeed(privateKey) && noPIN) {
+      logger.sentry('Got key succesfully');
+      return privateKey;
+    }
+
     if (Device.isAndroid && privateKey) {
       const hasBiometricsEnabled = await getSupportedBiometryType();
       // Fallback to custom PIN
       if (!hasBiometricsEnabled) {
         try {
           const userPIN = await authenticateWithPIN();
-
           if (userPIN) {
+            if (isValidSeed(privateKey)) {
+              logger.sentry('Got key succesfully with PIN');
+              return privateKey;
+            }
+
             const decryptedPrivateKey = await encryptor.decrypt(
               userPIN,
               privateKey
             );
-
+            logger.sentry('Got decruypted key succesfully');
             return decryptedPrivateKey;
           }
         } catch (e) {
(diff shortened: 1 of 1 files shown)


Ab110055:
Good project strong project

src/model/wallet.ts
```diff
@@ -975,32 +976,28 @@ export const loadSeedPhrase = async (
...
 
-    if (seedPhrase) {
+    if (isValidSeedPhrase(seedPhrase) && noPIN) {
       logger.sentry('got seed succesfully');
-    } else {
-      captureMessage(
-        'Missing seed for account - (Key exists but value isnt valid)!'
-      );
+      return seedPhrase;
     }
```
danibonilha, line 996: let's add couple of sentry logs here, so it's easier to track if we need to

src/model/wallet.ts
```diff
@@ -975,32 +976,28 @@ export const loadSeedPhrase = async (
...
-          if (userPIN) {
-            const decryptedSeed = await encryptor.decrypt(userPIN, seedPhrase);
-            return decryptedSeed;
-          }
-        } catch (e) {
-          return null;
+      try {
+        const userPIN = await authenticateWithPIN(promptMessage);
+        if (userPIN) {
+          if (isValidSeedPhrase(seedPhrase)) return seedPhrase;
```
danibonilha: logger.sentry('got seed succesfully from PIN');

src/model/wallet.ts
```diff
@@ -975,32 +976,28 @@ export const loadSeedPhrase = async (
...
+
+          const decryptedSeed = await encryptor.decrypt(userPIN, seedPhrase);
+          return decryptedSeed;
```
danibonilha: logger.sentry('got decrypetedSeed succesfully');

src/model/wallet.ts
```diff
@@ -975,32 +976,28 @@ export const loadSeedPhrase = async (
...
         }
+      } catch (e) {
+        return null;
       }
     }
 
-    return seedPhrase;
+    return null;
```
danibonilha, line 1017: logger.sentry('No seed returned');

oidoug:
@danibonilha while testing, I found out we had the same issue for loading up the private key for all transactions.

So I've done the same to it, also another case for revisiting later with a better PIN implementation, and maybe doing so by reusing code for both key and seed loading.

Both cases, going from biometric to pin, and vice-versa, seems to work fine now, PIN is asked in both cases as we discussed.

danibonilha (approved):
LGTM, props on fixing the key too

src/model/wallet.ts
```diff
@@ -359,28 +360,39 @@ const loadPrivateKey = async (): Promise<
...
+              logger.sentry('Got key succesfully with PIN');
+              return privateKey;
+            }
+
             const decryptedPrivateKey = await encryptor.decrypt(
               userPIN,
               privateKey
             );
-
+            logger.sentry('Got decruypted key succesfully');
```
lukemelia:
```suggestion
            logger.sentry('Got decrypted key successfully');
```

lukemelia (approved):

PR #781, opened 2022-04-11
[CS-3551] Fix signing request loading by pressing push notification from app closed state

topwebtek7:
### Description
- In handling initial/generic WC notification only, reconnect web3Provider before fetching balances to make it properly 
- Refactor with useAssetBalance hook and got rid of useWalletBalances.js

- [x] Completes #CS-3551

### Checklist

### Screenshots

https://user-images.githubusercontent.com/16714648/162747108-128a53aa-9c05-4202-a0b4-c7b9029e6eaa.mp4

https://user-images.githubusercontent.com/16714648/162747132-d13da3cb-3cbc-411d-b08b-3b7f9f0c064f.mp4


src/hooks/useWalletBalances.js
```diff
@@ -38,21 +40,30 @@ const useWalletBalances = wallets => {
...
-      );
+      let balances = [];
+      try {
+        balances = await balanceCheckerContract.balances(keys(walletBalances), [
+          ETH_ADDRESS,
+        ]);
+      } catch (e) {
+        logger.log('error getting balances', e);
+      }
+
```
danibonilha: I recall we had some issues, using this type of function to get the balances, here is the reference: https://github.com/cardstack/cardwallet/pull/542, can you check if we can use the same approach here and maybe this already fixes the issue by itself

danibonilha (commented):
Can you elaborate on why this issue was happening ?

topwebtek7:
> Can you elaborate on why this issue was happening ?

We used to have this issue - `[Error: call revert exception (method="balances(address[],address[])", errorSignature=null, errorArgs=[null], reason=null, code=CALL_EXCEPTION, version=abi/5.0.7)]` when try to get balances at initial app loading and found it happens when Web3Provider is not connected properly, that's why I reconnected and addd exception handling as well.
I think we can get rid of `useWalletBalances.js` but some refactoring is needed. I suggest do that in another ticket, will create one for it. In this ticket I just fixed the mentioned bug.

danibonilha:
> > Can you elaborate on why this issue was happening ?
> 
> We used to have this issue - `[Error: call revert exception (method="balances(address[],address[])", errorSignature=null, errorArgs=[null], reason=null, code=CALL_EXCEPTION, version=abi/5.0.7)]` when try to get balances at initial app loading and found it happens when Web3Provider is not connected properly, that's why I reconnected and addd exception handling as well. I think we can get rid of `useWalletBalances.js` but some refactoring is needed. I suggest do that in another ticket, will create one for it. In this ticket I just fixed the mentioned bug.

Thanks for the context, so it's the same issue with the contract. What size of refactoring are we talking about ? If this is an urgent issue and the refactoring will be massive, then I agree we can have another ticket, but if it's a small to medium refactoring I don't see a reason to merge this as is and the next day merge the refactoring piece. I believe we can meet in the middle here, the ideal scenario would be to get rid of `useWalletBalances` and we probably already have the balances info stored, which we could use within GenericDisplay, but given that  `useWalletBalances` is used on diff places, we could change it's implementation to get the balances using `getOnChainAssetBalance`, this way the refactoring isn't huge and the bug is fixed.

topwebtek7:
ok, doing refactoring now

src/components/change-wallet/AddressRow.js
```diff
@@ -39,22 +37,11 @@ export default function AddressRow({ data, editMode, onPress, onEditWallet }) {
...
-    } else {
-      balanceAmount = `${balance || 0}`;
-    }
-
-    accountSubLabel = `${balanceAmount} ${nativeTokenSymbol}`;
-  }
+  const accountSubLabel = useMemo(
+    () => (isLayer2(network) ? getAddressPreview(address) : ''),
+    [address, network]
+  );
```
topwebtek7, line 44: removed balances to get rid of useWalletBalances. I think it's ok cause we don't show accounts list now in new design.

topwebtek7:
@danibonilha removed useWalletBalances and did some refactoring as well. please review again

danibonilha (approved):
LGTM,  thanks for refactoring!

PR #782, opened 2022-04-11
[CS-3620]: Fix crash on importing old nav

danibonilha:
### Description

`ScrollPagerWrapper` was throwing a ID error when it's replaced after importing the wallet,  but it seems we don't even need to use it anymore, my guess is that this implementation was to have the assetList etc, inside the scan screen, since we've moved away from those things, it's safe to remove, slide navigation still works on both platforms.

Simulator Screen Recording - iPhone 11 - 2022-04-11 at 12 12 53


lukemelia (approved):

oidoug (approved):
👍

PR #785, opened 2022-04-12
[CS-3618]: Format rewards tokens amount

danibonilha:
### Description
This PR adds a util function to convert wei values to ether with 2 decimals, in order to display all token amounts rounded, I don't have a claim amount of this type, so I added tests to make sure we will get the expected result.


topwebtek7 (approved):
LGTM

PR #786, opened 2022-04-12
[CS-3602]: Add withdraw transaction on  rewards history

danibonilha:
### Description

This PR adds the withdraw tx to rewards history section, next we should handle redirecting to blockscout on click and adding the from info, but I believe we will probably need to rethink the RewardRow's implementation in order to do that.

### Screenshots

PS: the token amount not fixed is pending on #785 
image

### DIFF
--- a/globals.d.ts
+++ b/globals.d.ts
@@ -14,3 +14,8 @@ declare module 'react-native-keychain' {
     results: rnKeychain.UserCredentials[];
   }>;
 }
+
+type OptionalUnion<T1, T2> = {
+  [P in keyof Omit<T1 & T2, keyof (T1 | T2)>]?: (T1 & T2)[P];
+} &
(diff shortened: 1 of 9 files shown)


globals.d.ts
```diff
@@ -14,3 +14,8 @@ declare module 'react-native-keychain' {
     results: rnKeychain.UserCredentials[];
   }>;
 }
+
+type OptionalUnion<T1, T2> = {
+  [P in keyof Omit<T1 & T2, keyof (T1 | T2)>]?: (T1 & T2)[P];
+} &
+  (T1 | T2);
```
oidoug, line 21: 👏

oidoug (commented):
Looks good!

oidoug (approved):

PR #787, opened 2022-04-13
[CS-3610][CS-3469] Updates gas estimation for max amount

oidoug:
### Description

Fixes issue when selecting max amount in Send Sheet for Depot and ~~EOA (not done yet)~~, and transactions couldn't complete for lack of funds. Gas estimation wasn't updated for the new amount so it wasn't enough at times.

- [x] Completes #(CS-3610)
- [x] Completes #(CS-3469)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/163181343-3063572a-04cc-4296-8fd7-b8b1d3194aa7.mp4

### DIFF
--- a/cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
+++ b/cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
@@ -87,7 +87,7 @@ export const useSendSheetDepotScreen = () => {
 
   const [amountDetails, setAmountDetails] = useState(amountDetailsInitialState);
 
-  const [maxInputBalance, updateMaxInputBalance] = useState('');
+  const maxInputBalance = useRef('');
 
   const isValidAddress = useSendAddressValidation(recipient);
   const [isAuthorizing, setIsAuthorizing] = useState(false);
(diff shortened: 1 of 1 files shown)


danibonilha (dismissed):
~~Let's merge this workaround until we the sdk provides the full amount option, then we revisit this, also let's keep on the radar that we need to update the gas estimate while the user is typing~~

cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
```diff
@@ -318,11 +319,30 @@ export const useSendSheetDepotScreen = () => {
...
+      // So, to keep the total value in bounds of balance, we need
+      // to subtract the difference of the new calculation from the previous one.
+      const gasDiff = new BN(gasEstimatedFee.amount).sub(
+        new BN(fullAmountGasEstimate)
+      );
+
+      const adjustedSendAmountInWei = new BN(amountInWei)
+        .add(gasDiff)
+        .toString();
+
```
danibonilha, line 340: I found an issue with this, we need to constraint this deduction to only when the user wants to send the max balance. For example, if I want to send  1 CARD, I expect 1 CARD to be sent and the gas fee to be deducted  from my remaining balance, with this approach, if I want to send 1 CARD, it will deduct the gas from my 1 CARD token and send only 0.86 CARD

oidoug:
Closing this PR since we decided in favor of fixing the SDK implementation instead.

PR #788, opened 2022-04-13
[CS-3619] Fix dismiss overlay on android when creating profile fails

topwebtek7:
### Description

Tested on several android devices and this fixed dismissing overlay for rarely failed cases on Android

- [x] Completes #CS-3619

### DIFF
--- a/cardstack/src/screens/ProfileScreen/useProfileForm.ts
+++ b/cardstack/src/screens/ProfileScreen/useProfileForm.ts
@@ -1,5 +1,5 @@
 import { useContext, useCallback, useEffect, useMemo, useState } from 'react';
-import { Alert } from 'react-native';
+import { Alert, InteractionManager } from 'react-native';
 import { validateMerchantId } from '@cardstack/cardpay-sdk';
 import { useNavigation } from '@react-navigation/native';
 import { ProfileFormContext, strings, exampleMerchantData } from './components';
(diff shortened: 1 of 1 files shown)


danibonilha:
@topwebtek7 Is there steps to repro this issue ? I can't repro, and it seems the ticket is about a successful state, that just froze

topwebtek7:
> @topwebtek7 Is there steps to repro this issue ? I can't repro, and it seems the ticket is about a successful state, that just froze

@danibonilha Hm I was also not able to replicate for successful state, tested on 3 different android devices and looked ok when created profile successfully, but experienced overlay does not dismiss when failed state and this pr for that bug.

cardstack/src/screens/ProfileScreen/useProfileForm.ts
```diff
@@ -71,6 +72,11 @@ export const useProfileForm = (params?: useProfileFormParams) => {
           status: isError,
           callback: () => {
             dismissLoadingOverlay();
+
+            if (Device.isAndroid) {
+              dismissLoadingOverlay();
+            }
+
```
danibonilha: instead of retry the call for the same function on Android, would using interactionManager work for this case ? just one dimissOverlay call wrapped in the intereaction ?

topwebtek7: @danibonilha good point. that fixed for me and updated pr

danibonilha (approved):
LGMT

PR #789, opened 2022-04-13
[CS-3659]: Attempt to fix rewards withdraws

danibonilha:
### Description

This PR uses the new approach to not send an amount to be able to withdraw the whole balance without deducting the gas fee on the app side.  I know we can simply not pass amount, but I decided to make it explicitly. 

Simulator Screen Recording - iPhone 11 - 2022-04-13 at 15 26 07


tintinthong:
Just want to note here. That my expectation is that if you ever want to pass in amount. You still need to (as before) deduct the gas from the amount. It should be backward compatible with the previous way you have done it. It has worked for most of our users so I kept it that way.

oidoug (approved):
LGTM

PR #790, opened 2022-04-14
[CS-3692] Improved font family usage | text on Tab Bar

oidoug:
### Description

This PR adds a better implementation for font-family (and weight) usage by using restyle `textVariants`. That's preferable because with it we can keep a library of font styles used throughout the app, moving font styling from components to the Theme.

It also fixes Tab Bar's height by giving it a fixed height, plus padding depending on bottom insets from safe-area-context. That's preferable to having it proportional to the screen's height because tall devices could present too big bars.

- [x] Completes #(CS-3692)

### Checklist

### Screenshots

OLD | NEW
--- | ---
Screenshot_20220413-232001_Card Wallet | Screenshot_20220413-232345_Card Wallet

OLD Tablet | NEW Tablet
--- | ---
Screenshot_20220413-170132_Card Wallet | Screenshot_20220413-231635_Card Wallet

### DIFF
--- a/cardstack/src/components/TabBarIcon/TabBarIcon.tsx
+++ b/cardstack/src/components/TabBarIcon/TabBarIcon.tsx
@@ -29,7 +29,12 @@ const TabBarIcon = ({ focused, iconName, label }: TabBarIconProps) => {
   );
 
   return (
-    <Container flex={1} alignItems="center">
+    <Container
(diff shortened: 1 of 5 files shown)


topwebtek7 (approved):
LGTM!

danibonilha (approved):
👏

PR #791, opened 2022-04-14
[CS-3700] Update profile flow and fix selecting color in profile

topwebtek7:
### Description

- Fixed ProfileScreen reloaded when colorpicker selected
- I think this can fix CS-3698 as well, we can see after new version published so Gabi can confirm on her side.

- [x] Completes #CS-3700
- [x] Completes #CS-3697

### Checklist

### Screenshots

https://user-images.githubusercontent.com/16714648/163401792-01e62734-dfac-48e7-9454-1b11de80fc37.mp4


oidoug (approved):
I don't see the loading happening, but the code looks good.

PR #792, opened 2022-04-14
 [CS-3696]: Fix crash on withdrawTo screen

danibonilha:
### Description

Safes query was not properly typed which could lead to undefined errors, specifically this error was on the withdrawaTo screen, I updated the types to avoid any upcoming issues, we could of course improve the selectFromResult, or even make the safe query keys default to `[]`, but it's safer and easier to do this way, considering we will move to persona safe soon and have a single data structure within safes.
The issue was happening when safes were not fully loaded while trying to withdraw, I added a loading state, not the best UI, but should do for now.

Completes CS-3694

Simulator Screen Recording - iPhone SE (2nd generation) - 2022-04-14 at 15 00 39

### DIFF
--- a/cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
+++ b/cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
@@ -2,6 +2,12 @@ import { convertRawAmountToBalance } from '@cardstack/cardpay-sdk';
 import React from 'react';
 import { ActivityIndicator } from 'react-native';
 import { SectionHeaderText } from '../components/SectionHeaderText';
+import {
+  MerchantOrDepotSafe,
+  IssuePrepaidCardDecodedData,
+  DepotType,
+  MerchantSafeType,
+} from '@cardstack/types';
 import { ContactAvatar } from '@rainbow-me/components/contacts';
 import {
   Container,
@@ -11,11 +17,7 @@ import {
   Text,
   TransactionConfirmationDisplayProps,
 } from '@cardstack/components';
-import {
-  IssuePrepaidCardDecodedData,
-  DepotType,
-  MerchantSafeType,
-} from '@cardstack/types';
+
 import {
   getSafeTokenByAddress,
   convertSpendForBalanceDisplay,
@@ -48,6 +50,11 @@ export const IssuePrepaidCardDisplay = (
   );
 };
 
+interface SelectedResult {
+  fromSafe?: MerchantOrDepotSafe;
+  isLoadingSafe: boolean;
+}
+
 const FromSection = ({
   safeAddress,
   tokenAddress,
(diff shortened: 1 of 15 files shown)


cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
```diff
@@ -67,20 +74,24 @@ const FromSection = ({
...
       }),
     }
   );
 
   const safeTypeText =
-    fromSafe?.merchantInfo?.name || fromSafe?.type.toUpperCase();
+    fromSafe?.merchantInfo?.name || fromSafe?.type?.toUpperCase();
 
-  const token = getSafeTokenByAddress(fromSafe, tokenAddress);
+  const token = getSafeTokenByAddress(fromSafe?.tokens || [], tokenAddress);
```
oidoug, line 94: I think the `|| [ ]` isn't necessary, because `getSafeTokenByAddress` will already return undefined if tokens is undefined.

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/RewardWithdrawToScreen.tsx
```diff
@@ -1,19 +1,34 @@
...
     <Container backgroundColor="white" flex={1}>
       <NavigationStackHeader title={strings.headerTitle} />
-      <SafeSelectionList
-        safes={availableSafesToWithdraw}
-        onSafePress={onSafePress}
-      />
+      {isLoading ? (
+        <CenteredContainer flex={1}>
+          <ActivityIndicator />
+        </CenteredContainer>
```
oidoug, line 25: nit: For a future PR it would be nice to have skeletons used here, and maybe move the loading into the SafeSelectionList comp.

cardstack/src/services/safes-api.ts
```diff
@@ -9,14 +10,21 @@ export enum CacheTags {
...
+  merchantSafes?: MerchantSafeType[];
+  timestamp: string;
+}
+
 export const safesApi = createApi({
   reducerPath: 'safesApi',
   baseQuery: fetchBaseQuery({ baseUrl: '/' }),
   tagTypes: [...Object.values(CacheTags)],
   endpoints: builder => ({
     // TODO: Add right return type
```
oidoug: comment not needed anymore

oidoug (approved):
All good, just some nits.

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/RewardWithdrawToScreen.tsx
danibonilha: Agreed, we just don't have any designs for this specific case at this time, and using on the SafeList right now, makes the InfoBanner jump on the screen

oidoug: Ah yeah, the info banner would need to be fixed at the bottom or before the list for it to work.

cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
danibonilha: getSafeTokenByAddress expects a value for tokens, I thought about making the param optional, but it doesn't make much sense to have both params of this function to be optional, so need this to avoid ts errors.

cardstack/src/services/safes-api.ts
danibonilha: good catch

cardstack/src/screens/RewardsCenterScreen/flows/RewardWithdraw/RewardWithdrawTo/RewardWithdrawToScreen.tsx
danibonilha: Exactly, so we can put the effort on this once the design team decides the best approach 🙂

PR #793, opened 2022-04-14
[CS-3610] Disables Max button on Depot transfers

oidoug:
### Description

This PR removes (comments) the MAX button on Depot transfer sheet, because it's not working as intended, so we need to improve the SDK implementation before re-adding it.

- [x] Completes #(CS-3610)

### Checklist


cardstack/src/screens/SendSheetDepot/useSendSheetDepotScreen.ts
```diff
@@ -401,7 +403,7 @@ export const useSendSheetDepotScreen = () => {
       onChangeAssetAmount,
       onChangeNativeAmount,
       onResetAssetSelection,
-      onMaxBalancePress,
+      // onMaxBalancePress,
```
danibonilha: Can we make this undefined ? Instead of changing in the sendSheet component ? Less comments and steps to change when we re-add it

danibonilha (approved):
Looks good just one small implementation suggestion

PR #794, opened 2022-04-14
[CS-3665]: Add structure and initial components for card-drop

danibonilha:
### Description

This PR adds the structure of request card screen, so we can start working on the other tickets, most components are just placeholders, for this task the InfoBanner is the only one completed, all the others have specific tickets to be handled. Also fixes the tabBar on iOS devices.

### Screenshots

// Placeholders
image


oidoug (approved):
all good

PR #798, opened 2022-04-18
[Chore] Cleaning unused font files 

oidoug:
### Description

I found out we had around 10MB of already compressed font files not used anywhere in the App. This PR removes most of Roboto's and all of Karbon files from:
- assets/fonts folder
- iOS's Info.plist and copy bundle 
- Android's assets folders

SF-PRO and SFMono were left behind because they are being used in Coin's Graph screen, which will need some refactoring to use OpenSans.


danibonilha (approved):
LGTM, nice cleanup 👏

PR #799, opened 2022-04-18
[CS-3706]: Fix find undefined error and reorder sentry logs

danibonilha:
### Description
Giving where this error was happening, it seems it's been happening for a while, but with the sentry update, sentry captures the exception better and now we are able to see, this fixes the issue and also reorders the logs, to log the breadcrumb before capturing some exceptions.

### DIFF
--- a/cardstack/src/redux/collectibles.ts
+++ b/cardstack/src/redux/collectibles.ts
@@ -171,8 +171,9 @@ const fetchNFTsViaOpenSea = () => (
       }
     } catch (error) {
       dispatch({ type: COLLECTIBLES_FETCH_FAILURE });
-      captureException(error);
+
       logger.sentry('COLLECTIBLES_FETCH_FAILURE');
+      captureException(error);
     }
   };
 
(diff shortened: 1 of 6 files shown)


cardstack/src/redux/collectibles.ts
```diff
@@ -172,6 +172,7 @@ const fetchNFTsViaOpenSea = () => (
     } catch (error) {
       dispatch({ type: COLLECTIBLES_FETCH_FAILURE });
       captureException(error);
+      logger.sentry('COLLECTIBLES_FETCH_FAILURE');
```
oidoug, line 175: Doesn't it make sense to have the capture before the log here too?

src/redux/fallbackExplorer.js
```diff
@@ -83,7 +83,7 @@ const fetchCoingeckoIds = async (network, coingeckoCoins) => {
     const honeyswapTokens = data.tokens;
 
     honeyswapTokens.forEach(({ address: tokenAddress, symbol }) => {
-      const coingeckoToken = coingeckoCoins.find(
+      const coingeckoToken = coingeckoCoins?.find(
```
oidoug, line 86: 👏

oidoug (approved):
LGTM, just a minor question.

cardstack/src/redux/collectibles.ts
danibonilha: it does! 😅, good catch.

PR #800, opened 2022-04-18
[CS-3531] Tags network on Sentry when initializing or switching networks.

oidoug:
### Description

PR exposes `Sentry.setTag` call through Logger so on wallet initialization and network switching we set the current network, for better debugging in Sentry.

- [x] Completes #(CS-3531)

### Screenshots

Screenshot

Screen Shot 2022-04-18 at 17 28 38


danibonilha (approved):
LGTM

PR #803, opened 2022-04-19
[CS-3667]: Add FormInput and email validation 

danibonilha:
### Description
This PR adds the FormInput component with variants and error handling which is then used on the RequestCard screen, where the email validation happens, to give better user experience, instead of saying it's invalid right in the first char, we actually allow partial values, if there's some really weird format, we trigger the error message, otherwise, we let the user type and we have a more enforced validation on the submit button, if it's invalid and the user tries to tap it even if disabled we show the error feedback. We are not tracking the onFocus borderColor state right now as it adds a complexity that's not worthy at the moment, so it defaults to teal. 

### Screenshots

Simulator Screen Recording - iPhone 12 Pro - 2022-04-19 at 13 53 21


cardstack/src/components/Input/FormInput.tsx
```diff
@@ -0,0 +1,128 @@
...
+      <Container
+        flexDirection="row"
+        justifyContent="space-between"
+        paddingBottom={2}
+      >
+        <Text color="borderGray" weight="bold" fontSize={14}>
+          {label}
+        </Text>
+        <Text color="borderGray" fontSize={12}>
+          {isRequired ? 'required' : ''}
```
oidoug, line 101: nit: might be a good candidate for i18n

cardstack/src/utils/__tests__/validators.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  it('should return false for invalid email even if is partial', () => {
+    invalids
+      .filter(email => !partialValids.includes(email))
+      .forEach(input => {
+        const result = isEmailPartial(input);
+
+        expect({ input, result }).toEqual({ input, result: false });
+      });
+  });
+});
```
oidoug, line 42: 👏

oidoug (approved):
Well done on the new Input, well rounded.
Left a minor nit.

cardstack/src/components/Input/FormInput.tsx
danibonilha: right, forgot to extract this one, will fix one latest PR

PR #806, opened 2022-04-20
[CS-3668]: Add non auth icon to request submit btn

danibonilha:
### Description

This PR adds the icon based on the auth flag, also adds a helper to handle the iconProps based on biometry type and fixes a bug on the rewards withdraw where face-id was hardcoded.

### Screenshots

image

### DIFF
--- a/cardstack/src/components/Input/FormInput.tsx
+++ b/cardstack/src/components/Input/FormInput.tsx
@@ -1,11 +1,14 @@
 import React, { memo, useEffect, useMemo, useRef } from 'react';
 import { Animated } from 'react-native';
 
+import { Device } from '@cardstack/utils/device';
+
 import { Container } from '../Container';
 import { IconProps } from '../Icon';
 import { Text } from '../Text';
 
 import { Input, InputProps } from './Input';
+import { strings } from './strings';
 
 const baseInputProps: InputProps = {
   borderWidth: 1,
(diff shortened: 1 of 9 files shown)


cardstack/src/components/Input/FormInput.tsx
```diff
@@ -19,7 +22,7 @@ const baseInputProps: InputProps = {
 type InputVariants = 'valid' | 'error' | 'default';
 
 const baseIconProps: Partial<IconProps> = {
-  top: 12,
+  top: Device.isIOS ? 12 : 13,
```
oidoug, line 25: Just a thought, I wonder if that minor difference isn't needed because of a rounding issue on our SVGs. I see some of them have decimals in their sizes.

oidoug (approved):
On my Android, I see the face icon where it should be a fingerprint, but that's understandable since getSupportedBiometryType always returns 'Face'.
Hopefully, it's something we can improve by updating the keychain lib.

Regardless, PR LGTM.

cardstack/src/components/Input/FormInput.tsx
danibonilha: I don't think so, giving that this one is not using local svgs, it's using icons from [rn-vector-icons](https://github.com/oblador/react-native-vector-icons) which I'm not sure why this was added, we could probably add to tech debt to map the icons we need, add to the project and remove this lib, it might also decrease the app size. I'll also try to play with percentages to see if we can get a nice behavior without relaying on the platform

PR #808, opened 2022-04-20
[CS-3727]: Handle FormInput focused/blurred/error borderColor

danibonilha:
### Description

This PR handles the borderColor states on the formInput, also removes autoFocus from request screen, so the user can read the conditions.

~~Pending on #806~~

### Screenshots

IOS:
Simulator Screen Recording - iPhone 12 Pro - 2022-04-20 at 12 20 02
Android:
android


oidoug (commented):
I found issues with Android email validation that were merged in a previous PR.

oidoug (approved):
LGTM

PR #809, opened 2022-04-20
[CS-3715] Updates configurations for fetching RPC Node and hub url from sdk

oidoug:
Updates cardpay-sdk and did-resolver to v1.0.0;

### Description

This PR updates how we generate the RPC node and RPC WSS node URLs, using new functions from the SDK.

- [x] Completes #(CS-3715)


danibonilha (approved):
Let's merge it and later we can talk about abstraction the hubConfig to a singleton

PR #810, opened 2022-04-22
[Chore]: Bump to 1.1.0 and remove business banner

danibonilha:
Removing the business banner now that we can create a profile from within the app


oidoug (approved):
LGTM!

PR #811, opened 2022-04-22
[Improvement]: Add new hub service structure

danibonilha:
### Description

This PR adds the initial new structure of the hub api using RTK Query, automatically handling token headers and auth errors. It migrates one of the services as example, CustodialWallet, most of the keys from the api are kebab case, which might make accessing the properties a bit annoying, so I added some type and function helpers to transform kebab to camel case, we can discuss it if it helps or it's just over engineering but, the custodialWallet was refactored with those helpers. In order to have the recursive types, a typescript update was needed, which probably fixes CS-3695 (Ts caught that we were passing a boolean where it should actually be a filename, so hopefully this issue is fixed).

### DIFF
--- a/cardstack/src/services/hub/hub-api.ts
+++ b/cardstack/src/services/hub/hub-api.ts
@@ -1,0 +1,30 @@
+import { createApi } from '@reduxjs/toolkit/query/react';
+
+import { CustodialWallet } from '@cardstack/types';
+import { transformObjKeysToCamelCase } from '@cardstack/utils';
+
+import { fetchHubBaseQuery } from './hub-service';
+import { GetCustodialWalletQueryResult } from './hub-types';
+
+const routes = {
+  custodialWallet: '/api/custodial-wallet',
+};
+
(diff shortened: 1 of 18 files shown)


cardstack/src/services/hub/hub-api.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export const hubApi = createApi({
+  reducerPath: 'hubApi',
+  baseQuery: fetchHubBaseQuery,
+  endpoints: builder => ({
+    getCustodialWallet: builder.query<GetCustodialWalletQueryResult, void>({
+      query: () => routes.custodialWallet,
+      transformResponse: (response: { data: CustodialWallet }) => {
+        const attributes = transformObjKeysToCamelCase(
```
oidoug, line 20: Nice

oidoug (approved):
LGTM. I see we now have a nice template with RTKQuery to move all hub calls over 👏

PR #812, opened 2022-04-25
[CS-3663] Card drop banner display control

oidoug:
### Description

- Adds hook useWelcomeCtaBanner to control the display of the welcome banner;
- Adds rtk-query for Hub Public API.

new:
- Refactored hub-public-api to conform to new HUB structure.
- Renamed helper fetchHubBaseQuery to fetchHubAuthenticatedQuery to avoid confusion. Base only sets baseURL and base headers.

~~There are some refactoring I could do here after we merge #811 , we can move the hub public API introduced here to the hub folder and extract hub-public-api `baseQuery` function to `hub-service.ts`, so we can reuse some bits from it, but I think it's preferable to handle the public part as a different API since it doesn't need authentication.~~

- [x] Completes #(CS-3663)

### Checklist

### Screenshots

Screenshot


cardstack/src/services/hub-public-api.ts
```diff
@@ -0,0 +1,32 @@
...
+    prepareHeaders: headers => {
+      headers.set('Accept', 'application/vnd.api+json');
+      headers.set('Content-Type', 'application/vnd.api+json');
+
+      return headers;
+    },
+  }),
+  tagTypes: [...Object.values(HubPublicCacheTags)],
+  endpoints: builder => ({
+    getEoaClaimed: builder.query<any, EoaClaimedParams>({
```
danibonilha: Let's type the query return, we can also transform the response to just return the claimed boolean

cardstack/src/services/hub-public-api.ts
```diff
@@ -0,0 +1,32 @@
...
+  reducerPath: 'hubPublicApi',
+  baseQuery: fetchBaseQuery({
+    baseUrl: HUB_URL,
+    prepareHeaders: headers => {
+      headers.set('Accept', 'application/vnd.api+json');
+      headers.set('Content-Type', 'application/vnd.api+json');
+
+      return headers;
+    },
+  }),
```
danibonilha: Agreed that this can be a diff api, but we could think about customizing the `fetchHubBaseQuery` to conditionally add the token, so we can reuse the logic on choosing the hubUrl and the common headers on baseQuery

cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
```diff
@@ -0,0 +1,56 @@
...
+      (account: any) => account.address === accountAddress
+    );
+
+    return currentWalletAddress?.index === 0 || false;
+  }, [wallets, selectedWallet, accountAddress]);
+
+  // Addresses that have already claimed a prepaidcard don't see the banner.
+  const hasClaimed = useMemo(() => {
+    return data?.data?.attributes?.claimed || false;
+  }, [data]);
```
danibonilha: By transforming the response on the api we may not even need this as the default value can be added on the query destructuring

cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!wallets?.[selectedWallet?.id]) return false;
+
+    const { addresses } = wallets[selectedWallet.id];
+
+    const currentWalletAddress = addresses.find(
+      (account: any) => account.address === accountAddress
+    );
+
+    return currentWalletAddress?.index === 0 || false;
+  }, [wallets, selectedWallet, accountAddress]);
```
danibonilha, line 47: I believe it might be safer to rely on [web3.eth.getAccounts()](https://web3js.readthedocs.io/en/v1.2.11/web3-eth.html#getaccounts)  and check if the first address is same as the current selected wallet, instead of using the indexes locally set on wallet creation

cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  // Addresses that have already claimed a prepaidcard don't see the banner.
+  const hasClaimed = useMemo(() => claimedResponse ?? true, [claimedResponse]);
+
+  return {
+    showBanner:
+      showBannerUserDecision && isFirstAddressForCurrentWallet && !hasClaimed,
```
danibonilha: we could create this const, just to cleanup and I guess we should probably check if it's layer2 to show the banner

cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const currentWalletAddress = addresses.find(
+      (account: any) => account.address === accountAddress
+    );
+
+    return currentWalletAddress?.index === 0 || false;
+  }, [wallets, selectedWallet, accountAddress]);
+
+  // Addresses that have already claimed a prepaidcard don't see the banner.
+  const hasClaimed = useMemo(() => claimedResponse ?? true, [claimedResponse]);
```
danibonilha: nit: maybe we don't need this, we can default to true on line 17, `const { data: hasClaimed = true } = `

cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
oidoug: My understanding is that we can't get addresses from the contract without loading up the HDProvider, which prompts the touch id check. I tried calling without setting the provider and it returns empty. So not sure it's ideal.
Maybe there's another way to use it I'm not aware of, plz let me know.

The index is saved when creating an address for the wallet, and its also used when creating new addresses, so it seams reasonable to use it.

oidoug:
@danibonilha, I think you'll like what I found :) by using `extraOptions` parameter for BaseQueryFn, I was able to merge the public and private endpoints, so we only have `hub-api` now.

cardstack/src/services/hub/hub-api.ts
```diff
@@ -16,6 +22,7 @@ export const hubApi = createApi({
   endpoints: builder => ({
     getCustodialWallet: builder.query<GetCustodialWalletQueryResult, void>({
       query: () => routes.custodialWallet,
+      extraOptions: { authenticate: true },
```
danibonilha: 👏

cardstack/src/services/hub/hub-service.ts
```diff
@@ -13,32 +13,42 @@ import logger from 'logger';
...
+import { BaseQueryExtraOptions } from './hub-types';
+
 // Helpers
 
 export const fetchHubBaseQuery: BaseQueryFn<
   string | FetchArgs,
   unknown,
-  FetchBaseQueryError
+  FetchBaseQueryError,
+  BaseQueryExtraOptions // extraOptions
```
danibonilha: Maybe authenticate can default to true ? Giving 90% of the call will be authenticated

cardstack/src/services/hub/hub-api.ts
```diff
@@ -7,11 +7,15 @@ import { fetchHubBaseQuery, hubBodyBuilder } from './hub-service';
 import {
   GetCustodialWalletQueryResult,
   RequestCardDropQueryParams,
+  EoaClaimedArg,
+  EoaClaimedResultType,
 } from './hub-types';
 
 const routes = {
-  custodialWallet: '/custodial-wallet',
+  custodialWallet: '/api/custodial-wallet',
```
danibonilha:
```suggestion
  custodialWallet: '/custodial-wallet',
```

cardstack/src/services/hub/hub-service.ts
oidoug: well noted, I think so, will update

cardstack/src/services/hub/hub-api.ts
```diff
@@ -7,11 +7,15 @@ import { fetchHubBaseQuery, hubBodyBuilder } from './hub-service';
...
   emailDrop: '/email-card-drop-requests',
+  emailCardDropRequest: ({ eoa }: EoaClaimedArg) =>
+    `/email-card-drop-requests?eoa=${eoa}`,
```
danibonilha: The idea of the routes object is to list all the possible routes, but here it might be confusing to  have the same route twice with the only diff being one accepts a param, for this case it might make more sense to construct the variable on the query, I'll add the suggestion.

cardstack/src/services/hub/hub-api.ts
```diff
@@ -31,16 +36,25 @@ export const hubApi = createApi({
...
         url: routes.emailDrop,
+        extraOptions: { authenticate: true },
         method: 'POST',
         body: hubBodyBuilder(routes.emailDrop, {
           email,
         }),
       }),
     }),
+    getEoaClaimed: builder.query<EoaClaimedResultType, EoaClaimedArg>({
+      query: routes.emailCardDropRequest,
```
danibonilha:
```suggestion
      query: ({ eoa }) => `${routes.emailDrop}?eoa=${eoa}`,
```

cardstack/src/services/hub/hub-types.ts
```diff
@@ -7,3 +7,14 @@ export type GetCustodialWalletQueryResult = KebabToCamelCaseKeys<CustodialWallet
...
 }
+
+// Note on baseQuery extraOptions, it works only with optional properties.
+export interface BaseQueryExtraOptions {
+  authenticate?: boolean;
+}
+
+export interface EoaClaimedArg {
+  eoa: string;
+}
```
danibonilha: nit: It would be nice to follow the pattern as the other types something like `GetEoaClaimedQueryParams`

danibonilha (approved):
Props on adding the extraParams 👏 , it's looking good, some minor nits, and one required change -> remove the  /api/ leftover on custodialWallet of hub routes.

cardstack/src/services/hub/hub-api.ts
oidoug: ah, left-over from the rebase, thanks for noticing it.

PR #813, opened 2022-04-25
[CS-3669] - Handle submit request to email card drop

danibonilha:
### Description

This PR adds the mutation to request a card drop, also it adds the `hubBodyBuilder` 💪  helper to facilitate building the body on next requests.

Missing pieces, we still need to update the data after posting to this URL, but it's not so trivial since this is a different api than the safes, or hub-public one, so, still need to think what is the best way to do that, we could also think about adding some transition animation, but that's a nice to have we can add later.

### Screenshots

Simulator Screen Recording - iPhone 11 - 2022-04-25 at 18 22 13

### DIFF
--- a/cardstack/src/services/hub/hub-service.ts
+++ b/cardstack/src/services/hub/hub-service.ts
@@ -21,16 +21,16 @@ export const fetchHubBaseQuery: BaseQueryFn<
   FetchBaseQueryError
 > = async (args, api, extraOptions) => {
   const network = await getNetwork();
-  const baseUrl = network === Network.xdai ? HUB_URL : HUB_URL_STAGING;
+  const hubUrl = network === Network.xdai ? HUB_URL : HUB_URL_STAGING;
 
   const result = await fetchBaseQuery({
-    baseUrl,
+    baseUrl: `${hubUrl}/api`,
     prepareHeaders: async (headers, { getState }) => {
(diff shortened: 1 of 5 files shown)


cardstack/src/services/hub/hub-service.ts
```diff
@@ -21,16 +21,16 @@ export const fetchHubBaseQuery: BaseQueryFn<
   FetchBaseQueryError
 > = async (args, api, extraOptions) => {
   const network = await getNetwork();
-  const baseUrl = network === Network.xdai ? HUB_URL : HUB_URL_STAGING;
+  const hubUrl = network === Network.xdai ? HUB_URL : HUB_URL_STAGING;
 
   const result = await fetchBaseQuery({
-    baseUrl,
+    baseUrl: `${hubUrl}/api`,
```
oidoug, line 27: 👍  I'll need to update #812 again after merging this one.

cardstack/src/services/hub/hub-service.ts
```diff
@@ -58,3 +58,11 @@ export const fetchHubBaseQuery: BaseQueryFn<
 
   return result;
 };
+
+export const hubBodyBuilder = <Attrs>(path: string, attributes: Attrs) =>
```
oidoug, line 62: Nice!

oidoug (approved):
LGTM!

PR #814, opened 2022-04-26
[CS-3555] OpenSea API Key

oidoug:
### Description

This PR adds a new `OPENSEA_API_KEY` env var (already sync'd) and uses it in `opensea-api` calls. Hopefully, this will avoid issues while pulling collections from Opensea (several reports on sentry issue CARDWALLET-MOBILE-4Q).

- [x] Completes #(CS-3555)


danibonilha (approved):
LGTM

PR #815, opened 2022-04-26
[CS-3767]: Check if valid auth exists to display correct UI 

danibonilha:
### Description
This PR checks if a valid token exists in order to show or not the authentication icon on request card, there's not much evidence to add for this case, as most of the time it will be a valid section.


oidoug (approved):
LGTM!

PR #817, opened 2022-04-27
[CS-3766] Update copy and add terms acceptance to Request PrepaidCard Screen.

oidoug:
### Description

- Adds new copy and terms acceptance checkbox to Request PrepaidCard Screen.
- Adds a new validation step to form for terms.
- Updates Checkbox component for new use cases and refactored current ones.

- [x] Completes #(CS-3766)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/165525579-e5bf77f7-3cbd-4f00-886b-f823111435f1.mp4

https://user-images.githubusercontent.com/129619/165525612-b5726190-3e35-4a0b-885f-e10c34e6589f.mp4

### DIFF
--- a/cardstack/src/components/Checkbox/Checkbox.tsx
+++ b/cardstack/src/components/Checkbox/Checkbox.tsx
@@ -1,11 +1,11 @@
 import React, { useCallback, useState, useEffect, ReactNode } from 'react';
 
-import { Touchable, Container, ContainerProps, Text, Icon } from '../.';
+import { Touchable, Container, Text, Icon } from '../.';
 import { IconProps } from '../Icon';
 
 type CheckboxPositionType = 'left' | 'right';
 
-interface CheckboxProps extends ContainerProps {
+interface CheckboxProps {
   onPress?: () => void;
   label?: string;
   isDisabled?: boolean;
@@ -22,7 +22,6 @@ export const Checkbox = ({
   isSelected = false,
   checkboxPosition = 'right',
   children,
-  ...rest
 }: CheckboxProps) => {
   const [selected, setSelected] = useState(isSelected);
   const [disabled] = useState(isDisabled);
@@ -47,7 +46,6 @@ export const Checkbox = ({
       alignItems="center"
       onPress={handleCall}
       disabled={disabled}
-      {...rest}
     >
       <Container
         alignItems="center"
--- a/cardstack/src/components/Checkbox/Checkbox.tsx
+++ b/cardstack/src/components/Checkbox/Checkbox.tsx
@@ -1,14 +1,18 @@
-import React, { useCallback, useState, useEffect } from 'react';
+import React, { useCallback, useState, useEffect, ReactNode } from 'react';
 
-import { Touchable, Container, Text, Icon } from '../.';
+import { Touchable, Container, ContainerProps, Text, Icon } from '../.';
 import { IconProps } from '../Icon';
 
-interface CheckboxProps {
+type CheckboxPositionType = 'left' | 'right';
+
+interface CheckboxProps extends ContainerProps {
   onPress?: () => void;
   label?: string;
   isDisabled?: boolean;
   iconProps?: IconProps;
   isSelected?: boolean;
+  checkboxPosition?: CheckboxPositionType;
+  children?: ReactNode;
 }
 
 export const Checkbox = ({
@@ -16,6 +20,9 @@ export const Checkbox = ({
   onPress,
   isDisabled,
   isSelected = false,
+  checkboxPosition = 'right',
+  children,
+  ...rest
 }: CheckboxProps) => {
   const [selected, setSelected] = useState(isSelected);
   const [disabled] = useState(isDisabled);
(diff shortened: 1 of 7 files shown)


cardstack/src/screens/RequestPrepaidCardScreen/strings.ts
```diff
@@ -2,11 +2,18 @@ export const strings = {
...
+    title: 'GDPR & Marketing Permissions',
     message:
-      'By clicking the “Submit” button, you are submitting your Card Wallet address and email address. You agree to receive a verification email and Cardstack newsletter opt-in request.\n\nFor more details visit',
-    link: 'https://support.cardstack.com/card-drop',
+      'Please let us know if you would like to receive our email newsletter with updates about the Cardstack Project.\n\nYou can unsubscribe at any time by clicking the link in the footer of our emails or by contacting us at contact@cardstack.com. For information about our privacy practices, please visit our website.',
+    messageNote:
+      'We use Mailchimp as our marketing platform. By clicking below to subscribe, you acknowledge that your information will be transferred to Mailchimp for processing.',
+    link: {
+      text: "Learn more about Mailchimp's privacy practices here.",
+      url: 'https://support.cardstack.com/card-drop',
```
danibonilha: Shouldn't this reference  `https://mailchimp.com/legal/terms/` ?

cardstack/src/components/Checkbox/Checkbox.tsx
```diff
@@ -32,33 +39,37 @@ export const Checkbox = ({
...
-            )}
-          </Container>
-        </Touchable>
+    <Touchable
+      flexDirection={flexDirection}
+      alignItems="center"
+      onPress={handleCall}
+      disabled={disabled}
+      {...rest}
+    >
```
danibonilha, line 51: I have mixed feelings about passing containerProps forward, as it give less predictability about the component behavior, but I do understand that sometimes it's necessary, for this case though it doesn't seem to be used, since it's receiving the flexDirection.

cardstack/src/components/Checkbox/Checkbox.tsx
```diff
@@ -32,33 +39,37 @@ export const Checkbox = ({
...
-            {selected && (
-              <Icon
-                color={disabled ? 'settingsGrayDark' : 'black'}
-                iconSize="small"
-                name="check"
-              />
-            )}
-          </Container>
-        </Touchable>
+    <Touchable
```
danibonilha, line 46: I'm not sure if the whole line should be wrapped in a `Touchable`, I mean, maybe we should use a `Pressable` so it won't have the opacity animation on the label.

cardstack/src/components/Checkbox/Checkbox.tsx
```diff
@@ -32,33 +39,37 @@ export const Checkbox = ({
...
+    >
+      <Container
+        alignItems="center"
+        backgroundColor={disabled ? 'underlineGray' : 'buttonPrimaryBackground'}
+        borderColor={disabled ? 'transparent' : 'black'}
+        borderRadius={5}
+        borderWidth={1}
+        height={22}
+        justifyContent="center"
+        width={22}
```
danibonilha: nit: we could extract width/height, to a size const

ios/Rainbow.xcodeproj/project.pbxproj
```diff
@@ -792,11 +792,11 @@
 				ASSETCATALOG_COMPILER_OPTIMIZATION = "";
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Rainbow/RainbowRelease.entitlements;
-				CODE_SIGN_IDENTITY = "iPhone Distribution";
-				CODE_SIGN_STYLE = Manual;
+				CODE_SIGN_IDENTITY = "Apple Development";
+				CODE_SIGN_STYLE = Automatic;
```
danibonilha: we should drop this change

cardstack/src/theme/textVariants.ts
```diff
@@ -57,4 +57,12 @@ export const textVariants = {
...
     ...fontFamilyVariants.regular,
   },
+  bannerSmallNote: {
+    fontSize: 10,
+    color: 'secondaryText',
+  },
+  checkboxLabel: {
+    fontSize: 13,
+    ...fontFamilyVariants.regular,
+  },
```
danibonilha: I think it's ok to add this right now, we can circle back on this discussion, but I'm not sure about creating textVariants for single component use, maybe we can start mapping which are the most common sizing and spacings and create a pattern, so we can use 4, 5 variants with most of the component, wdyt ?

cardstack/src/components/Checkbox/Checkbox.tsx
oidoug: I think it's nice to have the label be interactable, makes it easier to toggle the checkbox, so it makes sense to have the opacity animation, but I'm fine without it too, lmk.

cardstack/src/components/Checkbox/Checkbox.tsx
oidoug: Yes, you're right, but it also gives the component a bit more flexibility. I'll remove this one case.

cardstack/src/screens/RequestPrepaidCardScreen/strings.ts
oidoug: Ah right, thanks.

cardstack/src/theme/textVariants.ts
oidoug: IMO, ideally, we would only have a dozen variants for everything, and then these `textVariants` would make sense, but since that's not the case, I agree that it's only adding noise now.

One thing that `textVariants` does better is setting font weights, because setting the weight prop on Text does not use the correct font families, just apply line weight to "regular". But anyway, we may rather fix the weight instead.

danibonilha (approved):
LGTM

PR #818, opened 2022-04-27
[CS-3027]: Fix assets not loading on fresh start and other issues

danibonilha:
### Description
This PR fixes some issues related to importing, deleting and loading a new account 

- Fixes keyboard delayed dismissal while importing an account from fresh state, by handling the dismissOverlay differently for this case
- Resets the navigation state after deleting all the wallets
- Improves the delete wallet code by iterating only once to found accounts,  and using spread to handle the proprieties instead of the delete operator 
- Removes 'visible' property of account, as it seems this was used only for "deleting" the accounts, so instead of deleting  they were hidden, now we truly delete the accounts 
- Fixes the Create account modal padding, as it was covering the status bar 
- Fixes missing deleting keys 
- Fixes assets not loading on fresh start, (the address was not being while creating the wallet, so the first time it ran though initializeWallet, the address was empty, and the wallet was not properly initialized)

I validate most of the scenarios on iOS, like deleting multiple accounts, multiple wallets, initial state, so I do recommend testing ,  especially on Android as I did just small tests.

### Screenshots

Fresh assets import loading and keyboard handling:
Simulator Screen Recording - iPhone SE (3rd generation) - 2022-04-27 at 19 08 19

Create account modal, proper positioning:
image

### DIFF
--- a/src/model/wallet.ts
+++ b/src/model/wallet.ts
@@ -730,6 +730,9 @@ export const createOrImportWallet = async ({
     await saveAllWallets(allWallets);
     logger.sentry('[createWallet] - saveAllWallets');
 
+    await saveAddress(walletAddress);
+    logger.sentry('[createWallet] - saveAddress');
+
     if (walletResult && walletAddress) {
       // bip39 are derived from mnemioc
       const createdWallet =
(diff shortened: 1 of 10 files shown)


src/model/wallet.ts
```diff
@@ -526,7 +525,7 @@ const addAccountsWithTxHistory = async (
     let label = '';
 
     if (discoveredAccount && discoveredWalletId) {
-      if (discoveredAccount.visible) {
+      if (discoveredAccount) {
```
oidoug, line 528: nit: This if is redundant.

oidoug (approved):
One minor nit, otherwise LGTM.

PR #819, opened 2022-04-28
[CS-3775]: Fix and improve sentry logs

danibonilha:
### Description
There was a `safelyStringifyWithFormat` function  which was removing all the logging info for composed sentry logs, with the new version we don't need this, as sentry gives options to serialize the breadcrumbs, so this PR adds those new configs and removes old one, also I added a `skipPoller` flag, because I noticed the fallbackExplorer is being call 3 times on start and never again, which means the poller don't work as expected and it's just adding more "hardwork" on app init, so it's disabled until we decide to fix it. 

### Screenshots

image


lukemelia (approved):
Looks good 👍

oidoug (approved):
LGTM

PR #820, opened 2022-04-29
[CS-3752, CS-3632] Disabling forceful dark mode on android

oidoug:
### Description

Attempt to fix Xiaomi (MIUI) from forcing set dark mode. On tests, it did work on new installs but not on just updating.

Also, I noticed that this forceful dark mode only switches full white and full black colors, setting an off-white color to StyledQRCode will prevent the color switch even if the config change is bypassed.

- [x] Completes #(CS-3752)
- [x] Completes #(CS-3632)


danibonilha (approved):
LGTM

PR #821, opened 2022-05-02
Fix merchant claim 

tintinthong:
Fix merchant claim that returns `invalid character` error 

### Description

The function of `toString` within `bignumber.js` is problematic. If an explicit config is not set, it will return a string representation in exponential notation.

```
const { BigNumber, isBigNumber } = require("bignumber.js");
const web3 = require("web3")
let o = "1000000000000000000000" // 21 decimal places it turns into exponential notation 
o = new BigNumber(o)
console.log(o.toString()) //1e+21
console.log(o.toFixed()) //1000000000000000000000
```

See [here](https://mikemcl.github.io/bignumber.js/#exponential-at). 

- [x] [Completes #(Linear ticket)](https://linear.app/cardstack/issue/CS-3719/invalid-character-revenue-claim#comment-54377b6a)

### DIFF
--- a/cardstack/src/services/merchant/merchant-service.ts
+++ b/cardstack/src/services/merchant/merchant-service.ts
@@ -41,7 +41,7 @@ export const claimMerchantRevenue = async ({
 
     const claimAmount = new BigNumber(Web3.utils.toWei(token.balance.amount))
       .minus(new BigNumber(gasEstimate))
-      .toString();
+      .toFixed();
 
     await revenuePool.claim(
       merchantSafeAddress,
--- a/cardstack/src/services/merchant/merchant-service.ts
+++ b/cardstack/src/services/merchant/merchant-service.ts
@@ -30,7 +30,7 @@ export const claimMerchantRevenue = async ({
     // divide amount by 2 for estimate since we can't estimate the full amount
     // and the amount doesn't affect the gas price
     const claimEstimateAmount = Web3.utils.toWei(
-      new BigNumber(token.balance.amount).div(new BigNumber('2')).toString()
+      new BigNumber(token.balance.amount).div(new BigNumber('2')).toFixed()
     );
 
     const gasEstimate = await revenuePool.claimGasEstimate(
--- a/src/handlers/web3.js
+++ b/src/handlers/web3.js
@@ -169,7 +169,7 @@ export const estimateTransferNFTGas = async (
       params.id
     );
     if (!addPadding) {
-      return contractEstGas.toString();
+      return contractEstGas.toFixed();
     }
 
     return addPaddingToGasEstimate(contractEstGas, paddingFactor, provider);
@@ -245,7 +245,7 @@ export const estimateGasWithPadding = async (
  */
 export const toWei = ether => {
   const result = parseEther(ether);
-  return result.toString();
+  return result.toFixed();
 };
 
 /**
--- a/src/handlers/web3.js
+++ b/src/handlers/web3.js
@@ -169,7 +169,7 @@ export const estimateTransferNFTGas = async (
       params.id
     );
     if (!addPadding) {
-      return contractEstGas.toFixed();
+      return contractEstGas.toString();
     }
 
     return addPaddingToGasEstimate(contractEstGas, paddingFactor, provider);


tintinthong:
I am making a change to `claim` function in revenue pool sdk so as to not require you to supply the balance to claim. If timing appropriate I will include additional changes in this pr. Otherwise, the single line change atm can go in on its own

oidoug:
Should we look for other instances of `toString()` that might cause similar issues?

I see one potential case here: 
https://github.com/cardstack/cardwallet/blob/d2e9292805958b5efcd9870f90f889b01615b5c7/src/handlers/web3.js#L248

tintinthong:
> Should we look for other instances of `toString()` that might cause similar issues?
> 
> I see one potential case here:
> 
> https://github.com/cardstack/cardwallet/blob/d2e9292805958b5efcd9870f90f889b01615b5c7/src/handlers/web3.js#L248



> Should we look for other instances of `toString()` that might cause similar issues?
> 
> I see one potential case here:
> 
> https://github.com/cardstack/cardwallet/blob/d2e9292805958b5efcd9870f90f889b01615b5c7/src/handlers/web3.js#L248

yep. good idea

cardstack/src/services/merchant/merchant-service.ts
```diff
@@ -41,7 +41,7 @@ export const claimMerchantRevenue = async ({
 
     const claimAmount = new BigNumber(Web3.utils.toWei(token.balance.amount))
       .minus(new BigNumber(gasEstimate))
-      .toString();
+      .toFixed();
```
danibonilha, line 45: I'm wondering if we still need to subtract the gasEstimate on this case, can't we do the same as the rewards? By not providing an amount and claiming all available balance ?

src/handlers/web3.js
```diff
@@ -245,7 +245,7 @@ export const estimateGasWithPadding = async (
  */
 export const toWei = ether => {
   const result = parseEther(ether);
-  return result.toString();
+  return result.toFixed();
 };
```
danibonilha: It seems this function is only used [here](https://github.com/cardstack/cardwallet/blob/develop/cardstack/src/helpers/fallbackExplorerHelper.ts#L108), so maybe we can replace it  with ` Web3.utils.toWei` and delete this `toWei` function.

cardstack/src/services/merchant/merchant-service.ts
tintinthong: no, we dont. As long as the number claimed is not more than the revenueBalance, the claim will work. So to actually get a fix on the cardwallet its actually quite straight forward as long as `token.balance.amount`

But, for clarity sake I think its best to not rely on cardwallet to provide the correct balance of the safe. Similar to `withdraw` and `send-tokens`.  https://github.com/cardstack/cardstack/pull/2926. This may seem unnecessary but there seems to be lots of confusion when speicfying balance always.

src/handlers/web3.js
oidoug: I've now checked all instances of `toString` and besides what I pointed out above, we are good. There are lot's of toString coming from BN, but that's hows supposed to be.

@tintinthong , I did what Dani suggested above, removed the function, and refactor its uses. Do you prefer I push a commit to this PR or into a new one?

tintinthong: oh thanks @oidoug feel free to push in your changes into this branch

src/handlers/web3.js
```diff
@@ -245,7 +245,7 @@ export const estimateGasWithPadding = async (
  */
 export const toWei = ether => {
   const result = parseEther(ether);
-  return result.toString();
+  return result.toFixed();
```
tintinthong: I have discovered that this change is fundamentally wrong. Because `ethersproject` uses an alternative bignumber instance of `bignumber.js`. the method `toFixed` is non-existant within the `ethersproject` instance. daym must becareful of this esp inside js

oidoug: Yes, I found out the same after trying to switch some toString to toFixed, it was returning undefined.
I also found out that `toFixed` from `number` type will also return in scientific notation.

tintinthong: Anyway your changes have handled this already.

src/handlers/web3.js
```diff
@@ -169,7 +169,7 @@ export const estimateTransferNFTGas = async (
       params.id
     );
     if (!addPadding) {
-      return contractEstGas.toFixed();
+      return contractEstGas.toString();
```
tintinthong, line 172: @oidoug @danibonilha so I dont think there is any issues here so I revereted. This uses ethersproject BigNumber

danibonilha:
@tintinthong does it make sense to close this PR in favor of #822 ?

tintinthong:
closing pr in favor of 822

PR #822, opened 2022-05-02
Cleanup merchant claim

tintinthong:
### Description

`revenuePool.claim` api has been changed to not needing the user to specify amount if they want to claim all the revenue of the merchant. 

This avoids faulty use of cardwallet which finds out `amount` and deducts by `gasEstimate`. There is no need for this.

- the api changes [here](https://github.com/cardstack/cardstack/pull/2926)
- bumped up sdk

This change only assumes claiming of a non-zero balance (as I believe it was preeviously). If the balance is very very smol, there may be a failed error in the sdk. Im not too sure how to completely handle this in the wallet.  

### Checklist

- [x] no ui changes


danibonilha (approved):
LGTM! Thank you!

PR #824, opened 2022-05-02
[CS-3813]: Use ethers signer for sdk requests

danibonilha:
### Description
This PR adds the ethers signer to the sdk calls, instead of using the HDProvider, to hopefully mitigate the signing errors, if this proves to be successful we can remove the HDProvider, for now there's a todo and some things are disabled, also updated the `useWallets` hook to return more useful and formatted information, we can however think about adding a new hook to handle those things, but I'd wait for the new wallet/account structure in order to be worth it. There are no screenshots for this, probably a smoke test is needed, I tested paying merchant, sending token from safe to EOA, merchant claim.

Note: In order to remove the HDProvider we also need to change the getHubToken, so probably makes sense to first migrate all auth hub apis, then remove the Provider.


cardstack/src/models/ethers-wallet.ts
```diff
@@ -0,0 +1,49 @@
...
+    const derivedNode = hdNode.derivePath(`${DEFAULT_HD_PATH}/${accountIndex}`);
+
+    return new Wallet(derivedNode);
+  } catch (e) {
+    logger.sentry('Error getting ethersWallet' + e);
+  }
+};
+
+export const getWeb3ProviderWithEthSigner = (params?: EthersSignerParams) =>
+  Promise.all([Web3Instance.get(), getEthersWallet(params)]);
```
oidoug, line 49: Pretty clean, good job.

cardstack/src/models/hd-provider.ts
```diff
@@ -8,6 +8,7 @@ import logger from 'logger';
 import Web3Instance from './web3-instance';
 import Web3WsProvider from './web3-provider';
 
+// TODO: Remove HDProver once ethers signer proves to be better
```
oidoug:
```suggestion
// TODO: Remove HDProvider once ethers signer proves to be better
```

cardstack/src/screens/PayMerchant/usePayMerchant.ts
```diff
@@ -83,21 +79,19 @@ const usePayMerchantRequest = ({
     });
 
     payMerchant({
-      walletId: selectedWallet.id,
-      network: qrCodeNetwork as Network,
+      signerParams,
```
oidoug, line 82: So just to clarify if I understand the change correctly. Does the SDK support a signer from which it derives the network?

cardstack/src/services/merchant/merchant-api.ts
```diff
@@ -18,7 +18,7 @@ const merchantApi = safesApi.injectEndpoints({
           params,
           {
             errorLogMessage: 'Error claiming merchant revenue',
-            resetHdProvider: true,
+            resetHdProvider: false,
```
oidoug, line 21: nit: Is the `options?.resetHdProvider` in `query-promise-wrapper` still needed after these changes? Looks like we can remove it along with these options here.

cardstack/src/services/prepaid-cards/prepaid-card-api.ts
```diff
@@ -41,7 +41,7 @@ const prepaidCardApi = safesApi.injectEndpoints({
           PrepaidCardPayMerchantQueryParams
         >(payMerchant, params, {
           errorLogMessage: 'Error while paying merchant',
-          resetHdProvider: true,
+          resetHdProvider: false,
```
oidoug, line 44: nit: same as above.

cardstack/src/services/rewards-center/rewards-center-api.ts
```diff
@@ -94,7 +94,7 @@ const rewardsApi = safesApi.injectEndpoints({
           RewardsRegisterMutationParams
         >(registerToRewardProgram, params, {
           errorLogMessage: 'Error while registering to reward program',
-          resetHdProvider: true,
+          resetHdProvider: false,
```
oidoug, line 97: nit: same as above

src/hooks/useWallets.ts
```diff
@@ -33,14 +40,16 @@ export default function useWallets() {
     wallets,
   } = useSelector(walletSelector);
 
-  const walletReady = useSelector(state => state.appState.walletReady);
+  const walletReady = useRainbowSelector(state => state.appState.walletReady);
+
+  const { accountAddress } = useAccountSettings();
```
oidoug, line 45: Good idea to have the address here, nice refactor.

oidoug (approved):
Nice refactor! Hopefully the ether signer works better. Having to reset HDProvider was always a bit odd.

cardstack/src/screens/PayMerchant/usePayMerchant.ts
danibonilha: The web3 provider already contains the network information, in order to do the tx it will still use the web3, but to sign it instead of using web3 signer which can be a pain, it will use the ethers wallet one.

cardstack/src/services/merchant/merchant-api.ts
danibonilha: We don't, but I'd rather remove all hdProvider references after the migrating is successful

PR #825, opened 2022-05-03
[Refactor] Removes toWei function in web3

oidoug:
### Description

Removes function that was being used once in favor of `Web3.utils.toWei`, it also avoids issues with Wei toString conversions.


cardstack/src/helpers/fallbackExplorerHelper.ts
```diff
@@ -105,7 +105,7 @@ export const addPriceByCoingeckoIdOrOracle = async ({
     try {
       const priceUnit = await getNativeBalanceFromOracle({
         ...tokenInfo,
-        balance: toWei('1'),
```
tintinthong, line 108: Yea this was probably failing the test on my pr.

tintinthong (approved):

PR #826, opened 2022-05-04
[CS-3818] UX improvements on Asset Lists and assorted screens on Tab bar

oidoug:
### Description

This PR improves the loading state of assets, occurring especially on Android, by not updating the loading state on `DATA_UPDATE_ASSETS` reducer and by not calling dataLoadState twice if not on layer 1.

It also fixes some general screen paddings that were wrong after the switch to tabbar (Activity, Wallet, MerchantContent).

- [x] Completes #(CS-3818)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/166613041-86e580fb-25a1-4a72-8858-1fa66a522712.mp4

### DIFF
--- a/src/redux/data.js
+++ b/src/redux/data.js
@@ -98,7 +98,6 @@ export const DATA_UPDATE_PREPAIDCARDS = 'data/DATA_UPDATE_PREPAIDCARDS';
 
 // -- Actions ---------------------------------------- //
 export const dataLoadState = () => async (dispatch, getState) => {
-  console.log(':::dataLoadState');
   const { accountAddress, network } = getState().settings;
   try {
     const assetPricesFromUniswap = await getAssetPricesFromUniswap(
@@ -157,7 +156,6 @@ export const dataResetState = () => (dispatch, getState) => {
 };
 
 export const dataUpdateAssets = assets => (dispatch, getState) => {
-  console.log(':::dataUpdateAssets');
   const { accountAddress, network } = getState().settings;
   if (assets.length) {
     saveAssets(assets, accountAddress, network);
@@ -254,7 +252,6 @@ export const addressAssetsReceived = (
   change = false,
   removed = false
 ) => async (dispatch, getState) => {
-  console.log(':::addressAssetsReceived');
   const isValidMeta = dispatch(checkMeta(message));
   if (!isValidMeta) return;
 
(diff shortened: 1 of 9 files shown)


src/redux/data.js
```diff
@@ -733,7 +732,7 @@ export default (state = INITIAL_STATE, action) => {
         draft.isLoadingAssets = false;
         break;
       case DATA_LOAD_ASSETS_FAILURE:
-        draft.isLoadingAssets = true;
```
danibonilha, line 736: omg 😆

src/redux/data.js
```diff
@@ -698,7 +698,6 @@ export default (state = INITIAL_STATE, action) => {
         break;
       case DATA_UPDATE_ASSETS:
         draft.assets = action.payload;
-        draft.isLoadingAssets = false;
```
danibonilha: We still need this one though, after the fallbackExplorer finishes updating, it  calls `addressAssetsReceived` which dispatches this `DATA_UPDATE_ASSETS` if we do not set to false, Etherum net will be on a infinite load

danibonilha (commented):
Overall it's looking good, just need to revert the ` draft.isLoadingAssets = false;`  removal of `DATA_UPDATE_ASSETS` and it should be good to go

src/redux/data.js
oidoug: I see 🤔 I removed it because it also causes double loadings. It sets loading to false before dataLoadState runs.
So I think I'll split these loading status, since it doesn't refer to the same loading flow.

danibonilha: Yeah I guess what makes this so confusing is that this DATA_UPDATE_ASSETS is called for fallbackExplorer, and it actually  fetched the  data while `DATA_LOAD_ASSETS_SUCCESS` is related to retrieving the data from the storage,  and they both use the same loading state

oidoug:
So, `dataLoadState` is being called after fallbackExplorer finishes because it's also being called to populate the local storage at the end of `fetchSafes` (in gnosis-services) call.

Thus, setting isLoadingAssets in dataLoadState to true is not very important at that point.

Removing it from the redux will make it all more confusing, ideally, we want to refactor redux data so it doesn't need to handle different safe info.

danibonilha (approved):
LGTM

PR #827, opened 2022-05-04
[CS-3805]:  Use exchange-rates api instead of fixer

danibonilha:
### Description

This PR adds the exchange-rate path to the hubAPi which automatically controls the cache, there are some small improvements how we use conversionRates, I tried to remove everything at once from redux etc, but it's proven to be challenge so I'll be splitting into smaller PRs to be easier to review and test it. Fo now it's just the new api, and it sets on old redux state, upcoming PRs will make sure that we don't need need the old state anymore.
Also as an improvement the payMerchant flow will be changed a little, so we avoid the kind of issue were the nativeCurrency is EUR but the input defaults to USD.
The hub will cache the data for 15 min, caching on the app we are starting with 10 min, but we can evaluate if it needs less time
### Screenshots

Simulator Screen Recording - iPhone SE (3rd generation) - 2022-05-04 at 13 16 35

### DIFF
--- a/cardstack/src/services/exchange-rate-service.ts
+++ b/cardstack/src/services/exchange-rate-service.ts
@@ -1,51 +1,8 @@
-import {
-  getSDK,
-  nativeCurrencies,
-  NativeCurrency,
-} from '@cardstack/cardpay-sdk';
-import { FIXER_API_KEY } from 'react-native-dotenv';
+import { getSDK, NativeCurrency } from '@cardstack/cardpay-sdk';
 
 import Web3Instance from '@cardstack/models/web3-instance';
-import { CurrencyConversionRates } from '@cardstack/types';
 
-import { logger } from '@rainbow-me/utils';
-
-// Currency conversions
-const CurrencyConsts = {
-  fixerApi: 'https://data.fixer.io/api',
-  usdToSpendRate: 100,
-  currencySymbols: Object.keys(nativeCurrencies),
-};
-
-let cachedCurrencyConversionRates: CurrencyConversionRates | null = null;
-
-export const getCurrencyConversionsRates = async () => {
-  const defaults = CurrencyConsts.currencySymbols.reduce(
-    (accum, symbol) => ({
-      ...accum,
-      [symbol]: 0,
-    }),
-    {}
-  );
-
-  try {
-    const request = await fetch(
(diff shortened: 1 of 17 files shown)


cardstack/src/services/exchange-rate-service.ts
```diff
@@ -1,69 +1,39 @@
...
 
-  try {
-    const request = await fetch(
-      `${CurrencyConsts.fixerApi}/latest?access_key=${FIXER_API_KEY}&base=USD&symbols=${CurrencyConsts.currencySymbols}`
-    );
+  return layerTwoOracle;
+};
 
-    const data = await request.json();
+export const getValueInNativeCurrency = async (
```
oidoug, line 14: Nice separation.

cardstack/src/services/hub/hub-service.ts
```diff
@@ -99,3 +100,19 @@ export const checkHubAuth = async ({
...
   return isAuthenticated;
 };
+
+// External Queries
+
+const cacheExpiration = {
+  tenMinutes: 60 * 10,
+};
+
+export const getExchangeRatesQuery = () => {
```
oidoug, line 110: Question: This is called durant explorer's init. But that is not "reactive", so it will only pull this up once. How will this function be called again, so data is updated?
Or is it something coming up in the next PR?

src/redux/fallbackExplorer.js
```diff
@@ -398,6 +398,11 @@ export const fallbackExplorerInit = () => async (dispatch, getState) => {
   const { accountAddress, network } = getState().settings;
   const { latestTxBlockNumber, assets } = getState().fallbackExplorer;
   const coingeckoCoins = getState().coingecko.coins;
+
+  const { data: conversionsRates } = await getExchangeRatesQuery();
```
oidoug, line 402: Follow up question: This will be called once, right?

oidoug (approved):
Just a couple of questions. Nice upgrade!

cardstack/src/services/hub/hub-service.ts
danibonilha: it works just like the other rtk queries, with the forceRefresh, every time this data is asked it will refetch if it's not within the cache expiration time, now it may not make much sense, but in the upcoming PR, you'll see this will be called on multiple places

src/redux/fallbackExplorer.js
danibonilha: in this instance yes, but if it already has a cache it will retrieve the information right away

PR #828, opened 2022-05-05
[CS-3825] Upgraded Flipper and using default setup for ios package

oidoug:
### Description

Double-conversion was set to use an old version, because new Flipper versions need to strip unused code in dev builds, as in prod.

- [x] Completes #(CS-3825)


danibonilha (approved):
LGTM, nice discovery!

PR #829, opened 2022-05-05
[CS-3806]: Use same func to convert currencies and skip if it's USD

danibonilha:
### Description
This PR aims to make the app more resilient on conversions in case fixer is out, the ideia is to use the same convert function across the app instead of passing the rates on multiple components and each making a diff calculation, in order to do that `convertSpendToNativeBalance` has been modified to retrieve the rates when needed, giving that now it's an async function, an `useSpendToNativeDisplay` hook was created to be used inside the components. Besides that there are some cleanups, some loading states were added and `useAccountSettings` was migrate to .ts, the final step to get rid of the old conversions is to update the pay merchant flow which I'll be doing on a following PR(which will be based on this branch). I realized based on the amount of files I touched that we still have improvements  to do, but I'm focusing on the conversion stuff.


cardstack/src/components/ChoosePrepaidCard/ChoosePrepaidCard.tsx
```diff
@@ -61,49 +61,32 @@ export const ChoosePrepaidCard = memo(
...
+      network,
+    ]);
 
-    const onSelect = useCallback(
-      (item: PrepaidCardType) => {
-        onSelectPrepaidCard(item);
-      },
-      [onSelectPrepaidCard]
-    );
+    const onSelect = useCallback(onSelectPrepaidCard, [onSelectPrepaidCard]);
```
oidoug, line 72: onSelectPrepaidCard is already a callback, this seems redundant.

cardstack/src/components/TransactionConfirmationSheet/displays/PrepaidCard/IssuePrepaidCardDisplay.tsx
```diff
@@ -172,17 +155,9 @@ const LoadSection = ({ data }: { data: IssuePrepaidCardDecodedData }) => {
...
-
-  const { nativeBalanceDisplay } = convertSpendForBalanceDisplay(
-    '0',
-    nativeCurrency,
-    currencyConversionRates,
-    true
-  );
+  const { nativeBalanceDisplay } = useSpendToNativeDisplay({
+    spendAmount: 0,
+  });
```
oidoug, line 160: This is so much cleaner 👏

oidoug (approved):
LGTM! Some parts got a nice trim, pretty good, a lot clearer now.
Only one minor find.

PR #831, opened 2022-05-06
[CS-3757] Updates Notice banner with new style

oidoug:
### Description

PR updates Notice and ServiceStatusNotice components with new styles for when there's an onPress action present.

It also changes the radius to match the ones on our designs.

- [x] Completes #(CS-3757)

### Checklist

### Screenshots

Screenshot

Screenshot


danibonilha (approved):
LGTM, although the white text on the red background seems a bit odd

oidoug:
Yes, black text looks better, but it's a bit less visible.
We only have designs for the yellow warning banner, and we're only using that for now.

Screenshot

PR #833, opened 2022-05-09
[CS-3729] Fixes gesture transparency issue on android

oidoug:
### Description

PR adds a workaround for an Android issue with gestures being relayed on modals that are fully transparent. It adds a tiny bit of black to the background so the view can capture touches.

- [x] Completes #(CS-3729)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/167418610-050d5911-9f36-488b-a5f8-5452e858f0f3.mp4

### DIFF
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -74,7 +74,6 @@ const nativeStackiOSLoadingConfig = {
   transitionDuration: 0,
   onWillDismiss: null,
   dismissable: false,
-  gesturedEnabled: false,
 };
 
 // Shareable component,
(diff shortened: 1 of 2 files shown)


cardstack/src/navigation/screens.ts
```diff
@@ -74,7 +74,6 @@ const nativeStackiOSLoadingConfig = {
   transitionDuration: 0,
   onWillDismiss: null,
   dismissable: false,
-  gesturedEnabled: false,
 };
```
danibonilha: nit: We can delete this `nativeStackiOSLoadingConfig`  since the new nav doesn't use the Native navigator anymore and leave the `LoadingOverlay` just on the mainScreens

danibonilha (approved):
LGTM, just a nit we can take leverage

PR #834, opened 2022-05-10
[CS-3843] Ethersproject library refactor in favor of ethers

oidoug:
### Description

This PR refactors uses of `ethersprojects` in favor of `ethers` parent module so we don't get submodule's version mismatch issues anymore. `ethers` installed version is now `5.6.4`, same as SDK's.

- [x] Completes #(CS-3843)


danibonilha (approved):
LGTM, great cleanup!

lukemelia (approved):

PR #835, opened 2022-05-10
[CS-3817]: Use authenticated rates and fix token refresh

danibonilha:
### Description
This PR adds the token to the exchange rate api call that was merged [here](https://github.com/cardstack/cardstack/pull/2944), and I noticed the token was never being removed, bc it's added with an id and a key, and the removal was key only.

## UPDATE
- Removes SPD filtering, depends on [PR on the sdk](https://github.com/cardstack/cardstack/pull/2949)


cardstack/src/services/hub/hub-api.ts
```diff
@@ -63,14 +63,7 @@ export const hubApi = createApi({
     }),
     getExchangeRates: builder.query<Record<NativeCurrency, number>, void>({
       query: () => routes.exchangeRates,
-      extraOptions: { authenticate: false },
-      transformResponse: ({
-        data: {
-          attributes: { rates },
-        },
-      }) => {
-        return { ...rates, SPD: 100 };
```
lukemelia, line 72: Why was { SPD: 100 } there before, and why was it removed?

danibonilha: We don't use spend as currency on the wallet anymore, and every conversion of spend we need, we use sdk methods to be more reliable.

lukemelia: 👍

oidoug (approved):
LGTM

PR #836, opened 2022-05-10
[CS-3428]: Fix rewards crash on mainnet

danibonilha:
### Description

Skips layer2 calls if it's on Layer1

Simulator Screen Recording - iPhone SE (3rd generation) - 2022-05-10 at 16 02 00


lukemelia (approved):
I think this is fine for a fix. It would be good to find a model for layer1/2 conditional checks that is more about asking if the current network has a particular feature/capability rather than sprinkling these layer1/2 checks everywhere

PR #839, opened 2022-05-12
[Refactor]: Remove tabFeature flag and delete dead code 

danibonilha:
### Description

This PR removes the tabBar context and makes the tab as default, by committing to the new nav, multiple components and files were removed.
- Migrates `react-navigation/core` to `react-navigation/native`
- Deletes old Profile, wallet, addCash, exchange screens along side components  those screens used
- Migrates useful navigation utils to cardstack side 
- Simplify routes and delete old navigators

There's still code to be removed, but it's best to split into smaller PRs so it's easier to review.

This will only be merged after cutting this sprint release


src/components/sheet/SlackSheet.js
```diff
@@ -16,7 +17,6 @@ import SheetHandleFixedToTop, {
 } from './SheetHandleFixedToTop';
 import { Container } from '@cardstack/components';
 import { useDimensions } from '@rainbow-me/hooks';
-import { useNavigation } from '@rainbow-me/navigation';
 import { position } from '@rainbow-me/styles';
 
 const { event } = Animated;
```
oidoug, line 22: I just added a ticket for removing SlackSheet, I don't think it's needed anymore now that we have the more generic Sheet:

https://linear.app/cardstack/issue/CS-3887/[debt]-refactor-out-slacksheet-in-favor-of-newer-sheet-component

oidoug (approved):
Amazing work ✨
Massive!

src/components/sheet/SlackSheet.js
danibonilha: Thanks!

PR #840, opened 2022-05-13
[CS-3762] Firebase Remote Config setup

oidoug:
### Description

This PR adds Firebase's Remote Config packages and Analytics (without Ad ID tracking) and implementation for fetching a few config properties.

- New remote-config-service that handles fetching firebase's remote configs;
- Adds hook for tracking remote config loading state;
- Refactors App's requirements check to a sep. component;
- Removes firebase function calls since they are no longer needed.

- [x] Completes #(CS-3762)


package.json
```diff
@@ -61,8 +61,10 @@
     "@react-native-community/clipboard": "git+https://github.com/gnardini/clipboard#01d1c13",
     "@react-native-community/masked-view": "^0.1.5",
     "@react-native-community/netinfo": "^6.1.0",
+    "@react-native-firebase/analytics": "^14.9.3",
```
danibonilha, line 64: is the analytics package required to have remote config working ?

package.json
```diff
@@ -61,8 +61,10 @@
...
     "@react-native-firebase/app": "^12.1.0",
     "@react-native-firebase/messaging": "^12.1.0",
+    "@react-native-firebase/remote-config": "^14.9.3",
     "@react-navigation/bottom-tabs": "^5.x",
```
danibonilha, line 68: We need to  follow the same version of the other packages, to avoid these warnings

image

cardstack/src/services/remote-config/remote-config-service.ts
```diff
@@ -0,0 +1,33 @@
+import remoteConfig from '@react-native-firebase/remote-config';
+
+import remoteConfigDefaults from './remote_config_defaults.json';
+
+const CACHE_INTERVAL_MILLIS = 3600; // 3600000ms == 1 hr.
+
+export enum ConfigKey {
+  requiredMinimumVersion = 'requiredMinimumVersion',
+  maintenanceStatus = 'maintenanceStatus',
+  appReviewFeature = 'appReviewFeature',
```
danibonilha: We can remove this reviewFeature, it's a rainbow leftover

cardstack/src/services/remote-config/remote-config-service.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export const fetchRemoteConfigs = async () => {
+  await remoteConfig().setConfigSettings({
+    minimumFetchIntervalMillis: CACHE_INTERVAL_MILLIS,
```
danibonilha, line 15: one hour is still a lot to be refetching something that will hardly change, but also too much for development, but we can set a developer flag to update the data faster, something like an easter egg, let's discuss it later.

cardstack/src/services/remote-config/remote_config_defaults.json
```diff
@@ -0,0 +1,5 @@
+{
+  "requiredMinimumVersion": "1.1.1",
+  "maintenanceStatus": "{\"active\":false,\"message\":\"Card Wallet is going through scheduled maintenance, please try again later.\"}",
```
danibonilha: This is not so intuitive while updating in the console for example, let's make it a single key, `maintenanceStatusMessage` idk, and if the message exists we render the component OR  we split into two diff keys, being, `maintenanceActive`, and `maintenanceStatusMessage`

src/App.js
```diff
@@ -39,24 +37,22 @@ import {
...
-import useHideSplashScreen from './hooks/useHideSplashScreen';
 import { loadAddress } from './model/wallet';
 import store, { persistor } from './redux/store';
 import { walletConnectLoadState } from './redux/walletconnect';
-import MaintenanceMode from './screens/MaintenanceMode';
+import { AppRequirementsCheck } from '@cardstack/components';
 import ErrorBoundary from '@cardstack/components/ErrorBoundary/ErrorBoundary';
-import { MinimumVersion } from '@cardstack/components/MinimumVersion';
 import { apolloClient } from '@cardstack/graphql/apollo-client';
+// import { useInitRemoteConfigs } from '@cardstack/hooks';
```
danibonilha: leftover ?

cardstack/src/components/AppRequirementsCheck/AppRequirementsCheck.tsx
```diff
@@ -0,0 +1,46 @@
...
+    return compareVersions(minVersion, appVersion) > 0;
+  }, [isReady]);
+
+  if (maintenance?.active) {
+    return <MaintenanceMode message={maintenance.message} />;
+  }
+
+  if (forceUpdate) {
+    return <MinimumVersion />;
+  }
```
danibonilha: not hiding the splashScreen on these states, still works as expected ? If so, is the hideSplashScreen really doing something at all?

src/App.js
```diff
@@ -245,53 +242,14 @@ class App extends Component {
   );
 }
 
-const CheckSystemReqs = ({ children }) => {
+const HideSplashScreen = () => {
```
danibonilha: it seems a little odd to have this as a component, if this is to handle the AppRequirements calls, it seems to make more sense to live within the AppReqs component

cardstack/src/services/remote-config/remote-config-service.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export const getConfigBoolean = (key: ConfigKey) =>
+  remoteConfig().getValue(key).asBoolean();
+
+export const getConfigString = (key: ConfigKey) =>
+  remoteConfig().getValue(key).asString();
+
+export const getConfigJSON = (key: ConfigKey) =>
+  JSON.parse(remoteConfig().getValue(key).asString());
```
danibonilha: supernit: if this was a remoteConfig model, we could export these functions within it, and use it like. `remoteConfig.getString` so it's clear that it's related to remoteConfig, cause right now, getConfigJson is too abstract, or maybe, `getRemoteConfigString`

cardstack/src/services/remote-config/remote-config-service.ts
```diff
@@ -0,0 +1,33 @@
...
+  await remoteConfig().setConfigSettings({
+    minimumFetchIntervalMillis: CACHE_INTERVAL_MILLIS,
+  });
+
+  await remoteConfig().setDefaults(remoteConfigDefaults);
+  const fetchedRemotely = await remoteConfig().fetchAndActivate();
+
+  return {
+    fetchedRemotely,
+  };
```
danibonilha: What does it return ? it doesn't seem that the return of this function is being used at all, if the retrieval will always happen within the below functions, this one could be called something like, `loadRemoteConfig` or `setupRemoteConfig` which says that something is happening, but there's not actually values to be used

cardstack/src/hooks/remote-configs/useRemoteConfigs.ts
```diff
@@ -0,0 +1,30 @@
+import remoteConfig from '@react-native-firebase/remote-config';
+import { useEffect, useCallback } from 'react';
+
+import { useBooleanState } from '@cardstack/hooks';
+import { fetchRemoteConfigs } from '@cardstack/services';
+
+import logger from 'logger';
+
+export const useRemoteConfigs = () => {
```
danibonilha: this name might be misleading as it gives the impression that we can use the remote config values, when it's actually just initializing it

cardstack/src/hooks/remote-configs/useRemoteConfigs.ts
```diff
@@ -0,0 +1,30 @@
...
+  const [isReady, setIsReady] = useBooleanState();
+
+  const initRemoteConfigs = useCallback(async () => {
+    try {
+      await fetchRemoteConfigs();
+    } catch (error) {
+      logger.sentry("Couldn't fetch remote config", error);
+    } finally {
+      console.log(remoteConfig().getAll());
```
danibonilha: whats the console.log for ?

danibonilha (commented):
There are somethings that need to be addressed, i'd also have one general suggestion for future tasks like this, to split into smaller chunks, for example, one, setting the library, and creating the models and another for implementing and refactoring.

package.json
oidoug: Yes, unfortunately. But I found in the docs a secondary analytics package that doesn't require the user to have the AD ID enabled, so it's pretty anonymous.

package.json
oidoug: Ah sorry, didn't notice the warnings.

cardstack/src/hooks/remote-configs/useRemoteConfigs.ts
oidoug: it's a leftover, not needed, I'll remove it.

cardstack/src/services/remote-config/remote_config_defaults.json
oidoug: There's a specific type for JSON properties and the editor validates the input, so it's safe to edit it. But I agree it's more convoluted.

I don't mind splitting into two though, so let's do that then.

cardstack/src/components/AppRequirementsCheck/AppRequirementsCheck.tsx
oidoug: hideSplashScreen was being called regardless of the state, we need to hide the splash if we want anything to display.
So I prefer to have it outside the requirements check.

src/App.js
oidoug: As said above, it's not really part of the AppReqs. It's in a sep. component because our App is a class and can't call our `useHideSplashScreen` hook.

I don't mind doing it differently though.

cardstack/src/services/remote-config/remote-config-service.ts
oidoug: Default is 12 hrs. I think we can have let it be a few minutes in DEV and a few hours in prod.
Mind that it will not automatically refresh, that's the time the cache is kept as valid.

src/App.js
oidoug: Ah, never mind, I see now what you meant. I'll move the call to be handled by the requirement component again. Strange that it was working.

package.json
oidoug: Updated all firebase packages, I looked through the docs, and APIs didn't change.

package.json
oidoug: I updated all firebase packages to the latest. I looked through the docs and APIs are the same.

cardstack/src/services/remote-config/remote-config-service.ts
oidoug: Set to 1min in DEV and 6hrs in prod.

src/components/settings-menu/SettingsSection.js
```diff
@@ -70,12 +69,6 @@ export default function SettingsSection({
 
   const onSendFeedback = useSendFeedback();
 
-  const [isReviewEnabled, setReviewEnabled] = useState(false);
-
-  useEffect(() => {
-    setReviewFeature();
-  }, []);
-
   const onPressReview = useCallback(async () => {
```
danibonilha: We can remove this, as it's not being used anymore

danibonilha (approved):
Nice update! Good to go after deleting  the `onPressReview`

src/components/settings-menu/SettingsSection.js
oidoug: Ah, yeah, I merged with dev and that got obsolete.

PR #842, opened 2022-05-16
[CS-3836]: Remove dead code 

danibonilha:
### Description

It cleans old functionality (partially like swap, uniswap, savings), add cash, rainbow context, old components, old  coing hidden/pinned gesture and animations from old navigation. Migrates stuff that used context unnecessarily, like `InvalidPaste`, fixes the wallet connect non navigating issue, updates firebase pod deps.
There are lot's of things still to be improved, but the goal of this one, is to remove as much as possible, without breaking anything else.


src/components/send/SendAssetList.js
```diff
@@ -54,47 +54,15 @@ export default class SendAssetList extends React.Component {
...
-      smallBalances = assets.pop();
-      shitcoins = smallBalances.assets;
-    }
+    const { allAssets, nativeCurrency, network, savings, collectibles } = props;
 
-    if (assets[assets.length - 1].coinDivider) {
-      assets.pop(); // removes not needed coin divider
-    }
+    const { assets } = buildCoinsList(allAssets, nativeCurrency, true);
+    let shitcoins = [];
```
oidoug, line 60: nit: Is `shitcoins` referring to altcoins? It would be nice to rename them.

src/hooks/useInvalidPaste.ts
```diff
@@ -0,0 +1,30 @@
...
+  }, [isInvalidPaste, unsetInvalid, startTimeout, stopTimeout]);
+
+  // 🚪️ Reset isInvalidPaste when we leave the screen
+  useEffect(() => () => unsetInvalid(), [unsetInvalid]);
+
+  return {
+    isInvalidPaste,
+    onInvalidPaste,
+  };
+}
```
oidoug: nit: since you converted to TS, isn't it better to move to cardstack's hooks folder?

oidoug (approved):
Amazing! I'm impressed by how much stuff you removed, excellent job!
I left a minor nit.

src/components/send/SendAssetList.js
danibonilha: I've actually have no idea what are `shitcoins`,  we might not even need it , but the conditions on this sendAssetList are pretty complex, so I'll add a ticket to refactor the way the coins list is built and we can simplify and remove this.

PR #843, opened 2022-05-17
[Fix] Padding issue when Notice not visible.

oidoug:
### Description

On the last Service Notice update, the padding was moved and it was always present even when the banner wasn't visible.

### Checklist

### Screenshots

Screenshot


paulinhapenedo (approved):
Makes sense to me by looking at the code. I mean, if the item can be hidden (assuming that by the `isVisible` prop), then the padding for it should be put in the component itself, not in its container.

danibonilha (approved):

PR #844, opened 2022-05-17
[CS-3880] Beta test remote config and feature prepaid card drop config value

oidoug:
### Description

This PR adds a way for setting a user property that enables BETA configurations in Remote Config and adds a new feature flag for prepaid card drop.

ATM the flag is set to true for when version 1.1.3

- [x] Completes #(CS-3880)

### Checklist

### Screenshots

How to enable BETA in Firebase console.

https://user-images.githubusercontent.com/129619/168864601-4be6ab3e-2bc4-4b3d-8d89-b3a060539150.mov


src/components/AppVersionStamp.tsx
```diff
@@ -2,27 +2,52 @@ import React, { memo, useCallback, useRef } from 'react';
...
+
     numberOfTaps.current++;
 
     if (numberOfTaps.current === parseInt(VERSION_TAP_COUNT)) {
-      // trigger some feature
+      numberOfTaps.current = 0;
+      try {
+        await setUserAccessType(UserAccessType.BETA);
+        await forceFetch();
+        setToast({ label: `User now part of ${UserAccessType.BETA} access.` });
```
danibonilha: Nice extraction to create a reusable bottomToast, I'm not so sure though about this `setToast` method, couldn't it follow the same pattern as the copyToast one ? 

```tsx
const { ToastComponent, showToast: showBetaToast } = useToast({ label: `User now part of ${UserAccessType.BETA} access.`});

// then 

await forceFetch();
showBetaToast();
```

Or even better, we could make the `useToast` consistent with not receiving any params, and add a `showToast` function that receives the label param, but this would require to update the useCopyToast too, if there's any doubts DM me

danibonilha (approved):
Overall it looks good, personally I think the `setToast`  might be a unnecessary, but if there's a good reason to keep it, it's okay

cardstack/src/hooks/useToast.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+interface ToastProps {
+  label?: string;
+  duration?: number;
+}
+
+const toastBottomPosition = screenHeight * 0.1;
+const defaultDuration = 2000;
+
+export const useToast = ({
```
danibonilha: nit: we could consider renaming this to `useBottomToast`  since it's hardcoded to come from the bottom

src/components/AppVersionStamp.tsx
oidoug: I wanted it to be flexible to use both ways, but I see it's a bit too much. So it's better to set it in the showToast call, otherwise, some components using it would need to keep a local state for the params, to update the toast.

PR #845, opened 2022-05-18
[CS-2378] Remove fixed face-id icon in favor of BiometryType icon

paulinhapenedo:
### Description

This PR removes the hardcoded `face-id` icon from both the `TransactionConfirmationSheet` component and the SecretDisplay sheet in favor of using the icons mapped by the `BiometricType` enum. Devices with no FaceID, like the iPhone SE in the screenshot, show the touch ID icon. While in the area, I moved the button text into the `strings` file.

There's a known issue with `react-native-keychain` where, in some devices with facial recognition but no actual FaceID capability, the library returns `face-id` as supported biometric type (tested this with a Moto G8: the device doesn't use face as auth but has face recognition capabilities). Will be dealing with that in another PR since it will probably involve trying to update both `react-native-keychain` and `react-native-device-info` or maybe use [Expo Local Authentication](https://github.com/expo/expo/tree/main/packages/expo-local-authentication). Let me know if y'all have tried this before and if it's a path worth exploring :)

- [x] Completes #CS-2378

### Checklist

### Screenshots

Screenshot

### DIFF
--- a/cardstack/src/components/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
+++ b/cardstack/src/components/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
@@ -12,6 +12,7 @@ import {
   Text,
   Touchable,
 } from '@cardstack/components';
+import { useBiometricIconProps } from '@cardstack/hooks/useBiometricIconProps';
 import { TransactionConfirmationData } from '@cardstack/types';
 import { layoutEasingAnimation } from '@cardstack/utils';
 
@@ -19,6 +20,7 @@ import {
   DisplayInformation,
   transactionTypeMap,
 } from './displays/DisplayInformation';
+import { strings } from './strings';
 
 export interface TransactionConfirmationDisplayProps {
   dappUrl?: string;
@@ -124,6 +126,8 @@ const SheetFooter = ({
   onCancel,
   onConfirmLoading = false,
 }: TransactionConfirmationDisplayProps) => {
+  const iconProps = useBiometricIconProps();
+
   return (
     <Container
       width="100%"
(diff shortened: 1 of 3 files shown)


cardstack/src/components/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
```diff
@@ -141,15 +146,15 @@ const SheetFooter = ({
...
         </Button>
         <Button
           loading={onConfirmLoading}
           variant="small"
           onPress={onConfirm}
-          iconProps={{ name: 'face-id' }}
+          iconProps={{ name: biometryIconName }}
         >
-          Confirm
+          {strings.buttons.submit}
```
oidoug, line 156: 👏

oidoug (approved):
Congrats on your first PR!
LGTM 👍

src/components/secret-display/SecretDisplaySection.js
```diff
@@ -100,7 +102,7 @@ export default function SecretDisplaySection({
               {`You need to authenticate in order to access your recovery ${typeLabel}`}
             </Text>
             <Button
-              iconProps={{ name: 'face-id' }}
+              iconProps={{ name: biometryIconName }}
```
danibonilha: nit: we could simplify this step, by using the `useBiometricIconProps` hook, the result can be passed straight to iconProps.

danibonilha (approved):
Congrats on figuring out the biometric stuff and thanks for extracting strings into it's own file, just one small nit  🚀

lukemelia:
Yay, first PR!

src/components/secret-display/SecretDisplaySection.js
paulinhapenedo: Good catch, thanks for the guidance! Updated to use the hook.

PR #846, opened 2022-05-18
[CS-3861]: Create PinInput component

danibonilha:
### Description
This PR adds the PinInput component, with "theme"  options dark | light, since it's a PIN component, we are keeping it simple, without handing paste/coping  nor handling the edition of a single cell. 

### Screenshots

Simulator Screen Recording - iPhone 11 - 2022-05-18 at 16 31 19

// Example: 
```tsx
  const [value, setValue] = useState('');

  return (
    <>
      <CenteredContainer flex={0.3} backgroundColor="white">
        <PinInput
          variant="light"
          value={value}
          onChangeText={setValue}
          secureTextEntry={false}
        />
      </CenteredContainer>
      <CenteredContainer flex={0.3} backgroundColor="backgroundDarkPurple">
        <PinInput variant="dark" value={value} onChangeText={setValue} />
      </CenteredContainer>
    </>
  );
```

### DIFF
--- a/cardstack/src/components/Input/PinInput/PinInput.tsx
+++ b/cardstack/src/components/Input/PinInput/PinInput.tsx
@@ -1,0 +1,75 @@
+import React, { memo, useMemo, useRef } from 'react';
+import { StyleSheet, TextInput } from 'react-native';
+
+import { Container, Input, InputProps } from '@cardstack/components';
+
+import CharCell, { cellLayout } from './CharCell';
+import { PinLayoutVariant } from './types';
+
+const DEFAULT_PIN_LENGTH = 6;
+const PIN_LENGTH = new Array(DEFAULT_PIN_LENGTH).fill('');
+
+const styles = StyleSheet.create({
+  hiddenInput: {
+    ...StyleSheet.absoluteFillObject,
+    height: cellLayout.height,
+    top: -cellLayout.height,
+    opacity: 0,
+  },
+});
+
+interface PinInputProps extends InputProps {
+  variant: PinLayoutVariant;
+  value: string;
+}
+
+const PinInput = ({
+  variant,
+  value = '',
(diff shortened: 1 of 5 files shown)


cardstack/src/components/Input/PinInput/types.ts
```diff
@@ -0,0 +1 @@
+export type PinLayoutVariant = 'dark' | 'light';
```
oidoug, line 1: I think we should define global theme variants, but this is fine for now, we can refactor once everything is in place.

oidoug (approved):
LGTM

We'll need a more consistent way to define dark and light themes, but up for discussion.

cardstack/src/components/Input/PinInput/PinInput.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { memo, useMemo, useRef } from 'react';
+import { StyleSheet, TextInput } from 'react-native';
+
+import { Container, Input, InputProps } from '@cardstack/components';
+
+import CharCell, { cellLayout } from './CharCell';
+import { PinLayoutVariant } from './types';
+
+const DEFAULT_PIN_LENGTH = 6;
```
oidoug, line 9: suggestion: Include PIN length in PinInputProps.

oidoug (approved):

cardstack/src/components/Input/PinInput/PinInput.tsx
danibonilha: We don't have a scenario for less digits right now, once we do, we can add it, maybe for validating the old PIN ? 🤔

oidoug: Probably... I was thinking about the "beta" access key.

PR #848, opened 2022-05-19
[CS-3862] Biometrics switcher component

oidoug:
### Description

This PR adds a new Biometric Switcher component that handles UI, has a hook `` for pulling the user selection globally that uses redux slice, and is persisted with redux-persist.

- [x] Completes #(CS-3862)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/169331611-f3d85559-760b-44e3-aab8-328bf5ea052e.mov

### DIFF
--- a/cardstack/src/components/BiometricSwitch/BiometricSwitch.tsx
+++ b/cardstack/src/components/BiometricSwitch/BiometricSwitch.tsx
@@ -23,11 +23,12 @@ export const BiometricSwitch = ({ variant }: BiometricSwitchProps) => {
     iconProps,
     biometryLabel,
     isBiometryEnabled,
+    biometryAvailable,
     toggleBiometrySwitch,
   } = useBiometricSwitch();
 
-  // iconProps can only be undefined if biometry isn't ready or isn't available.
-  if (!iconProps) return null;
+  // iconProps is undefined until biometry is ready.
+  if (!biometryAvailable || !iconProps) return null;
 
   return (
     <Container flexDirection="row" alignItems="center">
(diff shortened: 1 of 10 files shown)


cardstack/src/redux/biometryToggleSlice.ts
```diff
@@ -0,0 +1,31 @@
...
+type SliceType = {
+  enabled: boolean;
+};
+
+const initialState: SliceType = {
+  enabled: false,
+};
+
+const slice = createSlice({
+  name: 'biometryToogle',
```
danibonilha: typo: toggle

cardstack/src/components/BiometricSwitch/BiometricSwitch.tsx
```diff
@@ -0,0 +1,46 @@
...
+type ThemeType = 'light' | 'dark';
+
+const colorVariant: Record<ThemeType, ColorTypes> = {
+  light: 'black',
+  dark: 'white',
+};
+
+interface BiometricSwitchProps {
+  variant: ThemeType;
+}
```
danibonilha, line 19: I'm adding a colorThemeVariant on my next PR so we can use across the components

danibonilha (approved):
LGTM, just a small typo and minimum concern

cardstack/src/components/BiometricSwitch/BiometricSwitch.tsx
```diff
@@ -0,0 +1,46 @@
...
+export const BiometricSwitch = ({ variant }: BiometricSwitchProps) => {
+  const {
+    iconProps,
+    biometryLabel,
+    isBiometryEnabled,
+    toggleBiometrySwitch,
+  } = useBiometricSwitch();
+
+  // iconProps can only be undefined if biometry isn't ready or isn't available.
+  if (!iconProps) return null;
```
danibonilha: Shouldn't we return null for the case  where `biometryLabel`  is falsy instead of iconProps? Bc if there's no biometric sensor nor device PIN, the icon when it's `none` defaults to `lock`, which is a truthy value

PR #849, opened 2022-05-19
[CS-3863]: Adds PinScreen UI

danibonilha:
### Description
 This PR creates the PINScreen UI, the goal is to be able to reuse it across the most flows, there's some hardcorded strings and params for now, just to show functionality, but the implementation to be fully generic is going to be done on the logic ticket it also adds shareable dark | light styles, updates biometric, input and cell to use it and the ability to show/hide the pin 

### Screenshots

image

image

image

image

// up-to-date info icon
image

### DIFF
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -32,6 +32,7 @@ import {
   TransactionConfirmationSheet,
   ColorPickerModal,
   RequestPrepaidCardScreen,
+  PinScreen,
 } from '@cardstack/screens';
 import {
   RewardWithdrawConfirmationScreen,
@@ -206,6 +207,10 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
   COLOR_PICKER_MODAL: {
     component: ColorPickerModal,
     options: expandedPreset as StackNavigationOptions,
+  },
+  PIN_SCREEN: {
+    component: PinScreen,
+    options: horizontalInterpolator,
   },
 };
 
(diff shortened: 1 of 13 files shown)


cardstack/src/navigation/screens.ts
```diff
@@ -207,6 +208,10 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
     component: ColorPickerModal,
     options: expandedPreset as StackNavigationOptions,
   },
+  PIN_SCREEN: {
+    component: PinScreen,
+    options: horizontalInterpolator,
```
oidoug, line 213: `horizontalInterpolator` allows for horizontal gestures to close the screen on iOS.

Maybe we can discuss a new appear animation for the PIN screen too.

paulinhapenedo (approved):
LGTM 🚀

cardstack/src/screens/PinScreen/PinScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+                color={colorStyleVariants.textColor[variant]}
+                textAlign="center"
+              >
+                {feedbackProps.label}
+              </Text>
+              <Icon
+                name={feedbackProps.iconName}
+                size={20}
+                paddingLeft={2}
+                backgroundColor="white"
```
oidoug, line 106: shouldn't this follow colorStyleVariants too?

oidoug (approved):
LGTM, I just added some comments for later evaluation.

cardstack/src/screens/PinScreen/PinScreen.tsx
danibonilha: This is the color of the ! point, and on both themes the color is the same

cardstack/src/navigation/screens.ts
danibonilha: just a placeholder for now, once we build the flows we can decide the correct animations

PR #850, opened 2022-05-19
[CS-3913] Handle /email-card-drop request failure for 503 status

paulinhapenedo:
### Description

This PR adds a custom error handling for the `POST` request to the `/email-card-drop` endpoint. Because this is a particular error, it's being treated in the component itself not on the `fetchHubBaseQuery` function.

Let me know if there's a need to cover this use case with a test 🙂

- [x] Completes #CS-3913

### Checklist

### DIFF
--- a/cardstack/src/screens/RequestPrepaidCardScreen/useRequestPrepaidCardScreen.ts
+++ b/cardstack/src/screens/RequestPrepaidCardScreen/useRequestPrepaidCardScreen.ts
@@ -1,3 +1,4 @@
+import { FetchBaseQueryError } from '@reduxjs/toolkit/dist/query/fetchBaseQuery';
 import { useCallback, useMemo, useState, useEffect } from 'react';
 import { Linking } from 'react-native';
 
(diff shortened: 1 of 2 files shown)


cardstack/src/screens/RequestPrepaidCardScreen/useRequestPrepaidCardScreen.ts
```diff
@@ -33,18 +34,29 @@ export const useRequestPrepaidCardScreen = () => {
 
   const [
     requestCardDrop,
-    { isError, isSuccess, isLoading },
+    { isError = true, isSuccess, isLoading, error },
```
danibonilha: is `isError = true` a leftover ?

cardstack/src/screens/RequestPrepaidCardScreen/useRequestPrepaidCardScreen.ts
```diff
@@ -33,18 +34,29 @@ export const useRequestPrepaidCardScreen = () => {
...
   ] = useRequestEmailCardDropMutation();
 
+  const errorMessage = useMemo(() => {
+    const { status } = error as FetchBaseQueryError;
+
+    return status === 503
+      ? {
+          title: strings.customError.title,
+          message: strings.customError.message,
+        }
```
danibonilha: nit: we could use the customError directly 

```tsx

return status === 503 ? strings.customError ? defaultErrorAlert

```

danibonilha (approved):
Good to go after removing leftover! 🚀 , it does seem a good opportunity to write some tests for this hook, but it can be done on a separate PR

cardstack/src/screens/RequestPrepaidCardScreen/useRequestPrepaidCardScreen.ts
paulinhapenedo: It is 🙈

PR #851, opened 2022-05-20
[FIX] Removes useMemo for featurePrepaidCardDrop config value in useWelcomeCtaBanner hook.

oidoug:
### Description

Removes useMemo from featurePrepaidCardDrop remote config get call, so config value is not cached in the hook. Since this memo doesn't have a dependency to update itself, it was holding to an old value.


danibonilha (approved):
LGTM

PR #852, opened 2022-05-20
[CS-3926]: Share same tags across remote-config

danibonilha:
### Description

Updates the firebase remote-config module to use the same tag, types, from default value, so we don't forget to update anywhere in case of a new flag etc, also adds a function that returns all the values, so it's easier to access. 
We could also improve the remoteFlags function in the future, to automatically map if it's a string, number, or boolean, and return the correct flag.


oidoug (approved):
Nice upgrade, always better to have typed properties!

PR #855, opened 2022-05-23
[Test]: Add max balance first test

danibonilha:
### Description
This PR adds the first test to the sendFullBalance flow, it gives the structure to add more tests for the other scenarios as this one is just testing if onMaxPress the amount is undefined.


oidoug (approved):
🚀 
Thanks Dani!

PR #856, opened 2022-05-23
[CS-3865] Add Change Pin item on Settings menu

paulinhapenedo:
### Description

This PR adds the Change Pin option to the Settings Menu. Right now, we don't have a route to redirect when tapping the item, so I've left a `console.log`.

While on the area, I noticed that we have two flags for development mode: `IS_DEV` and `__DEV__`. Seems that the latter is the approach the team decided to use moving forward, so I replaced the instances where we were using `IS_DEV` and removed it from the `globals.d.ts` file.

- [x] Completes #CS-3865

### Checklist

### Screenshots

Screenshot 

Screenshot

### DIFF
--- a/globalVariables.js
+++ b/globalVariables.js
@@ -7,15 +7,12 @@ import {
   useState,
 } from 'react';
 import { Platform } from 'react-native';
-import { ENABLE_DEV_MODE } from 'react-native-dotenv';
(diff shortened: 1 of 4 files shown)


danibonilha (approved):
LGTM

PR #857, opened 2022-05-24
[Test] Adds a couple more tests to send sheet depot screen hook

oidoug:
### Description

Adds test for asserting that max selection is cleared after an asset change and that `sendTokens` is called with the right amount set.

### DIFF
--- a/cardstack/src/screens/SendSheetDepot/__tests__/useSendSheetDepotScreen.test.ts
+++ b/cardstack/src/screens/SendSheetDepot/__tests__/useSendSheetDepotScreen.test.ts
@@ -296,8 +296,6 @@ describe('useSendSheetDepotScreen', () => {
       expect(result.current.amountDetails.assetAmount).toBeTruthy()
     );
 
-    console.log('::: ', result.current.amountDetails);
-
     await act(async () => {
       await result.current.onSendPress();
     });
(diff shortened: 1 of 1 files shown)


danibonilha (approved):
🚀

cardstack/src/screens/SendSheetDepot/__tests__/useSendSheetDepotScreen.test.ts
```diff
@@ -223,4 +224,46 @@ describe('useSendSheetDepotScreen', () => {
...
+
+    await waitFor(() =>
+      expect(result.current.selectedGasPrice.amount).toBeTruthy()
+    );
+
+    await act(async () => {
+      await result.current.setRecipient(recipient);
+      await result.current.onMaxBalancePress();
+      await result.current.onChangeAssetAmount(updatedAssetAmount);
+    });
```
danibonilha, line 249: I guess we don't need to use async/await here, I added on my test in an attempt to remove the act console warning, but it doesn't seem to make any diff, it doesn't hurt to leave it like this, but it's good to know that is not required

oidoug: I have another test coming up, I'll change it for the next PR.

PR #858, opened 2022-05-24
[CS-3868]: Add PinScreen flows

danibonilha:
### Description

This PR adds the PinScreen with a hook to handle the different PIN flows 

- [x] Completes CS-3875
- [x] Completes CS-3869
- [x] Completes CS-3868

### Screenshots

Simulator Screen Recording - iPhone 11 - 2022-05-24 at 10 28 35

Simulator Screen Recording - iPhone 11 - 2022-05-24 at 10 26 13

### DIFF
--- a/cardstack/src/screens/PinScreen/PinScreen.tsx
+++ b/cardstack/src/screens/PinScreen/PinScreen.tsx
@@ -1,4 +1,4 @@
-import React, { memo, useEffect, useMemo, useState } from 'react';
+import React, { memo, useEffect, useMemo } from 'react';
 import { StatusBar, NativeModules } from 'react-native';
 
 import {
@@ -13,12 +13,7 @@ import { colorStyleVariants } from '@cardstack/theme/colorStyleVariants';
 import { Device } from '@cardstack/utils';
 
 import { strings } from './strings';
-
-// To be replaced with states
-const variant = 'light';
-const showFeedback = false;
-const showBiometricSwitcher = true;
-const isValidPin = true;
+import { usePinScreen } from './usePinScreen';
 
 const feedbackStatusProps = {
   success: {
@@ -34,7 +29,15 @@ const feedbackStatusProps = {
 };
 
 const PinScreen = () => {
-  const [inputPin, setInputPin] = useState('');
+  const {
+    canGoBack,
+    showBiometricSwitcher,
+    isValidPin,
+    variant,
+    setInputPin,
+    inputPin,
+    flow,
+  } = usePinScreen();
 
   useEffect(() => {
     Device.isAndroid && NativeModules?.AndroidKeyboardAdjust.setAdjustPan();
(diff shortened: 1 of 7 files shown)


cardstack/src/screens/PinScreen/PinScreen.tsx
```diff
@@ -117,7 +120,7 @@ const PinScreen = () => {
         {showBiometricSwitcher && (
           <Container
             justifyContent="flex-end"
-            flex={Device.isIOS ? 0.3 : 0.4}
+            flex={Device.isIOS ? 0.25 : 0.4}
```
oidoug, line 123: I see that the switcher is fitting the screen alright even on ios small. But I wonder if using a KeyboardAvoidingView isn't simpler. You could then remove the device-specific code.

cardstack/src/screens/PinScreen/__tests__/usePinScreen.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      expect(result.current.isValidPin).toBeTruthy();
+    });
+
+    expect(onSuccess).toBeCalledWith(PIN);
+
+    jest.runAllTimers();
+
+    expect(mockNavDispatch).toBeCalledWith(StackActions.popToTop());
+  });
+});
```
oidoug, line 177: 👏  pretty comprehensive!

cardstack/src/screens/PinScreen/usePinScreen.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  useEffect(() => {
+    if (isValidPin) {
+      // Probably gonna need to extract to async func
+      // If so, sending a dimiss callback might be better
+      // then using setTimeout
+      params?.onSuccess?.(inputPin);
+
+      setTimeout(() => {
+        navDispatch(StackActions.popToTop());
```
oidoug, line 75: This may need refactoring later since we might need to use PIN from screens already stacked up, so `popToTop` would close the calling screen.

oidoug (approved):
LGTM. I left some comments for your consideration.

oidoug:
I'm thinking the unlock screen should also be a flow in PinScreen, lots to reuse here.

cardstack/src/screens/PinScreen/usePinScreen.tsx
paulinhapenedo: Yeah, I agree with Doug and your comment on this one. Maybe receiving a callback that handles what happens after the pin flow is successful (that will mean different things in different scenarios) is a better approach than having a `setTimeout`.

cardstack/src/screens/PinScreen/PinScreen.tsx
danibonilha: Using KeyboardAvoidingView also requires to set conditions based on the platform, and we know that usually doesn't work out of box as expected, also it makes the items move down while dismissing the keyboard, this approach seemed easier to handle, but if you think it's worth to use KeyboardAvoidingView feel free to make the improvement.

danibonilha:
> I'm thinking the unlock screen should also be a flow in PinScreen, lots to reuse here.

I'm not sure, the only common things, are the input and the switcher, the remaining components are pretty different, we could share one hook is that was case, just to handle the input of the PIN, but need to check if it's worth I don't want to overcomplicate the PinScreen

cardstack/src/screens/PinScreen/usePinScreen.tsx
danibonilha: I added the comment so we won't forget

PR #859, opened 2022-05-24
[CS-3937] Prepaid Card: disable Submit button if both email and checkbox are not valid

paulinhapenedo:
This PR enhances the `disablePress` prop from the Submit button on the request prepaid card screen. The button should be disabled, both visually and in functionality, until the user enters a valid email and checks the checkbox. 

- [x] Completes #CS-3937

### Checklist

### DIFF
--- a/cardstack/src/screens/RequestPrepaidCardScreen/RequestPrepaidCardScreen.tsx
+++ b/cardstack/src/screens/RequestPrepaidCardScreen/RequestPrepaidCardScreen.tsx
@@ -96,7 +96,7 @@ const RequestPrepaidCardScreen = () => {
                 variant={!canSubmit ? 'disabledBlack' : undefined}
                 onPress={onSubmitPress}
                 iconProps={!isAuthenticated ? iconProps : undefined}
-                disablePress={isLoading}
+                disablePress={isLoading || !canSubmit}
                 loading={isLoading}
               >
                 {strings.button.submit}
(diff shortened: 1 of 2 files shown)


oidoug (approved):
👏 LGTM!

cardstack/src/screens/RequestPrepaidCardScreen/RequestPrepaidCardScreen.tsx
```diff
@@ -96,7 +96,7 @@ const RequestPrepaidCardScreen = () => {
                 variant={!canSubmit ? 'disabledBlack' : undefined}
                 onPress={onSubmitPress}
                 iconProps={!isAuthenticated ? iconProps : undefined}
-                disablePress={isLoading}
+                disablePress={isLoading || !canSubmit}
```
danibonilha: There was a reason for not using canSubmit on the disablePress and have the early return, but I can't recall right now what it was.

danibonilha: Oh I found this PR #803, it seems the idea to keep it pressable it's to handle the case where the user inputs something, that doesn't really seems like an email, and still tries to proceed, so we give the feedback, in order to keep this behavior we should actually change the onSubmit to skip setting the error if the checkbox is not set yet.

paulinhapenedo: Oh, thank you for the explanation! I'll update the code and let a comment explaining the quirk for future reference.

danibonilha (approved):

PR #861, opened 2022-05-25
[CS-3941]: Setup expo on project and remove dead pkgs

danibonilha:
### Description

This PR adds the expo sdk to the project so we can leverage its packages, in order to make it work with our project, some dependencies were updated, and some removed. 

- Updates kotlin, gradle and java version (Make sure you have Java 11 installed)
- Updates supported API to 31 to support Android 12 changes 
- Removes outdated react-native-number-version in favor of deviceInfo
- Removes native user settings keychain wipe, in favor of adding the new reset wallet btn 
- Removes native screens in favor of using react-navigation only
- Removes safe-area-view in favor of safe-area-view-context, which is used more broadly in the app

Please checkout this branch and build locally on both devices, to make sure everything is good.


.github/workflows/release-android.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       - name: Set up JDK 1.8
```
paulinhapenedo: Forgot to change the name here to reflect the JDK version.

react-native.config.js
```diff
@@ -10,5 +10,10 @@ module.exports = {
         ios: null,
       },
     },
+    '@rainbow-me/react-native-payments': {
+      platforms: {
+        android: null,
+      },
+    },
```
oidoug, line 17: Can you elaborate on why this was necessary?

src/components/savings/APYPill.js
```diff
@@ -62,20 +51,12 @@ const TextComponent = ios ? GradientText : Text;
 function APYPill({ small, value }) {
   return (
     <Centered style={small ? sx.containerSmall : sx.container}>
-      {IS_TESTING === 'true' ? null : small ? (
-        <RadialGradient
-          {...radialGradientProps}
-          borderRadius={21}
-          radius={81}
-        />
```
oidoug, line 70: I see this style can still be shown on SavingSheet, but I'm with you if you think it's not worth having the `react-native-radial-gradient` lib just for this.

package.json
```diff
@@ -162,7 +153,7 @@
     "react-native-image-crop-picker": "^0.32.2",
     "react-native-indicators": "0.17.0",
     "react-native-iphone-x-helper": "^1.2.1",
-    "react-native-keyboard-area": "^1.0.5",
+    "react-native-keyboard-area": "thomasgazzoni/react-native-keyboard-area#749746a18f84d8567620efacf3ad9f4a07f8fb74",
```
oidoug, line 156: I've added a ticket for removing this lib:
https://linear.app/cardstack/issue/CS-3963/[debt]-remove-custom-lib-react-native-keyboard-area

src/react-native-cool-modals/Portal.js
```diff
@@ -5,21 +5,15 @@ import React, {
...
 
 export function usePortal() {
   return useContext(NativePortalContext);
 }
 
-const NativePortal =
-  Platform.OS === 'ios' ? requireNativeComponent('WindowPortal') : View;
+const NativePortal = View;
 
 const Wrapper = Platform.OS === 'ios' ? ({ children }) => children : View;
```
oidoug, line 19: I created a ticket to remove the final bits of this custom lib, LMK if you think it's relevant or not.
https://linear.app/cardstack/issue/CS-3964/[debt]-remove-custom-local-lib-react-native-cool-modals

src/screens/SavingsSheet.js
```diff
@@ -34,7 +33,7 @@ import {
 import { position } from '@rainbow-me/styles';
 
 export const SavingsSheetEmptyHeight = 313;
-export const SavingsSheetHeight = android ? 424 - getSoftMenuBarHeight() : 352;
+export const SavingsSheetHeight = android ? 400 : 352;
```
oidoug, line 36: nit: It would be preferable to have this height be affected by safe area insets (bottom maybe?).

oidoug (commented):
I'm still building/testing the PR, but I'm leaving my review comments for your appreciation.

Generally, it's great that you removed all this unused or barely unused code and libs, but I feel your PR got out of scope a bit.

react-native.config.js
danibonilha: We don't use the payments api on android, buy prepaid card is only for iOS

src/screens/SavingsSheet.js
danibonilha: It seems this is going to be remove on @paulinhapenedo PR's

src/react-native-cool-modals/Portal.js
danibonilha: The main blocker, is to first migrate all the states that uses `setIsWalletLoading`

package.json
danibonilha: It's hardcoded to a commit, bc the maintainer didn't publish 1.0.5 correctly, so it doesn't have the gradle update

src/components/savings/APYPill.js
danibonilha: All the savings stuff is going to be removed, so not worrying to much about that

danibonilha:
> I'm still building/testing the PR, but I'm leaving my review comments for your appreciation.
> 
> Generally, it's great that you removed all this unused or barely unused code and libs, but I feel your PR got out of scope a bit.

@oidoug  I don't see what you mean by, being out of scope a bit, Expo uses the most up-to-date packages and apis, which makes lots of old packages incompatible with it, instead of trying to update all the packages that were breaking the build, it felt safer to research the usage and decide between removing, updating or replacing it, same goes for the native code that was not being used and it was causing the update to fail.

src/screens/SavingsSheet.js
paulinhapenedo: `SavingsSheet` is not being used anywhere in the code, so I'm removing the file as part of the work in [CS-3887](https://linear.app/cardstack/issue/CS-3887/[debt]-refactor-out-slacksheet-in-favor-of-newer-sheet-component).

oidoug:
Hum.. ok, got it. I didn't think these packages would affect expo's ones, since you're only installing expo core. But if that's the case then it makes sense your refactor.

paulinhapenedo (approved):
Seems like it's working after some deep cache clean 🎉 
Do you think it's necessary to add the `EXErrorRecovery` module to the `Podfile.lock` or not?

PR #864, opened 2022-05-27
[CS-3962] - Remove dead pkgs and improve wallet init

danibonilha:
### Description

As part of the heating issue, it was identified a bunch of unnecessary subscriptions for swaping, which was removed on this PR, to improve the app performance some init methods were changed in order to init faster.

- Removes uniwasp/swap features
- Updates Loading to use same component across application (This is one is a bit out of scope, but the loadingOverlay update makes the UI faster )
- Removes unused iOS internal packages
- Migrates coingecko call to fallbackback explorer and caches it, in order to improve wallet init
- Removes unnecessary data clean up (we were cleaning up, loading, then cleaning up again theen loading the final piece).
- Improves safes fetch, no need to wait for wallet to be ready now, as we already have conversionRates, refer to #528 for more context why we needed that before.
- Removes unused  charts reducer

The start of the application has improved overall from 11-12 to  8 seconds.


src/redux/fallbackExplorer.js
```diff
@@ -95,15 +96,31 @@ const fetchCoingeckoIds = async (network, coingeckoCoins) => {
...
+    const request = await fetch(
+      'https://api.coingecko.com/api/v3/coins/list?include_platform=true&asset_platform_id=ethereum'
+    );
+    const coingeckoCoins = await request.json();
+
+    return coingeckoCoins;
+  } catch (error) {
+    return coingeckoIdsFallback;
+  }
+};
```
danibonilha, line 112: The TODO initially it said to move rtk, which made sense at the time, but now that is only user here, it makes more sense to scope this call within the fallbackExplorer. This is just a list of available coins, so it's ok to fetch just once in a while

cardstack/src/components/LoadingOverlay/LoadingOverlay.tsx
```diff
@@ -1,32 +1,17 @@
...
 
-const transition = (
-  <Transition.Sequence>
-    <Transition.Out durationMs={500} interpolation="easeOut" type="fade" />
-    <Transition.Change durationMs={500} interpolation="easeOut" />
-    <Transition.In durationMs={500} interpolation="easeOut" type="fade" />
-  </Transition.Sequence>
-);
-
 const OverlyStyle = {
```
oidoug: Possibly old typo, OverlyStyle => overlayStyle

cardstack/src/components/LoadingOverlay/LoadingOverlay.tsx
```diff
@@ -42,37 +27,29 @@ const OverlyStyle = {
...
+}: {
+  title: string;
+  subTitle?: string;
+}) => (
+  <CenteredContainer style={styles.overlayWrapper}>
+    <CenteredContainer
+      backgroundColor="white"
+      borderColor="whiteOverlay"
+      shadowColor="overlay"
+      {...OverlyStyle}
```
oidoug: ditto

cardstack/src/components/LoadingOverlay/LoadingOverlay.tsx
```diff
@@ -42,37 +27,29 @@ const OverlyStyle = {
...
-  ({ title, subTitle }: { title: string; subTitle: string }) => {
-    return (
-      <Transitioning.View transition={transition} style={styles.overlayWrapper}>
-        <CenteredContainer
-          backgroundColor="white"
-          borderColor="whiteOverlay"
-          shadowColor="overlay"
-          {...OverlyStyle}
-        >
-          {Device.isAndroid ? (
```
oidoug, line 55: 👏

cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
```diff
@@ -59,13 +58,11 @@ export const usePaymentMerchantUniversalLink = () => {
...
   const [infoDID, setInfoDID] = useState<string | undefined>();
 
-  const walletReady = useRainbowSelector(state => state.appState.walletReady);
-
   const { isLoading = true, prepaidCards } = useGetPrepaidCardsQuery(
     { accountAddress, nativeCurrency },
     {
       refetchOnMountOrArgChange: 60,
-      skip: isLayer1(accountNetwork) || !accountAddress || !walletReady,
+      skip: isLayer1(accountNetwork) || !accountAddress,
```
oidoug, line 65: Can you elaborate on this change? Does this mean the Wallet doesn't need to be ready before fetching safes?

src/redux/fallbackExplorer.js
```diff
@@ -95,15 +96,31 @@ const fetchCoingeckoIds = async (network, coingeckoCoins) => {
...
+const findAssetsToWatch = async (address, latestTxBlockNumber, dispatch) => {
   // 1 - Discover the list of tokens for the address
   const network = store.getState().settings.network;
-  const coingeckoIds = await fetchCoingeckoIds(network, coingeckoCoins);
+
+  coingeckoCoinsCache.current =
+    coingeckoCoinsCache.current || (await fetchCoingeckoCoins());
```
oidoug, line 118: Note: You mentioned above that fetching once in a while was ok, but by saving the fetch result in this global variable you'll only refresh values on app re-runs.

oidoug (approved):
LGTM! Impressive work. I left some notes for your appreciation.

cardstack/src/hooks/merchant/usePaymentMerchantUniversalLink.ts
danibonilha: Yep, we already have the exchange rates now, this old implementation was add on this PR #528, you can find more context there.

src/redux/fallbackExplorer.js
danibonilha: Even though i'm not sure about the coins update frequency, on app re-runs should be enough to get new coins, if any is added.

PR #865, opened 2022-05-30
[CS-3871]: Add SecureStore package and initial models

danibonilha:
### Description

This PR adds the SecureStore package with Pin and seedPhrase models.

- Configured jest to work with expo
- Setup new env var to secure keys, remember to sync app_vars 
- Setup test env to be able to mock implementation 
- Added test to Pin model
- Added initial auth model (lot's of todo still, will add test later)

PS: Didn't add tests to the seedPhrase piece, bc I'm having some issues mocking the encryptor, will do on following up PRs

### DIFF
--- a/cardstack/src/models/secure-storage.ts
+++ b/cardstack/src/models/secure-storage.ts
@@ -1,6 +1,7 @@
 import * as SecureStore from 'expo-secure-store';
 import { SECURE_STORE_KEY } from 'react-native-dotenv';
 
+import AesEncryptor from '@rainbow-me/handlers/aesEncryption';
 import logger from 'logger';
 
 const keys = {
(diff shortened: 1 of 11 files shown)


paulinhapenedo (approved):
Looks good 👏

cardstack/src/models/secure-storage.ts
```diff
@@ -0,0 +1,92 @@
...
+    logger.sentry(`No value for key: ${secureKey}`, e);
+
+    return null;
+  }
+};
+
+// PIN
+const savePin = (pin: string) => SecureStore.setItemAsync(keys.AUTH_PIN, pin);
+
+const getPin = () => getSecureValue(keys.AUTH_PIN);
```
oidoug, line 36: Are we going to keep the unlock pin in a global state? I'm assuming we'll need that or we'll need the user to unlock the pin every time we need it.

oidoug (approved):
Pretty good! A lot cleaner than our current model.
Just left a comment / question for you.

cardstack/src/models/secure-storage.ts
danibonilha: We'll need in two flows, but we can talk on DM about it

PR #866, opened 2022-05-31
[CS-3971] Keyboard adjust policy change

oidoug:
### Description

This PR changes android soft input mode to `adjustPan`, this means it will behave more similarly to iOS, but it creates new challenges, since views that need a confirmation button to be visible above the keyboard, need to use KeyboardAvoidingView with `panning` or `height` (only android) behavior modes set.

PR also updates `Device.keyboardBehavior` to 'panning' for both android and ios, as it's now a better default for android,
and adds an option to set keyboard behavior in Sheet comp.

- [x] Completes #(CS-3971)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/171177558-5a6c211a-fec7-4cdb-bd91-4d4f96862659.MOV

https://user-images.githubusercontent.com/129619/171177580-86e13975-b449-4b6f-b1c4-fa0f400e6a46.MOV

https://user-images.githubusercontent.com/129619/171179416-04f586a1-8b57-4956-a609-a37f20bee882.mp4

https://user-images.githubusercontent.com/129619/171179408-52dbf9eb-8924-42e3-a1f7-c95caf117285.mp4

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -24,7 +24,7 @@
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
         android:launchMode="singleTask"
         android:exported="true"
-        android:windowSoftInputMode="adjustResize">
+        android:windowSoftInputMode="adjustPan">
         <intent-filter>
             <action android:name="android.intent.action.MAIN" />
             <category android:name="android.intent.category.LAUNCHER" />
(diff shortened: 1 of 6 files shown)


danibonilha (commented):
It's looking good, I tested on a device, didn't find any major issues, one thing it would be nice to improve is when we have a keyboard open and we call the Profile modal, the cancel button is hidden by the keyboard:
image
I did find some issues on my local emulator which I'll DM you about.

danibonilha (approved):
I believe we can merge this, and work on the fixes if needed along the sprint, so we can unblock the screens on Android, nice job on figuring the issue

oidoug:
Agreed, we'll probably find more issues like the above.

PR #867, opened 2022-05-31
[CS-3976]: Add custom config to android switch on light theme 

danibonilha:
### Description

The switch track on android when the screen was white, was not showing, this PR adds a config to handle that case

### Screenshots

image


paulinhapenedo (approved):
Nice catch!

oidoug (commented):
Thanks!

oidoug (approved):

PR #869, opened 2022-05-31
[CS-3523] Rewards: move claim functions into own hook

paulinhapenedo:
### Description
This PR is the first part of the `useRewardsCenter` hook refactor into smaller bits, work that's being tracked in [CS-3523](https://linear.app/cardstack/issue/CS-3523/split-userewardscenter-into-smaller-hooks).

Here, we are moving all things claim into its own hook. For now, we're passing params to the hook (accountAddress, mainPoolTokenInfo, rewardSafes) to avoid changing the current hook structure, but this may change in the future.

### Checklist

- [ ] (Trying) Adding Tests

### Screenshots

No visual or behavior changes should be seen from this PR.

### DIFF
--- a/cardstack/src/screens/RewardsCenterScreen/useRewardsCenterScreen.ts
+++ b/cardstack/src/screens/RewardsCenterScreen/useRewardsCenterScreen.ts
@@ -258,36 +258,11 @@ export const useRewardsCenterScreen = () => {
     });
   }, [navigate, onPrepaidCardSelection]);
 
-  const {
-    onClaimPress,
-    isLoadingClaimGas,
-    isClaimSuccess,
-    isClaimError,
-  } = useRewardsClaim({
+  const { onClaimPress, isLoadingClaimGas } = useRewardsClaim({
     accountAddress,
     mainPoolTokenInfo,
     rewardSafes,
   });
-
-  useMutationEffects(
-    useMemo(
-      () => ({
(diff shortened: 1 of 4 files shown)


cardstack/src/screens/RewardsCenterScreen/useRewardsCenterScreen.ts
```diff
@@ -286,74 +289,6 @@ export const useRewardsCenterScreen = () => {
     )
   );
```
danibonilha: This seems to be in the same claiming scope, which could be extract into the new claiming hook, the only two properties that we would need from the claim hook,  would be the isLoadingClaimGas and onClaimPress, is there any specific reason you can think of where it wouldn't make sense to move this useMutationEffects inside the useRewardsClaim ?

paulinhapenedo: When abstracting, I thought of leaving the `useMutationEffect` on the root hook, but after second thought, I think it's best to move it into the specific hook where it's being triggered.

paulinhapenedo:
Closing this PR because I've found an issue when trying to test this custom hook. It looks like I started from the wrong place.

I'll be extracting the `safes` and `pool` part to its own hook first to be able to reuse it across the other custom hooks.

PR #870, opened 2022-06-01
Profile modal keyboard panning responder

oidoug:
### Description

This PR simplifies ProfileModal and KeyboardFixedOpenLayout to use the new `keyboardavoidingview` panning behavior.

ProfileModal and components that use it could be rebuilt with `restyle` containers, I started for WalletProfileModal but it's not trivial because there are more components using it.

### Checklist

### Screenshots

Screenshot

Screenshot

Screenshot


paulinhapenedo (approved):
Looks good! 👏

src/components/layout/KeyboardFixedOpenLayout.js
```diff
@@ -1,48 +1,18 @@
...
 
-const Container = styled(Transitioning.View)`
-  height: ${({ height }) => height};
-  left: 0;
-  position: ${android ? 'relative' : 'absolute'};
-  right: 0;
-  top: 0;
+const Container = styled(View)`
+  height: 100%;
+  width: 100%;
```
danibonilha: nit: Would it be possible to replace this styled view with the restyle Container ?

src/components/layout/KeyboardFixedOpenLayout.js
```diff
@@ -1,48 +1,18 @@
...
   return (
-    <Container height={containerHeight} ref={ref} transition={transition}>
-      <KeyboardAvoidingView behavior="height" enabled={!keyboardHeight}>
-        <InnerWrapper {...props} insets={insets} />
+    <Container>
+      <KeyboardAvoidingView behavior="padding">
+        <Centered {...props} />
       </KeyboardAvoidingView>
     </Container>
   );
```
danibonilha: Actually, can we just get rid of this `KeyboardFixedOpenLayout` component ? it's used  only in the `ProfileModal`, which is a pretty small component, couldn't  we just add the KeyboardAvoidingView and the container there ?

src/components/layout/KeyboardFixedOpenLayout.js
```diff
@@ -1,48 +1,18 @@
...
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    useEffect(() => ref.current.animateNextTransition(), [containerHeight]);
-
+export default function KeyboardFixedOpenLayout({ ...props }) {
   return (
-    <Container height={containerHeight} ref={ref} transition={transition}>
-      <KeyboardAvoidingView behavior="height" enabled={!keyboardHeight}>
-        <InnerWrapper {...props} insets={insets} />
+    <Container>
+      <KeyboardAvoidingView behavior="padding">
```
danibonilha: nit: We could use the `Device.keyboardBehavior`

danibonilha (approved):
Nice fix, I've just added a suggestion to cleanup a bit more

src/components/layout/KeyboardFixedOpenLayout.js
oidoug: Yes, I tried but was having issues making the centered view stretch in width, some component is messing with the layout. I'll give it another go.

oidoug: Agh, I was trying too hard before... went back to it now and found an easy way around it, will push soon.

PR #871, opened 2022-06-01
[CS-3978]: Fix android failing build and loadingOverlay layout

danibonilha:
### Description

Expo-secure store uses the same lib as keychain for biometric auth, but theirs are most up-to-date, so we had a conflict going on since only the mos recent one is installed and the newer version doesn't need to  use `DeviceCredentialHandlerTheme`, so the keychain patch was updated to reflect that. I also slipped the loadingOverlay fix 😬 in this one it was having too much spacing when it didn't had the subtitle. 
Latest Android build was triggered from this Branch in order to make sure it's working.


paulinhapenedo (approved):
Good work with finding the solution! 👏

PR #872, opened 2022-06-01
[CS-3870] hook logic for unlock screen

oidoug:
### Description

This PR adds part of the logic for pin unlocking, but it deviated a bit from the ticket it refers to because it's still dependent on improvements on our auth flow. 

It also includes color and size variations for CardstackLogo comp. and a first try at the "reset wallet" copy, the action itself needs some discussion, but right now it uses `wipeKeychain` that apparently isn't working anymore and removes the pin saved in secure-store. 

- [x] Completes #(CS-3870)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/171480078-a3151718-2339-49a5-b009-a20f9f03b615.mp4


danibonilha (approved):
it would be nice to add some tests for this hook, but since it's still raw, it may make sense to wait, also it's odd to have the Incorrect feedback without typing anything

oidoug:
In the video I had already tried a wrong pin before recording. It'll not show the incorrect pin feedback on the first open.

PR #873, opened 2022-06-01
[CS-3872] - Create Pin on new wallet creation

danibonilha:
### Description

This PR adds the ability to create a PIN while setting a new wallet from scratch.

### Screenshots

Simulator Screen Recording - iPhone 11 - 2022-06-01 at 16 58 39

### DIFF
--- a/cardstack/src/models/secure-storage.ts
+++ b/cardstack/src/models/secure-storage.ts
@@ -7,11 +7,14 @@ import logger from 'logger';
 const keys = {
   AUTH_PIN: `${SECURE_STORE_KEY}_AUTH_PIN`,
   SEED: `${SECURE_STORE_KEY}_SEED`,
+  PKEY: `${SECURE_STORE_KEY}_PKEY`,
 } as const;
 
 type KeysType = typeof keys;
 
 type Keys = KeysType[keyof KeysType];
+
+const trimSecureKey = (key: Keys) => key.replace(`${SECURE_STORE_KEY}_`, '');
 
 const getSecureValue = async (
   key: Keys,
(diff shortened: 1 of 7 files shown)


cardstack/src/models/secure-storage.ts
```diff
@@ -22,14 +25,47 @@ const getSecureValue = async (
...
+    }
+  } catch (e) {
+    const secureKey = trimSecureKey(key);
+    logger.sentry(`Error getting ${secureKey} encrypted item`, e);
+
+    return null;
+  }
+};
+
+const setEncryptedItem = async (item: string, key: Keys, password: string) => {
```
oidoug, line 54: Is password here our pin? If so better call it accordingly.

cardstack/src/models/secure-storage.ts
```diff
@@ -40,53 +76,61 @@ const deletePin = () => SecureStore.deleteItemAsync(keys.AUTH_PIN);
...
 
 // SEED
-const getSeedKey = (walletId: string) => `${keys.SEED}_${walletId}`;
+const saveSeedPhrase = async (seed: string, walletId: string, pin: string) => {
+  const seedKey = buildKeyWithId(keys.SEED, walletId);
 
-const saveSeedPhrase = async (seed: string, walletId: string) => {
-  try {
-    const pin = await getPin();
+  await setEncryptedItem(seed, seedKey as Keys, pin);
```
oidoug, line 82: nit: I understood why you had to cast `seedKey`, but it looks forced because in reality `seedKey` is not there in the Keys type.

One idea is to have setEncryptedItem accept a second type that buildKeyWithId returns, a string, or something like KeyWithID which is a string. Not sure though.

oidoug (approved):
LGTM, just some minor nit you can skip.

cardstack/src/models/secure-storage.ts
danibonilha: will address on upcoming PR's

cardstack/src/models/secure-storage.ts
danibonilha: ditto

PR #874, opened 2022-06-01
[CS-3523] Rewards: create custom hook for safes and pool

paulinhapenedo:
### Description
This ticket is part of the effort to solve [CS-3523](https://linear.app/cardstack/issue/CS-3523/split-userewardscenter-into-smaller-hooks).

On #869, my approach was wrong. I tried to abstract all the claim functions, but bumped into an issue with the data fetching part (`rewardPoolTokenBalances` e `rewardSafes`) so I decided to abstract that first to be easier to have access to this info.

### DIFF
--- a/cardstack/src/screens/RewardsCenterScreen/useRewardsDataFetch.ts
+++ b/cardstack/src/screens/RewardsCenterScreen/useRewardsDataFetch.ts
@@ -55,11 +55,16 @@ const useRewardsDataFetch = () => {
     [rewardPoolTokenBalances, defaultRewardProgramId]
   );
 
-  return {
-    isLoading:
+  const isLoading = useMemo(
+    () =>
       isLoadingSafes ||
       isLoadingTokens ||
       (isUninitialized && !isLayer1(network)),
+    [isLoadingSafes, isLoadingTokens, isUninitialized, network]
+  );
+
+  return {
+    isLoading,
     rewardSafes,
     rewardPoolTokenBalances,
     mainPoolTokenInfo,
--- a/cardstack/src/screens/RewardsCenterScreen/useRewardsDataFetch.ts
+++ b/cardstack/src/screens/RewardsCenterScreen/useRewardsDataFetch.ts
@@ -1,0 +1,70 @@
+import { useMemo } from 'react';
+
+import {
+  useGetRewardPoolTokenBalancesQuery,
+  useGetRewardsSafeQuery,
+} from '@cardstack/services/rewards-center/rewards-center-api';
+import { isLayer1 } from '@cardstack/utils';
+
+import { networkTypes } from '@rainbow-me/helpers/networkTypes';
+import { useAccountSettings } from '@rainbow-me/hooks';
+
+const rewardDefaultProgramId = {
+  [networkTypes.sokol]: '0x5E4E148baae93424B969a0Ea67FF54c315248BbA',
+  // TestID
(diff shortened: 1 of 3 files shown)


cardstack/src/screens/RewardsCenterScreen/useRewardsDataFetch.ts
```diff
@@ -0,0 +1,70 @@
...
+          Number(amount) > 0 && rewardProgramId === defaultRewardProgramId
+      ),
+    [rewardPoolTokenBalances, defaultRewardProgramId]
+  );
+
+  return {
+    isLoading:
+      isLoadingSafes ||
+      isLoadingTokens ||
+      (isUninitialized && !isLayer1(network)),
```
danibonilha: nit: would be nice to extract to it's own memoized variable, to cleanup the hook return

cardstack/src/services/rewards-center/rewards-center-api.ts
```diff
@@ -140,4 +157,5 @@ export const {
   useLazyGetRegisterRewardeeGasEstimateQuery,
   useWithdrawRewardBalanceMutation,
   useGetRewardWithdrawGasEstimateQuery,
+  useGetClaimRewardsGasEstimateQuery,
 } = rewardsApi;
```
danibonilha, line 160: I believe this is not entirely related with rewards safes and pool, but it's ok to add it here, since it's just adding the new endpoint.

danibonilha (approved):
Nice abstraction!

cardstack/src/services/rewards-center/rewards-center-api.ts
```diff
@@ -140,4 +157,5 @@ export const {
   useLazyGetRegisterRewardeeGasEstimateQuery,
   useWithdrawRewardBalanceMutation,
   useGetRewardWithdrawGasEstimateQuery,
+  useGetClaimRewardsGasEstimateQuery,
```
oidoug, line 159: I liked you moved the gas for reward here, to me it makes sense, but maybe you should move this change to a new PR and also refactor the use of `getClaimRewardsGasEstimate` in `useRewardsCenterScren` alongside it.

cardstack/src/services/rewards-center/rewards-center-api.ts
```diff
@@ -114,6 +116,21 @@ const rewardsApi = safesApi.injectEndpoints({
...
+      string,
+      RewardsClaimGasEstimateParams
+    >({
+      async queryFn(params) {
+        return queryPromiseWrapper<string, RewardsClaimGasEstimateParams>(
+          getClaimRewardsGasEstimate,
+          params,
+          {
+            errorLogMessage: 'Error fetching reward claim gas fee',
+            timeout: 60000, // 1 min
```
oidoug: nit: Is it necessary to set a timeout here? If so, extract timeout to a constant at the top of the file.

oidoug (approved):
Good job! I see this new data fetch hook being used in the Wallet screen for showing rewards counting.

Just some minor comments.

cardstack/src/services/rewards-center/rewards-center-api.ts
paulinhapenedo: This is the next PR I'm about to open after I merge this one :)

cardstack/src/services/rewards-center/rewards-center-api.ts
paulinhapenedo: You're right, the previous implementation didn't have one. I copied and pasted from one the endpoints in the file and didn't pay attention to it. Removing it right now!

PR #875, opened 2022-06-02
[CS-3986] Rewards: move all claim functions to custom hook

paulinhapenedo:
### Description
This PR moves all claim functions into its own hook with test coverage.

One minor thing I changed visually was to add a loading skeleton to the claim sheet instead of the default `ActivityIndicator` (video below).

Again, thanks @danibonilha for the help with testing.

- [x] Completes #CS-3986

### Screenshots

https://user-images.githubusercontent.com/690904/171737728-8619f94f-b38d-43d3-a97c-1a16fa545b73.mp4

### DIFF
--- a/cardstack/src/screens/RewardsCenterScreen/RewardsClaimSheet/useRewardsClaim.ts
+++ b/cardstack/src/screens/RewardsCenterScreen/RewardsClaimSheet/useRewardsClaim.ts
@@ -1,0 +1,145 @@
+import { StackActions, useNavigation } from '@react-navigation/native';
+import { useCallback, useMemo } from 'react';
+
+import { defaultErrorAlert } from '@cardstack/constants';
+import { useMutationEffects } from '@cardstack/hooks';
+import { useLoadingOverlay } from '@cardstack/navigation';
+import {
+  useClaimRewardsMutation,
+  useGetClaimRewardsGasEstimateQuery,
+} from '@cardstack/services/rewards-center/rewards-center-api';
+import {
+  RewardsClaimData,
+  TransactionConfirmationType,
+} from '@cardstack/types';
+
+import { Alert } from '@rainbow-me/components/alerts';
+import { useWallets } from '@rainbow-me/hooks';
+
+import { strings } from '../strings';
+import useRewardsDataFetch from '../useRewardsDataFetch';
+
+interface onClaimCallbackProps {
+  title: string;
+  message: string;
+  popStackNavigation?: number;
+}
+
+const useRewardsClaim = () => {
+  const { goBack, dispatch } = useNavigation();
+  const { signerParams, accountAddress } = useWallets();
+
+  const {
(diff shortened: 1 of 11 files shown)


cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
```diff
@@ -40,10 +41,16 @@ export const ClaimContent = ({
...
-  isLoadingClaimGas,
 }: ClaimContentProps) => {
   const { TabHeader, currentTab } = useTabHeader({ tabs });
 
+  const { navigate } = useNavigation();
+
+  const onClaimPress = useCallback(
+    item => () => navigate(Routes.REWARDS_CLAIM_SHEET, item),
+    [navigate]
+  );
```
danibonilha: Since the info is being extracted from mainPoolToken directly, there's no need to pass the item thru navigation, it can be a simple callback, If we eventually have more tokens then it makes sense to pass foward

cardstack/src/screens/RewardsCenterScreen/RewardsClaimSheet/useRewardsClaim.ts
```diff
@@ -0,0 +1,147 @@
...
+    });
+  }, [
+    showLoadingOverlay,
+    claimRewards,
+    partialClaimParams,
+    signerParams,
+    defaultRewardProgramId,
+  ]);
+
+  const onClaimCallback = useCallback(
```
danibonilha: superNit: the name of the function could be related with alert, or something, `onClaimFulfilledAlert` , idk something like that.

cardstack/src/screens/RewardsCenterScreen/RewardsClaimSheet/useRewardsClaim.ts
```diff
@@ -0,0 +1,147 @@
...
+        message,
+        title,
+        buttons: [
+          {
+            text: strings.defaultAlertBtn,
+            onPress: popStackNavigation
+              ? () => {
+                  dispatch(StackActions.pop(popStackNavigation));
+                }
+              : undefined,
```
danibonilha: superNit: I'm wondering now,  if we still need to use pop here, considering its always 1 pop, we could probably just use goBack and one dismiss flag. Since there are still refactoring for these features, you can check that on a next PR if you'd like.

danibonilha (approved):
This is so much cleaner and better scoped! Nice job! 🚀

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
paulinhapenedo: Good catch! If we ever need to pass a different token, the whole approach will have to change because the `item` prop today just passes display info about the token: coinSymbol and text.

PR #876, opened 2022-06-03
[CS-3990] Logic model for local authentication

oidoug:
### Description

This PR adds expo's LocalAuthentication lib for our biometry check, also:
- Slice to save authentication value in redux;
- Hook to check for auth, authenticate and revoke more easily;
- AppStateUpdate dummy component to control local auth revoke when the app goes to background state.

Notes on the flow:
- When Unlock screen is opened, if authenticated it navigates straight away to Wallet screen. This looks odd and we should use `isAuthenticated` from `useLocalAuth` to decide the route before showing Unlock screen.
- Auth is stored in memory, so it clears on app refreshes or relaunches.
- Auth is revoked when app enters background mode.
- So in the video below, I first test going into background state to see if auth is revoked and then a second time to see if state is still authenticated.

- [x] Completes #(CS-3990)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/171870581-37ffa1bb-9d91-4fe1-944f-f3b41a283bdb.mp4

### DIFF
--- a/cardstack/src/components/AppStateUpdate/AppStateUpdate.tsx
+++ b/cardstack/src/components/AppStateUpdate/AppStateUpdate.tsx
@@ -1,0 +1,18 @@
+import { useEffect } from 'react';
+
+import { useLocalAuth } from '@cardstack/hooks/useLocalAuth';
+
+import useAppState from '@rainbow-me/hooks/useAppState';
+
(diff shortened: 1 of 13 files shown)


oidoug:
Superseded by merged PR #877

PR #877, opened 2022-06-03
[CS-3992]: Add auth/non-auth routes and app lock based on PIN 

danibonilha:
### Description
This PR adds the ability to lock the app when it goes to background.

- Adds redux slice to handle auth 
- Updates the UnlockScreen,  removing the login btn 
- Makes cardLogo handle proportional size, to cleanup welcome/unlock screen
- Adds the logic that authorizes the user to use the app with the PIN 
- Splits the navigation based on auth flags, there's a quirk here using shared routes, ideally it would not be required to call nav reset, but since the routes are shared, once we have the address, the navigator doesn't navigate automatically, this can be simplified, using the `Group` component and `navigationKey` of react-navigation, but unfortunately it's only available on version 6. (which might be trick to update right now based on the breaking changes).
- Handles create/delete flows

### Screenshots

[GIF on Linear](https://linear.app/cardstack/issue/CS-3992#comment-9a57ad08)

### DIFF
--- a/cardstack/src/components/AppRequirementsCheck/AppRequirementsCheck.tsx
+++ b/cardstack/src/components/AppRequirementsCheck/AppRequirementsCheck.tsx
@@ -29,14 +29,16 @@ export const AppRequirementsCheck = ({ children }: Props) => {
   }, [isReady]);
 
   if (isReady && remoteFlags().maintenanceActive) {
+    hideSplashScreen();
+
     return <MaintenanceMode message={remoteFlags().maintenanceMessage} />;
   }
 
   if (forceUpdate) {
+    hideSplashScreen();
+
     return <MinimumVersion />;
   }
-
-  if (isReady) hideSplashScreen?.();
 
   return children;
 };
--- a/cardstack/src/components/CardwalletLogo/CardwalletLogo.tsx
+++ b/cardstack/src/components/CardwalletLogo/CardwalletLogo.tsx
@@ -1,4 +1,4 @@
-import React, { memo } from 'react';
+import React, { memo, useMemo } from 'react';
 
 import { CenteredContainer, Icon, Text } from '@cardstack/components';
 import {
@@ -6,11 +6,14 @@ import {
   ThemeVariant,
 } from '@cardstack/theme/colorStyleVariants';
 
+import { useDimensions } from '@rainbow-me/hooks';
+
 type CardwalletLogoSizes = 'big' | 'medium';
 
 interface Props {
   size?: CardwalletLogoSizes;
   variant?: ThemeVariant;
+  proportionalSize?: boolean;
 }
 
 export const DEFAULT_CARDWALLET_ICON_SIZE = 100;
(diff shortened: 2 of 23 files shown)


cardstack/src/components/AppRequirementsCheck/AppRequirementsCheck.tsx
```diff
@@ -29,14 +29,16 @@ export const AppRequirementsCheck = ({ children }: Props) => {
...
     return <MaintenanceMode message={remoteFlags().maintenanceMessage} />;
   }
 
   if (forceUpdate) {
+    hideSplashScreen();
+
     return <MinimumVersion />;
   }
 
-  if (isReady) hideSplashScreen?.();
```
oidoug, line 39: If neither maintenance nor minimum version conditions are met, where does it hide the slash then?

cardstack/src/components/CardwalletLogo/CardwalletLogo.tsx
```diff
@@ -42,24 +45,35 @@ const sizes = {
...
-      <Text
-        color={colorStyleVariants.textColor[variant]}
-        weight="bold"
-        {...sizes[size].subtitle}
-      >
-        {strings.subtitle}
-      </Text>
-    </CenteredContainer>
-  )
+  ({ size, variant = 'dark', proportionalSize = false }: Props) => {
```
oidoug, line 48: You are only using the logo with `proportinalSize`, might just make it the default.

cardstack/src/screens/UnlockScreen/useUnlockScreen.tsx
```diff
@@ -1,48 +1,47 @@
...
+
   const { biometryAvailable } = useBiometricSwitch();
-  const appVersion = useAppVersion();
   const [inputPin, setInputPin] = useState('');
   const [pinInvalid, setPinInvalid, setPinValid] = useBooleanState();
 
-  const { variant = 'dark' } = params;
+  const { setUserAuthorized } = useAuthActions();
+
+  // Fetch on init so login it's faster
```
oidoug, line 25:
```suggestion
  // Fetch on init so login is faster
```

src/App.js
```diff
@@ -195,6 +195,12 @@ class App extends Component {
   };
 
   handleAppStateChange = async nextAppState => {
+    const disabledStates = ['background', 'inactive'];
+
+    if (disabledStates.includes(this.state.appState)) {
+      store.dispatch(authSlice.actions.setUserUnauthorized());
+    }
+
```
oidoug, line 203: Might later refactor to use a component.

oidoug (approved):
Great work you did, I liked the design on having hooks with the slice.

I'll try to fit my changes on top of these.

cardstack/src/components/AppRequirementsCheck/AppRequirementsCheck.tsx
danibonilha: [here](https://github.com/cardstack/cardwallet/pull/877/files#diff-619a4b27328f68c441b67a715d4edb54ab89fb0b6afbab1278df249ad1212343R176), after fetching the initial authInfo

src/App.js
danibonilha: The ideal scenario is to migrate the `App.js` to `App.ts` and make it a functional component, until then, we need to balance the creation of new components in the tree, just to "work around" the main issue, so i'd rather add the background implementation, where it already exists.

cardstack/src/components/CardwalletLogo/CardwalletLogo.tsx
danibonilha: ya, I mostly went with defaulting to false because `proportionalSize` will overwrite any size that is eventually passed  `proportionalSize || !size`, and I thought it might be confusing, but let me know if think it's better default to true still and i can add on a upcoming PR

PR #878, opened 2022-06-03
[CS-3991]: Add pin flow while importing with seed

danibonilha:
### Description

This PR adds the ability to create a pin while importing a wallet, it also fixes the Show label which was being displayed on the wrong side of the screen

### Screenshots

Simulator Screen Recording - iPhone 11 - 2022-06-03 at 17 25 19


oidoug (approved):
LGTM

PR #879, opened 2022-06-06
[chore] iOS proj changes for building in intel and apple silicon

oidoug:
### Description

This PR includes a change in iOS's `project.pbxproj` to make the build dependent on active architecture only, so it works without project changes on both Intel and Apple silicon macs.

It also includes instructions on setting up the project in Apple silicon using brew since it already has a version with all packages built natively.


oidoug:
- Running action to validate new build config.

oidoug:
@danibonilha pointed out to me that it's important to use bundle to keep versions of tools locked down. So I'll need to revisit this setup.

oidoug:
I'll get back to this soon, not a priority now.

PR #882, opened 2022-06-07
[CS-4012]: Fix restore backup from init

danibonilha:
### Description
There was a lot of back and forth trying to add the createPin flow on the initial backup restore, and we need to rethink how the backup works in oder to add a consistent flow,  so for now, the backup restore will work with the same implementation of an "existing account", which is to prompt the Create flow once the wallet has load and no PIN is found. So this PR basically fixes the restore flow, which was on an infinite loop.

### Screenshots

https://user-images.githubusercontent.com/20520102/172478282-d0b47ea5-f8e6-4aa9-b081-2d77b9eeb04a.mov


oidoug (approved):
LGTM, works fine for me.

PR #884, opened 2022-06-08
[CS-3369] Show contextual rewards banner on home screen

paulinhapenedo:
### Description
This PR adds logic to the `RewardsPromoBanner` component in order to have contextual messages showing to the user depending on whether they have rewards to claim or not.

To get the rewards icon to be monochromatic, I had to add a custom prop to the `CustomIcon` component. The default rewards icon has two colors, so I couldn't use `color` in both the SVG and the path. I'm open to discussion on this approach, though.  

- [x] Completes [#CS-3369](https://linear.app/cardstack/issue/CS-3369/show-rewards-available-banner-on-home-screen-when-reward-is-available)

### Checklist

### Screenshots

Screenshot


cardstack/src/components/Icon/Icon.tsx
```diff
@@ -42,6 +42,7 @@ export interface IconProps extends Omit<TouchableProps, 'children'> {
   stroke?: ColorTypes | null;
   strokeWidth?: number;
   visible?: boolean;
+  pathFillColor?: ColorTypes | null;
```
oidoug: nit: If we are keeping this approach, better rename this prop to something more obvious, maybe `forceFillColor` and add a comment on why it's there.

cardstack/src/components/RewardsPromoBanner/__tests__/useRewardsPromoBanner.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  it(`should NOT show an icon inside the button if the user doesn't have rewards to claim`, () => {
+    (useRewardsDataFetch as jest.Mock).mockImplementation(() => ({
+      mainPoolTokenInfo: undefined,
+    }));
+
+    const { result } = renderHook(() => useRewardsPromoBanner());
+
+    expect(result.current.buttonIcon).toBe(undefined);
+  });
+});
```
oidoug, line 64: 👏

cardstack/src/components/RewardsPromoBanner/useRewardsPromoBanner.ts
```diff
@@ -0,0 +1,51 @@
+import { useNavigation } from '@react-navigation/native';
+import { useCallback, useMemo } from 'react';
+
+import { Routes } from '@cardstack/navigation';
+import useRewardsDataFetch from '@cardstack/screens/RewardsCenterScreen/useRewardsDataFetch';
+
+import rewardsPromoBanner from '../../assets/promo-rewards.png';
+import rewardsUnclaimedBanner from '../../assets/rewards-unclaimed.png';
```
oidoug: nit: It would be nice to have the text extracted from the image and added by code.

oidoug (approved):
LGTM!
Just a couple of nits.

About the icon color, instead of having a new prop for Icon, I would prefer a copied version of the icon that is monochromatic and respects the color prop. Less confusing long term.

cardstack/src/components/RewardsPromoBanner/useRewardsPromoBanner.ts
paulinhapenedo: I don't think I have the option on Zeplin to export only the background image. Here's how I can select the image there. And that would also mean changing what we currently have too... don't know if this would be a significant gain for the end user.

Screen Shot 2022-06-09 at 10 26 58

danibonilha: You can extract as an svg and delete the text group, or we can ask the design team to provide the asset, I agree that it might be better to have it as a reusable banner, where we can add contextual text and btns, instead of having diff `.pngs` for each case

cardstack/src/components/Icon/custom-icons/rewards.tsx
```diff
@@ -1,13 +1,17 @@
...
+}
+
+const SvgComponent = (props: RewardsIconProps) => (
   <Svg width={30} height={21.672} viewBox="0 0 25 21.672" {...props}>
     <Path
       data-name="Union 52"
       d="M3.824 15.288A3.829 3.829 0 0 1 0 11.464V8.921a1.274 1.274 0 1 1 2.548 0v2.543a1.276 1.276 0 0 0 1.276 1.278h14.013a1.275 1.275 0 0 0 1.272-1.278v-7.64a1.277 1.277 0 0 0-1.272-1.276h-3.186a1.274 1.274 0 1 1 0-2.548h3.186a3.83 3.83 0 0 1 3.82 3.824v7.64a3.828 3.828 0 0 1-3.82 3.824z"
       transform="translate(3.342 5.152)"
-      fill="#37eb77"
+      fill={props.pathFillColor || '#37eb77'}
```
danibonilha: I'm wondering if it's possible to do the reverse,  and have `#37eb77` to be received as props, bc if the goal is to be monochromatic, passing the same color twice is a bit odd, something like, overwriteFillColor, but even so, it only applies to the reward icon, which might be confusing, let's talk about this.

paulinhapenedo:
After some discussion on Discord, we decided to:

1. Remove the icon from the button and handle the monochromatic icon issue in the future.
2. Update the reward banner image to NOT have text in it. Contextual changes are textual and controlled by the `strings.ts` file.
3. Checking if the user has rewards to claim is being handled in the `AssetList` component, so the home screen loading takes into consideration that as well.

**User WITH rewards to claim**
https://user-images.githubusercontent.com/690904/172935781-91f68df6-dd9d-4016-ac88-e5b636fba2f2.mp4

**User WITHOUT rewards to claim**
https://user-images.githubusercontent.com/690904/172935951-e2a54b1b-0078-40ae-9818-c14a829d4e12.mp4

danibonilha (approved):
Awesome!

PR #885, opened 2022-06-09
[CS-4019] Tests to unlock screen hook

oidoug:
### Description

Add a comprehensive list of tests to Unlock Screen.

I had difficulties getting the `getPin` return to properly reflect in the hook before checking values. Eventually, I figure it out, but it took a while 😓 

- [x] Completes #(CS-4019)

### DIFF
--- a/cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
+++ b/cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
@@ -93,8 +93,6 @@ describe('useUnlockScreen', () => {
     expect(mockSetAuthorized).toBeCalledTimes(0);
   });
 
-  it('should automatically call auth check if biometry toggle enabled', () => {});
-
   it('should authorize if biometry check is successful', async () => {
     mockBiometryAvailableHelper(true);
     mockAuthAuthorizedHelper(true);
--- a/cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
+++ b/cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
@@ -62,7 +62,7 @@ describe('useUnlockScreen', () => {
     jest.clearAllMocks();
   });
 
-  it('should authorized be called when inputed PIN match stored one', async () => {
+  it('should call setAuthorized when inputed PIN matches stored one', async () => {
     const { result } = renderHook(() => useUnlockScreen());
 
     // We need to wait getPin be resolved so storedPin.current is valid.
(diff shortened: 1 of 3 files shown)


cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    mockBiometryAvailableHelper();
+    mockAuthAuthorizedHelper();
+    mockGetPinHelper();
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should authorized be called when inputed PIN match stored one', async () => {
```
danibonilha: nit: the it, it's a helper to make the test case readability better, so we can reorder the words here, `it('should call setAuthorized when inputed PIN matches stored one')`

cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    act(() => {
+      result.current.setInputPin(PIN);
+    });
+
+    expect(result.current.inputPin).toMatch(PIN);
+    expect(result.current.pinInvalid).toBeFalsy();
+    expect(mockSetAuthorized).toBeCalledTimes(1);
+  });
+
+  it('should PIN be invalid and authorized not be called when inputed PIN doest match stored one', async () => {
```
danibonilha: `it('should not call setAuthorized if inputed PIN does not match stored one and set PIN as invalid')`

cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    await waitFor(() => expect(getPin).toHaveBeenCalled());
+
+    act(() => {
+      result.current.setInputPin(wrongPIN);
+    });
+
+    expect(result.current.inputPin).toMatch('');
+    expect(result.current.pinInvalid).toBeTruthy();
+    expect(mockSetAuthorized).toBeCalledTimes(0);
```
danibonilha: FYI: we can also use `expect(mockSetAuthorized).not.toBeCalled();`

cardstack/src/screens/UnlockScreen/useUnlockScreen.tsx
```diff
@@ -97,11 +99,6 @@ export const useUnlockScreen = () => {
...
   }, [authenticateBiometrically, isBiometryEnabled]);
 
-  const retryBiometricLabel = useMemo(
-    () => `Try ${biometryLabel || 'device authentication'} again`,
-    [biometryLabel]
-  );
-
   return {
     inputPin,
     setInputPin,
```
danibonilha, line 107: I believe we should drop the changes on this file, besides the reordering stuff

oidoug: Thanks for the pointer, I forgot to revert those 👍

danibonilha (approved):
Way to go!

PR #886, opened 2022-06-09
[CS-4022] Adds bigger hit area for reset link.

oidoug:
### Description

Increased hit slop for reset link in unlock screen.

- [x] Completes #(CS-4022)

### Checklist

### Screenshots

https://user-images.githubusercontent.com/129619/172899412-8593bc3e-a039-4bfd-8f36-a55fd2c12416.mp4

### DIFF
--- a/cardstack/src/utils/layouts.ts
+++ b/cardstack/src/utils/layouts.ts
@@ -8,19 +8,15 @@ export const sectionStyle = StyleSheet.create({
 const SMALL = 5;
 const MEDIUM = 15;
 
+const hitSlopBuilder = (size: number) =>
+  ['top', 'bottom', 'left', 'right'].reduce(
+    (acc, value) => ({ ...acc, [value]: size }),
+    {}
+  );
+
 export const hitSlop = {
-  small: {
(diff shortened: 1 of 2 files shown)


cardstack/src/utils/layouts.ts
```diff
@@ -5,12 +5,21 @@ export const sectionStyle = StyleSheet.create({
...
+    left: SMALL,
+    right: SMALL,
+  },
+  medium: {
+    top: MEDIUM,
+    bottom: MEDIUM,
+    left: MEDIUM,
+    right: MEDIUM,
   },
 };
```
danibonilha, line 20: nit: we could add a helper function here, for the cases the hitSlop is the same 
```ts
const hitSlopBuilder = (size: number) =>
    ['top', 'bottom', 'left', 'right'].reduce(
      (acc, value) => ({ ...acc, [value]: size }),
      {}
    );
    
export const hitSlop = {
  small: hitSlopBuilder(SMALL),
...
```

image

danibonilha (approved):

PR #887, opened 2022-06-09
[CS-3370] Show rewards available count on MainHeader rewards icon

paulinhapenedo:
### Description
This PR adds the rewards available count to the Rewards icon that lives inside the MainHeader component.  To get the count number, I'm using the information that comes from the `rewardPoolTokenBalances` query for future-proofing. Let me know if you think that we should be using `mainPoolTokenInfo` instead.

- [x] Completes [#CS-3370](https://linear.app/cardstack/issue/CS-3370/show-reward-claims-available-count-on-reward-icon)

### Checklist

### Screenshots

Screenshot

### DIFF
--- a/cardstack/src/components/MainHeader/MainHeader.tsx
+++ b/cardstack/src/components/MainHeader/MainHeader.tsx
@@ -8,11 +8,13 @@ import {
   MainHeaderWrapper,
 } from '@cardstack/components';
 import { Routes } from '@cardstack/navigation';
+import useRewardsDataFetch from '@cardstack/screens/RewardsCenterScreen/useRewardsDataFetch';
 
 import { networkInfo } from '@rainbow-me/helpers/networkInfo';
 import { useAccountSettings } from '@rainbow-me/hooks';
 
 import { ContainerProps } from '../Container';
+import { Touchable } from '../Touchable';
 
 interface Props extends ContainerProps {
   title?: string;
@@ -32,6 +34,7 @@ const MainHeader = ({
 }: Props) => {
   const { navigate } = useNavigation();
   const { network } = useAccountSettings();
+  const { rewardPoolTokenBalances } = useRewardsDataFetch();
 
   const onMenuPress = useCallback(() => navigate(Routes.SETTINGS_MODAL), [
     navigate,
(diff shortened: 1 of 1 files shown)


cardstack/src/components/MainHeader/MainHeader.tsx
```diff
@@ -41,6 +44,8 @@ const MainHeader = ({
     navigate(Routes.REWARDS_CENTER_SCREEN);
   }, [navigate]);
 
+  const hasRewardsClaim = rewardPoolTokenBalances?.length;
+
```
danibonilha: nit: the name gives the impression that it's a boolean value, it could be renamed to `availableRewards`

cardstack/src/components/MainHeader/MainHeader.tsx
```diff
@@ -71,13 +76,19 @@ const MainHeader = ({
...
+        <Touchable
+          flexDirection="row"
+          alignItems="center"
+          alignContent="center"
           onPress={onRewardPress}
-        />
+        >
+          <Icon color="teal" iconSize="medium" size={26} name="rewards" />
+          {!!hasRewardsClaim && (
+            <Text color="teal" fontSize={16} weight="bold" marginLeft={2}>
```
danibonilha: We either need to increase the fontSize, or make the icon smaller to match the designs

cardstack/src/components/MainHeader/MainHeader.tsx
```diff
@@ -71,13 +76,19 @@ const MainHeader = ({
...
       {!rightIcon ? (
-        <Icon
-          color="teal"
-          iconSize="medium"
-          size={26}
-          name="rewards"
+        <Touchable
+          flexDirection="row"
+          alignItems="center"
+          alignContent="center"
```
danibonilha: Maybe aligning with flex-end, would be a better fit since the rewards icons is bigger than the number
image

paulinhapenedo: Even with `flex-end` and the font-size adjustment from the previous comment, the icon and the text won't align to the bottom because of how the font behaves:

Screen Shot 2022-06-10 at 13 03 14

In order to have it really aligned to the bottom, we would have to tweak negative margin-bottom or something, which I don't like.

Also it's important to notice that seems like `<Text>` isn't affected by its parent `flex` definition, so I had to wrap the `Text` with a `Container`. It's a bit confusing, but I can show on Tuple what's happening.

danibonilha: It looks better already

danibonilha (approved):

PR #889, opened 2022-06-10
[CS-3873]: Prompt pin screen setup when user already has wallet setup 

danibonilha:
### Description
This PR adds the wallet migration to use the PIN, when a user already has a wallet set, it also adds usage for the new pin while creating/importing an account, as a result some methods were deprecated. Be on the lookout for potential bugs, and reach out if needed. There are some TODO's which will be handled on other PRs, I feel this is a good stopping point to have this PR merged, even if it's not 100% functional. The sooner we start testing the better to find bugs.

PS: The invalid request will be handled on a diff PR, bc I still see it happening while adding a new account.

A nice way to test it, is to install an older version, import an account with assets, then install this new version and check if you can still make transactions, pay merchant, and send other tokens.

Completes CS-3874
Completes CS-3873

### Screenshots

https://user-images.githubusercontent.com/20520102/173125351-ce41e61a-fcd1-412d-9f08-7b9e26ef8b6c.mov


src/handlers/authentication.js
```diff
@@ -1,14 +1,12 @@
...
 import AesEncryptor from '../handlers/aesEncryption';
 import * as keychain from '../model/keychain';
 import { pinKey } from '../utils/keychainConstants';
 import { Navigation, Routes } from '@cardstack/navigation';
-import logger from 'logger';
 
 const encryptor = new AesEncryptor();
 
-export async function getExistingPIN() {
+export async function DEPRECATED_getExistingPIN() {
```
paulinhapenedo, line 9: Nice prefix 👏

cardstack/src/navigation/tabBarNavigator.tsx
```diff
@@ -172,14 +173,15 @@ const useNavigationAuth = () => {
...
+      isFirstMount.current = false;
+    }
   }, []);
 
   useEffect(() => {
-    if (!hasWallet || !hasPin) {
-      loadAuthInfo();
-    }
-  }, [loadAuthInfo, hasWallet, hasPin]);
+    loadAuthInfo();
```
oidoug, line 183: `loadAuthInfo` will called twice, first when initializing, second when `isAuthorized` is true. Is that intended?

src/hooks/useWalletManager.ts
```diff
@@ -56,69 +65,134 @@ export default function useWalletManager() {
...
+              const seed = (await loadSeedPhrase(walletId, 'migration')) || '';
+              // keychain needs a delay in order to retrieve all values
+              await delay(1000);
+              seeds.push({ seed });
+            }
+
+            //TODO: replace with screen
+            Alert.alert(
+              'Multiple wallets',
+              `Looks like you have more than one wallet, cardstack from now on supports only single wallet,
```
oidoug, line 131:
```suggestion
              `Looks like you have more than one wallet, cardstack from now on supports only a single wallet,
```

oidoug (approved):
I've issues running an older version to set it up with a pin (trying last tag 1.1.0 - getting WS connection errors), so I couldn't test the PR properly.
Code looks good, I'm approving so you can move ahead.

cardstack/src/navigation/tabBarNavigator.tsx
danibonilha: For now yes, once we have the pin set for all flows we can remove one of the calls

src/hooks/useWalletManager.ts
danibonilha: this will be removed once we have the new screen

PR #890, opened 2022-06-13
[Bug]: Fix broken import flow

danibonilha:
### Description

For the import flow the pin was not being passed forward, and we were stuck on the initial screen, it also disables temp the keychain check, as it needs to be replaced with a secure storage one.


paulinhapenedo (approved):
Quick fix 👏

PR #891, opened 2022-06-13
[CS-4058] Pin: create basic screen to list all the seed phrases for backup

paulinhapenedo:
### Description
Pretty average screen for when the user has multiple wallets. In order for them to not loose access to their wallets, they need to manually save each wallet's seed phrase.

@danibonilha I used the approach of having the seed phrases being passed as navigation params. Let me know if you prefer another approach.

- [x] Completes [#CS-4058](https://linear.app/cardstack/issue/CS-4058/[flow]-add-screen-to-show-seed-phrases-and-allow-copy)

### Checklist

### DIFF
--- a/cardstack/src/screens/SeedPhraseBackup/SeedPhraseBackup.tsx
+++ b/cardstack/src/screens/SeedPhraseBackup/SeedPhraseBackup.tsx
@@ -14,10 +14,8 @@ import {
   SafeAreaView,
   Text,
 } from '@cardstack/components';
-import { useBottomToast } from '@cardstack/hooks';
+import { useCopyToast } from '@cardstack/hooks';
 import { RouteType } from '@cardstack/navigation/types';
-
-import { useClipboard } from '@rainbow-me/hooks';
 
 import { strings } from './strings';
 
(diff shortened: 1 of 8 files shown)


cardstack/src/screens/SeedPhraseBackup/SeedPhraseBackup.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { strings } from './strings';
+
+interface SeedPhraseBackupParams {
+  seedPhrases: string[];
+}
+
+const SeedPhraseBackup = () => {
+  const { params } = useRoute<RouteType<SeedPhraseBackupParams>>();
+  const { setClipboard } = useClipboard();
+  const { ToastComponent, showToast } = useBottomToast();
```
danibonilha: nit: we have the `useCopyToast` component, which already has the clipboard in it, maybe it's missing just an optional  param on the `copyToClipboard` function, so you can choose which data to copy while calling the function

cardstack/src/screens/SeedPhraseBackup/SeedPhraseBackup.tsx
```diff
@@ -0,0 +1,104 @@
...
+        </CenteredContainer>
+      </TouchableOpacity>
+    ),
+    [copySeedPhrase]
+  );
+
+  const onPress = useCallback(() => {
+    showToast({
+      label: `Go Back`,
+    });
```
danibonilha: I think we can add a todo here, and remove this showToast, something like, // TODO: go to create pin flow

cardstack/src/screens/SeedPhraseBackup/strings.ts
```diff
@@ -0,0 +1,6 @@
+export const strings = {
+  button: 'OK, I copied all my seed phrases',
```
danibonilha: nit: `Ok, I've saved my seed phrases`

danibonilha (approved):
This screen looks great, nice job! Just added some small nits.

cardstack/src/screens/SeedPhraseBackup/SeedPhraseBackup.tsx
oidoug: I agree, I also don't like to leave "visual placeholders" in merged code, better just a to-do comment along with a ticket so we don't forget.

oidoug (approved):
Good! UI looks inspired in the new onboarding screens, we might just re-align some styling once that project starts.

PR #893, opened 2022-06-14
[CS-3954] Adds retry to hub-api; refactors push token queries to RTKQ

oidoug:
### Description

This PR adds a retry mechanism if the stored hub token is rejected, so a new token is issued. It also moves register and unregister PUSH token calls that uses the hub token to use the retry mechanism in RTK-Query.

- [x] Completes #(CS-3954)

(no UI changes)


android/app/src/main/AndroidManifest.xml
```diff
@@ -46,5 +46,8 @@
             <data android:host="wallet.cardstack.com" />
         </intent-filter>
       </activity>
+
+      <activity android:name="com.facebook.flipper.android.diagnostics.FlipperDiagnosticActivity"
```
danibonilha: What is this for ?

cardstack/src/services/hub/hub-api.ts
```diff
@@ -65,6 +68,26 @@ export const hubApi = createApi({
...
+      RegisterFCMTokenQueryParams
+    >({
+      query: ({ fcmToken }) => ({
+        url: routes.registerFCMToken,
+        method: 'POST',
+        body: hubBodyBuilder('push-notification-registration', {
+          'push-client-id': fcmToken,
+        }),
+      }),
+      transformResponse: (response: any) => ({ success: !!response.data }),
```
danibonilha: nit: maybe we don't need to transform the response, we can just check data directly in the component, or even better check, `if(!error)`  from the query, since the data doesn't matter to us

cardstack/src/services/hub/hub-service.ts
```diff
@@ -29,17 +29,22 @@ export const fetchHubBaseQuery: BaseQueryFn<
   unknown,
   FetchBaseQueryError,
   BaseQueryExtraOptions
-> = async (args, api, extraOptions = { authenticate: true }) => {
+> = async (args, api, extraOptions) => {
+  const extraOptionsOverwrite = {
+    authenticate: true,
+    ...extraOptions,
+  };
```
danibonilha, line 36: Can you elaborate the necessity of this overwrite ?

cardstack/src/services/hub/hub-api.ts
```diff
@@ -65,6 +68,26 @@ export const hubApi = createApi({
...
+    }),
+    unregisterFcmToken: builder.mutation<void, RegisterFCMTokenQueryParams>({
+      query: ({ fcmToken }) => ({
+        url: `${routes.registerFCMToken}/${fcmToken}`,
+        method: 'DELETE',
+        responseHandler: response => response.text(),
```
oidoug, line 83: Note: Both register and unregister hub endpoints responds with text (not json) but only unregister fails if rtk tries to parse it as json.

android/app/src/main/AndroidManifest.xml
oidoug: Sorry, flipper isn't connecting on my android device, so I was trying to fix it. These were leftovers, removed.

cardstack/src/models/firebase.ts
```diff
@@ -54,11 +54,11 @@ export const removeFCMToken = async (address: string) => {
...
       addressesByNetwork[network].includes(address)
     ) {
-      const unregisterResponse = await unregisterFcmToken(fcmToken);
+      const response = await unregisterFcmToken(fcmToken);
 
-      logger.sentry('UnregisterFcmToken response ---', unregisterResponse);
+      logger.sentry('UnregisterFcmToken response ---', response);
 
-      if (unregisterResponse?.success) {
+      if ('data' in response) {
```
danibonilha, line 59: same suggestion to check for `!error`

cardstack/src/services/hub/hub-service.ts
oidoug: This is necessary if and when we add more extra options, since the extraOptions obj passed in the query overwrites the default parameter we had before. I found this because I was trying to add an option for content type, that endup not being needed.

danibonilha (commented):
It's looking good, and props on migrating the push notification api

cardstack/src/services/hub/hub-api.ts
oidoug: Ok, it makes sense. Did that way to look closer to what we had for that call, but it isn't necessary.

danibonilha (approved):

cardstack/src/models/firebase.ts
oidoug: This contains assertion with a string key is necessary because the unregisterFcmToken returns a union type from error and data, so TS can't decide which object is present in the response.

PR #895, opened 2022-06-14
[CS-4087]: Fix push notification and migrate app to .ts

danibonilha:
### Description
It seems dapp notifications are not working, since the remote config migration, because the baseUrl was removed, this PR adds thats. Also all the notification handling was a bit confused on app nit, so, the App.js was migrated to typescrypt and it's now a functional component, which gives us more flexibility to reuse hooks.  it's not related with the ticket feature, but since I was migrating the App, I had to test a bunch of stuff, and noticed our maintenance screen was pretty odd, so I did some tweaking to make it better.

I tested the notifications, but didn't record bc I was having an issue with the device connected.

PS: with this change, the unlock state will be way easier
### Screenshots

Before:
image

After:
image


cardstack/src/components/AppRequirementsCheck/AppRequirementsCheck.tsx
```diff
@@ -28,17 +22,17 @@ export const AppRequirementsCheck = ({ children }: Props) => {
...
 
-  return children;
+  return {
+    forceUpdate,
+    maintenance: {
+      active: maintenanceMode,
+      message: remoteFlags().maintenanceMessage,
+    },
+  };
 };
```
oidoug, line 38: nit: Rename the file to `useAppRequirementsCheck` and move to hooks folder.

src/redux/walletconnect.js
```diff
@@ -48,6 +47,9 @@ const WALLETCONNECT_SET_PENDING_REDIRECT =
 const WALLETCONNECT_REMOVE_PENDING_REDIRECT =
   'walletconnect/WALLETCONNECT_REMOVE_PENDING_REDIRECT';
 
+const baseCloudFunctionsUrl =
+  'https://us-central1-card-pay-3e9be.cloudfunctions.net';
+
```
oidoug, line 52: Oh, I see... sorry about that. It's unfortunate the linter didn't catch the missing value.
Thanks for fixing it.

src/screens/MaintenanceMode.js
```diff
@@ -1,29 +1,20 @@
...
+      <CardwalletLogo />
+      <Container padding={8}>
+        <Text color="underlineGray" textAlign="center">
+          {message || DEFAULT_MESSAGE}
+        </Text>
       </Container>
-    </Page>
+    </CenteredContainer>
   );
 }
```
oidoug, line 20: Nice!

src/useAppInit.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  useEffect(() => {
+    const handleWcDeepLink = async () => {
+      try {
+        const initialUrl = await Linking.getInitialURL();
+        if (initialUrl) {
+          handleDeepLink(initialUrl);
+        }
+      } catch (e) {
+        Logger.log('Error opening deeplink', e);
```
oidoug: nit: Isn't better to send this to Sentry? Seems like something we should keep an eye at.

oidoug (approved):
LGTM, I appreciate the refactor and fix 👏

src/useAppInit.ts
danibonilha: good catch, hadn't notice this one

src/redux/walletconnect.js
danibonilha: ya, not sure why the linter didn't flag it, maybe bc of the index export idk

PR #897, opened 2022-06-15
[CS-4023] Enhance resetWallet with secure storage + keychain cleanup

paulinhapenedo:
### Description
This PR adds two new methods to handle the complete reset of the current wallet on the app. 

- `wipeSecureStorage` handles everything that's in the secure storage. 
- calling `resetHasWallet` handles the user redirection to the auth screen.
- added some logs after the deletion of pin, seed phrase and private key to help debug. Can be removed before merging to develop.

- [x] Completes CS-4023

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/models/secure-storage.ts
+++ b/cardstack/src/models/secure-storage.ts
@@ -78,14 +78,16 @@ const setEncryptedItem = async (item: string, key: Keys, pin: string) => {
 const buildKeyWithId = (key: Keys, id: string) => `${key}_${id}`;
 
 // PIN
-const savePin = (pin: string) => SecureStore.setItemAsync(keys.AUTH_PIN, pin);
+const savePin = async (pin: string) =>
+  await SecureStore.setItemAsync(keys.AUTH_PIN, pin);
 
-const getPin = () => getSecureValue(keys.AUTH_PIN);
+const getPin = async () => await getSecureValue(keys.AUTH_PIN);
 
-const deletePin = () =>
-  SecureStore.deleteItemAsync(keys.AUTH_PIN).then(() =>
-    logger.log('Deleted PIN')
-  );
+const deletePin = async () => {
+  await SecureStore.deleteItemAsync(keys.AUTH_PIN);
+
+  logger.log('Deleted PIN');
+};
 
 const encryptor = new AesEncryptor();
 
@@ -107,9 +109,9 @@ const getSeedPhrase = async (walletId: string) => {
 const deleteSeedPhrase = async (walletId: string) => {
   const seedKey = buildKeyWithId(keys.SEED, walletId);
 
-  await SecureStore.deleteItemAsync(seedKey).then(() =>
-    logger.log('Deleted seed phrase')
-  );
+  await SecureStore.deleteItemAsync(seedKey);
+
+  logger.log('Deleted seed phrase');
 };
 
 // PRIVATE_KEY
@@ -136,9 +138,9 @@ const getPrivateKey = async (walletAddress: string) => {
 const deletePrivateKey = async (walletAddress: string) => {
   const pKey = buildKeyWithId(keys.PKEY, walletAddress);
 
-  await SecureStore.deleteItemAsync(pKey).then(() =>
-    logger.log('Deleted private key')
-  );
+  await SecureStore.deleteItemAsync(pKey);
+
+  logger.log('Deleted private key');
 };
 
 const wipeSecureStorage = async (wallets: AllRainbowWallets) => {
--- a/cardstack/src/models/secure-storage.ts
+++ b/cardstack/src/models/secure-storage.ts
@@ -2,6 +2,7 @@ import * as SecureStore from 'expo-secure-store';
 import { SECURE_STORE_KEY } from 'react-native-dotenv';
 
 import AesEncryptor from '@rainbow-me/handlers/aesEncryption';
+import { AllRainbowWallets } from '@rainbow-me/model/wallet';
 import logger from 'logger';
 
 const keys = {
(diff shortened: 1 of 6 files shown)


oidoug (approved):
Tested on both, and all working, good job!

cardstack/src/models/secure-storage.ts
```diff
@@ -103,7 +107,9 @@ const getSeedPhrase = async (walletId: string) => {
 const deleteSeedPhrase = async (walletId: string) => {
   const seedKey = buildKeyWithId(keys.SEED, walletId);
 
-  await SecureStore.deleteItemAsync(seedKey);
+  await SecureStore.deleteItemAsync(seedKey).then(() =>
+    logger.log('Deleted seed phrase')
+  );
```
danibonilha: it seems we don't need to use `then` here, we can log after the await

cardstack/src/models/secure-storage.ts
```diff
@@ -130,7 +136,24 @@ const getPrivateKey = async (walletAddress: string) => {
 const deletePrivateKey = async (walletAddress: string) => {
   const pKey = buildKeyWithId(keys.PKEY, walletAddress);
 
-  await SecureStore.deleteItemAsync(pKey);
+  await SecureStore.deleteItemAsync(pKey).then(() =>
+    logger.log('Deleted private key')
+  );
```
danibonilha: ditto

cardstack/src/models/secure-storage.ts
```diff
@@ -81,7 +82,10 @@ const savePin = (pin: string) => SecureStore.setItemAsync(keys.AUTH_PIN, pin);
 
 const getPin = () => getSecureValue(keys.AUTH_PIN);
 
-const deletePin = () => SecureStore.deleteItemAsync(keys.AUTH_PIN);
+const deletePin = () =>
+  SecureStore.deleteItemAsync(keys.AUTH_PIN).then(() =>
+    logger.log('Deleted PIN')
+  );
```
danibonilha: nit: we mostly use await, it would be nice to use the same pattern

cardstack/src/models/secure-storage.ts
```diff
@@ -130,7 +136,24 @@ const getPrivateKey = async (walletAddress: string) => {
...
+
+    // loop through addresses to pass all .address to deletePrivateKey
+    for (const account of walletAddresses) {
+      await deleteSeedPhrase(walletId);
+      await deletePrivateKey(account.address);
+      await deletePin();
+    }
+  }
+
+  return;
```
danibonilha: what is this return for ?

cardstack/src/models/secure-storage.ts
```diff
@@ -130,7 +136,24 @@ const getPrivateKey = async (walletAddress: string) => {
...
+
+const wipeSecureStorage = async (wallets: AllRainbowWallets) => {
+  for (const walletId of Object.keys(wallets)) {
+    const walletAddresses = wallets[walletId].addresses;
+
+    // loop through addresses to pass all .address to deletePrivateKey
+    for (const account of walletAddresses) {
+      await deleteSeedPhrase(walletId);
+      await deletePrivateKey(account.address);
+      await deletePin();
```
danibonilha, line 153: we don't need to call deletePin and deleteSeedPhrase for every account of the wallet, those calls can be done outside the loop

danibonilha: nit: would Promise.allSettled work here ? Or trying to delete simultaneously errors ?

cardstack/src/screens/UnlockScreen/useUnlockScreen.tsx
```diff
@@ -73,7 +73,7 @@ export const useUnlockScreen = () => {
       {
         text: strings.reset.delete,
         onPress: async () => {
-          await wipeKeychain();
+          await resetWallet();
           RNRestart.Restart();
```
danibonilha: no nieed to call restart here, `onPress: resetWallet`

src/model/wallet.ts
```diff
@@ -845,3 +848,17 @@ export const migrateSecretsWithNewPin = async (
...
+
+  // clearing secure storage and keychain
+  if (allWallets) {
+    await wipeSecureStorage(allWallets);
+    await keychain.wipeKeychain();
+
+    logger.log('Wallet reset done!');
+
+    store.dispatch(authSlice.actions.resetHasWallet());
+  }
```
danibonilha, line 867: we should add try/catch here

cardstack/src/models/secure-storage.ts
```diff
@@ -130,7 +136,24 @@ const getPrivateKey = async (walletAddress: string) => {
...
-  await SecureStore.deleteItemAsync(pKey);
+  await SecureStore.deleteItemAsync(pKey).then(() =>
+    logger.log('Deleted private key')
+  );
+};
+
+const wipeSecureStorage = async (wallets: AllRainbowWallets) => {
+  for (const walletId of Object.keys(wallets)) {
+    const walletAddresses = wallets[walletId].addresses;
+
```
danibonilha, line 151: it would be nice to add try/catch here too.

cardstack/src/models/secure-storage.ts
paulinhapenedo: So, it's ok/helpful to leave the log (?)

cardstack/src/models/secure-storage.ts
paulinhapenedo: Leftover ✂️

cardstack/src/models/secure-storage.ts
paulinhapenedo: When I first tried, it wasn't passing but we can pair later to double-check. Maybe the order was wrong or something.

PR #898, opened 2022-06-15
[CS-4015]: Lock app keeping navigation state

danibonilha:
### Description

This PR fixes the missing navigation state issue while locking the app, it also updates the react-native types which were mismatched, and in other to set the proper listeners needed to be updated, bc the older implementation is being deprecated. 

~~UPDATE: it's a draft, because I identify that the initial flows are not navigating automatically.  🥲~~

UPDATE: it's ready to review, the initial flows are still funky, but it will be handled on a diff PR

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Simulator Screen Recording - iPhone 11 - 2022-06-15 at 18 20 09


danibonilha (commented):
.

cardstack/src/navigation/tabBarNavigator.tsx
```diff
@@ -180,8 +179,10 @@ const useNavigationAuth = () => {
   }, []);
 
   useEffect(() => {
-    loadAuthInfo();
-  }, [loadAuthInfo, isAuthorized]);
+    if (!hasPin || isFirstMount.current) {
+      loadAuthInfo();
```
oidoug, line 183: I wonder if `loadAuthInfo` will not be called twice, once `isFirstMount` is true and then if `hasPin` is false.

oidoug (approved):

paulinhapenedo (approved):
It's always nice to see unused files being removed and lodash references removed 👏

cardstack/src/navigation/tabBarNavigator.tsx
danibonilha: There's still improvements to be done, the right way is to call just once on init, and use something else to trigger the reload once we have a pin

PR #899, opened 2022-06-17
[CS-4082] Add Copy Seed Phrase to Pin creation flow

oidoug:
### Description

This PR updates SeedPhraseBackup with new copy and params to be used by the wallet manager before creating a new PIN.

- [x] Completes #(CS-4082)

### Checklist

- [x] Tested on Android
- [ ] Tested on iOS (I couldn't load wallet on an older tagged version to test migration)

### Screenshots

On linear ticket.

### DIFF
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -211,7 +211,7 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
   SEED_PHRASE_BACKUP: {
     component: SeedPhraseBackup,
     options: {
-      ...bottomSheetPreset,
+      ...horizontalInterpolator,
       gestureEnabled: false,
     } as StackNavigationOptions,
   },
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -210,7 +210,10 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
   },
   SEED_PHRASE_BACKUP: {
     component: SeedPhraseBackup,
-    options: horizontalInterpolator,
+    options: {
+      ...bottomSheetPreset,
+      gestureEnabled: false,
+    } as StackNavigationOptions,
   },
 };
 
--- a/cardstack/src/screens/SeedPhraseBackup/strings.ts
+++ b/cardstack/src/screens/SeedPhraseBackup/strings.ts
@@ -1,19 +1,8 @@
-import { SeedPhraseBackupFlow } from './types';
-
-export const strings: Record<
-  SeedPhraseBackupFlow,
-  { title: string; subtitle: string; button: string }
-> = {
-  [SeedPhraseBackupFlow.backup]: {
-    button: `OK, I've saved my seed phrases`,
-    title: 'Backup your seed phrases',
-    subtitle:
-      'Tap the seed phrases below to copy them. Save them in a secure place.',
-  },
-  [SeedPhraseBackupFlow.singlewallet]: {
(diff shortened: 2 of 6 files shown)


oidoug:
@danibonilha, please check this PR again, thanks!

src/hooks/useWalletManager.ts
```diff
@@ -82,73 +80,71 @@ export default function useWalletManager() {
     [navigate]
   );
 
+  const loadAllSeedPhrases = useCallback(async (wallets: RainbowWallet[]) => {
+    const seeds = [];
+
+    // Needs to be sequential
+    for (const walletId of Object.keys(wallets)) {
+      const seed = (await loadSeedPhrase(walletId, 'migration')) || '';
```
danibonilha: we can remove the `migration` string, I believe it's a left over from a previous pr

src/hooks/useWalletManager.ts
```diff
@@ -82,73 +80,71 @@ export default function useWalletManager() {
...
+  }, []);
+
   const migrateWalletIfNeeded = useCallback(
     async ({ selectedWallet, wallets }: WalletsState) =>
       new Promise<void>(async resolve => {
         try {
           const hasPin = !!(await getPin());
 
-          if (hasPin) {
+          if (hasPin || Object.keys(wallets).length < 1) {
```
danibonilha: not sure what is the purpose of this check -> `Object.keys(wallets).length < 1)` , if a user have no wallet, it shouldn't even enter the app

cardstack/src/navigation/screens.ts
```diff
@@ -210,7 +210,10 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
   },
   SEED_PHRASE_BACKUP: {
     component: SeedPhraseBackup,
-    options: horizontalInterpolator,
+    options: {
+      ...bottomSheetPreset,
+      gestureEnabled: false,
+    } as StackNavigationOptions,
```
danibonilha, line 216: Let's leave it with the horizontalInterpolator for now, just to avoid conflicts with the #901, since `bottomSheetPreset` has been removed

cardstack/src/screens/SeedPhraseBackup/strings.ts
```diff
@@ -1,6 +1,8 @@
 export const strings = {
   button: `OK, I've saved my seed phrases`,
   title: 'Backup your seed phrases',
-  subtitle:
-    'Tap the seed phrases below to copy them. Save them in a secure place.',
+  message:
+    'We are moving to a more robust PIN and biometrics management system, and your seed phrase will be migrated. Before we continue, please back up your wallet by tapping your seed phrase below to copy it to your clipboard, then paste it into a secure place. And please write it down on paper.',
+  multipleWallets:
+    'Additionally, Cardwallet will only support a single wallet from now on. Moving forward, only your selected wallet will be available to be used.',
 };
```
danibonilha, line 8: Title:
Seed Phrase migration is required.

Message:
We are moving to a more robust PIN and biometrics management system, and your seed phrase will be migrated. Tap on your seed phrase below to copy it to your clipboard. Then paste it in a secure place and  write it down on paper.

If user has multiple seeds, include:
Additionally, Card Wallet will support only a single wallet from now on. While you have multiple wallets in the app, only the currently-selected wallet will be available for use moving forward. Please back up the seed phrases below

danibonilha (approved):
Awesome, after addressing the comments, it's good to go. I've tested on iOS

src/hooks/useWalletManager.ts
oidoug: You're right, it's not needed.

PR #900, opened 2022-06-17
[CS-4059] Prevent asset list to show assets with zero balance amount

paulinhapenedo:
### Description
This PR solves an issue with the asset list on the `SendSheet` component. For unknown reasons, we were listing assets with zero balance, thus allowing the user to choose this option. When doing that, the app would break because the `getGasPriceEstimate` function would break.

This PR removes the asset from the list.

- [x] Completes #CS-4059

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

|  Before | After |
|--|--|
| Screenshot |  Screenshot |

### DIFF
--- a/src/components/send/SendAssetList.js
+++ b/src/components/send/SendAssetList.js
@@ -59,9 +59,14 @@ export default class SendAssetList extends React.Component {
     const { assets } = buildCoinsList(allAssets, nativeCurrency, true);
     let shitcoins = [];
 
-    const visibleAssetsLength = assets.length;
+    // assets with no balance shouldn't be visible
+    const assetsWithBalance = assets.filter(
+      asset => parseFloat(asset.balance.amount) > 0
+    );
 
-    this.data = assets;
+    const visibleAssetsLength = assetsWithBalance.length;
+
+    this.data = assetsWithBalance;
 
     if (savings && savings.length > 0 && network === networkTypes.mainnet) {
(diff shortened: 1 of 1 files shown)


src/components/send/SendAssetList.js
```diff
@@ -59,9 +59,14 @@ export default class SendAssetList extends React.Component {
     const { assets } = buildCoinsList(allAssets, nativeCurrency, true);
     let shitcoins = [];
 
-    const visibleAssetsLength = assets.length;
+    // assets with no balance shouldn't be visible
+    const assetsWithBalance = assets.filter(
+      asset => parseFloat(asset.balance.amount) > 0
```
oidoug, line 64: I think we should consider 0 also amounts under a "dust" threshold.

I tried to send an asset up to its decimals until the transaction fails for lack of funds for gas, and my asset was still enough dust on it to not be considered 0.

oidoug (commented):
We need to discuss if we should handle dust in this filter.

src/components/send/SendAssetList.js
paulinhapenedo: Can you elaborate, please?
Are you saying we should take decimals into consideration when filtering? Or is tis filter already taking decimals into consideration (assets with 0.01 balance are shown) and then the gas fee estimate isn’t working, thus breaking the app again?

danibonilha (approved):
As discussed on the mobile meeting, this approach should be enough.

PR #902, opened 2022-06-20
[CS-4094] Hide banner when GET /api/email-card-drop-requests has show-banner: false

oidoug:
### Description

PR updates welcome banner and getEoaClaimed response type to use new show-banner attribute from `email-card-drop-requests` endpoint call.

- [x] Completes #(CS-4094)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

No UI changes.

### DIFF
--- a/cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
+++ b/cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
@@ -17,7 +17,7 @@ export const useWelcomeCtaBanner = () => {
   const { accountAddress, network } = useAccountSettings();
 
   const {
-    data: emailDropGetData = { showBanner: true },
+    data: emailDropGetData = { showBanner: false },
   } = useGetEoaClaimedQuery(
     {
       eoa: accountAddress,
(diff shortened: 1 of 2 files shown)


cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
```diff
@@ -17,7 +17,7 @@ export const useWelcomeCtaBanner = () => {
   const { accountAddress, network } = useAccountSettings();
 
   const {
-    data: emailDropGetData = { claimed: true, rateLimited: true },
+    data: emailDropGetData = { showBanner: true },
```
paulinhapenedo: Shouldn't this default value be `false`? Based on the changes done here https://github.com/cardstack/cardwallet/pull/896, the default value should be `false` so, when offline, the user doesn't see the banner.

paulinhapenedo (approved):
Minor comment, but LGTM.

cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
oidoug: You're right, my mistake.

PR #903, opened 2022-06-20
[CS-4114] Remove biometric icons from confirmation buttons

paulinhapenedo:
### Description
This PR removes the biometric icon from the confirmation buttons across the app, leaving only the Unlock Screen one.
In the process, I deleted a component that wasn't being used.

- [x] Completes #4114

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


oidoug (approved):
LGTM :)

One consideration though is that once we launch the new PIN, we might get requests to re-add auth confirmations to transactions, and the `HoldToAuthorizeButtonIcon` would have to be recovered.

PR #904, opened 2022-06-21
[CS-3838] Replace CyptoCompare with api/exchange-rates endpoint

paulinhapenedo:
### Description
This PR replaces the CryptoCompare URL with the endpoint created on the Hub API (PR for reference https://github.com/cardstack/cardstack/pull/2992)

- [x] Completes #CS-3838

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots
No visual changes :)

### DIFF
--- a/cardstack/src/services/historical-pricing-service.ts
+++ b/cardstack/src/services/historical-pricing-service.ts
@@ -1,21 +1,23 @@
-import { cryptoCurrencies } from '@cardstack/cardpay-sdk';
-import { CRYPTOCOMPARE_API_KEY } from 'react-native-dotenv';
+import { cryptoCurrencies, NativeCurrency } from '@cardstack/cardpay-sdk';
+import { format } from 'date-fns';
 
+import { getExchangeRatesQuery } from '@cardstack/services/hub/hub-service';
 import { removeCPXDTokenSuffix } from '@cardstack/utils';
 
 import logger from 'logger';
 
-const CryptoCompareAPIBaseURL = `https://min-api.cryptocompare.com/data/pricehistorical?&api_key=${CRYPTOCOMPARE_API_KEY}`;
-
 export const fetchHistoricalPrice = async (
   symbol: string,
-  timestamp: string | number,
-  nativeCurrency: string
+  timestamp: number,
+  nativeCurrency: NativeCurrency | string
 ): Promise<number> => {
   try {
     if (!symbol) {
       return 0;
     }
+
+    // date needs to be formatted as yyyy-MM-dd
+    const formattedDate = format(timestamp, 'yyyy-MM-dd');
 
     // For cpxd tokens we chop off .CPXD in order to get mainnet's historical price
     const tokenSymbol = removeCPXDTokenSuffix(symbol);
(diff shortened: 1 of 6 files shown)


lukemelia (approved):

oidoug (approved):
Nice job, props on the move to rtk-query.

Do we have a documentation on this endpoint params and response? It's more complex than usual.

paulinhapenedo:
> Nice job, props on the move to rtk-query.
> 
> Do we have a documentation on this endpoint params and response? It's more complex than usual.

I don’t think we have, other than the Hub PR and the code comment, but I’m happy to write it. Let’s discuss on Discord where should I put it.

cardstack/src/services/hub/hub-types.ts
```diff
@@ -35,3 +36,10 @@ export interface CheckHubAuthQueryParams {
 export interface RegisterFCMTokenQueryParams {
   fcmToken: string;
 }
+
+export interface GetExchangeRatesQueryParams {
+  from?: NativeCurrency | string;
+  to?: NativeCurrency | string;
+  date?: string | number;
+  e?: 'kucoin' | string;
+}
```
danibonilha, line 45: Isn't just `e`, really optional when we decide to pass the params ?  The whole `GetExchangeRatesQueryParams` is optional, but it's params, from, to and date maybe should not be

cardstack/src/services/hub/hub-api.ts
```diff
@@ -63,8 +64,14 @@ export const hubApi = createApi({
...
-    getExchangeRates: builder.query<Record<NativeCurrency, number>, void>({
-      query: () => routes.exchangeRates,
+    getExchangeRates: builder.query<
+      Record<NativeCurrency | string, number>,
+      GetExchangeRatesQueryParams | undefined
+    >({
+      query: params => ({
+        url: routes.exchangeRates,
+        params,
+      }),
```
danibonilha, line 74: if we type like this:
image

we can remove the  empty obj `{}` param, from `useInputAmountHelper`

cardstack/src/services/historical-pricing-service.ts
```diff
@@ -1,51 +1,63 @@
...
+      });
 
-      return usdtPrice * nativeCurrencyPriceForUSDT;
+      const nativeCurrencyPriceForUSDT = USDTRate?.[tokenSymbol];
+
+      if (usdtPrice && nativeCurrencyPriceForUSDT) {
+        return usdtPrice * nativeCurrencyPriceForUSDT;
+      } else {
+        return 0;
+      }
```
danibonilha, line 49: superNit: we could default usdtPrice and nativeCurrencyPriceForUSDT to 0, and simply return the multiplication

cardstack/src/services/historical-pricing-service.ts
```diff
@@ -1,51 +1,63 @@
...
     }
 
-    const response = await fetch(
-      `${CryptoCompareAPIBaseURL}&fsym=${tokenSymbol}&tsyms=${nativeCurrency}&ts=${timestamp}`
-    );
+    const defaultParams = {
+      from: tokenSymbol,
+      to: nativeCurrency,
+      date: formattedDate,
+    };
```
danibonilha, line 56: nit: the defaultParams, could be added to the top, this way we could use it to spread on the inner `cryptoCurrencies.CARD.currency` condition and just replace what's different

cardstack/src/services/historical-pricing-service.ts
paulinhapenedo: The problem here is that both ustdPrice and nativeCurrencyPriceForUSDT are promises, so they both can be undefined.... what do you think if I switch to:

```suggestion
      if (!usdtPrice || !nativeCurrencyPriceForUSDT) {
        return 0;
      }

      return usdtPrice * nativeCurrencyPriceForUSDT;
```

PR #906, opened 2022-06-22
[Fix] Handle iconProps on OptionItem component

paulinhapenedo:
### Description
I missed some stuff with the work here https://github.com/cardstack/cardwallet/pull/903.
`OptionItem` component had logic expecting for `iconProps` that we're not gonna use anymore, so I simplified the component a bit and removed the only place where `iconProps.visible` was being set.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


danibonilha (approved):
LGTM

oidoug (approved):
LGTM

PR #907, opened 2022-06-23
[CS-4140] PrepaidCard screen workaround for Android

paulinhapenedo:
### Description
This PR adds a workaround for the PrepaidCard screen while we don't figure out what's happening after the `react-navigation` upgrade. On Android, we're removing the gesture on the View to allow the user to scroll inside the history section and adding a back button on the top for the modal to be closed.

On iOS the behavior is the same as before.
 
- [x] Completes #CS-4140

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

#### Android
https://user-images.githubusercontent.com/690904/175328970-c1f56515-b753-4545-9469-7172ddf25602.mov 

#### iOS
https://user-images.githubusercontent.com/690904/175330886-c89c4d7d-16b3-42c2-8c1f-a4c998b289a7.mp4

### DIFF
--- a/cardstack/src/navigation/screens.ts
+++ b/cardstack/src/navigation/screens.ts
@@ -38,6 +38,7 @@ import {
   RewardWithdrawConfirmationScreen,
   RewardWithdrawToScreen,
 } from '@cardstack/screens/RewardsCenterScreen/flows';
+import { Device } from '@cardstack/utils';
 
 import BackupSheet from '@rainbow-me/screens/BackupSheet';
 import SendSheetEOA from '@rainbow-me/screens/SendSheetEOA';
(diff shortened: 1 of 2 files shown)


cardstack/src/navigation/screens.ts
```diff
@@ -71,7 +72,7 @@ export const MainScreens: Record<keyof typeof MainRoutes, ScreenNavigation> = {
   },
   PREPAID_CARD_MODAL: {
     component: PrepaidCardModal,
-    options: sheetPreset(),
+    options: { ...sheetPreset(), gestureEnabled: Device.isIOS ? true : false },
```
danibonilha: nit: we can just user `Device.isIOS`, no need for the ternary

danibonilha (approved):
Nice workaround!

oidoug (approved):
Tested and working as expected. Sorry for the delayed review.

PR #908, opened 2022-06-23
[CS-4126] Prompt for permissions on Pay Tab has white overlay on Android

oidoug:
### Description

The old camera module we were using handles camera permission while trying to render, with the new rn-nav6 the screen was getting blank while permission was requested, and it is not being maintained anymore. 

This PR upgrades our camera module to the better-maintained expo-camera and expo-barcode-scanner (both packages are needed for QR reading).

- [x] Completes #(CS-4126)

* known issue: Camera blacks out after navigation "go backs" even though it should be rendered, something related to Camera only allowing one instance at a time. * - I'll handle this in another PR.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/175535292-005471ec-3278-4c5d-a81c-2b00236cc490.mp4

### DIFF
--- a/cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/components/CameraNotAuthorizedView.tsx
+++ b/cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/components/CameraNotAuthorizedView.tsx
@@ -10,10 +10,10 @@ import {
 import { strings } from '../strings';
 
 interface Props {
-  enableCameraPressed: () => void;
+  onEnableCameraPress: () => void;
 }
 
-const CameraNotAuthorizedView = ({ enableCameraPressed }: Props) => {
+const CameraNotAuthorizedView = ({ onEnableCameraPress }: Props) => {
   return (
     <AbsoluteFullScreenContainer
       backgroundColor="backgroundDarkPurple"
@@ -33,7 +33,7 @@ const CameraNotAuthorizedView = ({ enableCameraPressed }: Props) => {
       >
         {strings.authorize.info}
       </Text>
-      <Button onPress={enableCameraPressed}>
+      <Button onPress={onEnableCameraPress}>
         <Text>{strings.authorize.cameraBtn}</Text>
       </Button>
     </AbsoluteFullScreenContainer>
--- a/cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/components/CameraNotAuthorizedView.tsx
+++ b/cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/components/CameraNotAuthorizedView.tsx
@@ -1,5 +1,4 @@
-import React, { memo, useCallback } from 'react';
-import { Linking } from 'react-native';
+import React, { memo } from 'react';
 
 import {
   Text,
@@ -10,11 +9,11 @@ import {
 
 import { strings } from '../strings';
 
-const CameraNotAuthorizedView = () => {
-  const handlePressSettings = useCallback(() => {
-    Linking.openSettings();
-  }, []);
+interface Props {
+  enableCameraPressed: () => void;
+}
 
+const CameraNotAuthorizedView = ({ enableCameraPressed }: Props) => {
   return (
     <AbsoluteFullScreenContainer
       backgroundColor="backgroundDarkPurple"
@@ -34,7 +33,7 @@ const CameraNotAuthorizedView = () => {
       >
         {strings.authorize.info}
       </Text>
-      <Button onPress={handlePressSettings}>
+      <Button onPress={enableCameraPressed}>
         <Text>{strings.authorize.cameraBtn}</Text>
       </Button>
     </AbsoluteFullScreenContainer>
--- a/cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/useQRCodeScannerPage.tsx
+++ b/cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/useQRCodeScannerPage.tsx
@@ -1,0 +1,56 @@
+import { useIsFocused } from '@react-navigation/native';
+import { Camera } from 'expo-camera';
+import { useEffect, useCallback } from 'react';
+import { Linking } from 'react-native';
+import { useIsEmulator } from 'react-native-device-info';
+
+import { useBooleanState } from '@cardstack/hooks';
+
+import { useAccountSettings } from '@rainbow-me/hooks';
+
+import { useScanner, useScannerParams } from './useScanner';
+
+export const useQRCodeScannerPage = (props: useScannerParams) => {
+  const [isCameraAllowed, setCameraAllowed] = useBooleanState();
+  const { onScan, isLoading } = useScanner(props);
+  const { result: isEmulator } = useIsEmulator();
+  const [error, showError] = useBooleanState();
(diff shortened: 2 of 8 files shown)


cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/useQRCodeScannerPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export const useQRCodeScannerPage = (props: useScannerParams) => {
+  const [isCameraAllowed, setCameraAllowed] = useBooleanState();
+  const { onScan, isLoading } = useScanner(props);
+  const { result: isEmulator } = useIsEmulator();
+  const [error, showError] = useBooleanState();
+  const { network } = useAccountSettings();
+  const isFocused = useIsFocused();
+
+  const requestPersmission = useCallback(async () => {
```
danibonilha: typo `requestPersmission` -> `requestPermission`

cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/components/CameraNotAuthorizedView.tsx
```diff
@@ -34,7 +33,7 @@ const CameraNotAuthorizedView = () => {
       >
         {strings.authorize.info}
       </Text>
-      <Button onPress={handlePressSettings}>
+      <Button onPress={enableCameraPressed}>
```
danibonilha: nit: we could follow the pattern, like, `onEnableCameraPress`

cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/useQRCodeScannerPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+
+    requestPersmission();
+  }, [requestPersmission]);
+
+  useEffect(() => {
+    if (isFocused) requestPersmission();
+  }, [requestPersmission, isFocused]);
+
```
danibonilha: Shouldn't we check if it is not authorized also  ?  `  if (isFocused && !isCameraAllowed)`

src/useAppInit.ts
```diff
@@ -97,9 +101,29 @@ export const useAppInit = () => {
...
+      unauthorizeDelay.current = null;
+      return;
     }
+
+    // We'll only delay the unauthorization on android and in these specific screens.
+    if (
+      Device.isAndroid &&
+      (Navigation.getActiveRouteName() === Routes.QR_SCANNER_SCREEN ||
+        Navigation.getActiveRouteName() === Routes.PAY_MERCHANT)
+    ) {
```
danibonilha: why we need to check for PayMerchant ? if we do need both, there's some nit code to improve this
```tsx
const tempAuthorizedRoutes: string[] = [Routes.QR_SCANNER_SCREEN, Routes.PAY_MERCHANT]

const currentRoute = Navigation.getActiveRouteName() || ''
const isOnDelayedRoute = tempAuthorizedRoutes.includes(currentRoute)

if(Device.isAndroid && isOnDelayedRoute)
```

cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/components/CameraNotAuthorizedView.tsx
oidoug: yes, that's better.

src/useAppInit.ts
```diff
@@ -45,6 +48,7 @@ export const useAppInit = () => {
   useNotificationSetup();
 
   const initialDeepLink = useRef<string | null>(null);
+  const unauthorizeDelay = useRef<number | null>(null);
```
danibonilha: nite: Not sure naming this as a delay makes sense, maybe this should be called `androidLastAuthorizedDate` and isn't the type `useRef<Date | null>`

src/useAppInit.ts
```diff
@@ -97,9 +101,29 @@ export const useAppInit = () => {
   }, [isAuthorized, walletReady]);
 
   useEffect(() => {
-    if (movedToBackground) {
-      setUserUnauthorized();
+    // Resuming state, we check if we should unauthorize already.
+    if (!movedToBackground && unauthorizeDelay.current) {
```
danibonilha: not quite sure I understand this, checking against the movedBackground, if we want to check resuming the state, wouldn't justBecameActive work for this case ?

src/useAppInit.ts
```diff
@@ -97,9 +101,29 @@ export const useAppInit = () => {
...
+        Navigation.getActiveRouteName() === Routes.PAY_MERCHANT)
+    ) {
+      unauthorizeDelay.current = Date.now();
+
+      return;
+    }
+
+    // Otherwise, unauthorize right away.
+    setUserUnauthorized();
   }, [setUserUnauthorized, movedToBackground]);
```
danibonilha: lmk what you think about this `useEffect`, if it works for this case 
```ts
  useEffect(() => {
    if (movedToBackground) {
      const tempAuthorizedRoutes: string[] = [
        Routes.QR_SCANNER_SCREEN,
        Routes.PAY_MERCHANT,
      ];

      const isOnDelayedRoute = tempAuthorizedRoutes.includes(
        Navigation.getActiveRouteName() || ''
      );

      if (Device.isAndroid && isOnDelayedRoute) {
        const currentDate = Date.now();

        if (!unauthorizedDelay.current) {
          unauthorizedDelay.current = currentDate;
          return;
        }

        if (currentDate - unauthorizedDelay.current < UNAUTHORIZE_DELAY) {
          return;
        }
      }

      unauthorizedDelay.current = null;

      setUserUnauthorized();
    }
  }, [setUserUnauthorized, movedToBackground]);
```

cardstack/src/screens/QRScannerScreen/pages/QRCodeScanner/useQRCodeScannerPage.tsx
oidoug: Yes, that makes sense.

src/useAppInit.ts
oidoug: We do, once QR is read and we navigate to prepaid card selection, the app jumps to background again.

We could try finding out why that happens, but might be a bit more challenging.

src/useAppInit.ts
oidoug: `Date.now()` return the milliseconds since epoch.

src/useAppInit.ts
oidoug: `movedToBackground` is false when app becomes active again, I think it's safe to use it.

src/useAppInit.ts
oidoug: Hum, almost, but we need to check unauthorizedDelay when app gets active again (!movedToBackgroun) or we'll only be able to unauthorize when going to background again.

I see my algo is a bit confusing, I'll try merge both.

oidoug:
As suggested, I'm splitting android's background logout delay from camera lib upgrade PR.

oidoug:
Upgrading RN to latest fixed the white overlay this PR was addressing.
expo-camera is not working as well as our current module, so this PR is being dropped.

PR #909, opened 2022-06-23
[CS-4139] Remove handleFocus from SendSheet Header to avoid focus error

paulinhapenedo:
### Description
This PR is part of the workarounds we're doing after `react-navigation` upgrades.
For some unknown reason, the `autoFocus` prop isn't working on Android. Using the `magicAutoFocus()` makes things even buggier, so we're deleting this function. On Android, the user will have to tap on the Input to focus. On iOS, things are OK since `autoFocus` is working as expected.

- [x] Completes #CS-4139

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/src/components/fields/AddressField.js
+++ b/src/components/fields/AddressField.js
@@ -57,6 +57,8 @@ const AddressField = (
     []
   );
 
+  const onPress = useCallback(() => ref?.current?.focus(), [ref]);
+
   return (
     <Row flex={1}>
       <Input
@@ -83,7 +85,7 @@ const AddressField = (
       />
       {!inputValue && (
         <Container position="absolute" zIndex={1} {...androidProps}>
-          <Touchable onPress={ref?.current?.focus}>
+          <Touchable onPress={onPress}>
             <Text color="grayText" fontSize={15} weight="bold">
               Enter Address (0x...) or ENS
             </Text>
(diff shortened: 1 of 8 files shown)


src/components/fields/AddressField.js
```diff
@@ -57,6 +57,8 @@ const AddressField = (
     []
   );
 
+  const onPress = useCallback(() => ref?.current?.focus(), [ref]);
```
paulinhapenedo, line 60: For some unknown reason, putting the `focus` directly into the `Touchable` doesn't work.

oidoug: I think that happens because on first mount, Touchable will be instantiated with a null pointer to ref.current, and it doesn't update later because it gets memoized. A callback pointer works because it will get re-build when ref changes.

oidoug (approved):
LGTM.

At least the `handleFocus` works fine, so "send amount" field gets correctly auto-focused.

paulinhapenedo:
> LGTM.
> 
> At least the `handleFocus` works fine, so "send amount" field gets correctly auto-focused.

`handleFocus` doesn't do anything. It just saves the `ref` of the element that was focused, but doesn't do anything with it. It's not responsible for the autofocus itself. That's why I'm removing the `useMagicAutoFocus` too.

PR #911, opened 2022-06-24
[CS-4146] Prompt for permissions on pay tab logs user out

oidoug:
### Description

On Android and in QR code and Pay Merchant screens, de-authorization is being delayed after the app state changes to `background` by 5 seconds. This is a workaround a limitation in Android and rn-navigation 6 that leads to all new activity shown to move the app to background state - thus login out user when not desired.

- [x] Completes #(CS-4146)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/175611056-143ae1eb-71dd-4f34-a727-661bc492aa22.mp4

### DIFF
--- a/src/useAppInit.ts
+++ b/src/useAppInit.ts
@@ -24,9 +24,12 @@ import {
 } from '@cardstack/notification-handler';
 import { useAuthSelectorAndActions } from '@cardstack/redux/authSlice';
 import { requestsForTopic } from '@cardstack/redux/requests';
+import { Device } from '@cardstack/utils';
+
 import Logger from 'logger';
 
 const WALLETCONNECT_SYNC_DELAY = 500;
+const ANDROID_UNAUTHORIZE_DELAY = 5000;
 
 const skipKeychainCheck = true;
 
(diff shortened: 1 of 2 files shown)


src/useAppInit.ts
```diff
@@ -97,8 +101,34 @@ export const useAppInit = () => {
...
       setUserUnauthorized();
+    } else if (androidLastAuthorizedDate.current) {
+      // When returning to active
+      if (
+        currentDate - androidLastAuthorizedDate.current >
+        ANDROID_UNAUTHORIZE_DELAY
+      ) {
+        androidLastAuthorizedDate.current = null;
+        setUserUnauthorized();
+      }
```
danibonilha, line 131: nit: if we keep movedToBackground as before:  
```
  const movedToBackground = useMemo(
    () => prevAppState === AppStateType.background,
    [prevAppState]
  );
```
would this piece still be necessary ?

danibonilha (approved):
LGTM, just a small question

src/useAppInit.ts
oidoug: Yes, we need to act on both going in and out of background mode.

If movedToBg checks for prev state, it will be true once the app becomes active, and we would need another flag to handle when the app is backgrounded. So I prefer to have it reflect the actual state, although we could rename it to `appInBackground`.

PR #912, opened 2022-06-24
[CS-4011]: Fix invalid request using new signer on hubAuth

danibonilha:
### Description

This PR fixes the invalid request error, by using the new signer on hubAuth, by doing that we were able to remove simplify the signer params, by using the private key instead of account indexes. the hubAuth was the last module using the HDWallet,  with the migration I also removed the HDWallet remaining references. A little bit of context, but it also fixes one of the sheets, that were not closing properly.

Completes CS-3967

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Layout change
Before: 
Simulator Screen Recording - iPhone 11 - 2022-06-24 at 15 20 05
After:
Simulator Screen Recording - iPhone 11 - 2022-06-24 at 15 19 16

Invalid request fix:
Simulator Screen Recording - iPhone 11 - 2022-06-24 at 15 35 11
Simulator Screen Recording - iPhone 11 - 2022-06-24 at 15 35 31


paulinhapenedo (approved):
It's really good to not have the invalid request happening anymore. 👏

cardstack/src/services/hub-service.ts
```diff
@@ -113,22 +95,13 @@ export const getHubAuthToken = async (
...
-  const walletId =
-    getWalletByAddress({ walletAddress: address, allWallets })?.id || '';
-
   try {
-    const hdProvider = await HDProvider.get({
-      walletId,
-      network,
-      keychainAcessAskPrompt,
+    const [web3, signer] = await getWeb3ProviderWithEthSigner({
+      accountAddress: address,
```
oidoug, line 100: 👏

oidoug (approved):
Awesome Dani! Although I don't fully understand the details I really appreciate how much more understandable the signing process is now.

PR #914, opened 2022-06-27
[CS-2939]- Update react native to 0.68.2

danibonilha:
### Description

This PR upgrades RN to the latest compatible version with expo, it also updates react-native-reanimated (previous version was not compatible with latest rn one).

- [x] Completes CS-2939
- [x] Completes CS-4138

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

Release builds were triggered from this branch, and ran successfully


paulinhapenedo (approved):
So nice seeing this patches being deleted one by one. It sparks joy 🌟 
Thank you, again, for tackling these herculean updates 💪

android/app/build.gradle
```diff
@@ -192,6 +192,7 @@ android {
             abiFilters 'armeabi-v7a','arm64-v8a','x86','x86_64'
         }
 
+        buildConfigField "boolean", "IS_NEW_ARCHITECTURE_ENABLED", "false"
```
oidoug, line 195: It would be nice to try the new architecture soon! It's supposed to be a 10x improvement to RN. It's probably going to be the default in rn-0.69.

oidoug (approved):
LGTM!
Minor inconvenience so far is that in dev, fast-reloading some screens crashes the app, but we can check that later.

android/app/build.gradle
danibonilha: Yeah! That's why I added all the new modules, it was not a requirement for 0.68.2, but I wanted to have most of the things configured so we could eventually try on.

PR #915, opened 2022-06-27
[CS-4123] Remove Firebase Analytics ID collection from Android

paulinhapenedo:
### Description
This PR adds a configuration to `AndroidManifest.xml` that disables Firebase from collecting the Advertising ID. We already had configured that on iOS, but we missed on Android.

This will solve the policy error that Play Store is signaling to us.

- [x] Completes #CS-4123

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -17,6 +17,8 @@
       android:networkSecurityConfig="@xml/network_security_config"
     >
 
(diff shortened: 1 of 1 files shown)


oidoug (approved):
Awesome! Thanks for finding that out.

danibonilha (commented):
It would be nice to trigger a release from this branch to make sure it solves the issue, before merging, ping me if you need help triggering the action

PR #916, opened 2022-06-27
[CS-3999] Adds PIN migration to Change Pin UI

oidoug:
### Description

This PR adds new callback in useWalletManger for updating PIN for selected wallet.

- [x] Completes #(CS-3999)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/176020287-e87207cd-e659-40e2-b0d7-cfb06c95bbba.mp4

### DIFF
--- a/src/components/settings-menu/SettingsSection.js
+++ b/src/components/settings-menu/SettingsSection.js
@@ -1,5 +1,4 @@
 import { getConstantByNetwork } from '@cardstack/cardpay-sdk';
-import { useNavigation } from '@react-navigation/native';
 import React, { useCallback, useMemo } from 'react';
 import { Linking } from 'react-native';
 
@@ -13,7 +12,6 @@ import {
 } from '../list';
 import { CenteredContainer, Icon, ScrollView } from '@cardstack/components';
 import { SettingsExternalURLs } from '@cardstack/constants';
-import { Routes } from '@cardstack/navigation';
 import networkInfo from '@rainbow-me/helpers/networkInfo';
 import WalletTypes from '@rainbow-me/helpers/walletTypes';
 import {
@@ -59,7 +57,6 @@ export default function SettingsSection({
   onPressMyWalletAddress,
   onPressDS,
 }) {
-  const { navigate } = useNavigation();
   const { wallets } = useWallets();
   const { nativeCurrency, network, accountAddress } = useAccountSettings();
   const { updateWalletPIN } = useWalletManager();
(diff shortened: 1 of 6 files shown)


src/hooks/useWalletManager.ts
```diff
@@ -76,6 +79,18 @@ export default function useWalletManager() {
...
 
+  const updateWalletPIN = useCallback(async () => {
+    navigate(Routes.PIN_SCREEN, {
+      flow: PinFlow.new,
+      variant: 'light',
+      canGoBack: true,
+      dismissOnSuccess: true,
+      onSuccess: async (pin: string) => {
+        await migrateSecretsWithNewPin(selectedWallet, pin);
+      },
```
danibonilha: This is not the correct behavior, what we want here, is, before setting the pin, we need to decrypt the seed phrase, which uses the current pin, and then save it under the new PIN encryption. the `migrateSecretsWithNewPin`. should not be used anywhere outside the initial flow

danibonilha (commented):
Apologies for not adding a description on the ticket, I've updated it, ping me if you have any doubts

src/helpers/walletLoadingStates.ts
```diff
@@ -7,4 +7,5 @@ export default {
   RESTORING_WALLET: 'Restoring...',
   SWITCHING_ACCOUNT: 'Switching account...',
   MIGRATING_SECRETS: 'Migrating secrets...',
+  UPDATING_PIN: 'Updating PIN...',
```
danibonilha: This value is never used

oidoug: Right, the intention was to have a loading overlay but it's a quick process.

src/hooks/useWalletManager.ts
oidoug: Right, I got carried away by the existing method working 😅 . 
I created a function in secure-storage to iterate wallets and resave seed, private keys and pin with new pin, please review the PR again.

oidoug:
I just saw an issue, will update the PR soon.

src/components/settings-menu/SettingsSection.js
```diff
@@ -165,7 +158,7 @@ export default function SettingsSection({
             <ListItem
```
danibonilha: We need to move this ListItem, out of the  ` {__DEV__ && (` condition.

danibonilha (approved):
LGTM, just need to enable for all users, instead of just DEV

src/components/settings-menu/SettingsSection.js
oidoug: Ah yes, thanks.

I also removed the Seed Phrase Backup list item that has already been moved to a final implementation.

PR #918, opened 2022-06-28
[CS-3934]: Add exponential backoff  on too many attempts

danibonilha:
### Description

This PR adds the max attempts amount and an exponential backoff to secure the pin authentication, for now we are showing an alert with the remaining waiting time, but it would be best to have some type of timer, which can be done on a diff PR, also the tests, I'll be adding tomorrow on a diff PR, so we can get this shipped.

### Checklist

-[x] Tested on iOS
-[x] Tested on Android

### Screenshots

Simulator Screen Recording - iPhone 11 - 2022-06-28 at 17 29 43

First try: 
image

Second try: 
image


paulinhapenedo (approved):
It's looking good 👏

PR #919, opened 2022-06-28
[CS-3838] Fix api/exchange-rates errors

paulinhapenedo:
### Description
This PR solves an error that's being logged with the `fetchHistoricalPrice` function. Seems like some transactions that come from GraphQL have `timestamp` as a string instead of a number and that was breaking the `date-fns` format.

Also addressed some late comments from Dani on the other PR to improve code readability.

- [x] Completes #[CS-3838]

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


danibonilha (approved):
LGTM!

PR #920, opened 2022-06-29
[CS-4113] Re-enable KeychainIntegrityCheck + SecureStorage enhancement

paulinhapenedo:
### Description
This PR adds the `keychainIntegrityCheck` function back to the code. It's a function that we use mostly for debugging purposes, since it goes through all the keychain + secureStorage entries we use and see if they are set correctly.

One thing to notice is that now we are resetting the flag that runs the check whenever we reset the wallet.

I don't know if the team had already mapped out other changes to this file, but I'm happy to address them too.

- [x] Completes #CS-4113

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/src/handlers/localstorage/globalSettings.js
+++ b/src/handlers/localstorage/globalSettings.js
@@ -18,6 +18,9 @@ export const getKeychainIntegrityState = () =>
 export const saveKeychainIntegrityState = state =>
   saveGlobal(KEYCHAIN_INTEGRITY_STATE, state);
 
+export const deleteKeychainIntegrityState = () =>
+  removeLocal(KEYCHAIN_INTEGRITY_STATE);
+
 export const getAuthTimelock = () => getGlobal(AUTH_TIMELOCK, null);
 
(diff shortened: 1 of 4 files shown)


src/handlers/localstorage/globalSettings.js
```diff
@@ -18,6 +18,9 @@ export const getKeychainIntegrityState = () =>
 export const saveKeychainIntegrityState = state =>
   saveGlobal(KEYCHAIN_INTEGRITY_STATE, state);
 
+export const deleteKeychainIntegritySate = () =>
```
oidoug: typo:`Sate` instead of `State`

oidoug (approved):
Looks good. I'm not sure how to test it properly though.

paulinhapenedo:
> Looks good. I'm not sure how to test it properly though.



> Looks good. I'm not sure how to test it properly though.

One way of testing it is to reset the wallet and importing an account. You should see the checks being logged on the console (if you're using Flipper, filtering by [KeychainIntegrityCheck] helps).

danibonilha (approved):

PR #921, opened 2022-06-29
[CS-4165]:Add tests to unlock w/ exponential backoff

danibonilha:
### Description

This PR adds tests for the exponential backoff feature on the unlockScreen, it covers the cases like, persisting the values, blocking, unblocking and exponentially increasing the time on wrong attempts. I did notice some `act` warnings, tried a bunch of stuff, by it didn't work, if anyone has any ideas, lmk and we can get rid of the warning.

### DIFF
--- a/cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
+++ b/cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
@@ -1,16 +1,24 @@
 import { renderHook } from '@testing-library/react-hooks';
-import { waitFor } from '@testing-library/react-native';
+import { waitFor, act } from '@testing-library/react-native';
 import { Alert } from 'react-native';
-import { act } from 'react-test-renderer';
 
 import { useBiometry } from '@cardstack/hooks/useBiometry';
 import { biometricAuthentication } from '@cardstack/models/biometric-auth';
 import { getPin } from '@cardstack/models/secure-storage';
 
+import {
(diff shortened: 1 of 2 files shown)


cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
```diff
@@ -137,4 +181,203 @@ describe('useUnlockScreen', () => {
...
+    });
+
+    it('should block the user if wrong attempts are greater than MAX_WRONG_ATTEMPTS', async () => {
+      const wrongPIN = '444444';
+
+      const {
+        result: {
+          current: { setInputPin, nextAttemptDate },
+        },
+      } = renderHook(() => useUnlockScreen());
```
oidoug, line 257: I have a question. I understand that these tests need to be run sequentially, because of the added attempts for each test.
But I don't understand how values are maintained, shouldn't renderHook create a new clean instance for useUnlockScreen at each test?

oidoug (approved):
LGTM. Just left a question on something I don't understand.

cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
danibonilha: The values are stored because I've created a storage implementation, so every time the saveSomething is called, i'm assigning the value to the mockStorage, and on Gets, we get the values from the object which makes it possible to retrieve on next test suites.

PR #923, opened 2022-06-30
[CS-4061] Handles multiple biometrics possible on Android

oidoug:
### Description

This PR adds a more generic case for biometry text and icon when we can't know for sure which option is available to the user. I roamed a bunch of issues and PRs from Expo and apparently theres no api for knowing the type of biometry enabled, just if the user has one or not.

I also found a bug in expo's lib that was supposed to be fixed but I can still reproduce:
https://github.com/expo/expo/issues/12353

I tried to force cancel the auth request when `app_cancel` occurs but it causes other issues.

And theres a case where you can cancel the face id authentication but no error is returned, nothing we can do about it unfortunately.

- [x] Completes #(CS-4061)

### Checklist

- [ ] Tested on iOS (no faceid enabled device to test on)
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/176783002-b7dbe7eb-2834-44ab-9984-42cd322925f7.MOV

### DIFF
--- a/cardstack/src/hooks/useBiometry.ts
+++ b/cardstack/src/hooks/useBiometry.ts
@@ -17,7 +17,7 @@ const securityTypeToIcon = {
 };
 
 const securityTypeToLabel = {
-  [SecurityType.BIOMETRIC]: 'biometry',
+  [SecurityType.BIOMETRIC]: 'biometrics',
   [SecurityType.FACE]: 'Face ID',
   [SecurityType.FINGERPRINT]: 'Touch ID',
   [SecurityType.PIN]: 'PIN',
--- a/cardstack/src/hooks/useBiometry.ts
+++ b/cardstack/src/hooks/useBiometry.ts
@@ -9,6 +9,7 @@ import {
 import { useAppState, usePrevious } from '@rainbow-me/hooks';
 
 const securityTypeToIcon = {
+  [SecurityType.BIOMETRIC]: 'lock',
   [SecurityType.FACE]: 'face-id',
   [SecurityType.FINGERPRINT]: 'thumbprint',
   [SecurityType.PIN]: 'lock',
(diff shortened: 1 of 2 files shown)


cardstack/src/hooks/useBiometry.ts
```diff
@@ -9,13 +9,15 @@ import {
...
 const securityTypeToIcon = {
+  [SecurityType.BIOMETRIC]: 'lock',
   [SecurityType.FACE]: 'face-id',
   [SecurityType.FINGERPRINT]: 'thumbprint',
   [SecurityType.PIN]: 'lock',
   [SecurityType.NONE]: 'lock',
 };
 
 const securityTypeToLabel = {
+  [SecurityType.BIOMETRIC]: 'biometry',
```
danibonilha: biometry => `biometrics`

cardstack/src/hooks/useBiometry.ts
```diff
@@ -53,11 +55,13 @@ export const useBiometry = () => {
     const biometryAvailable =
       !!biometryType &&
       (biometryType === SecurityType.FACE ||
-        biometryType === SecurityType.FINGERPRINT);
+        biometryType === SecurityType.FINGERPRINT ||
+        biometryType === SecurityType.BIOMETRIC);
 
     // Transactions require longpress when biometry is not available or
     // when using Face ID to minimize user errors.
     const longPressToConfirm =
```
danibonilha, line 63: not totally related with this PR, but I'm wondering, if we need this `longPressToConfirm` condition, my understanding is that, if the person has biometrics, it will use the tap, otherwise will need to hold to confirm, given that we won't re-request auth on any case, maybe hold to send should be default, making this `longPressToConfirm` useless, we can discuss on a diff place though.

danibonilha (approved):
LGTM, minor adjustment

cardstack/src/hooks/useBiometry.ts
oidoug: I agree, long press should always be needed for confirming transactions now. We could improve the long-press a bit too, it doesn't work first try for me sometimes.

oidoug: I created a couple of tickets to handle this.

PR #924, opened 2022-07-01
[CS-4060] Hide Welcome screen when the app restarts

paulinhapenedo:
### Description
~~This PR doesn't solve the issue described in the ticket, but I think it helps with the "awkwardness" of restarting the app while being authenticated and seeing briefly the Welcome screen.~~

This PR replaces the library we've been using to handle the Splash Screen to one that's being actively maintained (`react-native-bootsplash`) in a fixed version (3.6.2) so we can show and hide the splash screen programatically. A new version of this library (4.0) removes the `show` API because there's no native support for this on Android 12.

- [x] Completes #CS-4060

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots
https://user-images.githubusercontent.com/690904/176951954-309ae10a-e492-450e-b7b7-988dfb3d1a3b.mov

https://user-images.githubusercontent.com/690904/176951968-181649eb-e0dc-465a-870c-9673c95239a6.mp4


oidoug:
Maybe the problem is that we are hiding the splash too soon, before even a wallet has been detected so we go straight to unauthenticated flow.

Your fix makes it better than before but I'm not sure it's the right answer, so I'll let Dani give her opinion on before approving.

danibonilha (commented):
I believe a more user friendly approach would be to show the splash screen after calling `Restart`, I'm aware that `SplashScreen.show()` doesn't seem to work on iOS, but I think it's worth investigating why, we can timebox to one day or so, I can help out if you need, lmk.

paulinhapenedo:
# Update

Replaced `react-native-splash-screen` that's not being actively maintained with `react-native-bootsplash` that has a good support. We've fixed version on 3.6.2 because we need the `show` method. After the version 4.0, the maintainer removed that because there's no native support on Android.

The first splash screen on Android, the one with a white background, will be handled in another PR to broadening the scope of this one.

## Android cold start
https://user-images.githubusercontent.com/690904/177347857-a80d7094-4e28-463a-8c74-6f5b07938fdc.mov

## Android restart after something went wrong
https://user-images.githubusercontent.com/690904/177347887-54f1e66b-636d-4891-a623-39182fe425f2.mov

## iOS cold start
https://user-images.githubusercontent.com/690904/177350603-c1e0906a-9162-4062-9604-bf9706f7d5ab.mp4

## iOS restart after something went wrong
https://user-images.githubusercontent.com/690904/177350621-c2b9b1d9-bded-4012-a011-4d7af8c6142b.mp4

cardstack/src/navigation/tabBarNavigator.tsx
```diff
@@ -191,11 +191,14 @@ const stackScreenOptions: StackNavigationOptions = {
...
-  const { hasWallet, isAuthorized, hasPin } = useNavigationAuth();
+  const { hasWallet, isAuthorized, hasPin, isFirstMount } = useNavigationAuth();
+
+  // firstMount is checked to avoid showing welcome screen when user restarts the app
+  const showWelcomeScreen = !hasWallet && !isFirstMount.current;
 
   return (
     <Stack.Navigator screenOptions={stackScreenOptions}>
-      {!hasWallet ? (
+      {showWelcomeScreen ? (
```
danibonilha: Do we still need this ?

src/components/settings-menu/DeveloperSettings.js
```diff
@@ -53,7 +54,13 @@ const DeveloperSettings = () => {
         onPress={resetWallet}
         testID="reset-keychain-section"
       />
-      <ListItem label="🔄 Restart app" onPress={() => RNRestart.Restart()} />
+      <ListItem
+        label="🔄 Restart app"
+        onPress={() => {
+          RNRestart.Restart();
+          RNBootSplash.show();
```
danibonilha: To avoid forgetting that right after a Restart we also need to call the splashScreen we could create a helper function to do that, this way we can use only one function across the app

android/app/src/main/res/drawable/bootsplash.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android" android:opacity="opaque">
+    <item android:drawable="@color/black" />
+
+    <item>
+        <bitmap android:src="@mipmap/bootsplash_logo" android:gravity="center" />
```
danibonilha, line 7: we have the app_logo_splash, which is pretty much the same as the bootsplash_logo, so we should remove the app_logo_splash, alongside with the file that's using it, `launch_screen.xml` since it's been replaced with this one

ios/Rainbow/Images.xcassets/BootSplashLogo.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
...
+      "idiom": "universal",
+      "filename": "bootsplash_logo@3x.png",
+      "scale": "3x"
+    }
+  ],
+  "info": {
+    "version": 1,
+    "author": "xcode"
+  }
+}
```
danibonilha: We already have the logo, it's `CardstackIcon` so we probably don't need to add this new one

ios/Rainbow/BootSplash.storyboard
```diff
@@ -0,0 +1,45 @@
...
+                </viewController>
+                <placeholder placeholderIdentifier="IBFirstResponder" id="iYj-Kq-Ea1" userLabel="First Responder" sceneMemberID="firstResponder"/>
+            </objects>
+            <point key="canvasLocation" x="0.0" y="0.0"/>
+        </scene>
+    </scenes>
+    <resources>
+        <image name="BootSplashLogo" width="100" height="92"/>
+    </resources>
+</document>
```
danibonilha: I'm confused on why we have a `LauchScreen.storyboard` and also this file that seems pretty much the same thing. it seems this BootSplash.storyboard, is not even being used, given that `LaunchScreen.storyboard ` is the one added on project.pbxproj.

cardstack/src/navigation/tabBarNavigator.tsx
paulinhapenedo: No, removing it.

ios/Rainbow/BootSplash.storyboard
paulinhapenedo: This was created by the command line interface the new library provides to create the assets and the splash screen. Will remove this.

oidoug (commented):
Code looks good, but I can't see the splash on my device, just a blank screen for a couple of secs.

https://user-images.githubusercontent.com/129619/177549055-93982eb6-4813-42e1-b9c5-06c775e4664d.mp4

oidoug (approved):
I re-re-visited this, and it is working as expected. I didn't have the latest commits 😓, sorry!

One minor thing to consider if it's possible to fix or not. On iOS the transition is pretty smooth, but on android it seems the Status Bar height changes and it makes the splash to jump a bit, as shown in one of the screenshots - not a blocker though.

danibonilha (approved):
LGTM, just make sure to test android's deep linking before merging it 😉

PR #925, opened 2022-07-01
[CS-4177] Changes ask for biometrics on app state change.

oidoug:
### Description

Fix a bug in Biometric request not showing up on app being minimized then reopened. It was querying for biometrics on app going to background and that got the check stuck.

- Changes ask for biometrics only for app state active;
- Renames `useAppState`'s `movedToBackground` to `isInBackground` and adds a new `movedFromBackground` value that checks for prevAppState - it is not being used, but I think it's good to have it there so we are more clear of the intention for prevAppState.

- [x] Completes task (CS-4177)

### Checklist

- [ ] Tested on iOS - sorry, no iOS with biometry available :(
- [x] Tested on Android

### Screenshots

No visual changes.

### DIFF
--- a/cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
+++ b/cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
@@ -39,6 +39,10 @@ jest.mock('@cardstack/hooks/useBiometry', () => ({
   useBiometry: jest.fn(),
 }));
 
+jest.mock('@rainbow-me/hooks', () => ({
+  useAppState: () => ({ isActive: true }),
+}));
+
 const mockStorage = {
   attempts: 0,
   nextDate: null,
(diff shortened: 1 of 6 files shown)


cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
```diff
@@ -39,6 +39,10 @@ jest.mock('@cardstack/hooks/useBiometry', () => ({
   useBiometry: jest.fn(),
 }));
 
+jest.mock('@rainbow-me/hooks', () => ({
+  useAppState: () => ({ isActive: true }),
+}));
+
```
danibonilha, line 51: nit: it would be good to add a test case, to make sure `authenticateBiometrically` is not called

danibonilha (approved):
LGTM

cardstack/src/screens/UnlockScreen/__tests__/useUnlockScreen.test.ts
oidoug: Ok, added a couple more tests and mocks!

PR #926, opened 2022-07-01
[CS-4175] Make long press to confirm a transaction default

oidoug:
### Description

Since we don't need to ask for authentication at every transaction, it's better to always require a long press to confirm a transaction.

- [x] Completes #(CS-4175)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/176968636-58898e45-7f06-45b6-b49f-245763e38bf6.mp4


danibonilha (approved):
LGTM

PR #927, opened 2022-07-01
[CS-4178]: Retry migration until valid seed 

danibonilha:
### Description

@gabiburger  caught a bug, where you might be able to create the pin but not have the account fully migrated, this PR adds a condition to retry the migration, until it finds a valid seed.


oidoug (approved):
LGTM

PR #928, opened 2022-07-04
[CS-4179]: Dismiss keyboard on old pin

danibonilha:
### Description

We need to make sure the keyboard is dismissed, on old auth screen, since that screen has it's own keyboard.

### Checklist

- [x] Tested on Android

### Screenshots

dismiss-keybaord


oidoug (approved):
LGTM.

PR #929, opened 2022-07-04
Adds unit tests to useAppState

oidoug:
### Description

Adds unit tests to assert different values expected from `useAppState` hook for different app states.

It uses a `spyOn` mock call to the hook's app state listener, so for each test we need to remove and reattach the spyOn or  the hook is not correctly unmounted between tests.

### DIFF
--- a/cardstack/src/hooks/__tests__/useAppState.test.ts
+++ b/cardstack/src/hooks/__tests__/useAppState.test.ts
@@ -29,49 +29,49 @@ const SECOND_OPEN = {
 };
 
 describe('useAppState', () => {
+  const appStateSpy = jest.spyOn(AppState, 'addEventListener');
+
+  const mockAppState = (state: 'active' | 'background') =>
+    appStateSpy.mock.calls[0][1](state);
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
   it('should match values for first open state', async () => {
     const { result } = renderHook(() => useAppState());
-    const appStateSpy = jest.spyOn(AppState, 'addEventListener');
 
     await act(async () => {
-      await appStateSpy.mock.calls[0][1]('active');
+      await mockAppState('active');
     });
 
     await waitFor(() => {
       expect(result.current).toEqual(FIRST_OPEN);
     });
-
-    appStateSpy.mockClear();
   });
 
   it('should match values for background if state changes to background', async () => {
     const { result } = renderHook(() => useAppState());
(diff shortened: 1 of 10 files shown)


cardstack/src/hooks/useBiometry.ts
```diff
@@ -59,10 +59,19 @@ export const useBiometry = () => {
...
         biometryType === SecurityType.BIOMETRIC);
 
+    // Transactions require longpress when biometry is not available or
+    // when using Face ID to minimize user errors.
+    const longPressToConfirm =
+      biometryType === SecurityType.BIOMETRIC ||
+      biometryType === SecurityType.FACE ||
+      biometryType === SecurityType.NONE ||
+      !biometryAvailable;
+
```
danibonilha: this is probably a conflict, we need to drop

cardstack/src/hooks/__tests__/useAppState.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    appStateSpy.mockClear();
+  });
+
+  it('should match values for second open if state changes from background to active', async () => {
+    const { result } = renderHook(() => useAppState());
+    const appStateSpy = jest.spyOn(AppState, 'addEventListener');
+
+    await act(async () => {
+      await appStateSpy.mock.calls[0][1]('background');
+      await appStateSpy.mock.calls[0][1]('active');
```
danibonilha: supernit: maybe the mockAppState, could be a function that receives the state, but then you'd probably need:
```
  afterEach(() => {
    jest.clearAllMocks();
  });

```

to use a single spy on the describe block

danibonilha (approved):

cardstack/src/hooks/__tests__/useAppState.test.ts
```diff
@@ -0,0 +1,77 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { waitFor, act } from '@testing-library/react-native';
+import { AppState } from 'react-native';
+
+import { useAppState } from '../../../../src/hooks/useAppState';
```
oidoug: @danibonilha, this is a bit ugly rn. What do you think is better, to leave the test and hook apart or to move useAppState to cardstack's hooks?

cardstack/src/hooks/useBiometry.ts
oidoug: got it, thanks.

PR #930, opened 2022-07-06
[CS-4176] Improve long press on transactions

oidoug:
### Description

Adds long press action to Button component and uses it on `TransactionConfirmationSheet`, `RewardWithdrawConfimationScreen`, and `SendSheet`.

Removes HoldToAuthorizeButton in favor of modern Pressable's `onLongPress`, the former has bugs on android and its logic is too complex for minimal benefit.

- [x] Completes #(CS-4176)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/177555233-1ffe3d3f-ae4f-4963-bcef-8b354aa4ba52.mp4

https://user-images.githubusercontent.com/129619/177555356-fbc9cb36-9820-4a81-aae6-b6a943ea048e.mp4


cardstack/src/components/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
```diff
@@ -148,7 +148,7 @@ const SheetFooter = ({
       <Button
         loading={onConfirmLoading}
         variant="small"
-        onPress={onConfirm}
+        onLongPress={onConfirm}
```
paulinhapenedo: Won't this change affect all of the transactions screens? There are screens that seems like an overkill to do such thing.

oidoug: Yes it will. My reasoning to change this one too was to keep all approvals consistent, but I'm open to different opinions.

danibonilha: I believe it's important to have the `hold to confirm` on the SendSheet, since it doesn't trigger a confirmation screen, so the user needs to make sure they inputed the right amount etc, for this Tx Sheet, and other Confirmation Sheets, i'd stick with the simple `Confirm` giving that the screen already shows the confirmation data .

oidoug:
@danibonilha @paulinhapenedo , I reverted the change on Transaction and Reward Withdraw, please review it again. Thanks.

paulinhapenedo (approved):
LGTM 👏

cardstack/src/components/AnimatedPressable/AnimatedPressable.tsx
```diff
@@ -89,6 +102,8 @@ const AnimatedPressable = ({
         onPressIn={onPressAnimate(Scale.shrink)}
         onPressOut={onPressAnimate(Scale.grow)}
         onPress={handleOnPress}
+        onLongPress={handleLongPress}
+        delayLongPress={1000}
```
danibonilha: nit: we have a `delayLongPressMs` const, it's 200ms less though

src/components/send/SendButton.js
```diff
@@ -32,14 +34,15 @@ const SendButton = ({
...
 
   return (
-    <HoldToAuthorizeButton
+    <Button
       {...props}
       disabled={disabled}
-      isAuthorizing={isAuthorizing}
-      label={label}
-      onPress={onPress}
+      loading={isAuthorizing}
```
danibonilha, line 40: nit: since we show the overlay, not sure if we need to show the btn spinner too, it's okay to leave it, but maybe we don't even need this isAuthorizing flag for UI purposes

danibonilha (approved):
LGTM, props on removing that btn.

cardstack/src/components/AnimatedPressable/AnimatedPressable.tsx
oidoug: Ah yeah, thanks for noticing.

src/components/send/SendButton.js
oidoug: It's important to block interactions after authorization starts to avoid "double tapping". 
We can use disabled or loading states, both disable the button.

PR #931, opened 2022-07-06
[CS-4185] Solve claim error

paulinhapenedo:
### Description
This PR solves an issue with the claim screen. To avoid the inevitable re-render that happens after clearing the rewardMainPool cache, we're storing the rewardMainPool info that the user used to claim their rewards and showing that instead of the query response.

We're also flagging the gas fee estimate query to not run after a claim was made because the rewardMainPool info is different now.

- [x] Completes #CS-4185

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/690904/177617889-d5eda8be-f702-49b7-bb06-e1b73ce63ba7.mp4


danibonilha (approved):
LGTM

PR #932, opened 2022-07-06
[Chore] Adds Husky pre-push scripts

oidoug:
### Description

Adds Husky pre-push `git-hook` script. It will not allow lint or test issues to be pushed to origin.

The pre-push script first runs `yarn ci` because it's the most common to have issues, then `yarn test`.

Husky hooks are "configured" during `yarn setup` step, so make sure to run that.


danibonilha (approved):
Nice! 🚀

PR #933, opened 2022-07-06
[CS-4184]: Fixes android auth issues

danibonilha:
### Description

Fixes the payment link, and authentication loop on Android 10 , unfortunately it reintroduces the permission locking issue, but better be safe than sorry, and it only happens one per app install.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/20520102/177630633-a2b2ea9f-112d-4c89-a0b4-595b3b38f429.mp4

### DIFF
--- a/cardstack/src/hooks/useAppState.ts
+++ b/cardstack/src/hooks/useAppState.ts
@@ -39,17 +39,11 @@ export const useAppState = () => {
 
   const isActive = useMemo(() => appState === AppStateType.active, [appState]);
 
-  const isActiveWithoutUpdate = useMemo(
-    () => AppState.currentState === AppStateType.active,
-    []
-  );
-
   return {
     appState,
     justBecameActive,
     movedFromBackground,
     isInBackground,
     isActive,
-    isActiveWithoutUpdate,
   };
 };
--- a/cardstack/src/hooks/useAppState.ts
+++ b/cardstack/src/hooks/useAppState.ts
@@ -39,11 +39,17 @@ export const useAppState = () => {
 
   const isActive = useMemo(() => appState === AppStateType.active, [appState]);
 
+  const isActiveWithoutUpdate = useMemo(
+    () => AppState.currentState === AppStateType.active,
+    []
+  );
+
   return {
     appState,
     justBecameActive,
     movedFromBackground,
     isInBackground,
     isActive,
+    isActiveWithoutUpdate,
   };
 };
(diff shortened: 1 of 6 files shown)


cardstack/src/hooks/useAppState.ts
```diff
@@ -39,11 +39,17 @@ export const useAppState = () => {
 
   const isActive = useMemo(() => appState === AppStateType.active, [appState]);
 
+  const isActiveWithoutUpdate = useMemo(
+    () => AppState.currentState === AppStateType.active,
```
oidoug: So if I understand this correctly, `isActiveWithoutUpdate` will be true on app open and then never change during app's life time, because there's no reactive dependency here and this hook isn't unmounted on app going to background.

Is that intentional? What is its purpose?

cardstack/src/redux/authSlice.ts
```diff
@@ -58,11 +66,21 @@ export const useAuthActions = () => {
...
   }, [dispatch]);
 
+  const authWithBiometricsStarted = useCallback(() => {
+    dispatch(authSlice.actions.authWithBiometricsStarted());
+  }, [dispatch]);
+
+  const authWithBiometricsFinished = useCallback(() => {
+    dispatch(authSlice.actions.authWithBiometricsFinished());
+  }, [dispatch]);
+
```
oidoug, line 76: I saw some posts online recommending this approach but decided on the timer, but looks like this will be more reliable. Might be worth trying to use it for the camera permission issue too.

cardstack/src/screens/UnlockScreen/useUnlockScreen.tsx
```diff
@@ -157,10 +177,10 @@ export const useUnlockScreen = () => {
   }, [checkUserTemporaryBlock, inputPin, validatePin]);
 
   useEffect(() => {
-    if (isBiometryEnabled && isActive) {
+    if (isBiometryEnabled && isActiveWithoutUpdate) {
```
oidoug: Checking for `isActiveWithoutUpdate` here seems like the same as just checking for `isBiometryEnabled` only, I don't think their value change, the effect is called because the whole useUnlockScreen hooks is re-rendered. I understand having it if it's handling a corner case I don't see.

oidoug (approved):
Good to merge once tests are passing. Just left some comments for minor concerns.

Additionally, I saw the biometric auth once when unlock screen wasn't showing and once more after minimizing the app, but couldn't replicate neither of the cases 🤷

PR #934, opened 2022-07-07
[CS-4222]: Alert user on damaged wallet while adding new acc

danibonilha:
### Description

Since we support one wallet only, if a wallet is damaged we don't want to try to create a new one anymore, we alert the user about it, so they can reset it and create a new one or re-import it. There's also a minor UI change on the WalletList, which could be better, but it will do for now.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/20520102/177828695-c75b35c8-f07d-492e-ad9a-53b82e0f6407.mov


paulinhapenedo (approved):
LGTM 👏

oidoug (approved):
LGTM, no issues found while testing.

PR #936, opened 2022-07-11
[CS-3929] Add cache invalidation once user request a card

paulinhapenedo:
### Description
This PR handles the use case when the user requests a card and goes back to the home screen. At this point, we're still not sure if the claim was successful or not, so we're adding a `pollInterval` to the claimed query based on the presence of a flag saved on the async storage.

- [x] Completes #CS-3929

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
```diff
@@ -15,14 +17,14 @@ const WELCOME_BANNER_KEY = 'WELCOME_BANNER_KEY';
...
-    data: emailDropGetData = { showBanner: false },
-  } = useGetEoaClaimedQuery(
+  const { data: emailDropGetData } = useGetEoaClaimedQuery(
+    { eoa: accountAddress },
     {
-      eoa: accountAddress,
-    },
-    { skip: !accountAddress }
+      skip: !accountAddress,
+      pollingInterval: hasClaimed ? undefined : 5000,
```
oidoug: nit: extract the constant value to top of the file.

cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
```diff
@@ -57,6 +59,36 @@ export const useWelcomeCtaBanner = () => {
     ]
   );
 
+  const handleClaimStatus = useCallback(async () => {
+    try {
+      if (emailDropGetData?.claimed) {
```
oidoug: Claimed will return true on all runs after first claim, so I think it isn't a good pattern to re-set it in storage every time. I think you want to first check if the key is saved or not.

cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
```diff
@@ -57,6 +59,36 @@ export const useWelcomeCtaBanner = () => {
...
+        !claimedKey ||
+        claimedKey !== accountAddress ||
+        !emailDropGetData?.claimed
+      ) {
+        setHasClaimed(false);
+      }
+    } catch (error) {
+      logger.log('[Async Storage] Error getting key', error);
+    }
+  }, [accountAddress, emailDropGetData]);
```
oidoug: For cases like this one, I prefer to use a redux slice + redux-persist instead of the storage api directly (even though its only used here), just because of separation of concerns.

Or maybe you could extract this save/restore action into a model, so this logic isn't mixed here in the hook.

oidoug (approved):
Looks good and functional, but I left some concerns that would be nice to address before merging.

cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
paulinhapenedo: I prefer the second option since we seem to not be adding new things to redux (maybe I'm wrong on that 🤔)

cardstack/src/components/CtaBanner/useWelcomeCtaBanner.ts
```diff
@@ -57,6 +60,40 @@ export const useWelcomeCtaBanner = () => {
...
+      }
+
+      // starts polling to get new status after user requested a card
+      if (claimedKey && !emailDropGetData?.claimed) {
+        setHasClaimed(false);
+
+        return;
+      }
+
+      // status changed to claim: true, disables polling
```
oidoug:
```suggestion
      // status changed to claimed: true, disables polling
```

oidoug (approved):
Good job :)
added a minor suggestion.

paulinhapenedo:
@oidoug @danibonilha asking again for feedback because I changed a few things.
Dani agrees with you @oidoug that the async key needs to be handled by redux, so I moved both keys there and deleted the `useCtaBanner` hook

cardstack/src/redux/welcomeBanner.ts
```diff
@@ -0,0 +1,42 @@
...
+const initialState: SliceType = {
+  dismissedByUser: false,
+  requestedCardDrop: false,
+};
+
+const slice = createSlice({
+  name: 'welcomeBanner',
+  initialState,
+  reducers: {
+    dismissBanner(state) {
```
oidoug, line 20: Nice! I really prefer this approach, much cleaner and standardize than having a worker hook using async storage like `useCtaBanner` used. Thanks :)

src/redux/store.ts
```diff
@@ -16,7 +17,11 @@ const persistConfig = {
   key: 'persist',
   version: 1,
   storage: AsyncStorage,
-  whitelist: [primarySafeSliceName, biometryToggleSliceName],
+  whitelist: [
+    primarySafeSliceName,
+    biometryToggleSliceName,
+    welcomeBannerSliceName,
```
oidoug, line 23: 👏

oidoug (approved):
Good job, thanks for refactoring useCtaBanner's hook stuff as well!

danibonilha (approved):
Nice! 🚀

PR #938, opened 2022-07-11
[CS-4161]: Re-enable backup flow with new storage structure

danibonilha:
### Description
This PR re-enables the backup flow by saving the new seed, and having a consistent restore flow using the `importWallet`, instead of rewriting stuff on the keychain, it also removes the ability to save the backup password, since it was not optional and it only worked for iOS. it moves the `backup.ts` to the cardstack side so I can add tests on the next PR. it fixes some minor issues while importing and deleting the wallet, it made sense to have it here, since I was testing these flows a lot. There are subtasks on this issue to be handled on a diff PR. 
 

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

I didn't record it, but I tested the following flows on both platforms:

- Previous backup from < 1.1.3 and restoring on this branch 
- Backing up and restoring from this branch


oidoug (approved):
LGTM, tested on both iOS and Android.

On Android I had an old backup that first time I loaded didn't work, it errored with 'invalid seed' log. I then deleted the cloud backup, re-imported the wallet with a seed phrase and back it up again successfully.

On iOS I didn't had a previous backup, doing a fresh one worked right away. Minor hick-up is that after deleting my wallet the icloud option didn't appear in the "Add Account" sheet, but it appeared after an app reload.

Also, another minor difference, on Android the cloud option appeared after the 'import seed' option, on iOS it appeared above, first line. By looking the code and PR this looks like an old issue.

ios:
Screenshot

android:
Screenshot

danibonilha:
> LGTM, tested on both iOS and Android.
> 
> On Android I had an old backup that first time I loaded didn't work, it errored with 'invalid seed' log. I then deleted the cloud backup, re-imported the wallet with a seed phrase and back it up again successfully.
> 
> On iOS I didn't had a previous backup, doing a fresh one worked right away. Minor hick-up is that after deleting my wallet the icloud option didn't appear in the "Add Account" sheet, but it appeared after an app reload.
> 
> Also, another minor difference, on Android the cloud option appeared after the 'import seed' option, on iOS it appeared above, first line. By looking the code and PR this looks like an old issue.

It is expected to have different behavior between Android and iOS, as Android requires the user to authenticate before we are can check if any backup exists, and iOS has a small delay, because it checks first if a backup exists, before showing the item.

PR #939, opened 2022-07-12
[CS-3964]: Remove rn-cool-modals, rn-masked-view and unused components

danibonilha:
### Description

This PR removes the `Portal` and all it's references and usages, like the walletLoading state from redux, it also removes the rn-masked-view lib, I did try to remove the old theme entirely but it's more tightly coupled than expected, so I just simplified a bit, while trying to removing the theme, I found a bunch of dead components, which is also removed in this PR. 

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


paulinhapenedo (approved):
Always nice to see old components and patches being deleted 🎉

PR #942, opened 2022-07-14
[CS-4086] In App Purchase base work

oidoug:
### Description

POC exploration work on In App Purchases and Hub's profile purchase query.

It's possible to purchase and create a profile from iOS, it logs hub's query output but it does not keep pulling for hub's profile creation task yet. There's a screen for testing this in Settings (DEV only).

To purchase on iOS in sandbox mode, you'll need:
- a physical device;
- to be on SOKOL;
- a sandbox apple connect account (there are good instructions here: https://www.iaphub.com/docs/set-up-ios/configure-sandbox-testing).

Unfortunately Google Play Store products are still not showing up in the app, I'm assuming it's related to banking connection issues, but it might be something else. Hard to say because theres no debugging info for this. We'll resume work on Play Store when our banking info clears.

I'll be writing tests for the two new hooks in a new PR.

Even though this is a proof of concept, I think it would be better to merge it if code is acceptable because this will be the base for the upcoming work on 4076 and others.

- [x] Completes #(CS-4086)

### Checklist

- [x] Tested on iOS
- [ ] ~~Tested on Android~~

### Screenshots

Screenshot


oidoug:
BTW, sorry for all the "wip" commits, I tried to squash those but I got a bunch of weird conflicts that were hard to properly merge. My bad for taking to long to push.

cardstack/src/screens/ProfilePurchaseScreen/ProfilePurchaseScreen.tsx
```diff
@@ -0,0 +1,112 @@
...
+            onChange={e =>
+              updateProfileInfo({
+                name: e.nativeEvent.text,
+              })
+            }
+            returnKeyType="next"
+            {...InputCommonProps}
+          />
+          <Text>Color:</Text>
+          <CardPressable onPress={onPressBusinessColor} position="relative">
```
paulinhapenedo: The name `CardPressable` seems a bit odd in the context that we're using. I mean, there's no card in this screen, neither in the Material Design concept (a box with shadow) or the Cardstack concept. 

I see that's the component handling the color picker and that it handles some sort of animation in a particular way, but does it make sense to use it here?

paulinhapenedo: Also the `position: relative` doesn't change anything here

cardstack/src/hooks/useProfileCreationData.ts
```diff
@@ -0,0 +1,61 @@
...
+    [profile, accountAddress]
+  );
+
+  return {
+    profile,
+    profileAttributes,
+    updateProfileInfo,
+    onPressBusinessColor,
+  };
+};
```
danibonilha: If this is just a POC, I'd rather not have this hook at all, which is just duplicating code from the ProfileForm, we could have the values "hardcoded"  or just changing in the dev mode manually. Specially because there's a task to handle the new ProfileScreen, colors etc

cardstack/src/screens/ProfilePurchaseScreen/ProfilePurchaseScreen.tsx
oidoug: Good catch!

CardPressable, it's simply a type of pressable. Yes, the naming could be better but it's a shared component I prefer to use instead of a touchable because it has a nice animation.

This screen is for testing and reference, it'll be discontinued when we get the new screens done.

cardstack/src/hooks/usePurchaseProfile.ts
```diff
@@ -0,0 +1,147 @@
...
+
+const PROVIDER = Device.isIOS ? 'apple' : 'google';
```
danibonilha: Instead of setting the provider here, we could add as a key of Device, something like, `IAPProvider`

cardstack/src/services/hub/hub-api.ts
```diff
@@ -90,6 +97,13 @@ export const hubApi = createApi({
         responseHandler: response => response.text(),
       }),
     }),
+    profilePurchases: builder.query<string, PostProfilePurchaseQueryParams>({
+      query: params => ({
+        url: routes.profilePurchases,
+        method: 'POST',
+        body: hubProfilePurchaseBody(routes.profilePurchases, params),
+      }),
+    }),
```
paulinhapenedo: Shouldn't this be a mutation since it's a `POST`?

cardstack/src/services/hub/hub-service.ts
```diff
@@ -138,3 +173,9 @@ export const registerFcmToken = (fcmToken: string) =>
 
 export const unregisterFcmToken = (fcmToken: string) =>
   store.dispatch(hubApi.endpoints.unregisterFcmToken.initiate({ fcmToken }));
+
+// Profile purchase and creation Queries
+
+export const profilePurchaseHubQuery = (
+  params: PostProfilePurchaseQueryParams
+) => store.dispatch(hubApi.endpoints.profilePurchases.initiate(params));
```
danibonilha: Why was this created as a separate query ?

cardstack/src/services/hub/hub-api.ts
danibonilha, line 113: If it's a POST it should actually be a `mutation` instead of a query

cardstack/src/hooks/usePurchaseProfile.ts
```diff
@@ -0,0 +1,147 @@
...
+    currentPurchaseError,
+    getProducts,
+    finishTransaction,
+    getAvailablePurchases,
+  } = useIAP();
+
+  // ** TEST CALL WITH STATIC RECEIPT **
+  // Receipt and associated data needs to be deleted from Hub's DB to retry this.
+  const fakeTestPurchase = useCallback(async () => {
+    const { data, error } = await profilePurchaseHubQuery({
```
danibonilha: by creating the ProfilePurchase endpoint as mutation you would already have this "callback" available, so we should replace this with the correct usage

cardstack/src/hooks/usePurchaseProfile.ts
```diff
@@ -0,0 +1,147 @@
...
+      iapReceipt: testReceipt,
+      provider: 'apple',
```
danibonilha: Shouldn't this be `PROVIDER` ?

cardstack/src/hooks/usePurchaseProfile.ts
```diff
@@ -0,0 +1,147 @@
...
+  const fetchProducts = useCallback(async () => {
+    logger.log('💰 Loading Products available for purchase.');
+
+    try {
+      const response = await getProducts(testSKUs);
+      logger.log('💰 Products response', response);
+    } catch (e) {
+      logger.sentry('Error fetching products:', e);
+    }
+  }, [getProducts]);
```
danibonilha, line 74: nit: we could abstract to a `useWorker`

cardstack/src/hooks/usePurchaseProfile.ts
```diff
@@ -0,0 +1,147 @@
...
+   * purchase but for some reason still doen't has the profile.
+   */
+  const fetchAvailablePurchases = useCallback(async () => {
+    try {
+      const response = await getAvailablePurchases();
+      logger.log('💰 Available Purchases response', response);
+    } catch (e) {
+      logger.sentry('Error fetching available purchases:', e);
+    }
+  }, [getAvailablePurchases]);
```
danibonilha, line 88: nit: we could abstract to a `useWorker`

cardstack/src/services/hub/hub-api.ts
oidoug: Yes, I think it should. Thanks.

cardstack/src/screens/ProfilePurchaseScreen/ProfilePurchaseScreen.tsx
```diff
@@ -0,0 +1,112 @@
...
+              </Button>
+            </Container>
+          ))}
+        <Container
+          width="100%"
+          height={2}
+          backgroundColor="grayBackground"
+          marginBottom={4}
+        />
+        {Device.isAndroid && (
```
danibonilha: why is this android only ?

cardstack/src/screens/ProfilePurchaseScreen/ProfilePurchaseScreen.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <Button onPress={() => fakeTestPurchase()}>Fake call purchase</Button>
```
danibonilha: `onPress={fakeTestPurchase}`

cardstack/src/screens/ProfilePurchaseScreen/ProfilePurchaseScreen.tsx
```diff
@@ -0,0 +1,112 @@
...
+        )}
+      </CenteredContainer>
+    </>
+  );
+};
+
+export default memo(ProfilePurchaseScreen);
```
danibonilha, line 53: I'll be ignoring the layout, giving that this is mostly copied from the profile form

cardstack/src/services/hub/hub-api.ts
oidoug: so, if it's a mutation maybe I don't need the imperative `profilePurchasesHubMutation` I added to hub-services. I think the mutation has a `updateValue` callback for that. Wdyt?

cardstack/src/screens/ProfilePurchaseScreen/ProfilePurchaseScreen.tsx
danibonilha: it's kinda of related to this topic, but not this PR, I was wondering, if we still need this component, after the RN update, maybe we could check that sometime.

cardstack/src/hooks/useProfileCreationData.ts
oidoug: True, this was valuable for me while experimenting with Hub's api. I don't mind removing.

I decided to "copy" parts of useProfileForm because that handles both appearance and data so I couldn't use it. But yeah, test data would be enough.

cardstack/src/services/hub/hub-service.ts
oidoug: Because I needed to do an imperative api call once profile info was ready and user pressed the buy button. But I should have used a mutation as pointed by Paula above. I'll remove this query.

cardstack/src/hooks/usePurchaseProfile.ts
oidoug: This fake call only works with apple provider bc it uses an apple store receipt.

cardstack/src/screens/ProfilePurchaseScreen/ProfilePurchaseScreen.tsx
oidoug: On iOS it's possible to test the correct flow with a sandbox user, on Android products are not showing up, so I added this fake call to check requests/ ui.

I'll leave it visible on ios too, just be mindful it will fail if you don't remove the test receipt from hub's DB.

cardstack/src/screens/ProfilePurchaseScreen/ProfilePurchaseScreen.tsx
oidoug: Yes, it's temporary.

cardstack/src/hooks/usePurchaseProfile.ts
oidoug: I just tried to use useWorker but since there are two workers the error logging gets more messy than is worth it, it will make more sense when we're using isLoading and showing errors in UI.

Leaving as is but with a note.

oidoug: Well, actually never mind 😂  I'll use the worker to show a loading in the mock screen.

cardstack/src/utils/device.ts
```diff
@@ -18,6 +20,7 @@ const Device = {
   scrollSheetOffset: isIOS ? -(screenHeight * 0.2) : 1,
   tabBarHeightSize: screenHeight * 0.1,
   enableBackup: true,
+  iapProvider: (isIOS ? 'apple' : 'google') as IAPProviderType,
```
danibonilha: Casting the whole value here, might mislead typescript in case of an "error", let's say we change 'apple' by mistake to `aple`, typescript will not complain that there is a mismatch with `IAPProviderType`.  We have two ways to make it better, the first one is to "cast" each string as a const, this basically says that `'apple'` is a type, and does not let ts, infer that it's a string: 

image

The other way is to make `IAPProviderType` an enum instead of type.

danibonilha (approved):
Great! LGTM, maybe it's worth to trigger releases from this branch to make sure nothing breaks on CI before merging it.

cardstack/src/utils/device.ts
oidoug: Humm ok, I think enum looks nicer. Thanks for the detailed info!

PR #943, opened 2022-07-14
[CS-4237] Handle deeplink when user dismiss biometric authentication once

paulinhapenedo:
### Description
This PR handles the scenario of a user that opens the app, authenticates, moves the app to the background in order to read a QR code or to click on a payment link and then dismisses the biometric authorization. Clicking on the payment link or scanning the QR code again would redirect the user directly into the PaymentScreen without asking again for authentication.

- We're only adding the deep link configuration whenever the user is authorized.
- We're calling the `handleDeepLink` function to the `Linking` listener so we can redirect the user properly.

- [x] Completes CS-4237

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/navigation/navigator.tsx
+++ b/cardstack/src/navigation/navigator.tsx
@@ -1,7 +1,7 @@
 import { NavigationContainer } from '@react-navigation/native';
 import React, { memo } from 'react';
 
-import { useAuthSelectorAndActions } from '@cardstack/redux/authSlice';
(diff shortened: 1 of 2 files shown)


danibonilha (approved):
LGTM, but since we did it together it would be nice for @oidoug to take a look too.

oidoug (approved):
👏  LGTM

PR #945, opened 2022-07-18
[CS-4233] Set wallet as backed up when importing a wallet from a cloud backup

paulinhapenedo:
### Description
When importing a wallet from a cloud backup, we're not setting the wallet as backed up in our state management. This PR aims to fix that by calling the same function we use on the backup flow inside the app.

- [x] Completes #CS-4233

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


oidoug (approved):
LGTM, no issues found :)

danibonilha (approved):
LGTM! 🚀

PR #946, opened 2022-07-19
[CS-4259] [Tests] Add test coverage to Profile Purchase POC

oidoug:
### Description

Adds test coverage for hook `usePurchaseProfile`.

- [x] Completes #(CS-4259)

No visual updates.

### DIFF
--- a/cardstack/src/hooks/__tests__/usePurchaseProfile.test.ts
+++ b/cardstack/src/hooks/__tests__/usePurchaseProfile.test.ts
@@ -2,8 +2,6 @@ import { renderHook } from '@testing-library/react-hooks';
 import { act } from '@testing-library/react-native';
 import { useIAP, Product, Purchase } from 'react-native-iap';
 
-// import { CreateBusinessInfoDIDParams } from '@cardstack/types';
-
 import { useAccountProfile } from '@rainbow-me/hooks';
 import logger from 'logger';
 
@@ -35,7 +33,6 @@ jest.mock('@rainbow-me/hooks', () => ({
 
 jest.mock('react-native-iap', () => ({
   useIAP: jest.fn(),
-  // requestPurchase: mockedRequestPurchase,
 }));
 
 describe('usePurchaseProfile', () => {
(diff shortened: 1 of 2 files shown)


cardstack/src/hooks/__tests__/usePurchaseProfile.test.ts
```diff
@@ -0,0 +1,156 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { act } from '@testing-library/react-native';
+import { useIAP, Product, Purchase } from 'react-native-iap';
+
+// import { CreateBusinessInfoDIDParams } from '@cardstack/types';
```
danibonilha: letfover

cardstack/src/hooks/__tests__/usePurchaseProfile.test.ts
```diff
@@ -0,0 +1,156 @@
...
+  accountAddress: '1234567890',
+};
+
+jest.mock('@rainbow-me/hooks', () => ({
+  useAccountProfile: jest.fn(),
+}));
+
+jest.mock('react-native-iap', () => ({
+  useIAP: jest.fn(),
+  // requestPurchase: mockedRequestPurchase,
```
danibonilha: leftover

danibonilha (approved):
LGTM, just some leftovers, we could  however add a couple more of test cases, to check the `validateReceiptCreateProfile ` which I believe is pretty important, to check if after a valid purchase, the mutation is called or add a new task to handle it eventually

oidoug:
@danibonilha , I'll try test the mutation before merging, I see we have something similar in `useRequestPrepaidCardScreen.test.ts`.

LMK if theres any other good example 👍

oidoug:
@danibonilha , I've added a couple more tests to ensure `currentPurchase` triggers the verification, but didn't add tests to different mutation response statuses. I think that is valuable but still too soon to add, because the hook currently doesn't handle that either. I'll add more as we expand.

PR #947, opened 2022-07-19
[CS-4263]: Create first version of profile name screen

danibonilha:
### Description

This is the initial version of the ProfileNameScreen, it doesn't handle any logic, besides animation, also the styles and a couple of magic numbers are not extract into constants yet, bc there's still inconsistency between different devices (with the keyboard opened some devices shows the white background and others are too far up), and it's easier to change while developing, but it seems best to finish all the screen requirements and then revisit the animation across devices, then I'll extract everything to it's own place. 

PS: shadows have not been implemented yet.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Android Pixel:
android-keybaord

iPhone11: 
Simulator Screen Recording - iPhone 11 - 2022-07-19 at 18 10 12

iPhone 12 mini
Simulator Screen Recording - iPhone 12 mini - 2022-07-19 at 18 11 13


cardstack/src/utils/device.ts
```diff
@@ -17,6 +17,12 @@ const Device = {
   supportsHapticFeedback: isIOS,
   scrollSheetOffset: isIOS ? -(screenHeight * 0.2) : 1,
   tabBarHeightSize: screenHeight * 0.1,
+  keyboardEventWillShow: isIOS
+    ? ('keyboardWillShow' as const)
+    : ('keyboardDidShow' as const),
+  keyboardEventWillHide: isIOS
+    ? ('keyboardWillHide' as const)
+    : ('keyboardDidHide' as const),
```
danibonilha, line 25: Android doesn't have the `keyboardWillShow` event, just the `keyboardDidShow` one, so the animation kinda of happens after the keyboard is shown, and on iOS it happens while the keyboard is opening.

cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
```diff
@@ -0,0 +1,176 @@
...
+    ];
+
+    if (Device.isIOS) {
+      transformArray.push({
+        translateY: animated.interpolate({
+          inputRange: [0, 1],
+          outputRange: [0, -screenHeight * 0.091],
+        }),
+      });
+    }
```
danibonilha, line 81: Since the keyboard events are diff for each platform, on iOS we need to translateY while opening the keyboard, whereas on Android we just scale since the view has already been updated with the keyboard height

cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  return (
+    <SafeAreaView
+      backgroundColor="backgroundDarkPurple"
+      flex={1}
+      paddingHorizontal={5}
+      justifyContent="space-between"
+    >
+      <Container flex={0.1} alignItems="flex-end" paddingVertical={2}>
+        <Text fontSize={13} color="teal" fontFamily="OpenSans-Bold">
```
paulinhapenedo, line 104: Quick question: this is the first time I'm seeing the font-family being declared. Is it okay to use it like this or should we be using the `fontFamilyVariants` file?

paulinhapenedo (approved):
Everything looks good. Animation that works good across multiple devices is always difficult, so kudos 👏 

Adding some evidences for future reference of the animation behavior on other Android devices (video is from a Pixel 4 and static images are from a Motorola G8 with Android 11).

**First interaction with the input field**


**After typing the first word** 


**Same behavior above, but video**
https://user-images.githubusercontent.com/690904/180029863-6bc8706d-ed83-49f5-b19b-3a096b98bacf.mov

cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
danibonilha: Actually we might have text variants for this, this was added just for testing

PR #948, opened 2022-07-20
[CS-4077]: Improve ProfileNameScreen w/ animation and hook

danibonilha:
### Description

This PR fixes the devices inconsistency, iOS seems to be running smoothly and with all the items positioned in the same spot regardless of device size, for Android I believe most of the cases are covered, with slight differences on smaller devices, also it hides the description once the value is inputed. It also adds the `useProfileNameScreen` hooks with empty functions until we have the whole flow, and adds the validation on the Continue btn. 

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Small android device - Nexus 5: 
image

Pixel 4a:
androidddd

iPhone 8 :
Simulator Screen Recording - iPhone 8 - 2022-07-20 at 16 19 34

Iphone 11 Pro:

image

### DIFF
--- a/cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
+++ b/cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
@@ -4,7 +4,9 @@ import {
   Keyboard,
   KeyboardAvoidingView,
   ScaleTransform,
+  StyleSheet,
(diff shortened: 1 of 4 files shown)


paulinhapenedo (approved):
This is a good solution for the animation issue on Android devices. It's looking good on all devices I tested 👏

PR #949, opened 2022-07-20
[CS-4262] Adds SuffixedInput component for username

oidoug:
### Description

Adds `SuffixedInput` component that renders a customizable input with a Text label after it that wraps to new line as text gets bigger.

Known issue: Android text input field starts to squeeze the text to the left after it starts expanding horizontally and I can't find a fix, I'll get back to it once we have a testable version ready.

- [x] Completes #(CS-4262)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/180076904-5095bf13-c970-4ec9-8e00-c9a30acf5357.mp4

https://user-images.githubusercontent.com/129619/180076988-bdd9181b-5b41-49d9-be00-ef562d395144.mp4

### DIFF
--- a/cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
+++ b/cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
@@ -8,16 +8,20 @@ import {
   Text,
   TextProps,
 } from '@cardstack/components';
+import { Device } from '@cardstack/utils';
 import { fontFamilyVariants, colors } from '@cardstack/theme';
 
 const styles = StyleSheet.create({
   input: {
-    top: 0.5, // Input comp. renders inner text slightly above Text comp.
     color: colors.teal,
-    minWidth: '32%',
-    fontSize: 24,
+    // Input's font on android is rendering a lot smaller, not sure why.
+    minWidth: Device.isAndroid ? '40%' : '32%',
+    fontSize: Device.isAndroid ? 32 : 24,
+    padding: 0, // Clears phantom padding on android. Keep it consistent on ios.
+    left: 0,
+    ...fontFamilyVariants.bold,
   },
-  textFont: {
+  suffix: {
     fontSize: 24,
     ...fontFamilyVariants.bold,
   },
@@ -40,25 +44,31 @@ const SuffixedInput = forwardRef(
     }: SuffixedInputProps,
     ref
   ) => (
-    <Container flexDirection="row" flexWrap="wrap">
+    <Container
+      width="100%"
+      flexDirection="row"
+      flexWrap="wrap"
+      justifyContent="flex-start"
+      alignItems="center"
+    >
       <Input
         ref={ref}
+        style={styles.input}
+        placeholderTextColor={colors.secondaryText}
+        paddingRight={1}
         autoFocus
-        autoCapitalize="none"
         multiline={false}
         spellCheck={false}
         autoCorrect={false}
         maxLength={maxLength}
-        style={[styles.input, styles.textFont]}
+        autoCapitalize="none"
         placeholder="username"
         textContentType="username"
-        underlineColorAndroid="transparent"
         keyboardType="twitter"
-        paddingRight={1}
-        placeholderTextColor={colors.secondaryText}
+        underlineColorAndroid="transparent"
         {...props}
       />
-      <Text style={styles.textFont} color="white" {...suffixTextProps}>
+      <Text style={styles.suffix} color="white" {...suffixTextProps}>
         {suffixText}
       </Text>
     </Container>
--- a/cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
+++ b/cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
@@ -1,15 +1,11 @@
-import React, { memo, forwardRef } from 'react';
-import { StyleSheet } from 'react-native';
+import React, { memo } from 'react';
+import { StyleSheet, TextInput, TextInputProps } from 'react-native';
 
-import {
-  Container,
-  Input,
-  InputProps,
-  Text,
-  TextProps,
-} from '@cardstack/components';
+import { Container, Text, TextProps } from '@cardstack/components';
 import { fontFamilyVariants, colors } from '@cardstack/theme';
 import { Device } from '@cardstack/utils';
+
+import { strings } from './strings';
 
 const styles = StyleSheet.create({
   input: {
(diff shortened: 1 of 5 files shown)


cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+        paddingRight={1}
+        autoFocus
+        multiline={false}
+        spellCheck={false}
+        autoCorrect={false}
+        maxLength={maxLength}
+        autoCapitalize="none"
+        placeholder="username"
+        textContentType="username"
+        keyboardType="twitter"
```
danibonilha: Is there any specific reason to use this type of keyboard ?

cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+        underlineColorAndroid="transparent"
+        {...props}
```
danibonilha: Is it intended to be able to override all the props above ?

cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <Input
+        ref={ref}
+        style={styles.input}
+        placeholderTextColor={colors.secondaryText}
+        paddingRight={1}
+        autoFocus
+        multiline={false}
+        spellCheck={false}
+        autoCorrect={false}
+        maxLength={maxLength}
```
danibonilha: Meybe this component could have a custom onChangeText that returns the whole value including the suffix.

cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+});
+
+interface SuffixedInputProps extends InputProps {
+  suffixText?: string;
+  suffixTextProps?: TextProps;
+}
+
+const DEFAULT_MAX_LENGTH = 25;
+
+const SuffixedInput = forwardRef(
```
danibonilha: any specific case to use `forwardRef` ?

cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+} from '@cardstack/components';
+import { fontFamilyVariants, colors } from '@cardstack/theme';
+import { Device } from '@cardstack/utils';
+
+const styles = StyleSheet.create({
+  input: {
+    color: colors.teal,
+    // Input's font on android is rendering a lot smaller, not sure why.
+    minWidth: Device.isAndroid ? '40%' : '32%',
+    fontSize: Device.isAndroid ? 32 : 24,
```
danibonilha: We definitely need to revisit this, this minWidth seems like a hack just to avoid cropping the `e`  from username on Android, on iOS it doesn't seem to make any diff, the fontSize however, at lear on my device, 24 works perfectly and 32 breaks.

cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+      flexDirection="row"
+      flexWrap="wrap"
+      justifyContent="flex-start"
+      alignItems="center"
+    >
+      <Input
+        ref={ref}
+        style={styles.input}
+        placeholderTextColor={colors.secondaryText}
+        paddingRight={1}
```
danibonilha: This adds a unnecessary spacing on IOS

cardstack/src/screens/Dev/DesignSystemScreen.tsx
```diff
@@ -39,6 +44,12 @@ const DesignSystemScreen = () => {
   const renderItem = useCallback(
     ({ item, section: { title } }) => {
       switch (title) {
+        case 'Input':
+          return (
+            <CenteredContainer padding={2}>
+              <SuffixedInput suffixText=".card.xyz" />
+            </CenteredContainer>
```
danibonilha: it's best to use a simple container here

danibonilha (approved):
Maybe trying to use the `Input` component might've added some challenges, since it has another container wrapping, there are a couple of issues on Android, besides the cropping line, which also happens on iOS if the user uses bigger characters, I believe we can keep this one as it is, to have the flow completed and then we might need to create are our Input, maybe something like the Pin Code one, also the TextInput component can receive children as props, so we might be able to change the styles and manually add the suffix

cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
oidoug: It presents a keyboard that is closer in character types to a username, but I see now that's only on iOS. Better use 'default'.

cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
oidoug: We'll only use the slug, so better keep it separate.

cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
oidoug: Though it would be good to leave it open for customization, but we'll probably not re-use this input anyways. I'll remove it.

cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
oidoug: Yes, agreed. Theres something odd with Input, maybe it's the masked input, but on android the font was rendering smaller.

The minWidth is to keep the .card.xyz fixed until text grows above min width. The difference in %s is related to font size being rendered differently.

cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
oidoug: To both actually, it's there because theres a small gap there in the design mocks. I think it's intentional.

cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
oidoug: Not really, since we'll only use it alone, no need to move focus or similar. I'll remove it.

oidoug:
I using RN's TextInput now but still have the same issues with input font size on android. That's for all inputs in android, I tested the one in your PR and it has the same behavior... Might be some issue with the font. I'll investigate this more soon.

oidoug:
Found out it's the device font scaling that is messing up the sizes. Our Text component turns it off, but text input doesn't.

PR #951, opened 2022-07-21
[CS-4285]: Fix backup restoring after first restore

danibonilha:
### Description

There was a missing property (backupDate) after restoring the backup, and the data was being overwritten which caused the backup structure to be invalid when we tried to restore the same backup again, on a different device for example. In other to keep the data as consistent as possible, I'm passing the whole backedUp wallet so we overwrite just the new fields and keep the backupDate.  

The best way to test it is to have an old backup from 1.1.3, restore on this version, and then try to restore it again.


oidoug (approved):
LGTM, no issues backing up and restoring.

paulinhapenedo (approved):

PR #950, opened 2022-07-21
[CS-4075] In-app purchase screen

paulinhapenedo:
### Description
This PR adds the In-app purchase screen without any logic implemented yet. 

- [x] Completes #CS-4075

### Checklist

- [x] Tested on a small device
- [x] Tested on iOS
- [x] Tested on Android

PR #952, opened 2022-07-22
[CS-4074] Profile slug input screen

oidoug:
### Description

Profile username (slug) screen. Has the basic UI, validation will be handled in separate PR (4297), after ticket 4085 is also cleared (refactoring `checkBusinessIdUniqueness` and `createBusinessInfoDID` api calls to use rtk-query).

- [x] Completes #(CS-4074)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot

Screenshot


cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
```diff
@@ -0,0 +1,90 @@
...
+    >
+      <Container
+        alignItems="center"
+        justifyContent="space-between"
+        flexDirection="row"
+        flex={0.15}
+      >
+        <Touchable onPress={onGoBackPressed}>
+          <Icon name="chevron-left" color="teal" size={30} />
+        </Touchable>
```
danibonilha, line 41: it seems we need to add a negative margin, or even better, check why this icon component has a "square" around it, bc its start positioning is inconsistent with the other components

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <Touchable onPress={onSkipPressed}>
+          <Text fontSize={13} color="teal" fontFamily="OpenSans-Semibold">
```
danibonilha: there's a textVariant for semibold

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
```diff
@@ -0,0 +1,90 @@
...
+      </Container>
+      <Container flex={1}>
+        <Container width="90%" paddingBottom={4}>
+          <Text
+            fontSize={layouts.titleTextSize}
+            color="white"
+            {...fontFamilyVariants.light}
+          >
+            {strings.header}
+          </Text>
```
danibonilha: we should add a textVariant for this one, if @paulinhapenedo didn't create it yet

cardstack/src/screens/Profile/ProfileSlugScreen/components/UsernameValidFeedback.tsx
```diff
@@ -0,0 +1,51 @@
...
+const UsernameValidFeedback = ({
+  invalidUsernameMessage,
+}: UsernameValidFeedbackProps) => {
+  const attr: UsernameFeedbackType = useMemo(
+    () =>
+      !invalidUsernameMessage
+        ? {
+            iconName: 'check',
+            iconColor: 'lightGreen',
+            description: strings.input.valid,
```
danibonilha: Giving the upcoming screens for the project, this description should be a custom prop, since we are going to have this type of usage 
image

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
```diff
@@ -0,0 +1,90 @@
...
+          <SuffixedInput
+            value={username}
+            onChangeText={onUsernameChange}
+            suffixText={strings.input.domainSuffix}
+          />
+        </Container>
+        <Container width="100%">
+          <UsernameValidFeedback
+            invalidUsernameMessage={invalidUsernameMessage}
+          />
```
danibonilha: I believe there's a missing step here, this `UsernameValidFeedback` handles if a name is valid or not, but there's the case where we didn't validate yet and should not show any feedback

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
```diff
@@ -0,0 +1,90 @@
...
+          <Text fontSize={layouts.smallTextSize} color="grayText">
```
danibonilha: nit: it would be nice to add some kind of animation, opacity whatever, while showing the feedback

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
```diff
@@ -0,0 +1,90 @@
...
+            {strings.input.description}
+          </Text>
+        </Container>
+      </Container>
+      <Container flex={0.2}>
+        <Button disabled={!!invalidUsernameMessage}>
+          {strings.buttons.continue}
+        </Button>
```
danibonilha, line 70: I just realized the disabled btns have a different design, we can leave as it is, since the other screen is the same way, and revisit it later

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
```diff
@@ -0,0 +1,90 @@
...
+          <UsernameValidFeedback
+            invalidUsernameMessage={invalidUsernameMessage}
+          />
+          <Text fontSize={layouts.smallTextSize} color="grayText">
+            {strings.input.description}
+          </Text>
+        </Container>
+      </Container>
+      <Container flex={0.2}>
+        <Button disabled={!!invalidUsernameMessage}>
```
danibonilha: disabling this should rely on error, and on empty field

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  const onSkipPressed = useCallback(() => {
+    // TODO
+  }, []);
+
+  // TODO (CS-4085): Validate username uniqueness and report if not.
+  const onUsernameChange = useCallback(async ({ nativeEvent: { text } }) => {
+    setUsername(text);
+  }, []);
+
```
danibonilha, line 26: there's no onContinuePress

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
```diff
@@ -0,0 +1,90 @@
...
+    invalidUsernameMessage,
+    onGoBackPressed,
+    onSkipPressed,
+  } = useProfileSlugScreen();
+
+  return (
+    <SafeAreaView
+      backgroundColor="backgroundDarkPurple"
+      flex={1}
+      paddingHorizontal={layouts.defaultPadding}
```
danibonilha: FYI: on ProfileNameScreen I added defaultPadding as a param, because I needed the value in order to create the negative margin to hide the phonePreview overflow, but it's not a pattern or anything like it

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
paulinhapenedo: It's the same discussion from https://github.com/cardstack/cardwallet/pull/950#discussion_r927654510. All icons have a box around them, it is by design. If we were to align copy with the icon, we must take that "space" into consideration.

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
paulinhapenedo: Haven't created one yet.

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
oidoug: Adding a negative "left" for now, we can work on a proper fix on CS-4293.

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
oidoug: Ah yes, nice catch. I would argue that we should update the disabled state for all uses, but on white screens gray seems to work better. I'll ping Gardella, get his input.

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
oidoug: I'll add some in the next PR 👍

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
oidoug: Yes, I realized that too, I'll improve it after I have validation working on upcoming PR.

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
oidoug: We'll have a "valid" flag after validation, better than checking for errors. Also on upcoming PR.

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
oidoug: Padding for this new screens is 5 and it should remain consistent, maybe we can have a "content" container with correct values for reuse. WDYT?

danibonilha (approved):
LGTM

PR #953, opened 2022-07-22
[CS-4296]: Fix payLink not clearing after first attempt

danibonilha:
### Description

This Pr fixes an issue, where if an user opens a payment link, closes the screen, goes to bg and back the same link keeps opening, this was happening because we have a listener to "manually" store the link, and once we have an Url, and call `openLink` the subscription is triggered again thus reopening the same link we just did. Now we have a condition to prevent this behavior.

### Screenshots

Simulator Screen Recording - iPhone 11 Pro - 2022-07-22 at 12 58 05

Simulator Screen Recording - iPhone 11 Pro - 2022-07-22 at 13 00 11

### DIFF
--- a/src/handlers/deeplinks.js
+++ b/src/handlers/deeplinks.js
@@ -8,7 +8,7 @@ import {
 import { decodeNestedURI } from '@cardstack/utils';
 import logger from 'logger';
 
-export default function handleDeepLink(url) {
+export default function handleWcDeepLink(url) {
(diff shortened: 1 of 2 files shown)


paulinhapenedo (approved):
LGTM!
The new function names convey better the scope of the function too.

oidoug (commented):
LGTM

PR #954, opened 2022-07-25
[CS-4293] Create custom header for IAP screens

paulinhapenedo:
### Description
This PR aims to create a custom header that covers all the cases we currently have for the In-app purchase screens. The only screen of this flow that's not using it yet is the `ProfileNameScreen` since Dani is changing things on this screen.

I'm not super fan of using negative margins to solve alignment issues, but I think it's okay to use it in this particular case.
Other thing that's worth noting is that I'm not using `SafeAreaView` on the component because all of the screens we are building already use that component.

- [x] Completes #CS-4293

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Android
Screenshot

Screenshot

Screenshot

Screenshot

### iOS
Screenshot

Screenshot

Screenshot

Screenshot

### DIFF
--- a/cardstack/src/components/MainHeader/ProfileHeader.tsx
+++ b/cardstack/src/components/MainHeader/ProfileHeader.tsx
@@ -1,0 +1,56 @@
+import { useNavigation } from '@react-navigation/native';
+import React from 'react';
+
+import {
+  Container,
+  Icon,
+  Text,
+  Touchable,
+  ContainerProps,
+  IconProps,
+} from '@cardstack/components';
+
+interface Props extends ContainerProps {
+  leftIconProps?: IconProps;
+  showLeftIcon?: boolean;
+  showSkipButton?: boolean;
+  onSkipPress?: () => void;
+}
+
+const ProfileHeader = ({
+  leftIconProps,
+  showLeftIcon = true,
+  showSkipButton = true,
+  onSkipPress,
+}: Props) => {
+  const { goBack } = useNavigation();
+
+  return (
+    <Container
+      flexDirection="row"
+      alignItems="center"
+      justifyContent={showLeftIcon ? 'space-between' : 'flex-end'}
+      flex={0.1}
+    >
+      {showLeftIcon && (
+        <Icon
+          color="teal"
+          name="chevron-left"
+          onPress={goBack}
+          size={30}
+          left={-8}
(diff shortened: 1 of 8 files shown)


cardstack/src/components/MainHeader/ProfileHeader.tsx
```diff
@@ -0,0 +1,56 @@
...
+} from '@cardstack/components';
+
+interface Props extends ContainerProps {
+  leftIconProps?: IconProps;
+  showLeftIcon?: boolean;
+  showSkipButton?: boolean;
+  onSkipPress?: () => void;
+}
+
+const ProfileHeader = ({
```
oidoug: nit: This header will also be used by all the new onboarding screens, I suggest a more broad name, like `InPageHeader`.

oidoug (approved):
LGTM!

cardstack/src/components/MainHeader/ProfileHeader.tsx
```diff
@@ -0,0 +1,56 @@
...
+      flex={0.1}
+    >
+      {showLeftIcon && (
+        <Icon
+          color="teal"
+          name="chevron-left"
+          onPress={goBack}
+          size={30}
+          left={-8}
+          {...leftIconProps}
```
danibonilha: nit: receiving only the iconName as props enforces a "better" pattern, giving that the size and position will always be the same, whereas passing through all icon props gives the ability to override everything . `chevron-left` could be the default iconName, and instead of having a `showLeftIcon` we could just check if iconName exists but we could argue both ways about the flag itself bc in order to hide it we'd pass iconName as undefined.

danibonilha (approved):

cardstack/src/components/MainHeader/ProfileHeader.tsx
paulinhapenedo: `OnboardingHeader`, maybe?

cardstack/src/components/MainHeader/ProfileHeader.tsx
paulinhapenedo: Size will be the same, but position may not be the same. When trying with the X icon, the `left` prop needed to be different than -8, as well as other icons I've tested. That's why I went with the approach of allowing the developer to overwrite props as needed.

PR #955, opened 2022-07-25
[CS-4085-4297] Username Validation

oidoug:
### Description

This PR adds improvements to slug validation, it also:

- Moves merchant id validation to RTKQuery;
- Adds local string validation using SDK's `validateMerchantId` call;
- Adds animation when validation feedback becomes visible.

To test it, `ProfileSlugScreen` needs to be called after wallet has been initialized.

It is still missing tests. I'll add test coverage in a new PR (CS-4300).

- [x] Completes #(CS-4085)
- [x] Completes #(CS-4297)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/180869196-235f54b3-f427-4fda-8774-f9f2f3a279a0.mp4

### DIFF
--- a/cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
+++ b/cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
@@ -6,20 +6,18 @@ import {
   SafeAreaView,
   Button,
   InPageHeader,
+  ValidationMessage,
 } from '@cardstack/components';
 import SuffixedInput from '@cardstack/components/Input/SuffixedInput/SuffixedInput';
 
-import ValidationMessage from './components/ValidationMessage';
 import { strings } from './strings';
 import { useProfileSlugScreen } from './useProfileSlugScreen';
 
 const ProfileSlugScreen = () => {
   const {
-    username,
-    onUsernameChange,
-    isUsernameValid,
-    showValidationMessage,
-    message,
+    slug,
+    onSlugChange,
+    slugValidation: { slugAvailable, detail: slugFeedback },
     onContinuePress,
   } = useProfileSlugScreen();
 
@@ -38,16 +36,16 @@ const ProfileSlugScreen = () => {
 
         <Container paddingBottom={1}>
           <SuffixedInput
-            value={username}
-            onChangeText={onUsernameChange}
+            value={slug}
+            onChangeText={onSlugChange}
             suffixText={strings.input.domainSuffix}
           />
         </Container>
         <Container width="100%">
           <ValidationMessage
-            isVisible={showValidationMessage}
-            isValid={isUsernameValid}
-            message={message}
+            isVisible={!!slugFeedback}
+            isValid={slugAvailable}
+            message={slugFeedback}
           />
           <Text variant="pageDescriptionSmall">
             {strings.input.description}
@@ -55,10 +53,7 @@ const ProfileSlugScreen = () => {
         </Container>
       </Container>
       <Container flex={0.2}>
-        <Button
-          onPress={onContinuePress}
-          disabled={!(isUsernameValid && showValidationMessage)}
-        >
+        <Button onPress={onContinuePress} disabled={!slugAvailable}>
           {strings.buttons.continue}
         </Button>
       </Container>
--- a/cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
+++ b/cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
@@ -18,7 +18,7 @@ const ProfileSlugScreen = () => {
     username,
     onUsernameChange,
     isUsernameValid,
-    showMessage,
+    showValidationMessage,
     message,
     onContinuePress,
   } = useProfileSlugScreen();
@@ -45,7 +45,7 @@ const ProfileSlugScreen = () => {
         </Container>
         <Container width="100%">
           <ValidationMessage
-            isVisible={showMessage}
+            isVisible={showValidationMessage}
             isValid={isUsernameValid}
             message={message}
           />
@@ -57,7 +57,7 @@ const ProfileSlugScreen = () => {
       <Container flex={0.2}>
         <Button
           onPress={onContinuePress}
-          disabled={!(isUsernameValid && showMessage)}
+          disabled={!(isUsernameValid && showValidationMessage)}
         >
           {strings.buttons.continue}
         </Button>
--- a/cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
+++ b/cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
@@ -53,7 +53,7 @@ const ProfileSlugScreen = () => {
         </Container>
       </Container>
       <Container flex={0.2}>
-        <Button onPress={onContinuePress} disabled={!slugAvailable}>
+        <Button onPress={onContinuePress} blocked={!slugAvailable}>
           {strings.buttons.continue}
         </Button>
       </Container>
--- a/cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
+++ b/cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
@@ -9,7 +9,7 @@ import {
 } from '@cardstack/components';
 import SuffixedInput from '@cardstack/components/Input/SuffixedInput/SuffixedInput';
 
-import UsernameValidFeedback from './components/UsernameValidFeedback';
+import ValidationMessage from './components/ValidationMessage';
 import { strings } from './strings';
 import { useProfileSlugScreen } from './useProfileSlugScreen';
 
@@ -44,10 +44,10 @@ const ProfileSlugScreen = () => {
           />
         </Container>
         <Container width="100%">
-          <UsernameValidFeedback
+          <ValidationMessage
+            isVisible={showMessage}
             isValid={isUsernameValid}
             message={message}
-            isVisible={showMessage}
           />
           <Text variant="pageDescriptionSmall">
             {strings.input.description}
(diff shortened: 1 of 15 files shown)


cardstack/src/screens/Profile/ProfileSlugScreen/components/UsernameValidFeedback.tsx
```diff
@@ -16,38 +18,42 @@ interface UsernameFeedbackType {
...
             iconName: 'check',
             iconColor: 'lightGreen',
-            description: validMessage || strings.input.valid,
+            description: message || strings.input.valid,
           }
         : {
             iconName: 'x',
             iconColor: 'red',
-            description: invalidMessage,
+            description: message || strings.input.invalid,
```
danibonilha: there's no difference between the red message and lightGreen message ? Shouldn't we two different messages ?

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
```diff
@@ -1,10 +1,21 @@
...
+  const [validateUsername, setValidateUsername] = useState('');
+  const [isUsernameValid, setIsUsernameValid] = useState(false);
+  const [message, setMessage] = useState('');
 
-  const [invalidUsernameMessage] = useState<string | undefined>(
-    'Already taken'
+  const { data, error } = useMerchantInfoValidateSlugQuery(
+    { slug: validateUsername },
+    {
+      skip: validateUsername === '',
```
danibonilha: nit: `!validateUsername`

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
```diff
@@ -1,10 +1,21 @@
...
+
+import { useMerchantInfoValidateSlugQuery } from '@cardstack/services';
+
+const MIN_USERNAME_LENGTH = 4;
 
 export const useProfileSlugScreen = () => {
-  const [username, setUsername] = useState<string>('');
+  const [username, setUsername] = useState('');
+  const [validateUsername, setValidateUsername] = useState('');
+  const [isUsernameValid, setIsUsernameValid] = useState(false);
```
danibonilha: I'm a bit confuse about these states, if we only set the userName once it's valid, why do we need a flag for that too, also `validUsername` seems to read better

cardstack/src/services/hub/hub-api.ts
```diff
@@ -30,6 +34,7 @@ const routes = {
   exchangeRates: '/exchange-rates',
   registerFCMToken: '/push-notification-registrations',
   profilePurchases: '/profile-purchases',
+  merchantInfoValidateSlug: '/merchant-infos/validate-slug',
```
danibonilha: even though the api is still related to merchant, maybe it's best for us to use profile references, like, validateProfileSlug

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
```diff
@@ -58,14 +60,21 @@ const ProfileSlugScreen = () => {
           />
         </Container>
         <Container width="100%">
-          <UsernameValidFeedback invalidMessage={invalidUsernameMessage} />
+          <UsernameValidFeedback
+            isValid={isUsernameValid}
+            message={message}
+            isVisible={showMessage}
+          />
```
paulinhapenedo: I know that we've already approved the name for this component, but with this new implementation, it's kind of confusing to call it `ValidFeedback` that receives a prop `isValid` that can be false.

What do you think about `ValidationMessage`?

cardstack/src/screens/Profile/ProfileSlugScreen/components/UsernameValidFeedback.tsx
oidoug: There's also a `isValid` prop, and the message is needed for both states so I think it's a better api this way. I don't mind changing though.

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
oidoug: We need to validate the input in two passes, one with sdk and one with the api. I'll only check the api if the first check passes. To check on the api, I use/set `validateUsername`. Having `validateUsername` set doesn't mean the username is actually valid, if for example, it's not available to use.

Thats why theres an extra `isUsernameValid` state. Hope that clarifies 😅

cardstack/src/services/hub/hub-api.ts
oidoug: True, I'll update it.

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
oidoug: You're right, that sounds better, I'll update it 👍

danibonilha: Nice thought on this one, I agree that to being more customizable makes sense to change the name, also it makes sense to receive all feedback messages as props, instead of having a default one related to the username.

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
```diff
@@ -19,17 +30,47 @@ export const useProfileSlugScreen = () => {
...
+  const onUsernameChange = useCallback(
+    async text => {
+      setUsername(text.trim());
+    },
+    [setUsername]
+  );
+
+  const showMessage = useMemo(() => username.length >= MIN_USERNAME_LENGTH, [
+    username,
+  ]);
```
danibonilha: nit: Is message used just for error cases ? if so using showErrorMessage might improve readability

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
danibonilha: right, maybe we can simplify that by using ref some kind of ref ? Ping me if you need more clarification

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
```diff
@@ -19,17 +30,47 @@ export const useProfileSlugScreen = () => {
...
+
+  useEffect(() => {
+    if (username.length >= MIN_USERNAME_LENGTH) {
```
danibonilha: we are checking this condition twice, maybe checking on the `onUsernameChange` once makes more sense ?

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
danibonilha: I'm thinking the validation api might work better here by being a lazyQuery, instead of a simple query, if that simplifies this path, that'd be great

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
oidoug: No, this is for showing/hiding valid and invalid messages.

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
oidoug: Yes, that makes sense, I'll try it out :)

cardstack/src/screens/Profile/ProfileSlugScreen/components/ValidationMessage.tsx
```diff
@@ -0,0 +1,60 @@
...
+  isValid,
+  isVisible = true,
+}: ValidationMessageProps) => {
+  const attr: UsernameFeedbackType = useMemo(
+    () =>
+      isValid
+        ? {
+            iconName: 'check',
+            iconColor: 'lightGreen',
+            description: message || strings.input.valid,
```
danibonilha: If this is reusable component, we shouldn't default the messages to username valid/ username invalid. Also we could extract this to the components folder

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
```diff
@@ -15,16 +21,48 @@ export const useProfileSlugScreen = () => {
...
+
+      if (sdkIDValidationError) {
+        setIsUsernameValid(false);
+        setMessage(sdkIDValidationError);
+
+        return;
+      }
+
+      validateSlugHub({ slug: username });
+    }
```
danibonilha: the validations on useEffect might be a bit confusing, I'm proposing a new solution on this branch `dani/slug-validation-suggestion`. feel free to cherry-pick or adapt anything you'd like, I also took the liberty to change the username references to slug, to avoid confusion. The main point of this new solution, is that there's a new "state" where we show a feedback about the 4 chars, since it's not very straightforward why the flow doesn't work if you type just 3 chars. You can compare the changes [here](https://github.com/cardstack/cardwallet/compare/cs-4085-4297-username-validation...dani/slug-validation-suggestion?expand=1)

oidoug: Ok, thanks for the effort on improving this, appreciated. I'll update my PR based on your suggestions, all good stuff.

I just don't agree with the min length "error" message, the user didn't even started typing and is seeing an error, seems too extreme 😅
Maybe it would look better with just the message without the X icon. wdyt @danibonilha ?

oidoug: On second thought, the error don't show immediately, so it's fine as is.

danibonilha (approved):
🚀

PR #957, opened 2022-07-26
[CS-4299] Adds blocked state to Button component

oidoug:
### Description

This PR adds the new `blocked` button state. It's a state instead of a variant because we want to be able to apply the state to any variant. It sets an opacity to the button and blocks interactions.

- [x] Completes #(CS-4299)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

### DIFF
--- a/cardstack/src/screens/Dev/DesignSystemScreen.tsx
+++ b/cardstack/src/screens/Dev/DesignSystemScreen.tsx
@@ -23,7 +23,7 @@ const DesignSystemScreen = () => {
       data: ['slug'],
     },
     {
-      title: 'Buttom States',
+      title: 'Button States',
       data: ['longPress', 'loading', 'blocked'],
     },
     {
@@ -75,7 +75,7 @@ const DesignSystemScreen = () => {
             </CenteredContainer>
           );
 
-        case 'Buttom States':
+        case 'Button States':
           return (
             <CenteredContainer padding={2}>
               {renderButtonStates(item)}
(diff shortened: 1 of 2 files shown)


cardstack/src/screens/Dev/DesignSystemScreen.tsx
```diff
@@ -23,8 +23,8 @@ const DesignSystemScreen = () => {
       data: ['slug'],
     },
     {
-      title: 'Hold to confirm Button',
-      data: ['1'],
+      title: 'Buttom States',
```
paulinhapenedo: Typo `button`

cardstack/src/screens/Dev/DesignSystemScreen.tsx
```diff
@@ -56,12 +74,11 @@ const DesignSystemScreen = () => {
               <Button variant={item}>{item}</Button>
             </CenteredContainer>
           );
-        case 'Hold to confirm Button':
+
+        case 'Buttom States':
```
paulinhapenedo: Same typo here: `button`

paulinhapenedo (approved):
I don't know if I agree with adding another state just because the button looks different. Disabled and blocked seem like the same to me.

I understand that creating a theme just for this use case could be a problem bigger than simply adding a state to the button, though... that's why the approval.

oidoug:
Yes, I agree blocked and disabled states reflect similar usa cases, but @cmgardella made a point that this blocked state will be helpful in new UI iterations - when we want to give a hint that the button will re-enable if the form is filled correctly.

danibonilha (approved):
LGTM

PR #958, opened 2022-07-26
[CS-4260] Create navigator for In-app purchase screens

paulinhapenedo:
### Description
This PR creates a navigator for the In-app purchase screens. There's no particular order defined yet, so I went with the order we have implemented the screens.

While in the area, I noticed that I forgot to delete the `skip` copy from the screens that are using the `InPageHeader` component, so I'm removing them too.

- [x] Completes #CS-4260

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/navigation/index.ts
+++ b/cardstack/src/navigation/index.ts
@@ -2,6 +2,6 @@ export * from './hooks';
 export * from './routes';
 export * from './screens';
 export * from './presetOptions';
-export * from './profileNavigator';
+export * from './profileScreenGroup';
 export { default as AppContainer } from './navigator';
 export { default as Navigation } from './Navigation';
--- a/cardstack/src/navigation/index.ts
+++ b/cardstack/src/navigation/index.ts
@@ -2,5 +2,6 @@ export * from './hooks';
 export * from './routes';
 export * from './screens';
 export * from './presetOptions';
+export * from './profileNavigator';
 export { default as AppContainer } from './navigator';
 export { default as Navigation } from './Navigation';
--- a/cardstack/src/navigation/profileNavigator.tsx
+++ b/cardstack/src/navigation/profileNavigator.tsx
@@ -1,4 +1,3 @@
-import { createStackNavigator } from '@react-navigation/stack';
 import React from 'react';
 
 import {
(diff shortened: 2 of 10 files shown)


oidoug (approved):
LGTM, but I'm missing some MVP logic to control when this flow will start.

Or we can create a new card to handle that specifically.

cardstack/src/navigation/profileNavigator.tsx
```diff
@@ -0,0 +1,42 @@
...
+} from '@cardstack/screens';
+
+import { horizontalInterpolator, Routes } from '.';
+
+export const ProfileNavigation = ({
+  Stack,
+}: {
+  Stack: ReturnType<typeof createStackNavigator>;
+}) => (
+  <Stack.Group>
```
danibonilha: I think this might have a screenOptions, where we could define `horizontalInterpolator` just once

paulinhapenedo:
> LGTM, but I'm missing some MVP logic to control when this flow will start.
> 
> Or we can create a new card to handle that specifically.

Do we already know when this flow will start? To me, it's still TBD as the screen order is still TBD too. If we already have that defined, it would be nice to have that written either on a ticket or the RFC (not sure how much we can change things in the RFC after it was approved). From what I understood from the ticket, there was no logic expected from it.

cardstack/src/navigation/profileNavigator.tsx
```diff
@@ -0,0 +1,42 @@
...
+  ProfileSlugScreen,
+  ProfilePurchaseCTA,
+} from '@cardstack/screens';
+
+import { horizontalInterpolator, Routes } from '.';
+
+export const ProfileNavigation = ({
+  Stack,
+}: {
+  Stack: ReturnType<typeof createStackNavigator>;
```
danibonilha: nit: it would be nice to create this as a "global" type, we use it in 3 diff places

cardstack/src/navigation/index.ts
```diff
@@ -2,5 +2,6 @@ export * from './hooks';
 export * from './routes';
 export * from './screens';
 export * from './presetOptions';
+export * from './profileNavigator';
```
danibonilha: superNit: it's not a navigator so I'd rename it as profileScreenGroup or something, but feel free to ignore it

danibonilha (approved):

oidoug:
@paulinhapenedo just relaying our private discussion.

Yes, you're right, the card didn't specified this extra logic. This should be handled on CS-4227.

PR #965, opened 2022-07-29
[CS-4341]: Add nonStacking nav anim on iap flow, and replace popping with opacity btn

danibonilha:
### Description
This PR adds a new nav animation without overlapping, it also replaces the scaling btn animation with an opacity one, this change though affects the whole app so let's be aware of possible bugs.

### Screenshots

Simulator Screen Recording - iPhone 11 Pro - 2022-07-29 at 08 49 48


oidoug (approved):
LGTM!

PR #962, opened 2022-07-28
[CS-4300] Adds unit tests to profile slug validation screen

oidoug:
### Description

Adds unit tests to `useProfileSlugScreen.ts` with good coverage of edge cases, including api erroring out.

- [x] Completes #(CS-4300)

No visual changes.

PR #966, opened 2022-07-29
Fix import and tests

paulinhapenedo:
### Description
This PR solves a few issues we had with imports not being correct and tests also breaking because we forgot to rebase and merge some of the open PRs.


oidoug (approved):
LGTM, thanks for fixing this up.
I wonder if there weren't any other tests deleted by accident from PR #962 .
I'll check later.

PR #967, opened 2022-08-01
[CS-4227] Show profile purchase when wallet created

oidoug:
### Description

This PR adds a hook that checks if a new wallet doesn't has a safe and if user is a beta tester to present profile purchase flow.

- [x] Completes #(CS-4227)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Please check linear ticket CS-4227.

### DIFF
--- a/cardstack/src/hooks/onboarding/useShowOnboarding.ts
+++ b/cardstack/src/hooks/onboarding/useShowOnboarding.ts
@@ -1,17 +1,12 @@
 import { useNavigation } from '@react-navigation/native';
-import { useCallback, useEffect, useRef, useMemo } from 'react';
+import { useCallback, useMemo } from 'react';
 
 import { Routes } from '@cardstack/navigation';
 import { usePrimarySafe } from '@cardstack/redux/hooks/usePrimarySafe';
 
-import { useWalletManager, useWallets } from '@rainbow-me/hooks';
-
 export const useShowOnboarding = () => {
   const { navigate } = useNavigation();
   const { primarySafe, isFetching } = usePrimarySafe();
-  // const { initializeWallet } = useWalletManager();
-  // const { walletReady } = useWallets();
-  // const initialized = useRef(false);
 
   const isOnboardingNeeded = useMemo(() => !isFetching && !primarySafe, [
     isFetching,
@@ -23,13 +18,6 @@ export const useShowOnboarding = () => {
       navigate(Routes.PROFILE_SLUG);
     }
   }, [isOnboardingNeeded, navigate]);
-
-  // useEffect(() => {
-  //   if (!initialized.current && !walletReady) {
-  //     initializeWallet();
-  //     initialized.current = true;
-  //   }
-  // }, [initializeWallet, walletReady]);
 
   return {
     isOnboardingNeeded,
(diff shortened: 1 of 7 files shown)


cardstack/src/hooks/onboarding/useShowOnboarding.ts
```diff
@@ -0,0 +1,20 @@
...
+export const useShowOnboarding = () => {
+  const { primarySafe, isFetching } = usePrimarySafe();
+
+  const shouldPresentOnboarding = useCallback(() => {
+    return (
+      !isFetching &&
+      !primarySafe &&
+      remoteFlags().featureProfilePurchaseOnboarding
+    );
+  }, [isFetching, primarySafe]);
```
danibonilha: Why is this a function instead of a memoized value ? I recall we've had some problems with memo and remote config, but it was related with memoizing the `getRemoteConfigAsBoolean` function, I believe this could be something like the `showBanner`on `useWelcomeCtaBanner` .

paulinhapenedo: What do you think about using `useMemo` instead of `useCallback`? Makes more sense to me, at least, to have this as a simple variable flag instead of a function.

danibonilha (commented):
Let's not forget to remove the Test-Drive from Settings panel, and make the condition on firebase to be 1.1.6 version and iOS only true

danibonilha (approved):

danibonilha:
Just one thing I noticed on the first appearance, maybe the ProfileSlug, should have a diff card interpolator, giving that it's showing a white screen on the background

cardstack/src/hooks/onboarding/useShowOnboarding.ts
oidoug: Problem with memo is that it doesn't reload if the remote config value changes, it's an imperative value.

We could improve on this by having a remove config hook that stores values in a state.

oidoug: It would be better, but remote config value changes don't trigger the memo update.

danibonilha: how about this? 

```ts
export const useShowOnboarding = () => {
  const { primarySafe, isFetching, isLoading } = usePrimarySafe();

  const { featureProfilePurchaseOnboarding } = remoteFlags();

  const shouldPresentOnboarding = useMemo(
    () =>
      !isLoading &&
      !isFetching &&
      !primarySafe &&
      featureProfilePurchaseOnboarding,
    [featureProfilePurchaseOnboarding, isFetching, isLoading, primarySafe]
  );

  return {
    shouldPresentOnboarding,
  };
};
```

we need to check the Loading flag for the first fetch

oidoug: I'll test it, but I think extracting the flag from remoteFlags will not make it reactive.
Thanks for pointing out about isLoading.

oidoug: Just confirmed, even if value changes memo is not updated.

PR #968, opened 2022-08-01
[CS-4168] Change app name to Cardstack Wallet

paulinhapenedo:
### Description
This PR renames the app from "Card Wallet" to be "Cardstack Wallet":

1. Replaces all the copy.
2. Updates the Splash screen images
3. Updates `info.plist` and `strings.xml` to update the app's display name.

- [x] Completes #CS-4168

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Splash screens

Screenshot

Screenshot


cardstack/src/components/Icon/custom-icons/phone-pc.tsx
```diff
@@ -69,11 +69,7 @@ function SvgComponent(props: SvgProps) {
             />
           </G>
         </G>
-        <G
-          id="prefix__Card_Wallet_app_icon_FINAL"
-          data-name="Card Wallet app icon FINAL"
-          transform="translate(128.674 480.029)"
-        >
```
paulinhapenedo, line 76: Both `id` and `data-name` were not being used as they were leftover from exporting the icon from the design tool, so I removed them instead of renaming it.

cardstack/src/components/Icon/custom-icons/cardstack.tsx
```diff
@@ -2,24 +2,24 @@
...
 
+import { appName } from '@cardstack/constants';
+
 function SvgComponent(props: SvgProps) {
   return (
     <Svg
       //@ts-ignore
       xmlns="http://www.w3.org/2000/svg"
-      data-name="Card Wallet app icon FINAL"
+      data-name={`${appName} Logo`}
```
danibonilha: is this property being used ?

cardstack/src/components/MinimumVersion/MinimumVersion.tsx
```diff
@@ -8,13 +8,14 @@ import {
...
 } from '@cardstack/components';
+import { appName } from '@cardstack/constants';
 import { Device, screenHeight, screenWidth } from '@cardstack/utils';
 
 import CardWalletLogo from '../../assets/cardstackLogo.png';
 import PeopleIllustrationBackground from '../../assets/people-ill-bg.png';
 
 const strings = {
-  title: 'There is a new version of Card Wallet.',
+  title: `There is a new version of ${appName}.`,
```
danibonilha, line 18: Let's not forget to update the remoteConfig message on firebase's console

cardstack/src/components/TransactionConfirmationSheet/displays/HubAuthenticationDisplay.tsx
```diff
@@ -76,7 +77,7 @@ const AuthenticateSection = () => {
           <Text fontSize={15} textAlign="center">
             I am signing this message to prove to Cardstack Hub that I am the
             owner of this address, so I can store and update information using
-            the Card Wallet
+            the {appName}
```
danibonilha, line 80: Let's verify if the hub has any mentions to Card Wallet that needs to be updated

ios/Rainbow/Info.plist
```diff
@@ -59,33 +59,33 @@
 		</dict>
 	</dict>
 	<key>NSAppleMusicUsageDescription</key>
-	<string>Card Wallet</string>
+	<string>Cardstack Wallet</string>
```
danibonilha: My guess is that we should be using `$(PRODUCT_NAME)`, but by looking at the code, this value was never updated from Rainbow to Cardstack Wallet, which we should definitely update, it doesn't need to be on this PR though, but if you do change it, it's best to trigger a release from your branch, to make sure the build won't fail

danibonilha (approved):

oidoug (approved):
LGTM!

cardstack/src/components/Icon/custom-icons/cardstack.tsx
paulinhapenedo: It will be shown as the SVG name if someone is using an assistive technology. Since it's the company logo, I decided to leave it. It's ok if we remove it, though.

cardstack/src/components/MinimumVersion/MinimumVersion.tsx
paulinhapenedo: Gonna need some assistance with that since I don't think I have access to Firebase's console.

ios/Rainbow/Info.plist
paulinhapenedo: I think we should do this on this PR too. Will do that and trigger a release.

PR #969, opened 2022-08-01
Moves profile purchase test-drive option under DEV

oidoug:
### Description

This PR removes Profile Purchase Test-Drive option from settings menu.

Onboarding will be accessible by iOS users on 1.1.6 version, defined in Remote Config.

*update* Reverted removal, now still present but only in DEV.


danibonilha (approved):
LGTM, although it might still be useful to have it under the Dev tag, but I guess we can just set the flag to true

PR #972, opened 2022-08-02
[FIX] New PRODUCT_NAME breaking pod install

oidoug:
### Description

The new PRODUCT_NAME value Cardstack Wallet without quotes was causing `pod install` to fail.

This PR adds quotes around new PRODUCT_NAME on pbxproj file.

Not sure why it didn't error on CI runs, maybe it's the architecture or variation in env.


oidoug:
There are other changes needed.

PR #975, opened 2022-08-02
[CS-4168] Change app name and show skip on first profile screen

paulinhapenedo:
### Description
This PR solves two things at the same time:

1. Chris asked us to change the name to be "Cardstack" when installing the app;
2. Show the "Skip" button instead of the back button on the first screen of the IAP because it makes more sense for the flow.

- [x] Completes #CS-4168

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


oidoug (approved):
LGTM

danibonilha (approved):

PR #976, opened 2022-08-02
Enhance showOnboarding hook to not show when importing an account

paulinhapenedo:
### Description
This PR adds a new flag to the `shouldPresentOnboarding` so that the IAP flow doesn't show when the user is importing a wallet.  

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/hooks/onboarding/useShowOnboarding.ts
+++ b/cardstack/src/hooks/onboarding/useShowOnboarding.ts
@@ -4,19 +4,22 @@ import { usePrimarySafe } from '@cardstack/redux/hooks/usePrimarySafe';
 import { remoteFlags } from '@cardstack/services/remote-config';
 
 export const useShowOnboarding = () => {
-  const { primarySafe, isFetching, isLoading } = usePrimarySafe();
+  const {
+    primarySafe,
+    isFetching,
+    isLoading,
+    isUninitialized,
+  } = usePrimarySafe();
 
   // Using a callback instead of memo to pull the feature flag value on call.
   // That is because the flag is stateles and does not causes react's side-effects,
   // so it will not update the memo value.
   const shouldPresentOnboarding = useCallback(() => {
     return (
-      !isLoading &&
-      !isFetching &&
-      !primarySafe &&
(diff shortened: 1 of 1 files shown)


oidoug:
Could you plz elaborate on how isUninitialized works and why is it false when importing?

paulinhapenedo:
> Could you plz elaborate on how isUninitialized works and why is it false when importing?

Sure! According to the [docs](https://redux-toolkit.js.org/rtk-query/usage/queries#frequently-used-query-hook-return-values), `isUninitialized` default value is `true`, which means that the query hasn't been called yet.

danibonilha (approved):
LGTM, the initial state of a query is:
```
isLoading: false
isUninitialized: true
isFetching: false
```
Once the first fetch starts : 

```
isLoading: true
isUninitialized: false
isFetching: false
```
And if there's a refresh: 
```
isLoading: false
isUninitialized: false
isFetching: true
```

So if we don't check for uninitialized on start, there's a gap where isLoading is still false, but it was expected to be true already. Which will show the iap flow for all imported accounts, even if it already has a profile, because isLoading, primarySafe, and isFetching will all be false, before the first run, resulting in a true value for a split second, until isLoading turns to true, by checking isUninitialized too this case is handled.

oidoug:
> LGTM, the initial state of a query is:
> 
> ```
> isLoading: false
> isUninitialized: true
> isFetching: false
> ```
> 
> Once the first fetch starts :
> 
> ```
> isLoading: true
> isUninitialized: false
> isFetching: false
> ```
> 
> And if there's a refresh:
> 
> ```
> isLoading: false
> isUninitialized: false
> isFetching: true
> ```
> 
> So if we don't check for uninitialized on start, there's a gap where isLoading is still false, but it was expected to be true already. Which will show the iap flow for all imported accounts, even if it already has a profile, because isLoading, primarySafe, and isFetching will all be false, before the first run, resulting in a true value for a split second, until isLoading turns to true, by checking isUninitialized too this case is handled.

Nice, thanks for the explanation.

PR #977, opened 2022-08-03
[CS-4083] Await for profile creation job (pooling)

oidoug:
### Description

This PR adds implementation for awaiting the profile creation process to finish, using a Lazy Query that starts pooling on first `jobQuery` call, and ends it once we get a successful response or an error.

UX will be improved once we have the loading / retry screen done in upcoming project. For now it requires users to wait until its done and to pull down to refresh in Wallet Screen to see the profile.

- [x] Completes #(CS-4083)

### Checklist

- [x] Tested on iOS

### Update 2

- Using job-ticked-id again to strict pooling to a specific job;
- Showing Profile Screen after profile creation flow;
- Also shows non-blocking loading in Profile Screen while profile being created.

### Screenshots (updated)

https://user-images.githubusercontent.com/129619/183513213-e5110b69-c3f8-493e-ae02-c6998ca52467.mp4


oidoug:
Updating it based on discussion.

oidoug:
PR updated, please take a look again @danibonilha and @paulinhapenedo

cardstack/src/services/hub/hub-api.ts
```diff
@@ -136,6 +138,22 @@ export const hubApi = createApi({
...
+      async onQueryStarted(arg, { dispatch, queryFulfilled }) {
+        const { data: profileId } = await queryFulfilled;
+
+        if (profileId) {
+          // TODO: Investigate cache invalidation configurations.
+          // This only invalidates cached results but doesn't automatically
+          // makes a new query, as it should per documented here:
+          // https://redux-toolkit.js.org/rtk-query/usage/automated-refetching
+          dispatch(hubApi.util.invalidateTags([SafeCacheTags.SAFES]));
+        }
```
danibonilha: Does invalidating tags from different api works?  hubApi has its own tags, and safeApi have another, I'm not sure if they communicate. if it doesn't we should probably just trigger a simple refresh

cardstack/src/services/hub/hub-api.ts
```diff
@@ -136,6 +138,22 @@ export const hubApi = createApi({
       }),
       transformResponse: ({ data }) => data?.attributes?.did,
     }),
+    createProfileJob: builder.query<string, { eoa: string }>({
```
danibonilha: create, gives a impression that is some kind of mutation, you could call it, getProfileJobStatus

cardstack/src/services/hub/hub-api.ts
```diff
@@ -136,6 +138,22 @@ export const hubApi = createApi({
...
+      query: ({ eoa }) => `${routes.profileInfo.jobTicket}?eoa=${eoa}`,
+      transformResponse: ({ data }) =>
+        data.find((job: any) => job.attributes?.state === 'success'),
```
danibonilha: This assumes the only job ticket existing is related to the profile, which might not be true, we should filter by the jobType if this query is related just with profile.

danibonilha: it would be nice to add the typing of job, this way we don't need to go back and forth on the api to take a look

danibonilha: Also this find, will return when it finds the first successful job-ticket, and if there's a pending on in the end of the array, it won't be accounted for

cardstack/src/components/AssetList/useAssetList.ts
```diff
@@ -106,12 +109,34 @@ export const useAssetList = ({
...
 
+  // Profile creation job takes time to complete, this hook keeps pooling until the profile is ready.
+  const {
+    isSafesRefreshNeeded,
+    handleAwaitForProfileCreation,
+  } = useProfileJobAwait();
+
+  // shouldAwaitForProfile will be set if screen was open from profile creation flow.
+  useEffect(() => {
+    if (params?.shouldAwaitForProfile && accountAddress) {
```
danibonilha: I'm not sure we need this nav param, it might be less confusing to just trigger the job ticket at start, if we change the screen in the end of the profile flow, for example, to go to Profile, this jobPolling will never be triggered. I believe the `useProfileJobAwait` should be smart enough to handle everything, in the case we want to change it's place, the only change would be to call it in the expected screen

oidoug: Yes, I considered and tested doing that, but it requires a separate global state to keep track when to start pooling. I did that to avoid having a slice with just that flag, but might be necessary anyway.

cardstack/src/services/hub/hub-api.ts
oidoug: That's probably why the invalidation it's not actually refreshing the safes. I can't find clear info in rtk-q docs about invalidating tags "across apis", so you're probably right. I'll remove this invalidation then.

cardstack/src/services/hub/hub-api.ts
oidoug: True, I forgot about the possibility of having more job types.

oidoug: This query is only considering successful jobs now. If we intend on using the pending jobs too, using a filter by job-type would be better. Until we make use of pending jobs, do you think getting the first successful is enough?

oidoug:
Known issue: on Profile tab, while loading the new safe, you see the "create a profile" state for a second.

I'm investigating but prefer to handle this in a new PR.

cardstack/src/hooks/usePurchaseProfile.ts
```diff
@@ -59,12 +59,14 @@ export const usePurchaseProfile = (profile: CreateProfileInfoParams) => {
 
               Alert({
                 title: 'Success',
-                message:
-                  'Your profile will appear shortly, pull to refresh in a minute or so',
+                message: 'Your profile will appear shortly',
```
paulinhapenedo: nit: move these into a strings file.

cardstack/src/hooks/usePurchaseProfile.ts
```diff
@@ -59,12 +59,14 @@ export const usePurchaseProfile = (profile: CreateProfileInfoParams) => {
...
                 buttons: [
                   {
                     text: 'Okay',
-                    onPress: () => navigate(Routes.WALLET_SCREEN),
+                    onPress: () =>
+                      navigate(Routes.PROFILE_SCREEN, {
+                        profileCreationJobID: profileJobInfo?.included[0]?.id,
```
danibonilha: nit: we could transform the query response, and return the job id

cardstack/src/screens/ProfileScreen/useProfileScreen.ts
```diff
@@ -0,0 +1,83 @@
...
+  const {
+    primarySafe,
+    isFetching,
+    refetch,
+    safesCount,
+    isLoading,
+    isUninitialized,
+  } = usePrimarySafe();
+
+  // Profile creation job takes time to complete, this hook keeps pooling until the profile is ready.
```
danibonilha: This comment is duped

cardstack/src/services/hub/hub-api.ts
```diff
@@ -46,7 +49,7 @@ enum CacheTag {
 export const hubApi = createApi({
   reducerPath: 'hubApi',
   baseQuery: fetchHubBaseQuery,
-  tagTypes: [...Object.values(CacheTag)],
+  tagTypes: [...Object.values(CacheTag), ...Object.values(SafeCacheTags)],
```
danibonilha: We should drop this change

cardstack/src/services/hub/hub-types.ts
```diff
@@ -77,3 +77,17 @@ export interface PostProfilePurchaseQueryResult {
...
+  type: string;
+  attributes: {
+    'job-type': string;
+    state: string;
+    result: {
+      id: string;
+    };
+  };
+}
+
```
danibonilha, line 92: nice

cardstack/src/hooks/__tests__/usePurchaseProfile.test.ts
```diff
@@ -45,6 +45,14 @@ jest.mock('@cardstack/services', () => ({
...
 }));
 
+jest.mock('@cardstack/hooks/useProfileJobPolling', () => ({
+  useProfileJobPolling: jest.fn(() => ({
+    handleAwaitForProfileCreation: jest.fn(),
+    profileSafeId: undefined,
+    error: undefined,
+  })),
+}));
+
```
danibonilha: This seems a leftover

cardstack/src/screens/ProfileScreen/useProfileScreen.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const redirectToSwitchNetwork = useCallback(() => {
+    navigate(Routes.SETTINGS_MODAL, {
+      initialRoute: SettingsPages.network.key,
+    });
+  }, [navigate]);
+
+  // When the new profile is done, we need to refresh the safes list.
+  useEffect(() => {
+    if (data?.attributes?.state === 'success') {
```
danibonilha: nit: instead of returning the data, we could return something like, isJobCompleted/Successful, and handle whatever we want, OR the `useProfileJobPolling` could receive a `onJobCompleted` callback then we wouldn't need this useEffect here

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -9,31 +8,24 @@ import {
...
-    safesCount,
-    isLoading,
-    isUninitialized,
-  } = usePrimarySafe();
-
-  const { network } = useAccountSettings();
-
-  const isRefreshingForNewAccount = useIsFetchingDataNewAccount(isFetching);
+    redirectToSwitchNetwork,
+  } = useProfileScreen();
```
danibonilha, line 26: nice job on creating the screen hook

cardstack/src/hooks/usePurchaseProfile.ts
```diff
@@ -59,12 +59,14 @@ export const usePurchaseProfile = (profile: CreateProfileInfoParams) => {
 
               Alert({
```
danibonilha: Wondering if we still need this alert, since this flow will be redirected to a loading screen

cardstack/src/hooks/usePurchaseProfile.ts
oidoug: This alert will be removed.

cardstack/src/hooks/usePurchaseProfile.ts
oidoug: I agree, there are too many alerts in this process currently, even the one for a successful purchase might not be needed.

For now, I'll remove this one only.

cardstack/src/hooks/usePurchaseProfile.ts
oidoug: yeah, makes sense.

cardstack/src/hooks/__tests__/usePurchaseProfile.test.ts
oidoug: Yes, thanks for pointing that out.

cardstack/src/screens/ProfileScreen/useProfileScreen.ts
oidoug: Yes, thats a good idea.

I was considering we would use the response data when pending somehow in the future, but we can refactor if needed later.

danibonilha (approved):
LGTM! 🚀

PR #978, opened 2022-08-03
[CS-4332]: Add option to buy profile with prepaidcard on new flow

danibonilha:
### Description

This PR updates the Profile tab layout, and adds the ability to buy a profile using prepaid cards on new flow, also it prevents the user to make the purchase based on the remote flag, and fixes Android products not appearing with correct localized value.

A nice way to test it, is to use card drop on sokol, and then using that card to buy the profile

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Whole flow is on Linear[ [CS-4332]](https://linear.app/cardstack/issue/CS-4332/[ui]-add-option-to-re-launch-profile-creation-flow-if-skipped-from)

Small android device: 

Screen Shot 2022-08-03 at 12 32 23


paulinhapenedo (approved):
LGTM 👏

oidoug (approved):
LGTM

PR #979, opened 2022-08-04
[CS-4370]: Fix historical price not showing

danibonilha:
### Description

This Pr fixes the historical price not appearing, it seems there's been a change on the timestamps, and now we need to multiply by 1000,  I'll be talking with the BE team to make sure this change is needed before merging it.

### Screenshots

Before:
image

After: 

image


paulinhapenedo (approved):
LGTM

PR #980, opened 2022-08-05
[CS-3987] Enhance isInvalid to not allow user to send value less than minimal

paulinhapenedo:
### Description
This PR solves an issue with the Pay screen when using a currency that's not USD. Because of conversion rates, the minimum amount wasn't being taken into consideration when typing the amount.

There's another issue happening here that's being discussed in the ticket, so please refer to that for further explanation.

- [x] Completes #CS-3897

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

### DIFF
--- a/cardstack/src/components/Input/InputAmount/useInputAmountHelper.tsx
+++ b/cardstack/src/components/Input/InputAmount/useInputAmountHelper.tsx
@@ -3,6 +3,7 @@ import {
   convertToSpend,
   convertStringToNumber,
   NativeCurrency,
+  spendToUsd,
 } from '@cardstack/cardpay-sdk';
 import { useMemo, useState } from 'react';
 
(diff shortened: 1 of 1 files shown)


cardstack/src/components/Input/InputAmount/useInputAmountHelper.tsx
```diff
@@ -40,10 +41,14 @@ export const useInputAmountHelper = () => {
...
 
   // input amount should be more than MIN_SPEND_AMOUNT (50)
-  const isInvalid = useMemo(
-    () => amountInNum > 0 && spendAmount < MIN_SPEND_AMOUNT,
-    [amountInNum, spendAmount]
-  );
+  const isInvalid = useMemo(() => {
+    const minSpendAmountToUSD = spendToUsd(MIN_SPEND_AMOUNT) || 0;
+
+    const convertedAmount =
```
danibonilha: supernit: `convertedAmount` -> `minNativeCurrencyAmount`

cardstack/src/components/Input/InputAmount/useInputAmountHelper.tsx
```diff
@@ -40,10 +41,14 @@ export const useInputAmountHelper = () => {
...
+      (rates?.[paymentCurrency] ?? 0) * minSpendAmountToUSD;
```
danibonilha: nit: Even though rates returns USD, we could do something similar as we do on `spendAmount`, this way, if the rates api is out, we are still able to make a payment using USD.

danibonilha (approved):
LGTM, just nit comments.

PR #985, opened 2022-08-09
[Fix]: Avoid showing onboarding on import flow

danibonilha:
### Description

This PR fixes the issue while importing an account which already contains a profile asking to create a new one, the parentheses on the negation matters, and we actually want to check each flag individually. It also adds tests to make sure the expected behavior is correct.


paulinhapenedo (approved):
Test addition 👏 
LGTM!

PR #986, opened 2022-08-09
[CS-4366] Customize ProfileName screen to edit profile

paulinhapenedo:
### Description
This PR changes the ProfileName component to also handle the edition of the profile.
For now, we don't have the mutation because the endpoint is not ready yet. 

- [x] Completes #CS-4366

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/690904/183724266-c156bbd0-a9bf-426b-9ed4-968adf019613.mp4

### DIFF
--- a/cardstack/src/hooks/profile/useProfileEdit.ts
+++ b/cardstack/src/hooks/profile/useProfileEdit.ts
@@ -1,0 +1,54 @@
+import { useNavigation } from '@react-navigation/native';
+import { useCallback, useMemo } from 'react';
+
+import { defaultErrorAlert } from '@cardstack/constants';
+import { useLoadingOverlay } from '@cardstack/navigation';
+
+import { Alert } from '@rainbow-me/components/alerts';
+import logger from 'logger';
+
+import { useMutationEffects } from '../useMutationEffects';
+
+const strings = {
+  loading: 'Updating your profile',
+};
+
+export const useProfileEdit = () => {
+  const { goBack } = useNavigation();
+  const { showLoadingOverlay, dismissLoadingOverlay } = useLoadingOverlay();
+
+  // TODO: update this with the correct mutation after hub is done
+  const updateProfile = useCallback(() => {
+    showLoadingOverlay({ title: strings.loading });
+
+    // TODO: call mutation
+  }, [showLoadingOverlay]);
+
+  useMutationEffects(
+    useMemo(
+      () => ({
+        success: {
+          status: false,
+          callback: () => {
+            dismissLoadingOverlay();
+            goBack();
+          },
+        },
+        error: {
+          status: false,
+          callback: () => {
+            dismissLoadingOverlay();
+            logger.sentry('Error updating profile');
+
+            Alert(defaultErrorAlert);
+          },
+        },
+      }),
+      [dismissLoadingOverlay, goBack]
+    )
+  );
+
+  return {
+    updateProfile,
+  };
+};
--- a/cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
+++ b/cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
@@ -191,7 +191,7 @@ export const ProfileNameScreen = () => {
     >
       <InPageHeader skipAmount={2} />
       <Animated.View style={animatedHeaderStyles}>
-        <Text variant="pageHeader">{strings.header}</Text>
+        <Text variant="pageHeader">{strings.header.create}</Text>
         <CenteredContainer>
           <Text fontSize={12} color="grayText" paddingBottom={2}>
             {strings.editColor}
--- a/cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
+++ b/cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
@@ -67,6 +67,7 @@ export const ProfileNameScreen = () => {
     onChangeText,
     onPressEditColor,
     isEditing,
+    isBlocked,
   } = useProfileNameScreen();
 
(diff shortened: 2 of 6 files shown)


cardstack/src/screens/Profile/ProfileNameScreen/useProfileNameScreen.ts
```diff
@@ -67,5 +85,7 @@ export const useProfileNameScreen = () => {
     onChangeText,
     onPressEditColor,
     profile,
+    isEditing: currentProfile,
```
danibonilha: nit: we are using update, and edit, should we constraint to one of those ?

cardstack/src/screens/Profile/ProfileNameScreen/strings.ts
```diff
@@ -1,8 +1,12 @@
 export const strings = {
-  header: 'Congrats! Choose a name\nfor your profile page',
+  header: {
+    create: 'Congrats! Choose a name\nfor your profile page',
```
oidoug: nit: I think we can take the opportunity to change the create header text by removing "Congrats!". I don't think it fits well in our current flow.

cardstack/src/screens/Profile/ProfileNameScreen/useProfileNameScreen.ts
```diff
@@ -53,9 +55,25 @@ export const useProfileNameScreen = () => {
     });
   }, [navigate, profileColor]);
 
+  const isBlocked = useMemo(
+    () =>
+      !profile.name ||
+      (currentProfile.name === profile.name &&
+        currentProfile.color === profile.color),
```
danibonilha, line 62: nit: we could use isMatch from lodash here to shrink this condition:

```
isMatch(profile, currentProfile)
```

oidoug (approved):
LGTM, well done.

I forgot the update api wasn't done yet. If you didn't already, please add a new card to add the update logic once the api part is cleared, thanks :)

paulinhapenedo:
> LGTM, well done.
> 
> I forgot the update api wasn't done yet. If you didn't already, please add a new card to add the update logic once the api part is cleared, thanks :)

I think you've already created one here https://linear.app/cardstack/issue/CS-4084/[logic]-implement-call-to-endpoint-card-space-to-update-profile-data

cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
```diff
@@ -182,16 +184,26 @@ export const ProfileNameScreen = () => {
...
+  const pageTitle = useMemo(
+    () => (isEditing ? strings.header.update : strings.header.create),
+    [isEditing]
+  );
+
+  const btnLabel = useMemo(
+    () => (isEditing ? strings.btns.save : strings.btns.continue),
+    [isEditing]
+  );
+
```
danibonilha, line 188: superNit: it's just a implementation suggestion, feel free to ignore, by using a key on the strings, we could check for isEditing just once, something like:

```ts
  const flow = useMemo(
    () => (isEditing ?  'update' :  'create'),
    [isEditing]
  );
-----------------------

<Text variant="pageHeader">{strings.header[flow]}</Text>

 <Button blocked={isBlocked} onPress={onContinuePress}>
 {strings.btns[flow]}
   </Button>
-------------------------

export const strings = {
...
  btns: {
    create: 'Continue',
    update: 'Save',
  },

```
as fas as the strings, you could also have in this order, whatever feels more natural, for example: 
```ts
  create: {
    header: 'Congrats! Choose a name\nfor your profile page',
    btn: 'Continue'
  },
  update: {
    header: 'Choose a new name and\ncolor for your profile page',
    bnt: 'Save',
  },
  ---------------------------------
   {strings[flow]btn]}
```


of course by doing this you might have some typing to handle, using enum or so.

cardstack/src/screens/Profile/ProfileNameScreen/strings.ts
```diff
@@ -1,8 +1,12 @@
...
+    update: 'Choose a new name and\ncolor for your profile page',
+  },
   btns: {
     skip: 'Skip',
```
danibonilha: we don't need this skip copy anymore

cardstack/src/hooks/profile/useProfileEdit.ts
```diff
@@ -0,0 +1,54 @@
...
+const strings = {
+  loading: 'Updating your profile',
+};
+
+export const useProfileEdit = () => {
+  const { goBack } = useNavigation();
+  const { showLoadingOverlay, dismissLoadingOverlay } = useLoadingOverlay();
+
+  // TODO: update this with the correct mutation after hub is done
+  const updateProfile = useCallback(() => {
```
danibonilha: This needs to receive the profileInfo to be updated it

danibonilha (approved):
LGTM, just nit comments

cardstack/src/screens/Profile/ProfileNameScreen/useProfileNameScreen.ts
paulinhapenedo: Uhm, I'll rename everything to update or updating since it better reflects the action the user is doing.

cardstack/src/screens/Profile/ProfileNameScreen/useProfileNameScreen.ts
paulinhapenedo: We cannot use `isMatch` because the objects are different by default. `currentProfile` has additional nodes (`did`, for example).

danibonilha: Nice catch!

PR #987, opened 2022-08-10
[CS-4391] Profile creation loading state improvement 

oidoug:
### Description

Minor change that improves the loading response to users.

There are delays between starting the creation, and changing from creating to refetching safes states, so relying only in API loading states was causing "jumps" in UI.

- [x] Completes #(CS-4391)

### Checklist

- [x] Tested on iOS

### Screenshots

https://user-images.githubusercontent.com/129619/183946395-2c83e141-99f8-4d11-812e-079238e2625d.MP4


oidoug:
Closed in favor of #988

PR #988, opened 2022-08-10
[CS-4353]: Enable iap flow on Android

danibonilha:
### Description

This PR adds the changes based on platform make iap work, it fixes the loading flicker and also makes sure the job type is profile.

Android needs the purchaseToken and iOS the txReceipt, also to be able to buy the same item again, we need to pass the consumable flag to finishTransaction

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/20520102/183997149-f657aaa7-1e46-4863-a996-f1dab9e8b424.mp4


oidoug (approved):
Good stuff!

PR #989, opened 2022-08-11
[CS-4397]: Refactor showOnboarding to handle multiple flows

danibonilha:
### Description

This PR extracts the navigation logic to the onboarding hook, by having it to handle just the flag we were not able to predict its behavior, because even though we can test what the flag should look like we were not able to make sure the safe state would be the expected one by the time we called the flag, so it made more sense to have it handled within the hook itself, with this change we are able to show the flow just for users that don't have a profile, either on import, or while creating a new account, the main change it that now it will ask all the time the app is restarted to create a profile, we can either leave as is, or persist the user skip option, which can be handled on a different ticket. 
The remoteFlag has been remove since it's not related with the onboarding itself, it's related to the iap status and this is already handled on the Buy btn (after the prepaidcard and android implementation), if we want to prevent the user to totally access this flow we should discuss about having a flag for onboarding only. Let me know if you find any issues with this implementation.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

New acc flow: 

Simulator Screen Recording - iPhone 11 - 2022-08-11 at 12 30 37

Import flow is on linear ticket


paulinhapenedo (approved):
Makes more sense to handle it this way.
Tested both import with both an account that already has a profile and one without, and creation of new account too.

👏

oidoug (approved):
This causes a bit more "lag" to open the flow, since it needs to open the wallet first, but I understand the reason and I'm fine moving forward with this approach.

Code is all good! 🚀

PR #990, opened 2022-08-11
[CS-4400] Update CardWalletLogo component to the new design guidelines

paulinhapenedo:
### Description
This PR updates the `CardWalletLogo` component to reflect the name changes to the app. We shouldn't be using `<Text>` to render the app name, but the whole logo should be used instead (imagery + text).

- [x] Completes #CS-4400

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot


oidoug (approved):
LGTM 👍

PR #991, opened 2022-08-11
[CS-4399] Avoid duplicate purchase calls

oidoug:
### Description

This PR adds a debounce hook to be used on buttons that we don't want to control a disabled state but want to avoid double taps.

It looks a bit too much for this specific use, but it doesn't need an extra state and a confusing enabling/disabling logic in the screen's logic hook. And it can be reused for other similar cases.

- [x] Completes #(CS-4399)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Before (hard to see, but by tapping 3x the button, 3 purchase alerts appered):

https://user-images.githubusercontent.com/129619/184243150-08c9dbf4-df90-4954-a683-0f1d295ef69c.MP4

Now (tapped several times, purchase only called once):

https://user-images.githubusercontent.com/129619/184243279-3e1ecf42-e0ba-47b7-913a-81d888f74f95.MP4


paulinhapenedo:
Why not just use the `blocked` state that you've created? This seems a bit too much 🤔

cardstack/src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,27 @@
...
+      if (!busy.current) {
+        busy.current = true;
+        callback();
+      }
+    },
+    [bounceRate]
+  );
+
+  return { debounce };
+};
```
danibonilha: What is the gain of having this as a hook ? We have lodash in the project which already provides a debounce method.

oidoug: Lodash's debounce does not work for this case because it calls one last time after its timer is over. Options trailing or leading still causes this. I think my debounce idea is more like a cooldown, and not debounce 😄 .

Anyways, I'll use a state because I notice the other buttons were also leading to flow errors if pressed right after the buy button.

oidoug:
You're right @paulinhapenedo, using a state and blocked prop is better 👍

cardstack/src/screens/Profile/PurchaseCTAScreen/usePurchaseCTAScreen.ts
```diff
@@ -51,19 +52,31 @@ export const usePurchaseCTAScreen = () => {
...
+      Alert({
+        title: 'Oops!',
+        message: 'This feature is currently unavailable',
+      });
     }
 
-    Alert({ title: 'Oops!', message: 'This feature is currently unavailable' });
-  }, [purchaseProfile]);
+    setPurchaseEnd();
+  }, [purchaseProfile, setPurchaseStart, setPurchaseEnd]);
```
paulinhapenedo, line 74: If we are adding an async/await expression, what do you think about wrapping it in a try/catch?

paulinhapenedo (approved):
LGTM 👏

cardstack/src/screens/Profile/PurchaseCTAScreen/usePurchaseCTAScreen.ts
oidoug: I don't see it's needed in this case. purchaseProfile return a promise but it also handles any exceptions inside it, it will not throw anything.

PR #992, opened 2022-08-11
fix: isUpdating needs to check for property inside currentProfile

paulinhapenedo:
### Description
`currentProfile`, when destructuring, is an empty object and it's a truthy value. This PR fixes that.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


oidoug (approved):
Looks good, working as expected now.

PR #993, opened 2022-08-11
[CS-4346] Add Staking label to CARD.CPXD balance on Depot

paulinhapenedo:
### Description
This PR adds a tag "Staking" to inform the users that staking is active in their CARD.CPXD balance.  

- [x] Completes #CS-4346

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Android
Screenshot

Screenshot

### iOS
Screenshot

Screenshot

### DIFF
--- a/cardstack/src/components/TokenBalance/TokenBalance.tsx
+++ b/cardstack/src/components/TokenBalance/TokenBalance.tsx
@@ -7,8 +7,13 @@ import {
   Text,
   Touchable,
   CoinIcon,
+  FloatingTag,
 } from '@cardstack/components';
 import { Theme } from '@cardstack/theme';
+
+const strings = {
+  staking: 'Staking',
+};
 
 export interface TokenBalanceProps extends ContainerProps {
   onPress?: () => void;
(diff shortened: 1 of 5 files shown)


lukemelia:
Looks good!

cardstack/src/theme/textVariants.ts
```diff
@@ -69,4 +69,11 @@ export const textVariants = {
     color: 'grayText',
     fontSize: 12,
   },
+  stakingTag: {
```
oidoug: nit: We have to try to keep `textVariants` as generic as possible. So maybe just call it `tag` or `floatingTag`.

cardstack/src/components/TokenBalance/TokenBalance.tsx
```diff
@@ -79,6 +91,18 @@ export const TokenBalance = (props: TokenBalanceProps) => {
               <Text variant="subText">{nativeBalance}</Text>
             </Container>
           </Container>
+          {isCardCPXD && (
+            <CenteredContainer
```
oidoug: nit: Even though this tag is only used here atm, I think it would be nice to have it in a sep. component.

oidoug (approved):
LGTM, just a couple nits.

PR #997, opened 2022-08-16
[CS-4131] UI for profile retry

oidoug:
### Description

This PR adds UI adapted to profile tab for retry and support when profile creation fails.

- [x] Completes #(CS-4131)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

### DIFF
--- a/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
+++ b/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
@@ -16,7 +16,7 @@ import { CreateProfile, CreateProfileError, strings } from './components';
 import { useProfileScreen } from './useProfileScreen';
 
 // Todo: Extract errors from job-ticket api call.
-const isError = true;
+const isError = false;
 
 const ProfileScreen = () => {
   const {
--- a/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
+++ b/cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React from 'react';
 import { ActivityIndicator, Alert } from 'react-native';
 
 import {
@@ -16,7 +16,7 @@ import { CreateProfile, CreateProfileError, strings } from './components';
 import { useProfileScreen } from './useProfileScreen';
 
 // Todo: Extract errors from job-ticket api call.
-const error = 'Error message';
+const isError = true;
 
 const ProfileScreen = () => {
   const {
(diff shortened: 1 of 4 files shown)


cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -10,9 +10,14 @@ import {
...
 
-import { CreateProfile, strings } from './components';
+import { useSendFeedback } from '@rainbow-me/hooks';
+
+import { CreateProfile, CreateProfileError, strings } from './components';
 import { useProfileScreen } from './useProfileScreen';
 
+// Todo: Extract errors from job-ticket api call.
+const error = 'Error message';
+
```
paulinhapenedo: Are we sure that the API will give us clear error messages for us to show to the user?

cardstack/src/screens/ProfileScreen/components/CreateProfileError.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <Icon name="info" size={20} color="error" />
+        <Text
+          color="white"
+          fontWeight="bold"
+          fontSize={20}
+          textAlign="center"
+          paddingLeft={3}
+        >
+          {strings.profileError.title}
+        </Text>
```
paulinhapenedo: I believe this should be using the `InPageHeader` text variant you created a few PRs ago. What you have on the screenshot is different than the Zeplin file.

cardstack/src/screens/ProfileScreen/components/CreateProfileError.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const CreateProfileError = ({
+  onPressRetry,
+  onPressSupport,
+  errorMessage,
+}: CreateProfileErrorProps) => (
+  <Container flex={1} justifyContent="space-between" paddingBottom={10}>
+    <Container
+      flex={1}
+      justifyContent="center"
+      alignItems="center"
```
paulinhapenedo: From the Zeplin file, the text isn't horizontally centered. It should follow the alignment of all the other screens on the Profile Creation flow.

cardstack/src/screens/ProfileScreen/components/CreateProfileError.tsx
```diff
@@ -0,0 +1,65 @@
...
+          textAlign="center"
+          paddingLeft={3}
+        >
+          {strings.profileError.title}
+        </Text>
+      </Container>
+      {errorMessage && (
+        <Text color="grayText" textAlign="center">
+          {errorMessage}
+        </Text>
```
paulinhapenedo: Same here. Text shouldn't be center aligned.

cardstack/src/screens/ProfileScreen/components/CreateProfileError.tsx
```diff
@@ -0,0 +1,65 @@
...
+      )}
+    </Container>
+    <Container>
+      <Button
+        paddingHorizontal={4}
```
paulinhapenedo: Shouldn't this padding be global like on all the other screens and be 5? 4 makes the button be not aligned with the rest of the content on the screen.

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
oidoug: I'm trying to map that now, but it looks like so.

cardstack/src/screens/ProfileScreen/components/CreateProfileError.tsx
oidoug: I don't think that fits well inside the Profile tab, so I took the liberty of using the style we already use for content in side tabs.

cardstack/src/screens/ProfileScreen/components/CreateProfileError.tsx
oidoug: Same as above.

cardstack/src/screens/ProfileScreen/components/CreateProfileError.tsx
oidoug: ditto

cardstack/src/screens/ProfileScreen/components/CreateProfileError.tsx
oidoug: Isn't the button aligned? weird.. this padding actually doesn't do much bc the way our button comp works, it always sets itself to the width of the screen - some fixed paddings.

I'll remove the horizontal padding altogether.

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -44,7 +82,7 @@ const ProfileScreen = () => {
   return (
     <Container
       backgroundColor={
-        primarySafe && !showLoading ? 'white' : 'backgroundDarkPurple'
+        primarySafe && !showLoading && !error ? 'white' : 'backgroundDarkPurple'
```
danibonilha: This is starting to be confusing, I'm wondering if we could default it to `backgroundDarkPurple` and add a wrapper on the MerchantContent, to be white, bc it's the only different component

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -68,21 +106,10 @@ const ProfileScreen = () => {
...
-                paddingBottom={4}
-              >
-                {strings.ongoingProfileCreation}
-              </Text>
-            )}
-            <ActivityIndicator size="large" />
-          </>
+          ProfileLoading
         ) : (
           ProfileBody
```
danibonilha: nit:  I'm not the biggest fan of creating a component just to "contain"  stuff that we could read right way on the tree, but with all these ternaries, it feels that, it might be better to have a `ProfileContent` component, so we could conditionally return each state, without relying on nested ternaries, something like this: 
```tsx
const ProfileScreen = () => {
  const {
    primarySafe,
    showLoading,
    isCreatingProfile,
    safesCount,
    isFetching,
    network,
    refetch,
    redirectToSwitchNetwork,
  } = useProfileScreen();

  const ProfileContent = () => {
    // LayerOneRedirect
    if (isLayer1(network)) {
      return (
        <>
          <Text color="white" fontSize={24}>
            {strings.switchToGnosisChain}
          </Text>
          <Button
            borderColor="buttonSecondaryBorder"
            marginTop={10}
            onPress={redirectToSwitchNetwork}
            variant="primary"
          >
            {strings.switchNetwork}
          </Button>
        </>
      );
    }

    if (error) {
      return (
        <CreateProfileError
          onPressRetry={() => Alert.alert('Not available yet')}
          onPressSupport={onSendSupport}
          errorMessage={error}
        />
      );
    }

    if (showLoading) {
      return (
        <>
          {isCreatingProfile && (
            <Text
              variant="semibold"
              color="white"
              textAlign="center"
              fontSize={16}
              paddingBottom={4}
            >
              {strings.ongoingProfileCreation}
            </Text>
          )}
          <ActivityIndicator size="large" />
        </>
      );
    }

    if (primarySafe) {
      return (
        <Container backgroundColor="white">
          <MerchantContent
            showSafePrimarySelection={safesCount > 1}
            isPrimarySafe
            merchantSafe={primarySafe}
            isRefreshingBalances={isFetching}
            refetch={refetch}
          />
        </Container>
      );
    }

    return <CreateProfile />;
  };

  return (
      <MainHeader title={strings.header.profile} />
      <Container
        justifyContent="center"
        flex={1}
        paddingHorizontal={isLayer1(network) ? 5 : 0}
      >
        <ProfileContent />
      </Container>
    </Container>
  );
};
```

oidoug: Thats a good suggestion, thanks 👍

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
```diff
@@ -10,9 +10,14 @@ import {
...
 
-import { CreateProfile, strings } from './components';
+import { useSendFeedback } from '@rainbow-me/hooks';
+
+import { CreateProfile, CreateProfileError, strings } from './components';
 import { useProfileScreen } from './useProfileScreen';
 
+// Todo: Extract errors from job-ticket api call.
+const isError = true;
+
```
paulinhapenedo: What do you think about changing this to `false` to avoid blocking this screen in the error state after merging? I know you're already working on the logic, but since it's not complete yet, better to play safe.

paulinhapenedo (approved):
Minor comment on the `isError` flag, but LGTM.

cardstack/src/screens/ProfileScreen/ProfileScreen.tsx
oidoug: Will do, my mistake.

PR #998, opened 2022-08-16
[CS-4435] Update Rewards notification to use same rule as Rewards screen

paulinhapenedo:
### Description
This PR fixes the Rewards notification icon we have on the Tab navigation header to use the same rule as the Rewards screen (checking for the existence of `mainPoolTokenInfo` instead of something in the `rewardPoolTokenBalances` (which can contain rewards not yet available for claiming for N reasons).

- [x] Completes #CS-4435

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/components/MainHeader/MainHeader.tsx
+++ b/cardstack/src/components/MainHeader/MainHeader.tsx
@@ -34,7 +34,7 @@ const MainHeader = ({
 }: Props) => {
   const { navigate } = useNavigation();
   const { network } = useAccountSettings();
-  const { rewardPoolTokenBalances } = useRewardsDataFetch();
+  const { mainPoolTokenInfo } = useRewardsDataFetch();
 
   const onMenuPress = useCallback(() => navigate(Routes.SETTINGS_MODAL), [
     navigate,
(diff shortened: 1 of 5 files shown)


cardstack/src/components/MainHeader/MainHeader.tsx
```diff
@@ -44,7 +44,7 @@ const MainHeader = ({
     navigate(Routes.REWARDS_CENTER_SCREEN);
   }, [navigate]);
 
-  const availableRewards = rewardPoolTokenBalances?.length;
+  const availableRewards = !!mainPoolTokenInfo;
```
danibonilha: There's now 3 places doing this boolean conversion, `useRewardsDataFetch` could return ` hasRewardsAvailable: !!mainPoolTokenInfo`

paulinhapenedo: Done here [930cc90](https://github.com/cardstack/cardwallet/pull/998/commits/930cc90f0b1938e8434461ce3e936accbdb254fd)

cardstack/src/components/MainHeader/MainHeader.tsx
```diff
@@ -82,10 +80,10 @@ const MainHeader = ({
           onPress={onRewardPress}
         >
           <Icon color="teal" iconSize="medium" size={26} name="rewards" />
-          {!!availableRewards && (
+          {hasRewardsAvailable && (
             <Container>
               <Text color="teal" fontSize={18} weight="bold" marginLeft={2}>
-                {availableRewards}
+                1
```
danibonilha: Let's assign this to a variable, defaultRewardsQty or something, so it's easier to update in the future

danibonilha (approved):

PR #999, opened 2022-08-16
[CS-3963] Remove react-native-keyboard-area and keyboardHeight

paulinhapenedo:
### Description
This PR removes `react-native-keyboard-area`. We were using only on one location without any practical effects to the app.

As mentioned in the ticket comments, we've used this to also remove `keyboardHeight` from Redux and all of its use. 

- [x] Completes #CS-3963

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


cardstack/src/screens/sheets/ImportSeed/ImportSeedSheet.tsx
```diff
@@ -85,10 +82,10 @@ const ImportSeedSheet = () => {
...
           </Container>
         </Container>
+        <ToastPositionContainer>
+          <InvalidPasteToast isInvalidPaste={isInvalidPaste} />
+        </ToastPositionContainer>
       </Sheet>
-      <ToastPositionContainer bottom={keyboardHeight}>
-        <InvalidPasteToast isInvalidPaste={isInvalidPaste} />
-      </ToastPositionContainer>
     </>
```
paulinhapenedo, line 89: I noticed that we weren't showing the toast on this screen because the component was outside the `Sheet` component and only visible if you dismissed the keyboard (possible on the simulator, not possible IRL). Nad this happened on all devices and OS.

**Before**
https://user-images.githubusercontent.com/690904/184957270-c475c0f5-13f5-4788-80d2-9f71f9481f2a.mp4


**After**

src/components/backup/BackupCloudStep.tsx
```diff
@@ -9,15 +9,22 @@ import React, {
   useState,
 } from 'react';
 
-import { Keyboard } from 'react-native';
+import { Keyboard, StatusBar } from 'react-native';
 import { TouchableWithoutFeedback } from 'react-native-gesture-handler';
 import { DelayedAlert } from '../alerts';
-import BackupSheetKeyboardLayout from './BackupSheetKeyboardLayout';
+import { Column } from '../layout';
```
danibonilha: I'd advise against using Rainbow components, we could use our Container instead of this Column

src/components/backup/BackupCloudStep.tsx
```diff
@@ -135,15 +142,8 @@ export default function BackupCloudStep() {
...
-    <BackupSheetKeyboardLayout
-      footer={
-        <BackupPasswordButtonFooter
-          buttonLabel="Confirm"
-          isValidPassword={validPassword}
-          onButtonPress={onConfirmBackup}
-        />
-      }
-    >
+    <Column>
```
danibonilha: Does it need to be a Column , though ?  If it's just to add the StatusBar, we could use a fragment, or even try the statusBar inside the Touchable (not sure though if works as expected). I'm not even sure if we need to redefine the StatusBar style on these screens, since it's wrapped on the sheet, and the sheet doesn't cover the bar, so we may be able to remove it completely

src/components/backup/BackupConfirmPasswordStep.tsx
```diff
@@ -84,15 +91,8 @@ export default function BackupConfirmPasswordStep() {
...
-      footer={
-        <BackupPasswordButtonFooter
-          buttonLabel={label}
-          isValidPassword={validPassword}
-          onButtonPress={onSubmit}
-        />
-      }
-    >
+    <Column>
+      <StatusBar barStyle="light-content" />
```
danibonilha: ditto

src/components/backup/BackupSheetKeyboardLayout.tsx
```diff
@@ -1,38 +0,0 @@
...
-        {footer}
-      </Container>
-      {Device.isAndroid ? (
-        <KeyboardSizeView>
-          <Container />
-        </KeyboardSizeView>
-      ) : null}
-    </Column>
-  );
-}
```
danibonilha, line 38: nice 👌

src/components/backup/RestoreCloudStep.js
```diff
@@ -103,28 +110,8 @@ export default function RestoreCloudStep({ userData, backupSelected }) {
...
-            }
-            onPress={onPasswordSubmit}
-          >
-            {label}
-          </Button>
-        </Container>
-      }
-    >
+    <Column>
+      <StatusBar barStyle="light-content" />
```
danibonilha, line 114: ditto

danibonilha (commented):
Nice cleanup, let's just verify the statusBar comments

src/components/backup/BackupCloudStep.tsx
paulinhapenedo: There's no need for this wrapper around the component, removed it.

src/components/backup/BackupCloudStep.tsx
paulinhapenedo: StatusBar was not needed here. Thanks for the heads-up!

danibonilha (approved):
LGTM! Always nice to cleanup 🚀

PR #1000, opened 2022-08-16
[CS-4427]: Block any type of navigation on lock screen

danibonilha:
### Description

This PR adds a custom router and navigator to be able to filter and conditionally update the navigation state, mainly to avoid any inner screens to pop-up on top of the lock screen,  the custom navigator is derived from the StackNavigator, the only difference is the getStateForAction, please make sure to test on your devices, since it's a global change.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/navigation/customNavigator.tsx
+++ b/cardstack/src/navigation/customNavigator.tsx
@@ -1,0 +1,94 @@
+import {
+  createNavigatorFactory,
+  DefaultNavigatorOptions,
+  ParamListBase,
+  StackActionHelpers,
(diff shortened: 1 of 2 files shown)


danibonilha:
There's a bug with the payment link, once it's fixed it will be ready for review

danibonilha:
> There's a bug with the payment link, once it's fixed it will be ready for review

Turns out the bug is not related with this change, so I'll be handling it on a different PR

paulinhapenedo (approved):
LGTM! 👏

PR #1001, opened 2022-08-17
[CS-4436]: Fix slug unsynchronized validation

danibonilha:
### Description

This PR fixes two issues on the slug screen:

- The local error being overwriting by the hubValidation, which was caused by the hub responding to a previous value being valid, the solution is to skip setting the hubValidation when slug is partially invalid already.

- Not waiting hubValidation to happen to allow/block next flow, for slower devices and Android specifically the response may take a while, and we don't have a loading state, so  now we wait hubValidation to resolve before allowing the user to continue.

Also removes some dead code.

Maybe we can add some kind of loading state in the message feedback, but for now this fix should do.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Evidence recorded on emulator, on dev mode is slower than release mode:

https://user-images.githubusercontent.com/20520102/185209875-417f933e-85ab-4427-ab23-90465290c046.mp4


oidoug (approved):
Wow that's slow... I noticed something similar using my old iPhone, just tested and it's working a lot better :) thanks for taking care of it!

Maybe debouncing the input (to around 200 ms) can also improve responsiveness.

paulinhapenedo (approved):
LGTM!
Always nice to delete old files too 🚀

PR #1002, opened 2022-08-17
[CS-4423] Logic for retry and support for profile creation

oidoug:
### Description

This PR adds support and retry handling when connection errors or profile creation job fails.

It's hard to test these states manually, so I'm writing unit tests (trying to). To test UI change the return value for the `error` memo in `ProfileScreen`.

- [x] Completes #(CS-4423)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot

### DIFF
--- a/cardstack/src/hooks/useProfileJobPolling.ts
+++ b/cardstack/src/hooks/useProfileJobPolling.ts
@@ -1,8 +1,11 @@
 import { SerializedError } from '@reduxjs/toolkit';
 import { FetchBaseQueryError } from '@reduxjs/toolkit/query/react';
-import { useEffect } from 'react';
+import { useCallback, useEffect, useMemo } from 'react';
 
-import { useGetProfileJobStatusQuery } from '@cardstack/services';
+import {
+  useGetProfileJobStatusQuery,
+  usePostProfileJobRetryMutation,
+} from '@cardstack/services';
 
 import { useBooleanState } from './useBooleanState';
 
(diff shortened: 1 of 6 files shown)


cardstack/src/hooks/useProfileJobPolling.ts
```diff
@@ -47,7 +50,28 @@ export const useProfileJobPolling = ({
...
+
+  const [retryJobID] = usePostProfileJobRetryMutation();
+
+  const retryCurrentCreateProfile = useCallback(() => {
+    if (isCreateProfileError) {
+      retryJobID({ jobTicketID: jobID });
+    }
+
+    // Resume polling, either after a job is restarted or a connection error has occorred.
+    startPolling();
```
danibonilha, line 68: typo: occurred

cardstack/src/screens/ProfileScreen/components/CreateProfileError.tsx
```diff
@@ -10,24 +10,34 @@ import {
...
+  error: ErrorInfoType;
 }
 
 export const CreateProfileError = ({
   onPressRetry,
   onPressSupport,
+  error = {
+    title: strings.profileError.title,
+    message: strings.profileError.message,
+  },
```
danibonilha, line 30: nit: `error = strings.profileError`

cardstack/src/hooks/useProfileJobPolling.ts
danibonilha: It seems we've never stopped polling on the failed state, and retryJobId is async, so if we want to stop polling, we would need to call stopPolling if `isCreateProfileError` , then get the retry mutation status, isSuccess, to call `startPolling` again

danibonilha (approved):

PR #1003, opened 2022-08-17
[CS-4454]: Fix deeplink locking user

danibonilha:
### Description

We were using `Linking.openURL` to handle inner app links, but it seems that it  triggers a new Activity on Android, which sets the AppState to `background` even though the app is still open, making the app lock itself. This PR fixes this issue by using react-navigation helpers to use the provided path and navigate to pay merchant, without locking the screen again.

`Navigation.linkTo` was created because `useAppInit` is outside of the NavigationContext, so we're not able to use the same hook we do on `useScanner`.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/20520102/185239699-65a1695f-6a84-4d3f-b0ae-0b2757d162a0.mp4

### DIFF
--- a/cardstack/src/navigation/Navigation.ts
+++ b/cardstack/src/navigation/Navigation.ts
@@ -1,11 +1,15 @@
 import {
   CommonActions,
+  getActionFromState,
+  getStateFromPath,
   NavigationContainerRef,
   StackActions,
 } from '@react-navigation/native';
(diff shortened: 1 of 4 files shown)


oidoug (approved):
Very nice! Tested and working good.

PR #1004, opened 2022-08-18
[CS-4456]: Fix profile tab freezing

danibonilha:
### Description

This PR fixes the profile tab freezing, react doesn't know how to handle inner function as a component and it's not recommended, it should be either outside the main component or as function, not a component.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Simulator Screen Recording - iPhone 11 - 2022-08-18 at 13 07 00


paulinhapenedo (approved):
LGTM!

PR #1005, opened 2022-08-18
[CS-4170] Handle Skip button on profile creation flow

paulinhapenedo:
### Description
This PR adds a new entry to the redux store called `persistedFlags`, which can be used in the future for other things, but for now it's being used to handle the "Skip" button pressing on the profile creation flow. Once the user skips the profile creation flow, it won't be shown again unless they go to the Profile tab.

- [x] Completes #CS-4170

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/redux/persistedFlagsSlice.ts
+++ b/cardstack/src/redux/persistedFlagsSlice.ts
@@ -1,0 +1,35 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { useSelector } from 'react-redux';
+
+import { AppState } from '@rainbow-me/redux/store';
+
+type SliceType = {
+  hasSkippedProfileCreation: boolean;
+};
+
+const initialState: SliceType = {
+  hasSkippedProfileCreation: false,
+};
+
+const slice = createSlice({
+  name: 'persistedFlags',
+  initialState,
+  reducers: {
+    skipProfileCreation(state, action: PayloadAction<boolean>) {
+      state.hasSkippedProfileCreation = action.payload;
+    },
+  },
+});
+
+export const {
+  name: persistedFlagsName,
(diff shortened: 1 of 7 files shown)


cardstack/src/redux/persistedFlagsSlice.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const initialState: SliceType = {
+  hasSkippedProfileCreation: false,
+};
+
+const slice = createSlice({
+  name: 'persistedFlags',
+  initialState,
+  reducers: {
+    skipProfileCreation(state, action: PayloadAction<boolean>) {
```
oidoug, line 18: Nice 👍 
I prefer this approach of passing the enabling / disabling boolean over what we did in the welcome flag, because having duplicate reducers for just setting true and false becomes impractical when we add more flags.

cardstack/src/redux/persistedFlagsSlice.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+  },
+});
+
+export const {
+  name: persistedFlagsName,
+  actions: { skipProfileCreation },
+} = slice;
+
+export const usePersistedFlagsSelector = () => {
```
oidoug, line 29: Suggestion for a team discussion:

Since we are adding a hook here for selectors, wouldn't be nice to also have the dispatch wrapped by it?

Or maybe we should not use a hook here at all and have the selectors just as functions, and let the components import and use redux's useSelector, just as we do with useDispatcher. I'm considering consistency.

danibonilha (approved):
LGTM

oidoug (approved):
LGTM, left a comment for the team to discuss, not a blocker.

PR #1007, opened 2022-08-19
[CS-4473]: Fix profile layout and profile creation pop-up

danibonilha:
### Description

For the "import" flow we were using primarySafe existing, which can be a little slower than checking the safesQuery result. 

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

image


paulinhapenedo (approved):
LGTM!

PR #1008, opened 2022-08-19
[CS-4455] Tests for polling hook

oidoug:
### Description

This PR adds tests to `useProfileJobPolling` hook, checking if job state responses and retry call triggers the right returns.

- [x] Completes #(CS-4455)

### DIFF
--- a/cardstack/src/hooks/__tests__/useProfileJobPolling.test.ts
+++ b/cardstack/src/hooks/__tests__/useProfileJobPolling.test.ts
@@ -49,7 +49,7 @@ describe('useProfileJobPolling', () => {
     mockUsePostProfileJobRetryMutation();
   });
 
-  it('should start hook waiting when jobID not provided', () => {
+  it('should do nothing if jobID is not provided', () => {
     const { result } = renderHook(() =>
       useProfileJobPolling({ ...mockJobParams, jobID: undefined })
     );
@@ -74,7 +74,7 @@ describe('useProfileJobPolling', () => {
     expect(result.current.isCreatingProfile).toBeFalsy();
   });
 
-  it('should return profile error flag on job has failed', () => {
+  it('should return profile error flag once job has failed', () => {
     const { result, rerender } = renderHook(() =>
       useProfileJobPolling(mockJobParams)
     );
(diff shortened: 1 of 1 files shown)


cardstack/src/hooks/__tests__/useProfileJobPolling.test.ts
```diff
@@ -0,0 +1,131 @@
...
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  beforeEach(() => {
+    mockUseGetProfileJobStatusQuery();
+    mockUsePostProfileJobRetryMutation();
+  });
+
+  it('should start hook waiting when jobID not provided', () => {
```
paulinhapenedo: This assertion seems a bit confusing since there's no `waiting` status. What do you think about changing it to something like `should do nothing if jobID isn't provided`?

paulinhapenedo (approved):
Minor comment, but seems to be covering the use cases for the hook!

cardstack/src/hooks/__tests__/useProfileJobPolling.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      useProfileJobPolling(mockJobParams)
+    );
+
+    mockUseGetProfileJobStatusQuery('success');
+    rerender();
+
+    expect(result.current.isCreatingProfile).toBeFalsy();
+  });
+
+  it('should return profile error flag on job has failed', () => {
```
danibonilha: `on job` -> `once job`

cardstack/src/hooks/__tests__/useProfileJobPolling.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    );
+
+    mockUseGetProfileJobStatusQuery('failed');
+    rerender();
+
+    act(() => {
+      result.current.retryCurrentCreateProfile();
+    });
+
+    expect(mockJobRetryMutation).toBeCalled();
```
danibonilha: we could assert this with:` toBeCalledWith` to make sure the params are correct

danibonilha (approved):

cardstack/src/hooks/__tests__/useProfileJobPolling.test.ts
oidoug: Yeah, really badly written 😂 , thanks for the suggestion.

PR #1009, opened 2022-08-19
[CS-4127] Updated UI for pin code creation

paulinhapenedo:
- [x] Completes #CS-4127

### Description
This PR updates the PinScreen to match the new design guidelines:

- Removal of the subtitle
- Updated copy
- Title with the same style as the onboarding screens

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### iOS
Screenshot

Screenshot

Screenshot

### Android with biometrics
Screenshot

Screenshot

Screenshot

### DIFF
--- a/cardstack/src/screens/PinScreen/PinScreen.tsx
+++ b/cardstack/src/screens/PinScreen/PinScreen.tsx
@@ -4,9 +4,8 @@ import { StatusBar } from 'react-native';
 import {
   Container,
   Icon,
-  InPageHeader,
+  NavigationStackHeader,
   PinInput,
-  SafeAreaView,
   Text,
 } from '@cardstack/components';
 import { BiometricSwitch } from '@cardstack/components/BiometricSwitch';
@@ -52,23 +51,33 @@ const PinScreen = () => {
   );
 
   return (
-    <SafeAreaView
+    <Container
       backgroundColor={colorStyleVariants.backgroundColor[variant]}
       flex={1}
-      paddingHorizontal={5}
     >
       <StatusBar barStyle={statusBarStyle} />
-      <InPageHeader showSkipButton={false} showLeftIcon={canGoBack} />
+      <NavigationStackHeader
+        canGoBack={canGoBack}
+        backgroundColor={colorStyleVariants.backgroundColor[variant]}
+      />
 
-      <Container flex={1} alignItems="center">
-        <Container flex={0.15} width="85%" alignSelf="flex-start">
-          <Text
-            fontSize={22}
-            variant="pageHeader"
-            color={colorStyleVariants.textColor[variant]}
-          >
-            {strings.flow?.[flow]}
-          </Text>
+      <Container flex={0.7} alignItems="center">
+        <Container flex={0.35} width="85%">
+          <Container flex={1} width="80%">
+            <Text
+              variant="pageHeader"
+              color={colorStyleVariants.textColor[variant]}
+              paddingBottom={2}
+            >
+              {strings.flow?.[flow].title}
+            </Text>
+            <Text
+              fontSize={16}
+              color={colorStyleVariants.secondaryTextColor[variant]}
+            >
+              {strings.flow?.[flow]?.subtitle}
+            </Text>
+          </Container>
         </Container>
         <Container
           justifyContent="center"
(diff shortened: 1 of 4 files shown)


cardstack/src/screens/PinScreen/strings.ts
```diff
@@ -2,23 +2,11 @@ import { appName } from '@cardstack/constants';
...
-    subtitle: 'Please re-type your PIN number',
-  },
-  [PinFlow.enter]: {
-    title: 'Enter PIN',
-    subtitle: '',
-  },
-  [PinFlow.new]: {
-    title: 'Enter new PIN',
-    subtitle: `Please enter yout new PIN for your ${appName}`,
-  },
```
danibonilha, line 21: Not sure if we should remove the subtitle, what if we use the style like the restore, to have the title on top, and then the subtitle with a smaller font ? 

something like: 
image

but then the copy being, Create PIN ---- Please choose a pin to protect your wallet on this device. We could also ask for copy for the other flows.

update:
I see you've merged some of the copy, I think it should be fine when we get the layout fixed

cardstack/src/screens/PinScreen/PinScreen.tsx
```diff
@@ -61,25 +61,13 @@ const PinScreen = () => {
...
           >
-            {strings.flow?.[flow]?.title}
-          </Text>
-          <Text
-            fontSize={16}
-            color={colorStyleVariants.secondaryTextColor[variant]}
-            textAlign="center"
-          >
-            {strings.flow?.[flow]?.subtitle}
+            {strings.flow?.[flow]}
```
danibonilha: The layout doesn't seem correct, in my understanding the title should be closer to the back button, and not almost centered as is.

cardstack/src/screens/PinScreen/strings.ts
```diff
@@ -2,23 +2,11 @@ import { appName } from '@cardstack/constants';
...
-    title: 'Confirm PIN',
-    subtitle: 'Please re-type your PIN number',
-  },
-  [PinFlow.enter]: {
-    title: 'Enter PIN',
-    subtitle: '',
-  },
-  [PinFlow.new]: {
-    title: 'Enter new PIN',
-    subtitle: `Please enter yout new PIN for your ${appName}`,
```
danibonilha, line 20: if we were to keep these specific subtitles, there's a typo here, `yout`

cardstack/src/screens/PinScreen/PinScreen.tsx
oidoug: We should match the spacing we used for profile creation, even though it'll leave more empty space between title and the input boxes.

paulinhapenedo: ## Android








## iOS







cc @oidoug @danibonilha

oidoug: Looks like your android sim has a massive screen 😂 
may look strange, but it's more consistent this way.

oidoug (approved):
LGTM

danibonilha (changes requested):
The alignment with the pinInput is a bit off, specially on iPhone SE,  this is what we have today: 
image
We should aim to keep the spacing consistent in the input, and it feels more natural to have the back button aligned with the boxes, if we had the smalller input, then it would make sense to have different alignments

oidoug (changes requested):
Looking more carefully now I see the paddings for the input are a bit strange, too clumped together.

now:


before:


I approved it before on account that this change is a bit complex to get it right. We are mix-matching old and new designs, we'll have to get back to this after we do a design review.

Easiest to do now is to just focus on the title but without changing how the input renders, keeping its original paddings.

cardstack/src/screens/PinScreen/PinScreen.tsx
```diff
@@ -56,31 +56,29 @@ const PinScreen = () => {
       flex={1}
     >
       <StatusBar barStyle={statusBarStyle} />
+
```
danibonilha: nit: no need for new line

danibonilha (approved):
LGTM

oidoug (approved):
Looking good now 🚀

PR #1012, opened 2022-08-22
[CS-4477]: Improve depot screen performance

danibonilha:
### Description

Depot screen was loading all activities on first rendering, even without switching tabs, for users with lots of activities the depot screen was freezing for a couple of seconds preventing any interaction, this PR prevents this behavior by only fetching activities when the tab is accessed. It also: 

- Adds a dark mode to TabHeader component
- Replaces the local tabHeader with newer one to keep it consistent 
- Replaces absolute positioning items to use flex, removing a bunch of useless containers
- Removes old backButton with shrinking animation in favor of opacity one
- Removes dead components
- Adds a custom chevron left icon, to avoid padding on the left, we could add a ticket to use this one on other screens, then there's no need to positioning using a negative margin

Why are depot tabs positioned using zIndex ?
The zIndex here is used to create a type of "stack"  so when you switch tabs the tab is not unmounted, instead it's sent to end of the stack, this way we can keep the scroll position and the data.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

image


cardstack/src/components/Icon/custom-icons/chevron-left-no-box.tsx
```diff
@@ -0,0 +1,20 @@
...
+        stroke={props.color}
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        strokeWidth={3}
+      />
+    </Svg>
+  );
+}
+
+export default SvgComponent;
```
oidoug, line 20: I wonder if it's not simpler to keep using Icon and just encapsulate the negative padding, since it's just this chevron icon we need that. I'm fine with your solution too.

cardstack/src/components/TabHeader/useTabHeader.tsx
```diff
@@ -6,14 +6,15 @@ import { TabHeaderButton } from '.';
...
+  key: string | number;
 }
 
 export interface TabsArrayProps {
   tabs: Array<TabType>;
+  lightMode?: boolean;
 }
 
-export const useTabHeader = ({ tabs }: TabsArrayProps) => {
+export const useTabHeader = ({ tabs, lightMode = true }: TabsArrayProps) => {
```
oidoug: nit: I would prefer to use `ThemeVariant` as defined in `colorStyleVariants` for light/dark variants.

cardstack/src/screens/DepotScreen.tsx
```diff
@@ -76,174 +106,110 @@ export default function DepotScreen() {
...
     isFetchingMore,
     onEndReached,
     refetchLoading,
     refetch,
     isLoadingTransactions,
-  } = useDepotTransactions(address);
+  } = useDepotTransactions(depotAddress);
 
   return (
     <Container marginTop={7} flexDirection="column" width="100%">
```
oidoug, line 177:
```suggestion
    <Container flexDirection="column" width="100%">
```
This margin adds a gap that is apparent when scrolling. I suggest removing this margin and adding `paddingTop: 20` contentContainerStyle.

Ideally we would add this paddingTop to layouts.ts -> contentContainerStyle, but then all components using that will need to be checked and updated. We could add a ticket to handle that too if you prefer.

cardstack/src/screens/DepotScreen.tsx
```diff
@@ -76,174 +106,110 @@ export default function DepotScreen() {
...
+          isLoadingTransactions ? (
+            <TransactionListLoading light />
+          ) : (
+            <ListEmptyComponent text="No transactions" />
+          )
+        }
+        ListFooterComponent={
+          isFetchingMore ? <ActivityIndicator color="white" /> : null
+        }
+        contentContainerStyle={sectionStyle.contentContainerStyle}
```
oidoug, line 190:
```suggestion
        contentContainerStyle={[
          sectionStyle.contentContainerStyle,
          { paddingTop: 20 },
        ]}
```
As mentioned above.

oidoug (approved):
So much better, faster and it looks better.
There's one minor padding issue in the Activities list, we can fix it in this PR or add a card to take care of it.

cardstack/src/components/Icon/custom-icons/chevron-left-no-box.tsx
danibonilha: I believe the current implementation is more straightforward, we are still using the Icon component, but with the option to use the chevron without box on the left, it also seems more reliable for cases we want to use the chevron icon with different sizes and without being a back button for example.

cardstack/src/components/TabHeader/useTabHeader.tsx
danibonilha: nice thought, I totally forgot about that one

cardstack/src/screens/DepotScreen.tsx
danibonilha: That's just a leftover from the previous implementation, the only change I made was the loading, and the props, but I'll take a look this gapping

danibonilha: This "gap" seems to only happen on Android, iOS has stickyHeader as default whereas Android doesn't, so I'm enabling it, to have the same behavior

danibonilha: https://user-images.githubusercontent.com/20520102/186183613-fd795924-77c3-45f8-b28e-cdb8cad1908f.mp4

oidoug (approved):
LGTM!

PR #1015, opened 2022-08-24
[CS-4499] Add returnKeyType DONE to Profile Creation inputs

paulinhapenedo:
### Description
This PR enhances the user experience on the Profile Creation flow, changing the keyboard action key to be "Done" (iOS) or a checkmark (Android), instead of the default action key.

- [x] Completes #CS-4499

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot

### DIFF
--- a/cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
+++ b/cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
@@ -57,6 +57,7 @@ const SuffixedInput = ({
       allowFontScaling={false}
       onChangeText={onChangeText}
       value={value}
+      returnKeyType="done"
     />
     <Text
       style={styles.textStyle}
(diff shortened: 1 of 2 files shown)


cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
```diff
@@ -57,6 +57,7 @@ const SuffixedInput = ({
       allowFontScaling={false}
       onChangeText={onChangeText}
       value={value}
+      returnKeyType="done"
```
oidoug, line 60: nit: "next" makes sense here too. I'll leave it to you to decide.

cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
```diff
@@ -263,6 +263,7 @@ export const ProfileNameScreen = () => {
             paddingBottom={4}
             onChangeText={onChangeText}
             value={profile.name}
+            returnKeyType="done"
```
oidoug, line 266: same as above

oidoug (approved):
"done" is fine, but maybe using "next" works better. Just a suggestion, up to you.

danibonilha (approved):
I don't think it matter much, and don't want a lot of back and forth, but in my personal opinion, `next` feels like I did an action and will be taken automatically to the next step, whereas `done` it's just stating that I finished an action and will remain on the same step until I take the action to go to next step. 
But if we are going with NEXT, we also need to update the PR's title

PR #1016, opened 2022-08-25
[CS-4482] Disabling confirm button while loading data

oidoug:
### Description

PR to improve data loading on RegisterMerchantDisplay transaction sheet.

- Changes `usePrepaidCard` hook to use `useGetSafesDataQuery` that has probably already cached safes, instead of `useGetPrepaidCardsQuery` that would be a new query;
- Adds isLoading for `PayThisAmountSection` on `RegisterMerchantDisplay` too;
- Adds skeleton to Profile header info;
- Passes `disabledConfirmButton` on Register to disable pay section until loading done.

Note:

As it is now, we can't determine for certain when individual sections finish loading, they don't have a way to report back. It's imperceptible, but confirm button is still being enabled before all is loaded. Fixing this is honestly not worth it since loading is shown in all sections now and it's usually fast.

- [x] Completes #(CS-4482)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/186911011-41d25b37-8d52-4570-ac8a-6aab37bdf882.mp4

### DIFF
--- a/cardstack/src/components/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
+++ b/cardstack/src/components/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
@@ -124,6 +124,7 @@ const Header = ({
 const SheetFooter = ({
   onConfirm,
   onCancel,
+  loading,
   onConfirmLoading = false,
   disabledConfirmButton,
 }: TransactionConfirmationDisplayProps) => (
@@ -146,7 +147,7 @@ const SheetFooter = ({
         {strings.buttons.cancel}
       </Button>
       <Button
-        loading={onConfirmLoading}
+        loading={onConfirmLoading || loading}
         variant="small"
         onPress={onConfirm}
         disabled={disabledConfirmButton}
--- a/cardstack/src/components/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
+++ b/cardstack/src/components/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
@@ -147,10 +147,10 @@ const SheetFooter = ({
         {strings.buttons.cancel}
       </Button>
       <Button
-        loading={onConfirmLoading || loading}
+        loading={onConfirmLoading}
         variant="small"
         onPress={onConfirm}
-        disabled={disabledConfirmButton}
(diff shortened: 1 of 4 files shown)


cardstack/src/components/TransactionConfirmationSheet/TransactionConfirmationSheet.tsx
```diff
@@ -146,7 +147,7 @@ const SheetFooter = ({
         {strings.buttons.cancel}
       </Button>
       <Button
-        loading={onConfirmLoading}
+        loading={onConfirmLoading || loading}
```
danibonilha: What are we trying to achieve here ? I thought we wanted to disabled the button, while the state was loading, so this prop should go on disabled

paulinhapenedo: Shouldn't the `loading` be passed to the `disabled` prop instead? This will solve the loading problem of the first screenshot on the ticket. 

About the second loading, @danibonilha and I were discussing about it and she gave an advice to change the `PrepaidCardTransactionSection` query from `usePrepaidCard` to the `useGetSafesDataQuery` like we have on `useChoosePrepaidCard`. Since safes info are cached, we may not even see the second loading happening (or it will be a bit faster, at least).

oidoug: I see, we use loading in the button only after the action was taken, is that it? I never consider that, that's why I tend to use loading in the button instead of disable.

The goal is to avoid confirming before data was all presented to the user to decide.

oidoug: Nice, that makes sense, I'll try that.

danibonilha: Yes, that feels more intuitive, loading on btn is the result of the action

oidoug:
@paulinhapenedo @danibonilha , I updated the PR and the description, please take a look at it again.
Thanks for the comments, they helped me find a path forward 😊 👍

paulinhapenedo (approved):
It's looking way better! 👏

PR #1017, opened 2022-08-25
[CS-4426]: Handle pending/failed profile job on init 

danibonilha:
### Description

This PR adds the query to check if there's any pending or failed job on profile init, it also updates the profilejobPolling to invalidate the cache on the api, instead of having a callback.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Video on [Linear ticket](https://linear.app/cardstack/issue/CS-4426/[logic]-get-job-tickets-for-type-creation-profile-for-a-eoa-on-opening)

### DIFF
--- a/cardstack/src/services/hub/hub-api.ts
+++ b/cardstack/src/services/hub/hub-api.ts
@@ -4,6 +4,7 @@ import { createApi } from '@reduxjs/toolkit/query/react';
 import { CustodialWallet, ProfileIDUniquenessResponse } from '@cardstack/types';
 import { transformObjKeysToCamelCase } from '@cardstack/utils';
 
+import { CacheTags as SafesCacheTags, safesApi } from '../safes-api';
 import { queryPromiseWrapper } from '../utils';
 
 import {
@@ -28,6 +29,7 @@ import {
   PostProfilePurchaseQueryResult,
   UpdateProfileInfoParams,
   JobTicketTypeResult,
+  JobStateType,
 } from './hub-types';
 
 const routes = {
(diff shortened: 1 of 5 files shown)


cardstack/src/services/hub/hub-api.ts
```diff
@@ -151,20 +153,39 @@ export const hubApi = createApi({
...
+      },
     }),
     postProfileJobRetry: builder.mutation<unknown, { jobTicketID: string }>({
       query: ({ jobTicketID }) => ({
         url: `${routes.profileInfo.jobTicket}/${jobTicketID}/retry`,
         method: 'POST',
       }),
     }),
+    getProfileUnfulfilledJob: builder.query<string | undefined, void>({
+      query: () => `${routes.profileInfo.jobTicket}`,
```
oidoug, line 178: Shouldn't this also be querying by EOA? Otherwise it will return another user's pending job.

oidoug (commented):
Apparently you're not querying pending jobs by EOA.
I think you need to either query by EOA or save the pending jobID to check if only that job is pending.

cardstack/src/services/hub/hub-api.ts
danibonilha: No, the api uses the hubAuthToken to extract the eoa, there's no eoa param for this api,  you can check in this PR https://github.com/cardstack/cardstack/pull/3121

oidoug: Ah cool, forgot about that.

All good then.

oidoug (approved):
LGTM!

Awesome that cache invalidation works across apis, even though the api is a bit strange.

PR #1018, opened 2022-08-25
[CS-4518] Staking tag only on Depot

oidoug:
### Description

PR adds flags isDepot to BalanceSection and isOnDepot on TokenBalance to only show staking tag when it's in a depot.

- [x] Completes #(CS-4518)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/components/BalanceSection/BalanceSection.tsx
+++ b/cardstack/src/components/BalanceSection/BalanceSection.tsx
@@ -5,6 +5,7 @@ import { ScrollView } from 'react-native';
 import { Container, TokenBalance } from '@cardstack/components';
 import { Routes } from '@cardstack/navigation';
 import { TokenType } from '@cardstack/types';
+import { isBridgedCardToken } from '@cardstack/utils';
 
 interface BalancesProps {
   tokens: TokenType[];
@@ -35,7 +36,7 @@ export const BalanceSection = ({
     () =>
       tokens.map(token => (
         <TokenBalance
-          isOnDepot={isDepot}
+          isStaking={isBridgedCardToken(token.token.symbol) && isDepot}
           key={token.tokenAddress}
           address={token.tokenAddress}
           includeBorder
--- a/cardstack/src/components/BalanceSection/BalanceSection.tsx
+++ b/cardstack/src/components/BalanceSection/BalanceSection.tsx
@@ -9,9 +9,14 @@ import { TokenType } from '@cardstack/types';
 interface BalancesProps {
   tokens: TokenType[];
   safeAddress: string;
+  isDepot?: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


cardstack/src/components/TokenBalance/TokenBalance.tsx
```diff
@@ -91,7 +93,7 @@ export const TokenBalance = (props: TokenBalanceProps) => {
               <Text variant="subText">{nativeBalance}</Text>
             </Container>
           </Container>
-          {isCardCPXD && <FloatingTag copy={strings.staking} />}
+          {isOnDepot && isCardCPXD && <FloatingTag copy={strings.staking} />}
```
danibonilha: What if we make a condition like, `isStaking` instead of isOnDepot
```tsx
isStaking && <FloatingTag copy={strings.staking} />
```

This way TokenBalance doesn't need to know about depot, or the type of the token

paulinhapenedo (approved):
My bad! 🙈

cardstack/src/components/Depot/Depot.tsx
```diff
@@ -51,6 +51,7 @@ const Bottom = ({ tokens }: DepotProps) => {
       {firstThreeTokens.map((item, index) => (
         <Container key={`token-balance-${index}`}>
           <TokenBalance
+            isOnDepot
```
danibonilha: then here we could have, 
```tsx
isStaking={isBridgedCardToken(item.tokenSymbol)}
.....

// utils function

const isBridgedCardToken  = (tokenSymbol: string) => tokenSymbol === 'CARD.CPXD'
```

cardstack/src/components/BalanceSection/BalanceSection.tsx
```diff
@@ -30,6 +35,7 @@ export const BalanceSection = ({ tokens, safeAddress }: BalancesProps) => {
     () =>
       tokens.map(token => (
         <TokenBalance
+          isOnDepot={isDepot}
```
danibonilha: and here:
```tsx
isStaking={isBridgedCardToken(token.tokenSymbol) && isDepot}
```

oidoug:
> My bad! 🙈

No worries, it slipped through reviews too.

danibonilha (approved):
I think it's ok to drill down the depot flag, my suggestion is just to split the concerns of the tokenBalance but it's up to you to decide if it makes sense.  Just a heads up, the last screenshoot doesn't "prove" the staking is gone for card, since the tokens there are DAI

cardstack/src/components/TokenBalance/TokenBalance.tsx
oidoug: Agreed, that's cleaner.

I took the screenshot of the wrong profile 🤦 😂  (updated).

PR #1020, opened 2022-08-26
[CS-4445] Reward claim step over

tintinthong:
### Description

In rewards, the gas fees is taken out from the reward amount. So we have an issue that claims fail when the reward amount < gas fees (crypto dust). We seek to still show users that they have rewards (crypto dust) but we want to ensure that the user can claim all rewards which are not crypto dust. This is important becos we anticipate with staking that there will be lots of crypto dust in ppls rewards.

- `claimSheetTokenInfo` is meant to include reward token balance and claim estimates only including proofs that CAN be claimed -- u will see reward token balance lower bcos it doesn't include crypto dust. This is displayed in `rewardsClaimSheet`. 
- `mainPoolTokenInfo` shows balances regardless of crypto dust and will be displayed in main rewards center screen
- changed text i rewards screen to state there is possibility that claiming is not allowed
- passed `canClaimAll` state which toggles `claim` button in rewards center screen. `canClaim` only means that there is >0 `rewardTokenBalancesWithoutDust`
- Previously, balances were not shown if their dust. Now, we have allowed that by introducing `acceptDust` since in rewards we want to allow that. 

- [x] Completes (#CS-4445)

TODO
- [ ] **Can possibly remove some old functions**
- [ ] **Need to add new sdk version when pr is approved**

### Checklist

- [x] Tested on a simulator

### Screenshots

**1. A person has 0.2 reward token balance (1 proof which is crypto dust). Does see notification for reward but no claim button to proceed.** 

Screenshot

Screenshot

**1. A person has 10.2 reward token balance (1 proof which is crypto dust, 1 proof which is claimable 10 card worth). The claim sheet only shows 10 card and 0.5 estimate which is the estimate for 1 proof only. Balance recovers to crypto dust**

Screenshot
Screenshot
Screenshot
Screenshot
Screenshot

### DIFF
--- a/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
+++ b/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
@@ -20,7 +20,7 @@ const RewardsCenterScreen = () => {
     hasRewardsAvailable,
     mainPoolTokenInfo,
     isLoading,
-    canClaimAll
+    canClaimAll,
   } = useRewardsCenterScreen();
 
   const mainPoolRowProps = useMemo(
@@ -29,9 +29,8 @@ const RewardsCenterScreen = () => {
       subText: mainPoolTokenInfo?.native.balance.display || '',
       coinSymbol: mainPoolTokenInfo?.token.symbol || '',
       showClaimBtn: canClaimAll,
-      
     }),
-    [mainPoolTokenInfo]
+    [mainPoolTokenInfo, canClaimAll]
   );
 
   return (
--- a/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
+++ b/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
@@ -27,7 +27,7 @@ const RewardsCenterScreen = () => {
       primaryText: fullBalanceToken?.balance.display || '',
       subText: fullBalanceToken?.native.balance.display || '',
       coinSymbol: fullBalanceToken?.token.symbol || '',
-      showClaimBtn: fullBalanceToken?.isClaimable,
+      isClaimable: !!fullBalanceToken?.isClaimable,
     }),
     [fullBalanceToken]
   );
--- a/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
+++ b/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
@@ -20,6 +20,7 @@ const RewardsCenterScreen = () => {
     hasRewardsAvailable,
     mainPoolTokenInfo,
     isLoading,
+    canClaimAll
   } = useRewardsCenterScreen();
 
   const mainPoolRowProps = useMemo(
@@ -27,6 +28,8 @@ const RewardsCenterScreen = () => {
       primaryText: mainPoolTokenInfo?.balance.display || '',
       subText: mainPoolTokenInfo?.native.balance.display || '',
       coinSymbol: mainPoolTokenInfo?.token.symbol || '',
+      showClaimBtn: canClaimAll,
+      
     }),
     [mainPoolTokenInfo]
   );
--- a/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
+++ b/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
@@ -18,19 +18,18 @@ const RewardsCenterScreen = () => {
   const {
     isRegistered,
     hasRewardsAvailable,
-    mainPoolTokenInfo,
+    fullBalanceToken,
     isLoading,
-    canClaimAll,
   } = useRewardsCenterScreen();
 
   const mainPoolRowProps = useMemo(
     () => ({
-      primaryText: mainPoolTokenInfo?.balance.display || '',
-      subText: mainPoolTokenInfo?.native.balance.display || '',
-      coinSymbol: mainPoolTokenInfo?.token.symbol || '',
-      showClaimBtn: canClaimAll,
+      primaryText: fullBalanceToken?.balance.display || '',
+      subText: fullBalanceToken?.native.balance.display || '',
+      coinSymbol: fullBalanceToken?.token.symbol || '',
+      showClaimBtn: fullBalanceToken?.isClaimable,
     }),
-    [mainPoolTokenInfo, canClaimAll]
+    [fullBalanceToken]
   );
 
   return (
--- a/cardstack/src/screens/RewardsCenterScreen/RewardsClaimSheet/useRewardsClaim.ts
+++ b/cardstack/src/screens/RewardsCenterScreen/RewardsClaimSheet/useRewardsClaim.ts
@@ -5,8 +5,8 @@ import { defaultErrorAlert } from '@cardstack/constants';
 import { useMutationEffects } from '@cardstack/hooks';
 import { useLoadingOverlay } from '@cardstack/navigation';
 import {
-  useClaimRewardsMutation,
-  useGetClaimRewardsGasEstimateQuery,
+  useClaimAllRewardsMutation,
+  useGetClaimAllRewardsGasEstimateQuery,
 } from '@cardstack/services/rewards-center/rewards-center-api';
 import {
   RewardsClaimData,
@@ -32,19 +32,19 @@ const useRewardsClaim = () => {
   const {
     rewardSafes,
     defaultRewardProgramId,
-    mainPoolTokenInfo,
+    claimSheetTokenInfo,
   } = useRewardsDataFetch();
 
   // flag to avoid rerendering the screen after the claim happened
   const isClaiming = useRef(false);
-  const mainPoolTokenRef = useRef(mainPoolTokenInfo);
+  const mainPoolTokenRef = useRef(claimSheetTokenInfo);
 
   const { dismissLoadingOverlay, showLoadingOverlay } = useLoadingOverlay();
 
   const [
-    claimRewards,
+    claimAllRewards,
     { isSuccess: isClaimSuccess, isError: isClaimError },
-  ] = useClaimRewardsMutation();
+  ] = useClaimAllRewardsMutation();
 
   const rewardSafeForProgram = useMemo(
     () =>
@@ -65,7 +65,7 @@ const useRewardsClaim = () => {
     data: estimatedGasClaim,
     isLoading: loadingEstimatedGasClaim,
     isFetching: fetchingEstimatedGasClaim,
-  } = useGetClaimRewardsGasEstimateQuery(claimParams, {
+  } = useGetClaimAllRewardsGasEstimateQuery(claimParams, {
     refetchOnMountOrArgChange: true,
     skip: isClaiming.current,
   });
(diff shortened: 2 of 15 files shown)


cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
```diff
@@ -51,6 +51,7 @@ export const ClaimContent = ({ claimList }: ClaimContentProps) => {
           subText={item.subText}
           paddingBottom={index + 1 < claimList.length ? 5 : 0}
           onClaimPress={onClaimPress}
+          showClaimBtn={item.showClaimBtn}
```
oidoug: nit: Instead of adding a new prop you could pass undefined to `onClaimPress` if `item.showClaimBtn` is false.

cardstack/src/screens/RewardsCenterScreen/useRewardsDataFetch.ts
```diff
@@ -44,22 +44,56 @@ const useRewardsDataFetch = () => {
...
+        refetchOnMountOrArgChange: true,
+      },
+    }),
+    [accountAddress, rewardSafes, nativeCurrency, network]
+  );
+
+  const {
+    isLoading: isLoadingTokensWithoutDust,
+    data: { rewardPoolTokenBalances: rewardPoolTokenBalancesWithoutDust } = {},
+  } = useGetRewardPoolTokenBalancesQuery(dustQuery.params, dustQuery.options);
```
danibonilha: instead of creating the `dustQuery` we can reuse the `query` object here, just adding the missing param: 
```tsx
useGetRewardPoolTokenBalancesQuery(
    { ...query.params, safeAddress: rewardSafes?.[0].address },
    query.options
  );
```

cardstack/src/screens/RewardsCenterScreen/useRewardsDataFetch.ts
```diff
@@ -44,22 +44,56 @@ const useRewardsDataFetch = () => {
...
+
   // Checks if available tokens matches default program and has amount
```
danibonilha: we should remove the last piece of this comment, since there's no amount check anymore

oidoug (approved):
This PR covers all cases described in card 4445 (https://linear.app/cardstack/issue/CS-4445/handle-crypto-dust-rewards-in-mobile-app) except the point 2. Rewards badge should not display as active if user only has dust.

If you prefer we can handle that specifically in a new ticket after this one is merged.

I think it's clear to merge once SDK is updated and lint issues resolved.

cardstack/src/screens/RewardsCenterScreen/useRewardsDataFetch.ts
```diff
@@ -68,13 +102,17 @@ const useRewardsDataFetch = () => {
...
   const hasRewardsAvailable = !!mainPoolTokenInfo;
 
+  const canClaimAll = !!claimSheetTokenInfo;
+
   return {
     isLoading,
     rewardSafes,
     rewardPoolTokenBalances,
     mainPoolTokenInfo,
+    claimSheetTokenInfo,
```
danibonilha: I believe mainPoolTokenInfo and claimSheetTokenInfo are a bit confusing now,  we could rename those to reflect their purposes, `mainPoolTokenInfo  => fullBalanceToken` and `claimSheetTokenInfo => claimableBalanceToken`

cardstack/src/screens/RewardsCenterScreen/RewardsClaimSheet/useRewardsClaim.ts
```diff
@@ -32,19 +32,19 @@ const useRewardsClaim = () => {
...
     rewardSafes,
     defaultRewardProgramId,
-    mainPoolTokenInfo,
+    claimSheetTokenInfo,
   } = useRewardsDataFetch();
 
   // flag to avoid rerendering the screen after the claim happened
   const isClaiming = useRef(false);
-  const mainPoolTokenRef = useRef(mainPoolTokenInfo);
+  const mainPoolTokenRef = useRef(claimSheetTokenInfo);
```
danibonilha: we should change the name of this ref to match the used variable

cardstack/src/screens/RewardsCenterScreen/useRewardsDataFetch.ts
```diff
@@ -68,13 +102,17 @@ const useRewardsDataFetch = () => {
    */
   const hasRewardsAvailable = !!mainPoolTokenInfo;
 
+  const canClaimAll = !!claimSheetTokenInfo;
```
danibonilha: Is just checking `!!claimSheetTokenInfo` existence enough ?  Isn't it safer to check `!!claimSheetTokenInfo.balance`  in case claimSheetTokenInfo is an empty array, using `!!claimSheetTokenInfo'  will return true.


1. Instead of having this prop here and then passing down to token balances, we could add this to `mainPoolTokenInfo` (soon to be ) `fullBalanceToken`: 
```tsx
  const fullBalanceToken = useMemo(() => {
    const balances = rewardPoolTokenBalances?.find(
      ({ rewardProgramId }) => rewardProgramId === defaultRewardProgramId
    );

    return { ...balances, isClaimable: !!claimableBalanceToken.balance };
  }, [rewardPoolTokenBalances, claimableBalanceToken, defaultRewardProgramId]);
  ```

File: cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx [hunk 7]
```diff
@@ -20,15 +20,17 @@ const RewardsCenterScreen = () => {
...
     isLoading,
+    canClaimAll,
   } = useRewardsCenterScreen();
 
   const mainPoolRowProps = useMemo(
     () => ({
       primaryText: mainPoolTokenInfo?.balance.display || '',
       subText: mainPoolTokenInfo?.native.balance.display || '',
       coinSymbol: mainPoolTokenInfo?.token.symbol || '',
+      showClaimBtn: canClaimAll,
```
danibonilha: 1.1 then here, you replace showClaimBtn with `isClaimable:  mainPoolTokenInfo?.isClaimable`

cardstack/src/screens/RewardsCenterScreen/components/RewardRow.tsx
```diff
@@ -72,7 +74,7 @@ export const RewardRow = ({
           )
         }
       </Container>
-      {!!onClaimPress && (
+      {!!onClaimPress && showClaimBtn && (
```
danibonilha: 1.3 this change could be dropped

cardstack/src/screens/RewardsCenterScreen/useRewardsCenterScreen.ts
```diff
@@ -38,5 +39,6 @@ export const useRewardsCenterScreen = () => {
     hasRewardsAvailable,
     mainPoolTokenInfo,
     isLoading,
+    canClaimAll,
   };
```
danibonilha, line 41: 1.4 this one too

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
danibonilha: 1.2 then here you could have:
```ts
  onClaimPress={item.isClaimable ? onClaimPress : undefined}
```

cardstack/src/services/gnosis-service.ts
```diff
@@ -274,9 +275,9 @@ export const addNativePriceToToken = async (
 
   const isAmountDust = nativeBalance < 0.01;
 
-  const amount = isAmountDust ? 0 : nativeBalance;
+  const amount = acceptDust ? nativeBalance : isAmountDust ? 0 : nativeBalance;
 
-  const tokenBalance = isAmountDust ? 0 : balance;
+  const tokenBalance = acceptDust ? balance : isAmountDust ? 0 : balance;
```
danibonilha: we could simply this with:
```ts
  const isAmountDust = !acceptDust && nativeBalance < 0.01;

  const amount = isAmountDust ? 0 : nativeBalance;

  const tokenBalance = isAmountDust ? 0 : balance;
```

cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -201,6 +206,32 @@ export const getClaimRewardsGasEstimate = async ({
...
+  const rewardPoolInstance = await getRewardsPoolInstance();
+
+  const totalEstimatedGas = await rewardPoolInstance.claimAllGasEstimate(
+    safeAddress,
+    rewardProgramId,
+    tokenAddress
+  );
+
+  const totalEstimatedGasInEth = fromWei(
+    totalEstimatedGas[0].amount.toString()
```
danibonilha: Does `claimAllGasEstimate`  returns a array ? I'm a bit confused by `totalEstimatedGas[0].amount`

danibonilha (changes requested):
Adding the changes request just to avoid merging, before sdk and ci issues are fixed

cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -201,6 +206,32 @@ export const getClaimRewardsGasEstimate = async ({
   return formattedTotalEstimatedGas;
 };
 
+export const getClaimAllRewardsGasEstimate = async ({
+  safeAddress,
+  tokenAddress,
+  rewardProgramId,
+  accountAddress,
+}: RewardsClaimGasEstimateParams) => {
+  console.log(accountAddress);
```
danibonilha: leftover console.log

cardstack/src/services/rewards-center/rewards-center-service.ts
tintinthong: Right. I should have left a comment here. This is related to an SDK change. 

So `claimAllGasEstimate` derives estimates from all proofs. But our proofs can have different tokens. And `gasToken` is paid in whatever tokens (either dai or card or whatever new gas token). So what I did was return an array becuase u can have `[GasEstimateCardToken, GasEstimateDaiToken, GasEstimateNewGasToken]`.  The advantage of this is that you can list totalEstimatedGas in multiple tokens in the confirmation screen. 

Now the SDK only accepts `GasEstimateCardToken` but I didn't want to force myself into making assumptions on what tokens the proofs are. But I guess this has forced me into taking the first element and assuming this is card token for now IN THE WALLET

Perhaps, I should add a comment???

cardstack/src/screens/RewardsCenterScreen/useRewardsDataFetch.ts
tintinthong: im not too certain how react works. 

`useMemo` seems to take a list of dependencies one of which is `rewardSafes`. I assumed the cached value will change dependening on if `rewardSafes` is available. Which is why I made another query.

danibonilha: The common way to passing params to those queries, are inline objects, for this hook specifically we are using memo, because multiple queries needed the same params, and options, but it's basically like a `static variable`, which will only update if one of the deps change.

tintinthong: ok. Was just wondering that `dustQuery` uses rewardSafe as a dependency which is non-existent in `query`. But even after removal I have not seen any changes to the output.

tintinthong:
I fixed bug here [476593c](https://github.com/cardstack/cardwallet/pull/1020/commits/476593c7473364fcaf56adaa84624931ae2e7c75) and improved the same bug fix here [f6f33b5](https://github.com/cardstack/cardwallet/pull/1020/commits/f6f33b53ed91402d3ae595923eb45742d222c5a0). It was breaking the app when user did not have reward safe! Can please help eyeball this for me

oidoug (commented):
I have nothing new to add.

danibonilha (approved):
Looks good to go

PR #1022, opened 2022-08-26
[CS-2299] Add longPress copy on QRCode

paulinhapenedo:
### Description
This PR adds a new feature to the QRCode preview. If the user taps and hold on top of the QRCode, the link will be copied to the clipboard.

- [x] Completes #CS-2299

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/screens/PaymentRequest/components/PaymentRequestConfirmation.tsx
+++ b/cardstack/src/screens/PaymentRequest/components/PaymentRequestConfirmation.tsx
@@ -1,5 +1,4 @@
 import React, { useCallback } from 'react';
-import { Pressable } from 'react-native';
 
 import { Button, Container, StyledQRCode, Text } from '@cardstack/components';
 import { useCopyToast } from '@cardstack/hooks';
@@ -34,9 +33,10 @@ export const PaymentRequestConfirmation = (
     <>
       <Container paddingHorizontal={5} width="100%">
         <Container alignItems="center" paddingTop={5}>
-          <Pressable onLongPress={handleLongPressQRCode}>
-            <StyledQRCode value={paymentRequestDeepLink} />
-          </Pressable>
+          <StyledQRCode
+            value={paymentRequestDeepLink}
+            onLongPress={handleLongPressQRCode}
+          />
         </Container>
         <Container
           backgroundColor="grayCardBackground"
--- a/cardstack/src/screens/PaymentRequest/components/PaymentRequestConfirmation.tsx
+++ b/cardstack/src/screens/PaymentRequest/components/PaymentRequestConfirmation.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useCallback } from 'react';
+import { Pressable } from 'react-native';
 
 import { Button, Container, StyledQRCode, Text } from '@cardstack/components';
 import { useCopyToast } from '@cardstack/hooks';
@@ -25,11 +26,17 @@ export const PaymentRequestConfirmation = (
     customCopyLabel: 'Payment Request Link',
   });
 
+  const handleLongPressQRCode = useCallback(() => copyToClipboard(), [
+    copyToClipboard,
+  ]);
+
   return (
     <>
       <Container paddingHorizontal={5} width="100%">
         <Container alignItems="center" paddingTop={5}>
-          <StyledQRCode value={paymentRequestDeepLink} />
+          <Pressable onLongPress={handleLongPressQRCode}>
+            <StyledQRCode value={paymentRequestDeepLink} />
+          </Pressable>
         </Container>
         <Container
           backgroundColor="grayCardBackground"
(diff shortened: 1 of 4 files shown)


cardstack/src/screens/QRScannerScreen/pages/RequestQRCode/RequestQRCodePage.tsx
```diff
@@ -49,10 +59,11 @@ const RequestQRCodePage = () => {
         justifyContent="space-around"
         paddingHorizontal={5}
       >
-        <Container justifyContent="space-around" flex={0.3}>
+        <Container justifyContent="space-between" flex={0.3}>
           <Text
             fontSize={18}
             color="white"
+            paddingBottom={3}
```
paulinhapenedo, line 67: This change is to adjust the empty state on smaller screens. The button was too close to the content.

cardstack/src/screens/PaymentRequest/components/PaymentRequestConfirmation.tsx
```diff
@@ -25,11 +26,17 @@ export const PaymentRequestConfirmation = (
...
+  const handleLongPressQRCode = useCallback(() => copyToClipboard(), [
+    copyToClipboard,
+  ]);
+
   return (
     <>
       <Container paddingHorizontal={5} width="100%">
         <Container alignItems="center" paddingTop={5}>
-          <StyledQRCode value={paymentRequestDeepLink} />
+          <Pressable onLongPress={handleLongPressQRCode}>
```
oidoug: Better use our global delay here too, it's defined in `constants` as `delayLongPressMs`. Add it to Pressable prop `delayLongPress`.

cardstack/src/screens/QRScannerScreen/pages/RequestQRCode/RequestQRCodePage.tsx
```diff
@@ -68,59 +79,64 @@ const RequestQRCodePage = () => {
...
-    </CenteredContainer>
+    <>
+      <CenteredContainer
+        paddingHorizontal={10}
+        flex={1}
+        backgroundColor="backgroundDarkPurple"
+        style={styles.container}
+      >
+        <Container flexShrink={1} justifyContent="flex-start">
+          <Pressable onLongPress={handleLongPressQRCode}>
```
oidoug: Same as above (add delayLongPressMs).

oidoug (approved):
LGTM, only thing I think it's better to have a bit longer long press delay.

Just a comment for another card:
I noticed the toast component gets pretty high up in this case. That's not what is usually expected for a Toast component, it should be always pinned to the screen bottom. I'm adding a linear ticket to investigate this.

paulinhapenedo:
> LGTM, only thing I think it's better to have a bit longer long press delay.
> 
> Just a comment for another card: I noticed the toast component gets pretty high up in this case. That's not what is usually expected for a Toast component, it should be always pinned to the screen bottom. I'm adding a linear ticket to investigate this.

We dont need a linear ticket for that. If you take a look at where we are using the ToastComponent for this ticket, we are not using it at the bottom of the screen.

cardstack/src/screens/QRScannerScreen/pages/RequestQRCode/RequestQRCodePage.tsx
danibonilha: Can we have screenshots ?

cardstack/src/screens/PaymentRequest/components/PaymentRequestConfirmation.tsx
danibonilha: I think it's ok to leave the default LongPress for this case, since the QRCode doesn't even have an action onPress

cardstack/src/screens/PaymentRequest/components/PaymentRequestConfirmation.tsx
```diff
@@ -25,11 +26,17 @@ export const PaymentRequestConfirmation = (
...
+            <StyledQRCode value={paymentRequestDeepLink} />
+          </Pressable>
```
danibonilha: As far ar making the QRCode pressable on the requestPayment screen, I think it's okay, but kind of useless since we have a button that already does that, but if we still want to keep the same behavior across the qrcodes, I'd suggest that we add a prop on  `StyledQRCode` to handle longPress, rather than adding a pressable on each

oidoug:
Toasts are generally placed at a fixed position related to the screen's bottom (not components). If we feel our Toast can also behave like a Tooltip, that's fine too, but I think the current intention was to have it always at the bottom. 
We can discuss this more offline too.

cardstack/src/screens/QRScannerScreen/pages/RequestQRCode/RequestQRCodePage.tsx
paulinhapenedo: **New version**


**Old version**

danibonilha (approved):
LGTM

PR #1023, opened 2022-08-29
[CS-4521]: Rewards indicator for claimable amount only and small fixes

danibonilha:
### Description

This PR makes the notification for rewards appear only when it's enough to be claimable, by using too different flags:
`hasRewards` =  any token balance, including crypto dust (used basically inside rewards center)
`hasClaimableRewards` = token balance with enough amount to be claimed 

Also adds a helper to find an item by rewardProgramId since we do that a lot.
While testing I found an issue right after registering, if the user taps on Claim, the app crashes. This was caused by using two different balances, before we had just the mainPoolTokenBalance, and it was always available, to be used on the claimSheet, now `mainPoolTokenBalance` is `fullBalanceToken` and we have `claimableBalanceToken` (which doesn't exists until we have a safeAddress to estimate the gas). So while trying to enter the claim screen, it was expecting the `claimableBalanceToken` to already exist, but it was actually being fetched and the data was undefined. To fix that now we skip the tokenBalancesWithoutDust until we have a rewardSafe, and once we do, we refresh the screen with the new data, we could discuss eventually to add a load on the claimSheet itself or something, but for now it seems to be working good just by refreshing the rewards center. 

Tested claim, withdraw and crypto dust, on both networks and registering on sokol only.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

PS: Registering and claiming flows are on linear ticket

Simulator Screen Recording - iPhone SE (2nd generation) - 2022-08-29 at 14 22 55


paulinhapenedo (approved):
LGTM!

oidoug (approved):
LTGM, nice improvements, no issues found.

PR #1024, opened 2022-08-30
[CS-4439] RTKQ errors parsing

oidoug:
### Description

This PR adds a starting point for parsing errors returned from RTKQuery service calls. It also changes Profile Screen to show the parsed error.

- [x] Completes #(CS-4439)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

Note: I'm trying to test error responses with Proxyman but hub staging is not accepting the intercepted SSL handshake. Will continue in a bit.

### DIFF
--- a/cardstack/src/services/utils/parse-error.ts
+++ b/cardstack/src/services/utils/parse-error.ts
@@ -1,0 +1,48 @@
+import type { SerializedError } from '@reduxjs/toolkit';
+import { FetchBaseQueryError } from '@reduxjs/toolkit/query/react';
+
+export type RTKErrorType = SerializedError | FetchBaseQueryError | undefined;
+
+export interface ConnectionError {
+  error: RTKErrorType;
+  errorMessage: string;
+  type: ErrorType;
+}
+
+export enum ErrorType {
+  connectionError = 'connectionError',
+  parsingError = 'parsingError',
+  unknownError = 'unknownError',
+}
+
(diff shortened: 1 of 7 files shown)


cardstack/src/services/utils/parse-error.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (!error) return undefined;
+
+  // FetchBaseQueryError properties
+  if ('status' in error) {
+    const errorMessage =
+      'error' in error ? error.error : JSON.stringify(error.data);
+
+    return { error, errorMessage, type: ErrorType.connectionError };
+  }
+
```
danibonilha, line 32: I'm a bit confused about what you're trying to achieve here, I thought we were aiming to have a reusable wrapper, not something tight to connecting errors

oidoug: The goal was to have a more reliable error message to display if necessary.
If we add a wrapper to base query, it can only handle `FetchBaseQueryError` because `SerializedError` happens after.

But, TBH, now that I understand this a bit better 😅... I think a generic error parser or wrapper will not work well, because there will be errors coming with `error.data` that we can't just stringify to show to users, It's too dependent on a case-by-case basis. It's not worth the trouble.
WDYT?

oidoug:
I'll close this PR for now to avoid clutter. I'll pick it up again once I return.

PR #1026, opened 2022-08-31
[CS-4519]:  Migrate hub buy PrepaidCard to RTK 

danibonilha:
### Description

This PR migrates the old axios api calls related to wyre, to use the new rtk hub wyre api. The timeout will be removed once we are able to get pending job for prepaid-cards. 

Wyre's api is a bit funky today and there's a Linear ticket to mitigate the issue, I've tested but was only able to get one successful card and didn't record it, I'll try to run another set of tests tomorrow to make sure everything is smooth.


paulinhapenedo (approved):
Tested on my device and worked, even with Wyre acting weird.

PR #1030, opened 2022-09-02
[CS-4545]: Use wyre api to get supported countries

danibonilha:
### Description

We were using a hardcoded countries list that was outdated, this PR uses wyre api to get the updated list of supported countries. I've added a new rtk api to handle this wyre step, is default to use the production non-authenticated wyre api, since the country list is the same between both, in case we want to migrate the other endpoints we would need to create a custom base query to change the baseUrl path. Now it's possible to test from Brazil without any additional changes.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot


paulinhapenedo (approved):
LGTM!
Adding support for `Intl` is great 👏

PR #1027, opened 2022-08-31
[CS-4489] Enhance NavigationStackHeader for Onboarding screens

paulinhapenedo:
### Description
This PR changes the `NavigationStackHeader` component to be used in the new onboarding screens:

1. Added Skip button logic to `NavigationStackHeader`.
2. Changed the chevron icon to use the one without the transparent box.
3. Replaced `InPageHeader` with `NavigationStackHeader` on all of the profile creation screens. For this one, a few adjustments were made on the screens to handle alignment differences. Each screen has its own commit to help the reviewer.
4. Created a helper for the profile creation screens to handle the skip and the SafeArea logic in just one place instead of repeating in all hooks.
5. Created the `BackupStatus` component that will be used in the backup flow.

For the item 5, I have some questions regarding my approach, but I'll reach out to the team on Discord to discuss.

- [x] Completes #CS-4489

### Checklist

- [x] Tested on a small device
- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screen Shot 2022-08-31 at 19 45 37
Screen Shot 2022-08-31 at 19 49 05
Screen Shot 2022-08-31 at 19 45 27
Screen Shot 2022-08-31 at 19 48 54

PR #1031, opened 2022-09-05
[CS-4487] Create PageWithStackHeader component

paulinhapenedo:
### Description
This PR creates a component that could be used in all of the screens of the new onboarding flow. It uses #1027 as base because the enhancement of the `NavigationStackHeader` was done in that branch.

~One important thing to notice is that `ProfileNameScreen` won't be using this component because of the animations that happen on that screen. Since we're moving the NavigationStackHeader to be outside, the animation would have to change a lot. Maybe in the future 🤔~

## Update for future reference
We added this component on the `ProfileNameScreen` in [e2ead00](https://github.com/cardstack/cardwallet/pull/1031/commits/e2ead00dc4a2114f36066f39d89869ffa62d7920), but the animations are not behaving as they should. We will create a ticket to handle that.
 
- [x] Completes #CS-4487

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


cardstack/src/screens/Profile/helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+      paddingBottom: bottom,
+    }),
+    [bottom]
+  );
+
+  return {
+    onSkipPress,
+    containerStyles,
+  };
+};
```
danibonilha: We can drop this

cardstack/src/components/OnboardingPage/useOnboardingPage.ts
```diff
@@ -0,0 +1,60 @@
...
+  const handleSkipPress = useCallback(() => {
+    if (customSkipPress) {
+      customSkipPress();
+
+      return;
+    }
+
+    if (flow === 'profile-creation') {
+      dispatch(skipProfileCreation(true));
+    }
```
danibonilha: nit: No sure it's a good ideia to handle multiple different flows logic on this OnboardingPage, i'd suggest it to have a customSkipPress for this show handled, or a callback on this customSkipPress, to trigger another action after the flow has been skipped

cardstack/src/components/OnboardingPage/useOnboardingPage.ts
```diff
@@ -0,0 +1,60 @@
...
+}: Pick<OnboardingPageProps, 'skipAmount' | 'flow' | 'customSkipPress'>) => {
+  // handles SafeAreaView bottom spacing for consistency
+  const { bottom } = useSafeAreaInsets();
+
+  const containerStyles = useMemo(
+    () => ({
+      paddingBottom: bottom,
+    }),
+    [bottom]
+  );
```
danibonilha: I'm going to make a suggestion on how to use the onboardingPage on the ProfileName one, with that, this style logic could be migrated to the onboardingPage component

cardstack/src/components/OnboardingPage/useOnboardingPage.ts
```diff
@@ -0,0 +1,60 @@
...
+      customSkipPress();
+
+      return;
+    }
+
+    if (flow === 'profile-creation') {
+      dispatch(skipProfileCreation(true));
+    }
+
+    navDispatch(StackActions.pop(skipAmount));
```
danibonilha: I also think we might not need have skipCount anymore,  it's worth to try `navDispatch(StackActions.popToTop());`

cardstack/src/components/OnboardingPage/OnboardingPage.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <Container
+      backgroundColor="backgroundDarkPurple"
+      flex={1}
+      justifyContent="space-between"
+      style={containerStyles}
+      paddingHorizontal={5}
+    >
+      <NavigationStackHeader
+        canGoBack={canGoBack}
+        onSkipPress={handleSkipPress}
```
danibonilha: We doesn't seem to be covering the case to not show skip, the navigationHeader has the condition, but in this case handleSkipPress will never be undefined, so if we want to hide the skip for any of the pages, we won't be able to

cardstack/src/components/OnboardingPage/useOnboardingPage.ts
paulinhapenedo: `popToTop` brings the user back to the Wallet screen. I believe we can use it. Thanks for the suggestion!

cardstack/src/components/OnboardingPage/OnboardingPage.tsx
paulinhapenedo: Created a new prop for the now `PageWithStackHeader` component called `showSkip` that receives a boolean and handles that.

cardstack/src/components/OnboardingPage/useOnboardingPage.ts
paulinhapenedo: Removed the hook for this component in order to leave it as generic as possible. Renamed from `OnboardingPage` to be `PageWithStackHeader` for the same purpose and added a custom callback to the Skip button for extendability.

cardstack/src/components/PageWithStackHeader/PageWithStackHeader.tsx
```diff
@@ -0,0 +1,78 @@
...
+      style={containerStyles}
+      paddingHorizontal={5}
+    >
+      <NavigationStackHeader
+        canGoBack={canGoBack}
+        onSkipPress={showSkip ? handleSkipPress : undefined}
+        backgroundColor="backgroundDarkPurple"
+        marginBottom={4}
+        leftIconProps={leftIconProps}
+        paddingHorizontal={0}
```
danibonilha: Let's add a comment to remind us why we need this, like overrides default padding

cardstack/src/components/PageWithStackHeader/PageWithStackHeader.tsx
```diff
@@ -0,0 +1,78 @@
...
+  skipPressCallback?: () => void;
+  footer?: ReactNode;
+  leftIconProps?: Omit<IconProps, 'name'> & {
+    name?: IconName;
+  };
+}
+
+const PageWithStackHeader = ({
+  canGoBack,
+  showSkip = false,
```
danibonilha: nit: showSkip could default to true, since it's needed on most of the screens

cardstack/src/redux/hooks/usePersistedFlags.ts
```diff
@@ -0,0 +1,16 @@
...
+  const dispatch = useDispatch();
+
+  const triggerSkipProfileCreation = useCallback(() => {
+    dispatch(skipProfileCreation(true));
+  }, [dispatch]);
+
+  return {
+    triggerSkipProfileCreation,
+  };
+};
```
danibonilha: supernit: maybe this could live inside `persistedFlagsSlice`

cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
```diff
@@ -146,16 +146,16 @@ export const ProfileNameScreen = () => {
...
-          outputRange: [0, translateYTo],
-        }),
-      });
-    }
+    transform.push({
+      translateY: animated.interpolate({
+        inputRange: [Animation.keyboardClosing, Animation.keyboardOpening],
+        outputRange: [0, translateYTo],
+      }),
+    });
```
danibonilha: If it's not conditioned by a device anymore, there's no need to push this, you can just add this info on the `transform` const above

cardstack/src/components/PageWithStackHeader/types.ts
```diff
@@ -0,0 +1 @@
+export type OnboardingFlows = 'profile-creation' | 'backup';
```
danibonilha: We can drop this

cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
```diff
@@ -37,7 +36,7 @@ enum Animation {
 
 const layouts = {
   defaultPadding: 5,
-  keyboardVerticalOffset: Device.isIOS ? 15 : 95,
+  keyboardVerticalOffset: Device.isIOS ? screenHeight * 0.12 : -55,
```
danibonilha, line 39: This offset is behaving too different across devices, some doesn't show, and others show the description text cropped, I don't think it's worth though to block this PR over this, so let's just add a ticket to fix this bug and you can assign it to me.
image

image

danibonilha (approved):
Looks good, just final touches, and let's handle the profileName screen on a different ticket, also update the PR title to reflect the new component name

cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
```diff
@@ -187,13 +187,10 @@ export const ProfileNameScreen = () => {
   const flow = useMemo(() => (isUpdating ? 'update' : 'create'), [isUpdating]);
 
   return (
-    <SafeAreaView
-      backgroundColor="backgroundDarkPurple"
-      flex={1}
-      paddingHorizontal={layouts.defaultPadding}
-      justifyContent="space-between"
+    <PageWithStackHeader
+      showSkip
```
danibonilha: This should be conditioned by `!Updating`

cardstack/src/screens/Profile/ProfileNameScreen/ProfileNameScreen.tsx
paulinhapenedo: Ticket CS-4576 created!

PR #1034, opened 2022-09-08
[CS-4542]: Add wyre authentication widget

danibonilha:
### Description

This PR adds the authentication widget on a wyre purchase, unfortunately their website is not completing the authentication, so, I'm in contact with their support to have that figured out, that's why this is step is only enabled on DEV right now. And we still need to handle the page loading.

### Checklist

- [x] Tested on iOS

IMG_7C1072B0D13F-1

### DIFF
--- a/cardstack/src/navigation/routes.ts
+++ b/cardstack/src/navigation/routes.ts
@@ -33,6 +33,7 @@ export const MainRoutes = {
   SUPPORT_AND_FEES: 'SupportAndFeesSheet',
   AVAILABLE_BALANCE_SHEET: 'AvailableBalanceSheet',
   TOKEN_WITH_CHART_SHEET: 'TokenWithChartSheet',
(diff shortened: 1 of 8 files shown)


cardstack/src/navigation/routes.ts
```diff
@@ -33,6 +33,7 @@ export const MainRoutes = {
   SUPPORT_AND_FEES: 'SupportAndFeesSheet',
   AVAILABLE_BALANCE_SHEET: 'AvailableBalanceSheet',
   TOKEN_WITH_CHART_SHEET: 'TokenWithChartSheet',
+  WYRE_AUTH_WIDGET: 'WyreAuthentiationWidget',
```
paulinhapenedo: Small typo `WyreAuthenticationWidget`

paulinhapenedo (approved):
LGTM

PR #1036, opened 2022-09-09
[CS-4490] Create PasswordInput component

paulinhapenedo:
- [x] Completes #CS-4490

### Description
This PR adds an input to handle passwords. Roadmap of features that need to be build:
- [x] Visibility toggle
- [x] Validation rule is always visible and its style changes if what the user typed matches the rule.

In case you want to test locally, here's a code snippet with one of the real validations we're gonna build to be quickly added to any screen:

```jsx
const validationRule = {
    rule: (text: string, minLength = 8) =>
      hasAtLeastOneDigit(text) && matchMinLength(text, minLength),
    message: 'At least 8 characters, with at least 1 number',
};

// on the component render function
<PasswordInput validation={validationRule} />

```

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

**iOS - iPhone SE**

https://user-images.githubusercontent.com/690904/189400258-8c958c63-25ab-498f-9b98-ea5492114ae8.mp4

**Android 10 - Moto G8 Plus**

https://user-images.githubusercontent.com/690904/189400471-048f13e3-673c-4a14-b052-c6671ad5ab43.mp4

**Android 12 - Nexus 4 (small device)**

https://user-images.githubusercontent.com/690904/189400828-68d1641a-c3dc-46a4-a409-467a4351fc6d.mov

**Validation**

https://user-images.githubusercontent.com/690904/189711039-6251f141-862a-4824-9aab-06f949530061.mp4

### DIFF
--- a/cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
+++ b/cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
@@ -59,14 +59,13 @@ const PasswordInput = ({
       >
         <Container flex={1}>
           <Input
-            placeholder="Enter password"
+            {...baseInputProps}
+            {...inputProps}
             placeholderTextColor={palette.blueText}
             secureTextEntry={!isPasswordVisible}
             autoCapitalize="none"
             onChangeText={onChangeText}
             value={password}
-            {...baseInputProps}
-            {...inputProps}
           />
         </Container>
         <Touchable
--- a/cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
+++ b/cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
@@ -1,19 +1,18 @@
-import React, { memo, useMemo } from 'react';
+import React, { memo, useCallback, useMemo, useState } from 'react';
 
 import {
-  BaseInputProps,
   Container,
   ContainerProps,
   Icon,
+  IconName,
   Input,
   InputProps,
   Text,
   Touchable,
+  BaseInputProps,
 } from '@cardstack/components';
 import { palette } from '@cardstack/theme';
 import { hitSlop } from '@cardstack/utils';
-
-import { usePasswordInput } from './usePasswordInput';
 
 const baseInputProps: InputProps = {
   paddingHorizontal: 5,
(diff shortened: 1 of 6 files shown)


cardstack/src/utils/__tests__/validators.test.ts
```diff
@@ -39,4 +43,18 @@ describe('Email validator', () => {
...
   });
+
+  it('should return true for a text with at least one digit', () => {
+    const text = 'test123abc123';
+    const expected = hasAtLeastOneDigit(text);
+
+    expect(expected).toBe(true);
+  });
+
+  it('should return false for a text with at least one digit', () => {
```
danibonilha: **without

cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,96 @@
...
+  containerProps,
+  ...inputProps
+}: Props) => {
+  const {
+    togglePasswordVisibility,
+    iconName,
+    isPasswordVisible,
+    onChangeText,
+    password,
+  } = usePasswordInput();
```
danibonilha: My 2cents on this is that we could handle the hide/show password and icon, within this `PasswordInput`,  and use the `usePasswordInput` hook inside the screens with `PasswordInput`, meaning `PasswordInput` would receive, `onChangeText`, `isValid`, `password` from the props, which would be created on `usePasswordInput`. The main motivation is to have access for example to the `isValid` property,  since we'll have to control the button state on the screen. Giving the current implementation we would need to check twice the rule, once inside the input component and another on the screen, and how would we have access to the password value itself ?

cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <Container flex={1}>
+          <Input
+            placeholder="Enter password"
+            placeholderTextColor={palette.blueText}
+            secureTextEntry={!isPasswordVisible}
+            autoCapitalize="none"
+            onChangeText={onChangeText}
+            value={password}
+            {...baseInputProps}
+            {...inputProps}
```
danibonilha: is it intentional to allow input props to be overwritten  ?

cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,96 @@
...
+        backgroundColor="black"
+        borderRadius={6}
+        flexDirection="row"
+        alignContent="center"
+        justifyContent="space-between"
+        marginBottom={2}
+      >
+        <Container flex={1}>
+          <Input
+            placeholder="Enter password"
```
danibonilha: This should be a prop

danibonilha:
Is there a ticket to handle this case https://reactnative.dev/docs/0.68/textinput#passwordrules-ios ?

paulinhapenedo:
> Is there a ticket to handle this case https://reactnative.dev/docs/0.68/textinput#passwordrules-ios ?

~No, I think we can handle this here since it's part of the password input rule.~
I'm thinking on adding this as a comment on the ticket for when we're building the UI for the backup screen. What do you think? We're gonna handle all the business logic there and this seems like a business logic. Not all password inputs will be `newPassword`.

cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
paulinhapenedo: ~It is. This would handle the case brought up in another comment about the placeholder text. This would allow the developer to override it. Do you think a more strict approach to what can be overwritten is a better approach here?~

Nvm... changed the approach.

cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
paulinhapenedo: Gotcha! Working on it.

paulinhapenedo: Is this the API you had in mind for `usePasswordInput`?

```jsx
const { onChangeText, isValid, password } = usePasswordInput({
    validation: (text: string) =>
      hasAtLeastOneDigit(text) && matchMinLength(text, 8),
  });
  
  // on render
  
  <PasswordInput
    validationMessage="At least 8 characters, with at least 1 number"
    onChangeText={onChangeText}
    isValid={isValid}
    value={password}
  />
```

cc @danibonilha

danibonilha (approved):
LGTM

oidoug (approved):
LGTM 🚀

PR #1038, opened 2022-09-12
[CS-4349] Refactor Remote Config to be stateful

oidoug:
### Description

Adds a redux slice that holds Remote Configs values and exposes init and refresh thunks used by (now renamed) `useRemoteConfigs` hook.

PR also:
- Refactors remote flags service function uses in favor of new stateful hook wherever possible;
- Adds a small disclaimer message to BETA users in settings screen's footer.

- [x] Completes #(CS-4349)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

On linear.


cardstack/src/hooks/remote-configs/useRemoteConfigs.ts
```diff
@@ -0,0 +1,42 @@
...
+import logger from 'logger';
+
+export const useRemoteConfigs = () => {
+  const dispatch = useDispatch();
+  const configs = useSelector((state: AppState) => state.remoteConfigSlice);
+  const [isReady, setIsReady] = useBooleanState();
+
+  const initRemoteConfigs = useCallback(async () => {
+    try {
+      dispatch(initRemoteConfigsThunk());
```
danibonilha: we either need to await and unwrap this thunk, OR handle the isReady state within the remoteConfigSlice , right now we are not actually waiting anything

cardstack/src/hooks/remote-configs/useRemoteConfigs.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const fetchRemoteConfigs = useCallback(
+    () => dispatch(forceFetchRemoteConfigsThunk()),
+    [dispatch]
+  );
+
+  useEffect(() => {
+    initRemoteConfigs();
+  }, [initRemoteConfigs]);
+
```
danibonilha: nit: I believe this fetch could be conditional, and not triggered every time the hook is used, now that is used on multiple places, it would be nice to have some sort of timestamp of the last time it was fetched before trying to re-fetch it, I think firebase handles the cache, but either way, it would avoid a bunch of redux updates

cardstack/src/services/remote-config/remote-config-service.ts
```diff
@@ -15,11 +15,15 @@ export const loadRemoteConfigs = async () => {
 
   await remoteConfig().setDefaults(remoteConfigDefaults);
   await remoteConfig().fetchAndActivate();
+
+  return remoteFlags();
```
danibonilha, line 19: wondering if  `getAll()`  would work, so we wouldn't need to map every single flag as we do today

cardstack/src/hooks/remote-configs/useRemoteConfigs.ts
oidoug: Great suggestion, I've moved isReady to the store too, makes more sense control all states in redux. Also added a isInitializing to avoid init calls while the first call is pending.

cardstack/src/hooks/remote-configs/useRemoteConfigs.ts
oidoug: responded above

cardstack/src/services/remote-config/remote-config-service.ts
oidoug: Hum, it would still require some sort of mapping, bc getAll returns 
```
ConfigValues {
  [key: string]: ConfigValue;
}
```
It wont directly match our RemoteConfigValues type, unless theres some TS wizardry I'm not aware of 😅

danibonilha (approved):
LGTM

PR #1040, opened 2022-09-13
[CS-4527]: Improve authorization navigation check

danibonilha:
### Description

This PR updates the way we check for authorization, by using the redux flag, instead of relying on the navigation state, also handles the inner stack navigation of settings modal.

Please make sure to test it. since it affects all the navigation

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Video is on Linear ticket


paulinhapenedo (approved):
LGTM

PR #1042, opened 2022-09-14
[CS-4613] Handle default exclamation point color in warning icon

paulinhapenedo:
### Description
In https://github.com/cardstack/cardwallet/pull/1031, we introduced a change on the `warning` component. This change didn't take into consideration other instances of where this component was being used. To fix the issue, we added a default color to the exclamation point `fill` property. 

- [x] Completes #CS-4613

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot


danibonilha (approved):
LGTM

PR #1044, opened 2022-09-15
[CS-4494] Seed Phrase Table component

oidoug:
### Description

New `SeedPhraseTable` component, shows in a table a given `seedPhrase`. It handles partial seed phrases, with blurring and revealing, copying the phrase and showing as error.

Important: BlurView is crashing on navigation on Android. There are workarounds I'm trying. To now hold off other cards, for now, on Android it shows an opaque view instead.

- [x] Completes #(CS-4494)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

#### Updated in comment below.

https://user-images.githubusercontent.com/129619/190502339-e2341ead-fc56-4495-9f63-b39dd8e374e7.mp4

https://user-images.githubusercontent.com/129619/190502396-f77254a7-4d04-4174-be09-37c3f067b414.mp4


cardstack/src/components/SeedPhraseTable/SeedPhraseTable.tsx
```diff
@@ -0,0 +1,181 @@
...
+              />
+            ))}
+          </Container>
+        ))}
+      </Container>
+
+      {allowCopy && (
+        <CenteredContainer>
+          <Button
+            iconProps={{ name: 'copy', color: 'white' }}
```
danibonilha: this could live outside the component

cardstack/src/components/SeedPhraseTable/SeedPhraseTable.tsx
```diff
@@ -0,0 +1,181 @@
...
+            marginBottom={7}
+            variant="linkWhite"
+            onPress={() => copyToClipboard(seedPhrase)}
```
danibonilha: this could use an useCallback

cardstack/src/components/SeedPhraseTable/SeedPhraseTable.tsx
```diff
@@ -0,0 +1,181 @@
...
+      backgroundColor="darkBoxBackground"
+      borderRadius={20}
+      borderColor="darkBoxBackground"
+      borderWidth={1}
+    >
+      <Container flexDirection="row" justifyContent="space-between" padding={8}>
+        {wordsColumns.map(column => (
+          <Container width="50%" height={250} justifyContent="space-between">
+            {column.map(item => (
+              <WordItem
```
danibonilha: nit: since there’s hiding and unhiding, which means more re-rendering we could use named functions on these maps, so it won't lose reference, or even better, this seems like a good place to use a Flatlist with columns it might improve readability

cardstack/src/components/SeedPhraseTable/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+  index: number;
+}> => {
+  const words = seedPhrase?.split(' ') || [];
+  const filler = new Array(12 - words.length).fill('');
+
+  return [...words, ...filler].map((word, index) => ({
+    word,
+    index,
+  }));
+};
```
danibonilha, line 10: Can you elaborate on why this is needed and what it does ? If it's just for adding the index as a key, I'm not sure it's relevant  since split already returns the array of words, and you can get it's indexes by iterate over the array on the component

cardstack/src/components/SeedPhraseTable/SeedPhraseTable.tsx
```diff
@@ -0,0 +1,181 @@
...
+  blurred: 1,
+  visible: 0,
+};
+
+// BlurView is crashing and spilling a dark opacity on Android for an unknown reason.
+// TODO: Investigate fix for crash and spillage on Android.
+const BlurViewWrapper = (props: BlurViewProps) =>
+  Device.isIOS ? (
+    <BlurView {...props} />
+  ) : (
```
danibonilha: I'm wondering, if BlueView is crashing on android, and we don't really need to blur other places, maybe it's worth trying to use an image on top of the component instead of adding a new lib, dm if you'd like to talk more about this

cardstack/src/components/SeedPhraseTable/SeedPhraseTable.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+      {!phraseVisible && (
+        <Animated.View style={[styles.blurView, animatedOpacity]}>
+          <BlurViewWrapper
+            style={styles.blurView}
+            reducedTransparencyFallbackColor="darkBoxBackground"
+          />
+          <CenteredContainer width="100%" height="100%">
+            <Button variant="tinyOpacity" onPress={hideOverlay}>
+              {strings.tapToReveal}
```
danibonilha, line 137: it seems the text color for this button is white

cardstack/src/components/SeedPhraseTable/SeedPhraseTable.tsx
oidoug: We currently don't use blur, but there are instances in our designs where there was a desire to use it, I think in some headers and in the wallet screen. Maybe this is an opportunity to learn more about it. Android issues can probably be fixed, but it will be time consuming.

Having a blurred image kinda works, but will not match the revealed phrase precisely because different screen sizes will position the words differently.

I just tested with react-native Image's blurRadius prop, but it doesn't show the background content as blurred, so no good.

I'll ping you to discuss this more.

cardstack/src/components/SeedPhraseTable/SeedPhraseTable.tsx
oidoug: Right, I'll try to improve readability here a bit 😅 

FlatList is not a good option here, because it has a limitation, it only position items in columns in a zig-zag pattern... I'm trying to find if there's something hidden in the docs I'm not aware of. Also, we render all items at once, Flatlist can only optimize rendering when there is reuse of cells, like when scrolling a list.

cardstack/src/components/SeedPhraseTable/utils.ts
oidoug: It's basically what you described, I did it here to make the rendering simpler, but it's confusing.. I'll improve the rendering (memoize it) and this can be simplified.

cardstack/src/components/SeedPhraseTable/SeedPhraseTable.tsx
```diff
@@ -0,0 +1,156 @@
...
+    ({ word, index }: { word: string; index: number }) => (
+      <WordItem word={word} index={index} showAsError={showAsError} />
+    ),
+    [showAsError]
+  );
+
+  const WordsColumns = useMemo(
+    () => (
+      <Container flexDirection="row" justifyContent="space-between" padding={8}>
+        {words.map((column, j) => (
```
danibonilha: nit: words seems more like columns, or wordsColumns, also we can add more meaningful name to `j`, is it columnIndex ?

cardstack/src/components/SeedPhraseTable/SeedPhraseTable.tsx
```diff
@@ -0,0 +1,156 @@
...
+          <Container width="50%" height={250} justifyContent="space-between">
+            {column.map((word, i) =>
+              renderWordItem({ word, index: i + j * column.length })
```
danibonilha: same here on giving context, also is there any specific reason to not return the `WordItem` component directly ?

cardstack/src/components/SeedPhraseTable/components/WordItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+  <Container flexDirection="row" alignItems="flex-end">
+    <Container width={30}>
+      <Text
+        textAlign="right"
+        variant="semibold"
+        size="body"
+        color="white"
+        paddingRight={2}
+      >
+        {index + 1}
```
danibonilha: superNit: instead of handling this as Index, it could be generic to just accept a number, so the parent handles the "logic", since this is not an array

cardstack/src/components/SeedPhraseTable/SeedPhraseTable.tsx
oidoug: No reason, it was there from my tests using Flatlist, using directly now.

ios/Podfile.lock
```diff
@@ -8,6 +8,8 @@ PODS:
     - ExpoModulesCore
   - EXConstants (13.1.1):
     - ExpoModulesCore
+  - EXErrorRecovery (3.1.0):
+    - ExpoModulesCore
```
paulinhapenedo: I think these are libraries that are only added on M1 devices (?). Not sure if we should be adding them to the project too.  I remember us having some on and off talks about this.

oidoug: hum yes, this isn't needed and I'm removing the blur lib. I'll revert this change.

oidoug:
It was a bit more work to make the pre-rendered images match the UI, but it's close enough and looks good. 
I'm removing the blur lib and using same approach for both iOS and Android.

SE:

https://user-images.githubusercontent.com/129619/190728443-146f8d2c-d24c-4b5b-abf4-1ce0e574e792.mp4

14 PRO:

https://user-images.githubusercontent.com/129619/190728487-66814fa5-eb30-4b8f-ad7b-6627a0592418.mp4

Android:

https://user-images.githubusercontent.com/129619/190728646-73179248-4e78-472a-a91b-245a95721fc9.mp4


For iOS 14 MAX the fit isn't great, but passable.

cardstack/src/components/SeedPhraseTable/components/WordItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <Text
+        variant="semibold"
+        size="body"
+        color={showAsError ? 'invalid' : 'teal'}
+      >
+        {word}
+      </Text>
+    ) : (
+      <Container
+        backgroundColor={showAsError ? 'invalid' : 'teal'}
```
paulinhapenedo: I believe the underline will never be red because the validation only occurs after all the words are filled out. We can simplify this by removing the ternary.

cardstack/src/theme/colors.ts
```diff
@@ -111,6 +111,7 @@ export const colors = {
   tertiaryText: palette.blueText,
   bannerText: palette.white,
   orange: palette.orange,
+  darkBoxBackground: palette.blueDarkest,
```
paulinhapenedo: There's an update on `develop` with a more generic color name for the `blueDarkest`. I was using it in other Backup screens too. I think it would be nice to replace this with that.

oidoug: Sure thing

paulinhapenedo (approved):
LGTM!

PR #1045, opened 2022-09-19
[CS-4497] Add KeyboardAvoidingView to BackupCloudPasswordScreen

paulinhapenedo:
### Description
This PR adds a `KeyboardAvoidingView` wrapping the password inputs to avoiding having the inputs behind the keyboard when interacting with it.  

- [x] Completes #CS-4497

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/690904/191057402-1b08b90d-227f-4583-b4d2-186f8c6e465e.mov

https://user-images.githubusercontent.com/690904/191057409-66d91948-2351-468b-9e41-c33515cb50e8.mov

### DIFF
--- a/cardstack/src/screens/Backup/BackupCloudPasswordScreen/BackupCloudPasswordScreen.tsx
+++ b/cardstack/src/screens/Backup/BackupCloudPasswordScreen/BackupCloudPasswordScreen.tsx
@@ -79,12 +79,14 @@ const BackupCloudPasswordScreen = () => {
               isValid={isValid}
               value={password}
               containerProps={styles.passwordInput}
+              returnKeyType="next"
             />
             <PasswordInput
               validationMessage={strings.confirmPasswordValidation}
               onChangeText={onChangeConfirmation}
               isValid={isValidConfirmation}
               value={confirmation}
+              returnKeyType="done"
             />
           </Container>
         </KeyboardAvoidingView>
--- a/cardstack/src/screens/Backup/BackupCloudPasswordScreen/BackupCloudPasswordScreen.tsx
+++ b/cardstack/src/screens/Backup/BackupCloudPasswordScreen/BackupCloudPasswordScreen.tsx
@@ -1,5 +1,6 @@
 import { eq } from 'lodash';
 import React, { memo, useCallback } from 'react';
+import { KeyboardAvoidingView, StyleSheet } from 'react-native';
 
 import {
   Button,
@@ -12,6 +13,7 @@ import {
   Text,
   usePasswordInput,
 } from '@cardstack/components';
+import { Device } from '@cardstack/utils';
 import {
   hasAtLeastOneDigit,
   matchMinLength,
@@ -27,6 +29,12 @@ const layouts = {
     marginBottom: 5,
   },
 };
+
+const styles = StyleSheet.create({
+  keyboardAvoidView: {
+    flex: 1,
+  },
+});
 
 const BackupCloudPasswordScreen = () => {
   const { onChangeText, isValid, password } = usePasswordInput({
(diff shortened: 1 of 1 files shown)


cardstack/src/screens/Backup/BackupCloudPasswordScreen/BackupCloudPasswordScreen.tsx
```diff
@@ -48,35 +56,41 @@ const BackupCloudPasswordScreen = () => {
 
   return (
     <PageWithStackHeader showSkip={false}>
-      <ScrollView
-        flex={1}
-        contentContainerStyle={layouts.scrollView}
-        showsVerticalScrollIndicator={false}
+      <KeyboardAvoidingView
+        style={styles.keyboardAvoidView}
+        contentContainerStyle={{ marginBottom: 200 }}
```
danibonilha: nit: this could be a style

cardstack/src/screens/Backup/BackupCloudPasswordScreen/BackupCloudPasswordScreen.tsx
```diff
@@ -48,35 +56,41 @@ const BackupCloudPasswordScreen = () => {
...
-            validationMessage={strings.confirmPasswordValidation}
-            onChangeText={onChangeConfirmation}
-            isValid={isValidConfirmation}
-            value={confirmation}
-          />
-        </Container>
-      </ScrollView>
+        <ScrollView
+          flex={1}
+          contentContainerStyle={layouts.scrollView}
```
danibonilha: nit: those layouts could be added to  the styleSheet now

danibonilha (approved):
It's good, although as an improvement  we could add some sort of padding, offset, so the user can see the feedback message while typing.

cardstack/src/screens/Backup/BackupCloudPasswordScreen/BackupCloudPasswordScreen.tsx
```diff
@@ -48,35 +56,41 @@ const BackupCloudPasswordScreen = () => {
...
   return (
     <PageWithStackHeader showSkip={false}>
-      <ScrollView
-        flex={1}
-        contentContainerStyle={layouts.scrollView}
-        showsVerticalScrollIndicator={false}
+      <KeyboardAvoidingView
+        style={styles.keyboardAvoidView}
+        contentContainerStyle={{ marginBottom: 200 }}
+        behavior={Device.isIOS ? 'padding' : undefined}
```
oidoug: It's safe to apply "padding" to both ios and android. Some older implementations needed that, but some time ago we changed how android keyboard pushes things to 'adjustPan' (in AndroidManifest) to allow padding on android to work too.

One problem though is that offsets (`keyboardVerticalOffset` prop) don't work equally in android and ios, but you can try play with some different values for each.

oidoug: nit: Not in the scope of the PR, maybe keep in mind for the logic one. I noticed we could have the first input jump to the confirmation with a "next" confirmation button type.

oidoug (approved):
LGTM, just some nits.

cardstack/src/screens/Backup/BackupCloudPasswordScreen/BackupCloudPasswordScreen.tsx
paulinhapenedo: > nit: Not in the scope of the PR, maybe keep in mind for the logic one. I noticed we could have the first input jump to the confirmation with a "next" confirmation button type.

I'll add this to the logic ticket. Thanks for the suggestion!

paulinhapenedo:
@danibonilha and @oidoug new videos with the updated approach to `KeyboardAvoidingView`. Setting `padding` didn't work as expected and I suspect it's because we have a `Scrollview` wrapping the content (at least that's what I understood after reading some SO/Github issues). Using `behavior=padding` works like expected in smaller and taller devices alike.

https://user-images.githubusercontent.com/690904/191141085-e07cec84-70ab-4dc5-9993-fb4be3b92ff8.mov

https://user-images.githubusercontent.com/690904/191141087-93303ed5-8ba9-4629-86ad-0040c6d60a8e.mp4

PR #1046, opened 2022-09-19
[CS-4504] Manual Backup Confirm Seed Phrase Screen UI

oidoug:
### Description

Adds UI (only) for confirming recovery seed phrase screen.

It also includes a new Tag component (`TagCloud` and `TagPressable`) for selecting words - the FloatingTag comp we already have is too different in scope and style, so I decided on create a new one.

- [x] Completes #(CS-4504)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

https://user-images.githubusercontent.com/129619/191114587-25edfc9c-a2ae-49a5-9e96-83396a479ff3.mp4

https://user-images.githubusercontent.com/129619/191114556-6e2f1a17-e3b6-43d4-ba83-7dacdc55fb60.mp4


cardstack/src/components/SeedPhraseTable/strings.ts
```diff
@@ -1,4 +1,5 @@
 export const strings = {
   tapToReveal: 'Tap to Reveal',
   copyToClipboard: 'Copy to clipboard',
+  clean: 'Clean',
```
paulinhapenedo: I believe `Clear` works best in this case because it's a verb whereas `Clean` is an adjective.

cardstack/src/components/Tag/TagCloud.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <Container
+      paddingVertical={5}
+      paddingHorizontal={2}
+      width="100%"
+      flexDirection="row"
+      justifyContent="center"
+      flexWrap="wrap"
+    >
+      {tags.map((tag: string, index: number) => (
+        <Container padding={1}>
```
paulinhapenedo: We can remove this `Container` and move this padding definition to the `TagPressable` component as a margin.

src/components/settings-menu/SettingsSection.js
```diff
@@ -93,7 +93,11 @@ export default function SettingsSection({
   }, [navigate]);
 
   const onPressNewBackup = useCallback(() => {
-    navigate(Routes.BACKUP_EXPLANATION);
+    // navigate(Routes.BACKUP_EXPLANATION);
+    navigate(Routes.BACKUP_SEEDPHRASE_CONFIRMATION, {
+      seedPhrase:
+        'bright sell trunk jalopy donut enemy car invest donut enemy car invest',
+    });
```
paulinhapenedo: Oops, let's revert this.

cardstack/src/components/Tag/TagCloud.tsx
oidoug: Right, I added ContainerProps to TagPressable to do just that but forgot 😅 

Just to note, as a thought: A minor trade off of adding a margin to TagPressable (either in CardPressable or the inner Container) is that it also keeps that extra area interactive for pressing, so the tap can happen outside the tag slightly. Adding another outer container to TagPressable to define this margin and avoid this effect isn't good either because it limits reusability of the component a bit. But it's imperceptible though, so no harm in doing so.

src/components/settings-menu/SettingsSection.js
oidoug: I left it there on purpose, to make it easier for you guys to test it. Removing now :)

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+ * 1. Each element in the array is put in an object, and is given a random sort key
+ * 2. Sort is applied using the random key
+ * 3. Unmap elements to get the original objects
+ * speed: O(N log N), same as quick sort
+ */
+const shuffleWords = (words: Array<string>) =>
+  words
+    .map(value => ({ value, sort: Math.random() }))
+    .sort((a, b) => a.sort - b.sort)
+    .map(({ value }) => value);
```
paulinhapenedo, line 13: What do you think about adding tests to this?

oidoug: Sure, I can add in the upcoming logic PR.

cardstack/src/components/SeedPhraseTable/SeedPhraseTable.tsx
```diff
@@ -23,6 +23,7 @@ interface SeedPhraseTableProps {
   hideOnOpen?: boolean;
   allowCopy?: boolean;
   showAsError?: boolean;
+  onCleanPressed?: () => void;
```
danibonilha: we should reflect the "clear" change on the functions name too

cardstack/src/components/Tag/TagCloud.tsx
```diff
@@ -0,0 +1,39 @@
...
+          selected={selectedTags.includes(index)}
+          onPress={() => onTagSelection?.(index)}
+          margin={1}
+        />
+      ))}
+    </Container>
+  );
+};
+
+export default memo(TagCloud);
```
danibonilha: Not sure if this needs to be it's own component, 🤔  and `TagCloud` gives a diff idea of what it really is.

cardstack/src/components/Tag/TagPressable.tsx
```diff
@@ -0,0 +1,41 @@
...
+import {
+  Container,
+  ContainerProps,
+  Text,
+  CardPressable,
+} from '@cardstack/components';
+
+interface TagComponentProps {
+  text: string;
+  selected: boolean;
```
danibonilha: nit: This prop might make more sense if it was called `disabled`, it's odd that once we tap on the item it takes the "gray" state, that's more like a disabled button, than a selected one

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/BackupSeedPhraseConfirmationScreen.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const shuffledWords = useMemo(() => shuffleSeedPhraseAsArray(seedPhrase), [
+    seedPhrase,
+  ]);
+
+  const selectedSeedPhraseAsString = useMemo(
+    () => selectedWords.map(value => shuffledWords[value]).join(' '),
+    [selectedWords, shuffledWords]
+  );
+
```
danibonilha, line 55: instead of mapping twice, we could track the indexes and the word on different variables, so it's easier to read and understand how the data is constructed

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface BackupManualSeedPhraseConfirmationParams {
+  seedPhrase: string;
+  onConfirm?: () => void;
+}
```
danibonilha: These types will be reused across screens ? If not, maybe it can leave in the same file as the screen

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/BackupSeedPhraseConfirmationScreen.tsx
```diff
@@ -0,0 +1,83 @@
...
+  return (
+    <PageWithStackHeader showSkip={false} leftIconProps={leftIconProps}>
+      <ScrollView flex={1}>
```
danibonilha: we should hide the scrollIndicator

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/BackupSeedPhraseConfirmationScreen.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <PageWithStackHeaderFooter>
+        <CenteredContainer>
+          <Button onPress={onConfirm}>{strings.doneBtn}</Button>
+        </CenteredContainer>
+      </PageWithStackHeaderFooter>
+    </PageWithStackHeader>
+  );
+};
+
+export default memo(BackupSeedPhraseConfirmationScreen);
```
danibonilha, line 88: This is the suggestion with the indexes, other data structure and without creating a tagCloud component:
```tsx
import { useRoute } from '@react-navigation/native';
import React, { memo, useCallback, useState, useMemo, useRef } from 'react';

import {
  Button,
  CenteredContainer,
  Container,
  PageWithStackHeader,
  PageWithStackHeaderFooter,
  Text,
  IconProps,
  ScrollView,
  SeedPhraseTable,
  TagPressable,
import { RouteType } from '@cardstack/navigation/types';

import { strings } from './strings';
import { BackupManualSeedPhraseConfirmationParams } from './types';
import { shuffleSeedPhraseAsArray } from './utils';

const leftIconProps: IconProps = { name: 'x' };

const BackupSeedPhraseConfirmationScreen = () => {
  const { params } = useRoute<
    RouteType<BackupManualSeedPhraseConfirmationParams>
  >();

  const { seedPhrase = '', onConfirm } = params;

  const selectedTagsIndexes = useRef<number[]>([]).current;
  const [selectedWords, setSelectedWords] = useState<string[]>([]);

  const onTagSelected = useCallback(
    (word: string, index: number) => () => {
      setSelectedWords([...selectedWords, word]);
      selectedTagsIndexes.push(index);
    },
    [selectedTagsIndexes, selectedWords]
  );

  const onClearSelection = useCallback(() => {
    selectedTagsIndexes.length = 0;
    setSelectedWords([]);
  }, [selectedTagsIndexes.length]);

  const selectedSeedPhraseAsString = useMemo(() => selectedWords.join(' '), [
    selectedWords,
  ]);

  const shuffledWords = useMemo(() => shuffleSeedPhraseAsArray(seedPhrase), [
    seedPhrase,
  ]);

  const renderTags = useMemo(
    () =>
      shuffledWords.map((tag: string, index: number) => (
        <Container padding={1}>
          <TagPressable
            text={tag}
            disabled={selectedTagsIndexes.includes(index)}
            onPress={onTagSelected(tag, index)}
          />
        </Container>
      )),
    [onTagSelected, selectedTagsIndexes, shuffledWords]
  );

  return (
    <PageWithStackHeader showSkip={false} leftIconProps={leftIconProps}>
      <ScrollView flex={1} showsVerticalScrollIndicator={false}>
        <Container width="90%" paddingBottom={7}>
          <Text variant="pageHeader" paddingBottom={4}>
            {strings.title}
          </Text>
          <Text color="grayText" letterSpacing={0.4}>
            {strings.description}
          </Text>
        </Container>
        <SeedPhraseTable
          seedPhrase={selectedSeedPhraseAsString}
          onClearPress={onClearSelection}
        />
        <Container
          paddingVertical={5}
          paddingHorizontal={2}
          flexDirection="row"
          justifyContent="center"
          flexWrap="wrap"
        >
          {renderTags}
        </Container>
      </ScrollView>
      <PageWithStackHeaderFooter>
        <CenteredContainer>
          <Button onPress={onConfirm}>{strings.doneBtn}</Button>
        </CenteredContainer>
      </PageWithStackHeaderFooter>
    </PageWithStackHeader>
  );
};

export default memo(BackupSeedPhraseConfirmationScreen);

```

cardstack/src/components/Tag/TagCloud.tsx
oidoug: I prefer to leave it extracted, but I'll move it to be a component specific to this screen. What about renaming `TagPressable` to `WordPressable` and `TagCloud` to `WordPressableGroup`? Any other suggestion?

paulinhapenedo: Makes more sense, I guess. TagCloud reminded me of Blogspot/Blogger early 2000 haha 😅

cardstack/src/components/Tag/TagPressable.tsx
oidoug: Agreed, for this case it makes sense be marked as selected but in a reuse later it may not. I'll change it.

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/types.ts
oidoug: No, also the onConfirm don't make much sense here as pointed by Paula, removing it too.

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/BackupSeedPhraseConfirmationScreen.tsx
oidoug: Will do, but something for us to consider: 

We are removing the indicator because of the padding (`paddingHorizontal={5}`) is in the scrollview parent already (`PageWithStackHeader`).

We could have a PageWithStackHeaderContent or ...ScrollableContent comp. or similar that would contain the paddings, so we could compose the Page, with correct padding, but in the right components.

paulinhapenedo: The `PageWithStackHeader` is supposed to solve alignment issues between onboarding screens. Creating new `PageWithStackHeaderSomething` components adds complexity to something that: 1. Isn't a standard practice in the design team and 2. should be easy to use, like a page template. If we keep going that way, `PageWithStackHeaderSomething` will be just a `Container` with a set of style rules. If that's the case, we could just create a `layouts.ts` file and put an object with these rules there.

To me, there's no value in creating components if the only thing in them is a `Container` or a `Scrollview` with styles applied.

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/BackupSeedPhraseConfirmationScreen.tsx
oidoug: Not sure Dani, I think it's more confusing to have two structures to convey the same data just to avoid an extra occasional map.

I'm renaming selectedWords to `selectedWordsIndexes` to make it clearer, let me know if that's good enough or if you have a strong opinion on your approach, I can go with that too, np.

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/utils.ts
oidoug: I end up adding a couple of test in this PR

cardstack/src/components/WordPressable/WordPressable.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface TagComponentProps {
+  text: string;
+  disabled: boolean;
+  onPress: () => void;
+}
+
+export const WordPressable = memo(
+  ({
+    text,
+    disabled: selected,
```
danibonilha: any specific reason to keep as selected ? Or was it the auto refactor ?

cardstack/src/components/WordPressable/WordPressable.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { memo } from 'react';
+
+import {
+  Container,
+  ContainerProps,
+  Text,
+  CardPressable,
+} from '@cardstack/components';
+
+interface TagComponentProps {
```
danibonilha: We could call this just Props, or reflect the new name

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/components/WordPressableGroup.tsx
```diff
@@ -0,0 +1,31 @@
...
+    onWordPressed,
+  }: WordPressableGroupProps) => {
+    return (
+      <Container flexDirection="row" justifyContent="center" flexWrap="wrap">
+        {words.map((tag: string, index: number) => (
+          <Container padding={1}>
+            <WordPressable
+              text={tag}
+              disabled={selectedWordsIndexes.includes(index)}
+              onPress={() => onWordPressed?.(index)}
```
danibonilha, line 20: nit: this could use useCallback

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/components/WordPressableGroup.tsx
```diff
@@ -0,0 +1,31 @@
...
+            />
+          </Container>
+        ))}
+      </Container>
```
danibonilha, line 24: nit: still don't see much reuse of this  to make it a component, but I understand if you'd like to keep as is.

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/BackupSeedPhraseConfirmationScreen.tsx
```diff
@@ -0,0 +1,90 @@
...
+  ScrollView,
+  SeedPhraseTable,
+} from '@cardstack/components';
+import { RouteType } from '@cardstack/navigation/types';
+
+import { WordPressableGroup } from './components/WordPressableGroup';
+import { strings } from './strings';
+import { shuffleSeedPhraseAsArray } from './utils';
+
+export interface BackupManualSeedPhraseConfirmationParams {
```
danibonilha: nit: I guess we could drop the export word and rename this to NavParams, since it's scoped to this screen

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/__tests__/utils.test.ts
```diff
@@ -0,0 +1,18 @@
+import { splitSeedPhrase, shuffleSeedPhraseAsArray } from '../utils';
+
+const mockSeedPhrase = 'first second third fourth';
+
+describe('Shuffle seed phrase funtions', () => {
```
paulinhapenedo: Typo `functions`

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/__tests__/utils.test.ts
```diff
@@ -0,0 +1,18 @@
...
+  it('should split phrase into array with correct words and order', () => {
+    const splittedPhrase = splitSeedPhrase(mockSeedPhrase);
+
+    expect(splittedPhrase).toEqual(['first', 'second', 'third', 'fourth']);
+  });
+
+  it('should contain the same amount of words but not be the same as original phrase', () => {
```
paulinhapenedo: The phrasing here is a bit confusing. What do you think about: "should scramble the words in the seed phrase" ?

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/__tests__/utils.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    const shuffledPhrase = shuffleSeedPhraseAsArray(mockSeedPhrase);
+
+    expect(shuffledPhrase.length).toEqual(mockSeedPhrase.split(' ').length);
```
paulinhapenedo: Since you're exporting the `splitSeedPhrase`, why not use it here instead?

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/BackupSeedPhraseConfirmationScreen.tsx
danibonilha: Renaming it definitely makes it better, I don't think it matters much or one approach is way better than the other, it's just a different way of thinking, so it's ok if you want to keep as is.
I just want to explain why I suggested two data structures, rather than just to avoid another map, it was more related to the way things are rendered and my line of thought.

- Why do we need the indexes of words ? To track if one of the shuffledWords has been selected or not. 
- What do we render on the screen ? The words itself.

Giving this two statements, it makes sense for the words to live in the state, bc it's exactly what we see when we interact with a word, and having a ref for the indexes is enough to check if the word has been selected or not, and shuffledWords exists just to render the "WordGroup".

On the current approach, the indexes are tracked on the state, then we need to think for a while to understand the relation between selectedWordsIndexes and shuffledWords, bc it's not straightforward that the selectedIndexes are actually from the shuffledWords, and it needs to be mapped back into an array of strings, then concat the words in order to form the potential seed phrase string.

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/__tests__/utils.test.ts
danibonilha: typo: `functions`

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/__tests__/utils.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    expect(shuffledPhrase.join(' ')).not.toEqual(mockSeedPhrase);
```
danibonilha: nit: We could use more declarative expectations instead of using functions to modify the mockSeedPhrase, since thats also relies on functions.

so `expect(shuffledPhrase.length).toEqual(4)`
same for the second expect

danibonilha (approved):

cardstack/src/components/WordPressable/WordPressable.tsx
oidoug: no, the automatic refactor does that..

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/components/WordPressableGroup.tsx
oidoug: But then wouldn't I need to pass the index to the callback?

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/components/WordPressableGroup.tsx
oidoug: Maybe not reuse, but I think it's better for legibility.

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/__tests__/utils.test.ts
oidoug: Nice suggestion 👍

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/BackupSeedPhraseConfirmationScreen.tsx
oidoug: I appreciate the explanation, I understand your perspective better now. I'll commit to this change in the logic PR.

PR #1050, opened 2022-09-20
[CS-4496] Backup recovery phrase screen

paulinhapenedo:
### Description
This PR adds the recovery phrase screen. There's one thing worth mentioning here that's the change on the `PageWithStackHeader`. We need to be able to pass the `BackupStatus` component to the `NavigationStackHeader` via children.

- [x] Completes #CS-4496

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

**iOS - Tall**
Screenshot

**iOS - Small**

https://user-images.githubusercontent.com/690904/191345523-e890a954-b2de-4ff6-a665-d6e843e090ed.mov

**Android - Tall**
Screenshot

**Android - Small**

https://user-images.githubusercontent.com/690904/191345427-0416999c-1f27-4bbc-8f41-3e316e598406.mov

### DIFF
--- a/cardstack/src/screens/Backup/BackupRecoveryPhraseScreen/useBackupRecoveryPhraseScreen.ts
+++ b/cardstack/src/screens/Backup/BackupRecoveryPhraseScreen/useBackupRecoveryPhraseScreen.ts
@@ -1,0 +1,27 @@
+import { useNavigation } from '@react-navigation/native';
+import { useCallback } from 'react';
+
+import { Routes } from '@cardstack/navigation';
+
+import { useWallets } from '@rainbow-me/hooks';
+
+export const useBackupRecoveryPhraseScreen = () => {
+  const { navigate } = useNavigation();
+  const { wallets } = useWallets();
+
+  const handleCloudBackupOnPress = useCallback(
+    () => navigate(Routes.BACKUP_CLOUD_PASSWORD),
+    [navigate]
+  );
+
+  const handleManualBackupOnPress = useCallback(() => {
+    // TBD
+  }, []);
+
+  console.log('[LOG] wallets', wallets);
+
+  return {
+    handleCloudBackupOnPress,
+    handleManualBackupOnPress,
+  };
+};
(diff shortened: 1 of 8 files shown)


cardstack/src/screens/Backup/BackupRecoveryPhraseScreen/useBackupRecoveryPhraseScreen.ts
```diff
@@ -0,0 +1,39 @@
...
+import { useCallback, useEffect, useState } from 'react';
+
+import { getSeedPhrase } from '@cardstack/models/secure-storage';
+import { Routes } from '@cardstack/navigation';
+
+import { useWallets } from '@rainbow-me/hooks';
+
+export const useBackupRecoveryPhraseScreen = () => {
+  const { navigate } = useNavigation();
+  const { latestBackup, selectedWallet } = useWallets();
```
danibonilha: I think we need to use a different approach to check if the user has backed up, this latestBackup if I'm not mistaken is only for backupFiles from icloud. The wallet has a backup flag, `backedUp`,   if this is for the UI I'd suggest to create just a flag locally, and later we can search for this wallet flag. Let me know what you prefer.

cardstack/src/screens/Backup/BackupRecoveryPhraseScreen/useBackupRecoveryPhraseScreen.ts
```diff
@@ -0,0 +1,39 @@
...
+  const [seedPhrase, setSeedPhrase] = useState('');
+
+  const loadSeedPhrase = useCallback(async () => {
+    const seedphrase = await getSeedPhrase(selectedWallet.id);
+
+    setSeedPhrase(seedphrase);
```
danibonilha: nit: maybe we should add a try catch block here

danibonilha (approved):
Looks good, just the latestBackup flag issue, but it can be handled on the logic PR

cardstack/src/screens/Backup/BackupRecoveryPhraseScreen/useBackupRecoveryPhraseScreen.ts
paulinhapenedo: There's a `backedUp` flag on the `selectedWallet` that could handle that case. I totally missed that. Thanks for the heads-up.

PR #1051, opened 2022-09-21
[CS-4504] Manual Backup screen

paulinhapenedo:
### Description
This PR adds the Manual Backup screen to the flow. It's a relatively straightforward screen.

- [x] Completes #CS-4504

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshots in Linear.

### DIFF
--- a/cardstack/src/screens/Backup/BackupManualScreen/BackupManualScreen.tsx
+++ b/cardstack/src/screens/Backup/BackupManualScreen/BackupManualScreen.tsx
@@ -1,0 +1,58 @@
+import { useRoute } from '@react-navigation/native';
+import React, { memo, useCallback } from 'react';
+import { StyleSheet } from 'react-native';
+
+import {
+  Button,
+  Container,
+  PageWithStackHeader,
+  PageWithStackHeaderFooter,
+  ScrollView,
+  SeedPhraseTable,
+  Text,
+} from '@cardstack/components';
+import { RouteType } from '@cardstack/navigation/types';
+
+import { strings } from './strings';
+
+const style = StyleSheet.create({
+  scrollview: {
+    paddingBottom: 30,
+  },
+});
+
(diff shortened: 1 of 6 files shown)


cardstack/src/screens/Backup/BackupManualScreen/BackupManualScreen.tsx
```diff
@@ -0,0 +1,58 @@
...
+  SeedPhraseTable,
+  Text,
+} from '@cardstack/components';
+import { RouteType } from '@cardstack/navigation/types';
+
+import { strings } from './strings';
+
+const style = StyleSheet.create({
+  scrollview: {
+    paddingBottom: 30,
```
oidoug, line 20: I wonder if having a page's content composable component wouldn't help with this scroll paddings as well.

I'll probably put a PR with a suggestion.

oidoug (approved):
LGTM, besides the minor new conflicts 🚀

cardstack/src/screens/Backup/BackupManualScreen/BackupManualScreen.tsx
paulinhapenedo: I don't think adding more components is the solution here. I really don't see the the value of a component that only has `<ScrollView contentContainerStyle={{ paddingBottom: 30 }}>`. If we want to abstract this padding, we could use a `layouts.ts` file in the root of the `Backup` folder and keep it there. 

Not every piece of code needs to be a component.

oidoug: Sorry if I wasn't clear, but I was also referring to the problem we have with the scrollview being horizontally padded and the need of hiding the scroll indicator because of that. We can discuss it more offline.

PR #1053, opened 2022-09-26
[CS-4506] Backup Success Screen UI

oidoug:
### Description

This adds a new screen for showing the backup success state.

- [x] Completes #(CS-4506)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot

### DIFF
--- a/cardstack/src/navigation/routes.ts
+++ b/cardstack/src/navigation/routes.ts
@@ -77,6 +77,7 @@ const BackupRoutes = {
   BACKUP_RECOVERY_PHRASE: 'BackupRecoveryPhrase',
   BACKUP_SEEDPHRASE_CONFIRMATION: 'BackupSeedPhraseConfirmationScreen',
   BACKUP_MANUAL_BACKUP: 'BackupManualBackup',
+  BACKUP_SUCCESS: 'BackupSuccessScreen',
 };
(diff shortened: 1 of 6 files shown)


danibonilha (commented):
This screen has minimum differences when compared with `BackupRecoveryPhraseScreen ` is there a specific reason to not make it more generic and share the screen for both cases ? Maybe we'll need to change the name, but I was thinking the entry point to `backup` would be the same regardless of the status, and based on the status we would render the options, rather than check for backup to conditionally navigate the user. I'm not very close to this project though, so excuse me if I said something that does not make sense, please let me know.

oidoug:
No, you are right. I was going to bring that up offline. It's very possible to use a shared template.
I'll convert this to a draft and make an attempt on extracting the UI from the recovery screen tomorrow.

oidoug:
@danibonilha on a second thought, I believe it's cleaner if we maintain these two screen distinct.

They pretty much look the same, but there are small differences and the majority of the duped code is already extracted in reusable components.

Let's discuss this more offline.

oidoug:
We changed course to improve the backup flow and this screen is being replaced by a more generic Recovery Screen.

PR #1054, opened 2022-09-27
[CS-4498] Add logic to backup to cloud password screen

paulinhapenedo:
### Description
Opening this PR as a draft because there are a lot of changes already made and would be nice to have an input from the team. There's a checklist on the Linear ticket with the everything that was done and things that are still missing.

- [x] Completes #CS-4498

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


cardstack/src/models/backup.ts
```diff
@@ -4,34 +4,16 @@ import { captureException } from '@sentry/react-native';
...
-  wallets: AllRainbowWallets;
-}
-
-export interface ICloudBackupData {
-  createdAt: number;
-  updatedAt: number;
-  secrets?: { [key: string]: string };
-  seedPhrase?: string;
-}
+export const cloudBackupPasswordMinLength = 8;
```
danibonilha, line 15: we could use this value on `passwordValidation` string

cardstack/src/screens/Backup/BackupCloudPasswordScreen/useBackupCloudPasswordScreen.ts
```diff
@@ -0,0 +1,57 @@
...
+  hasAtLeastOneDigit,
+  matchMinLength,
+} from '@cardstack/utils/validators';
+
+export const useBackupCloudPasswordScreen = () => {
+  const [checked, setChecked] = useState(false);
+
+  const { onChangeText, isValid, password } = usePasswordInput({
+    validation: (text: string) =>
+      hasAtLeastOneDigit(text) && matchMinLength(text, 8),
```
danibonilha: we should use `cloudBackupPasswordMinLength` here

cardstack/src/screens/Backup/BackupCloudPasswordScreen/useBackupCloudPasswordScreen.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  const {
+    onChangeText: onChangeConfirmation,
+    isValid: isValidConfirmation,
+    password: confirmation,
+  } = usePasswordInput({
+    validation: (text: string) => matchMinLength(text, 1) && eq(password, text),
```
danibonilha: nit: `!!text &&  text === password`

cardstack/src/screens/Backup/BackupCloudPasswordScreen/useBackupCloudPasswordScreen.ts
```diff
@@ -0,0 +1,57 @@
...
+  const onCheckboxPress = useCallback(() => {
+    setChecked(!checked);
+  }, [checked]);
+
+  const isSubmitDisabled = useMemo(
+    () => !isValid || !isValidConfirmation || !checked,
+    [isValid, isValidConfirmation, checked]
+  );
+
+  const passwordRef = useRef<TextInput>();
```
danibonilha: Do we need this `passwordRef` ?

cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
```diff
@@ -19,23 +20,32 @@ const baseInputProps: InputProps = {
   paddingVertical: 3,
   color: 'white',
   selectionColor: 'teal',
+  autoCompleteType: 'password',
+  blurOnSubmit: false,
+  secureTextEntry: true,
```
danibonilha: isn't this prop duped ?

cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
```diff
@@ -19,23 +20,32 @@ const baseInputProps: InputProps = {
...
+  selectTextOnFocus: true,
+  textContentType: 'password',
+  autoCapitalize: 'none',
+  placeholderTextColor: palette.blueText,
+  passwordRules: iCloudPasswordRules,
 };
```
danibonilha: I'm a little bit confused why do we have those input properties on a const, instead of defining it on the component. It'd make sense if it was reused though, but I don't think it's the case here

cardstack/src/models/rn-cloud.ts
```diff
@@ -0,0 +1,239 @@
...
+  INTEGRITY_CHECK_FAILED: 'Backup integrity check failed',
+  KEYCHAIN_ACCESS_ERROR: `Couldn't read items from keychain`,
+  SPECIFIC_BACKUP_NOT_FOUND: 'No backup found with that name',
+  UKNOWN_ERROR: 'Unknown Error',
+  WALLET_BACKUP_STATUS_UPDATE_FAILED: 'Update wallet backup status failed',
+  USER_CANCELED_DRIVE_API_AUTH: 'User canceled Google Drive OAuth',
+};
+
+/**
+ * Goes throught the cloud backup directory and deletes all files.
```
danibonilha: typo: `through`

cardstack/src/models/rn-cloud.ts
```diff
@@ -0,0 +1,239 @@
...
+  Device.isIOS ? RNCloudFs.syncCloud() : true;
+
+/**
+ * isCloudBackupAvailable is an iOS-only method that returns a boolean value.
+ * When on an Android device, it will always return true.
+ * @returns boolean
+ */
+export const isCloudBackupAvailable = async (): Promise<boolean> =>
+  Device.isIOS ? RNCloudFs.isAvailable() : true;
+
```
danibonilha: I think we could improve this by calling `syncCloudIOS` and `isIOSCloudBackupAvailable`, then there's no need to check the device here, and wherever we use this function we need to check the device before, which I think we are already doing in most of the places, I'd argue that even though we will need to check before usage, I believe it makes things less "magical" and will avoid confusion on long term

cardstack/src/models/rn-cloud.ts
```diff
@@ -0,0 +1,239 @@
...
+    captureException(error);
+    throw error;
+  }
+
+  return document;
+};
+
+/**
+ * Searchs the cloud backup directory to find a file that matches the filename param.
```
danibonilha: typo: `Searches`

cardstack/src/models/rn-cloud.ts
```diff
@@ -0,0 +1,239 @@
...
+          encryptedData
+        );
+
+        if (backedUpDataStringified) {
+          const backedUpData = JSON.parse(backedUpDataStringified);
+          return backedUpData;
+        }
+
+        throw new Error(CLOUD_BACKUP_ERRORS.ERROR_DECRYPTING_DATA);
+      } catch (error) {
```
danibonilha: I'm not sure if `decrypt` forwards the error message, maybe it makes more sense to add the log there and leave this function as it was before, let's talk later

cardstack/src/models/rn-cloud.ts
```diff
@@ -0,0 +1,239 @@
...
+
+    return filename;
+  } catch (e) {
+    if (e.code === 'canceled') {
+      throw new Error(CLOUD_BACKUP_ERRORS.USER_CANCELED_DRIVE_API_AUTH);
+    }
+
+    logger.sentry('[BACKUP] Error during encryptAndSaveDataToCloud', e);
+    captureException(e);
+  }
```
danibonilha, line 218: I think we are missing the general error, we should also consider how we are tracking the errors and what to show, I think it's  handled on useWalletCloud

cardstack/src/models/rn-cloud.ts
```diff
@@ -0,0 +1,239 @@
...
+};
+
+/**
+ * Saves or updates the UserData.json file in the cloud directory.
+ * It calls `encryptAndSaveDataToCloud`.
+ * @returns the backup filename
+ */
+export const backupUserDataIntoCloud = async (data: BackupUserData) =>
+  await encryptAndSaveDataToCloud(data, CARDWALLET_MASTER_KEY, USERDATA_FILE);
```
danibonilha: Why there's this await here ? we could just return it

cardstack/src/models/rn-cloud.ts
```diff
@@ -0,0 +1,239 @@
...
+
+/**
+ * Gets the UserData.json file in the cloud directory.
+ * It calls `getDataFromCloud`.
+ * @returns ICloudBackupData
+ */
+export const fetchUserDataFromCloud = async () =>
+  await getDataFromCloud(CARDWALLET_MASTER_KEY, USERDATA_FILE);
```
danibonilha: ditto

cardstack/src/components/Input/PasswordInput/PasswordInput.tsx
paulinhapenedo: It's because they are the PasswordInput's default props. The only one I think we could remove from here is `passwordRules`, but all the others are as default as the padding settings.

cardstack/src/hooks/backup/useWalletCloudBackup.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export const useWalletCloudBackup = () => {
+  const { showLoadingOverlay, dismissLoadingOverlay } = useLoadingOverlay();
+  const { selectedWallet } = useWallets();
+  const dispatch = useDispatch();
+
+  const backupToCloud = useCallback(
+    async ({ password }: BackupToCloud) => {
+      try {
+        const isAvailable = await isIOSCloudBackupAvailable();
```
oidoug: I'm probably missing something, but wont this call on Android return undefined? 

Shouldn't you skip this check on android, as it was doing before?

cardstack/src/models/backup.ts
```diff
@@ -4,34 +4,16 @@ import { captureException } from '@sentry/react-native';
...
+export const iCloudPasswordRules = `minlength: ${cloudBackupPasswordMinLength}; required: digit;`;
```
oidoug, line 16: Although we're using icloud's minimum password requirements, apparently it is the same rule we set for both ios and android, right?

The naming here suggests only icloud will have password rules. I recommend leaving the name generic and adding a comment that the rules comes from icloud req.

cardstack/src/models/backup.ts
```diff
@@ -153,3 +135,8 @@ export async function findAndParseOldSeed(
     captureException(e);
   }
 }
+
+export const isCloudBackupPasswordValid = (password: string) =>
```
oidoug, line 140: Same as above if applies.

cardstack/src/models/rn-cloud.ts
```diff
@@ -25,112 +25,44 @@ export const CLOUD_BACKUP_ERRORS = {
...
-}
-
-function getICloudDocument(filename: string): Promise<any> {
-  return RNCloudFs.getIcloudDocument(filename);
-}
-
-function getGoogleDriveDocument(id: string): Promise<any> {
-  return RNCloudFs.getGoogleDriveDocument(id);
-}
+};
```
oidoug, line 50: oh god, git did an awful job matching lines here, just because of an added try/catch 😂

cardstack/src/models/rn-cloud.ts
```diff
@@ -25,112 +25,44 @@ export const CLOUD_BACKUP_ERRORS = {
...
-    return RNCloudFs.syncCloud();
-  }
-  return true;
-}
+/**
+ * syncCloudIOS is an iOS-only method that returns a boolean value.
+ * There's no sync method on Android.
+ * @returns boolean
+ */
+export const syncCloudIOS = async (): Promise<boolean> => RNCloudFs.syncCloud();
```
oidoug, line 57: If theres no syncCloud method on android, shouldn't you be handling an undefined reference call here? Maybe with `syncCloud?.() ?? true` or similar?

oidoug (commented):
Good job so far!

cardstack/src/models/rn-cloud.ts
paulinhapenedo: It was the first approach, but as Dani mentioned [here](https://github.com/cardstack/cardwallet/pull/1054#discussion_r981750443) `syncCloudIOS` is only called after we check if the device is an iOS (implementation [here](https://github.com/cardstack/cardwallet/blob/faef8cc748944650e50af2c6042bc49035d137a7/cardstack/src/screens/WelcomeScreen/hooks.tsx)).

cardstack/src/models/backup.ts
paulinhapenedo: But `passwordRules` is an iOS-only prop, isn't? That itself makes it clear that this rule will only be applied for iOS devices.

cardstack/src/models/backup.ts
paulinhapenedo: I think you've misread this one 🙈 ... it says `isCloud`, not `iCloud`.

cardstack/src/hooks/backup/useWalletCloudBackup.ts
paulinhapenedo: My bad, done in https://github.com/cardstack/cardwallet/pull/1054/commits/20a35d8d7a97a604261291ecc00964eee3766f01

cardstack/src/models/backup.ts
oidoug: Right, I didn't remember that detail. Since the input prop name doesn't make that obvious, maybe adding a comment above `iCloudPasswordRules` stating that helps.

cardstack/src/models/backup.ts
oidoug: hehe yeah, sorry 🙊

cardstack/src/models/rn-cloud.ts
oidoug: I see, no issues then.

oidoug: I see, no issues then.

cardstack/src/hooks/backup/useWalletCloudBackup.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        if (!isAvailable) {
+          Alert(iCloudAlertConfig);
+
+          return;
+        }
```
oidoug: nit: Another approach here would be to wrap this block to be specific to iOS, like:
```
if (Device.isIOS) {
    const isAvailable = await isIOSCloudBackupAvailable();
    if (!isAvailable) {
        Alert(iCloudAlertConfig);
        return;
    }
}
```

To me that will make it clearer overall.

cardstack/src/models/rn-cloud.ts
```diff
@@ -176,65 +108,129 @@ export async function getDataFromCloud(
...
-  if (Device.isIOS) {
-    return RNCloudFs.isAvailable();
+    if (!exists) {
+      logger.sentry(`[BACKUP] Backup doesn't exist after completion`);
+      const error = new Error(CLOUD_BACKUP_ERRORS.INTEGRITY_CHECK_FAILED);
+      captureException(error);
+      throw error;
+    }
+
+    await RNFS.unlink(path);
```
oidoug, line 207: I noticed that the path is not removed if we get an exception above. Wouldn't this better be in a finally statement?

oidoug (commented):
Added some minor comments, but it is looking good 👏 

I'll leave without an approval for now so we can wait for Dani's review.

oidoug (commented):
Added some minor comments, but it is looking good 👏 

I'll leave without an approval for now so we can wait for Dani's review.

cardstack/src/models/rn-cloud.ts
paulinhapenedo: If we were to move this into a finally statement, we would have to wrap it around another try/catch because this is an async function. Any error here wouldn't be handled, and being a native implementation, I fear that would case the app to break. What do you think?

oidoug: Well this was there before and it doesn't appear we had issues so far. Maybe leave as is to decrease risk for now.

About it being prone to crashing, I would assume that could only happen if unlink is called on an undefined RNFS or undefined path objects, so it's possible to mitigate. Adding another try/catch would make sense too.

cardstack/src/hooks/backup/utils.ts
```diff
@@ -0,0 +1,11 @@
...
+
+const userFriendlyErrorMsg = {
+  [CLOUD_BACKUP_ERRORS.KEYCHAIN_ACCESS_ERROR]:
+    'You need to authenticate to proceed with the Backup process',
+  [CLOUD_BACKUP_ERRORS.USER_CANCELED_DRIVE_API_AUTH]: `Not authorized to interact with Google Drive.`,
+};
+
+export const getFriendlyErrorMessage = (
+  errorMessage = CLOUD_BACKUP_ERRORS.WALLET_BACKUP_STATUS_UPDATE_FAILED
+) => userFriendlyErrorMsg[errorMessage];
```
danibonilha, line 11: The default error here is a key that is not mapped on `userFriendlyErrorMsg`, also in case some other error is thrown the app might crash, we should add a default error message, in case the errorMessage is not a valid key. Let's adda ticket to handle that

cardstack/src/models/backup.ts
```diff
@@ -153,3 +135,8 @@ export async function findAndParseOldSeed(
...
+  password &&
+  password !== '' &&
+  password.length >= cloudBackupPasswordMinLength;
```
danibonilha, line 143: This is just for the old implementation right ? Let's not forget to remove it once we migrate to the new backup flow

src/screens/RestoreSheet.js
```diff
@@ -33,14 +33,14 @@ export default function RestoreSheet() {
...
-        } else {
-          setNoBackupsFound(true);
         }
       }
 
       // Animate transforming into backup sheet
       layoutEasingAnimation();
       setParams({ step: WalletBackupStepTypes.cloud });
+    } catch {
+      setNoBackupsFound(true);
```
oidoug, line 45: Good catch, but isn't setting this if data is possible null? JSON.parse can return null without exceptions.

src/hooks/useWalletCloudBackup.ts
```diff
@@ -52,7 +52,7 @@ export default function useWalletCloudBackup() {
       password,
       walletId,
     }) => {
-      const isAvailable = await isCloudBackupAvailable();
+      const isAvailable = await isIOSCloudBackupAvailable();
```
danibonilha: The ios check here is missing, it will probably crash android

src/screens/RestoreSheet.js
danibonilha, line 45: we may need this on the catch block, but I don't think we should remove from the else block, since it might return successfully but data can be undefined which is actually the most common case

oidoug (commented):
Added two more comments. Finishing tests before approval.

danibonilha (approved):

oidoug (approved):

PR #1055, opened 2022-09-28
[CS-4651] Adjustments to manual backup flow

paulinhapenedo:
### Description
- [x] Completes #CS-4651

This PR has the refactoring of the manual backup flow to be the priority when on the onboarding flow. Screenshot on Linear.

Meaningful changes:
- ~Seed phrase loading inside the `Settings` component since we need that to render the SeedPhraseTable on the `ManualBackupScreen`.~
- Seed phrase is accessible without promises via `useWallets` hook.
- Added new button at the Confirmation screen to redirect the user to the cloud backup in case they want while keeping the "Done" button if they want to skip this process.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/src/components/settings-menu/SettingsSection.js
+++ b/src/components/settings-menu/SettingsSection.js
@@ -13,6 +13,7 @@ import {
 } from '../list';
 import { CenteredContainer, Icon, ScrollView } from '@cardstack/components';
 import { SettingsExternalURLs } from '@cardstack/constants';
+import { getSeedPhrase } from '@cardstack/models/secure-storage';
 import { Routes } from '@cardstack/navigation';
 import networkInfo from '@rainbow-me/helpers/networkInfo';
 import {
@@ -20,6 +21,7 @@ import {
   useSendFeedback,
   useWallets,
 } from '@rainbow-me/hooks';
+import { logger } from '@rainbow-me/utils';
 
 export default function SettingsSection({
   onPressDev,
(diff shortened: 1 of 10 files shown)


src/components/settings-menu/SettingsSection.js
```diff
@@ -59,8 +61,23 @@ export default function SettingsSection({
...
+
+      setSeedPhrase(seedphrase);
+    } catch (error) {
+      logger.log('Error getting seed phrase', error);
+    }
+  }, [selectedWallet]);
+
+  useEffect(() => {
+    loadSeedPhrase();
+  }, [loadSeedPhrase]);
```
oidoug: It would be nice to have this in it's own hook. It's kinda odd to see a state and imperative call for the phrase in the settings screen implementation.

paulinhapenedo: We could add this to one of the hooks we already have, maybe `useWallets` or `useAccountSettings`. That would be very helpful.

oidoug: Sure, that sounds good.

src/hooks/useWallets.ts
```diff
@@ -57,6 +59,20 @@ export default function useWallets() {
...
+      setSeedPhrase(seedphrase);
+    } catch (error) {
+      logger.log('Error getting seed phrase', error);
+    }
+  }, [selectedWallet]);
+
+  useEffect(() => {
+    loadSeedPhrase();
+  }, [loadSeedPhrase]);
+
```
danibonilha, line 75: this useWallets hook is used across multiple screen, that don't necessarily need the seed phrase, if we want to abstract this somehow, I think it would make sense to have its own hook otherwise a bunch of re-calls and re-renderings might happen, we already have a function called `loadSeedPhrase`, we could have a hook that returns the seedpharse, and if the the wallet is already backed up, this way you won't need to know anything about the selectedWallet, something like, useSelectedWallet, or usuBackupWallet, I think it's ok to leave as it is to unblock the tasks, but I don't think it's a good idea to access the seedPhrase on this hook.

danibonilha (approved):

src/hooks/useWallets.ts
paulinhapenedo: I'll create a ticket with this comment. Can we sync tomorrow offline to discuss the optimal approach for this hook?

PR #1056, opened 2022-09-29
[CS-4509] Updated Recovery Phrase screen to work from settings

oidoug:
### Description

PR for updating `BackupRecoverySeedPhraseScreen` to be used from settings.

Buttons for backing up manually or to cloud depend on wallet's stored values `backedUp` and `backupType`, but they only store the last backup info, not both. We'll need to improve this bit too. My thinking is to keep the above properties for backwards compatibility and add two new properties, `manualBackup` and `cloudBackup` as booleans. Will do in a new PR to not mix things up.

I intentionally reused the text "Backup successful" when manual backup is valid, just because I think it looks nice, but let me know if you disagree.

- [x] Completes #(CS-4509)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot

Screenshot

Screenshot


cardstack/src/screens/Backup/BackupRecoveryPhraseScreen/BackupRecoveryPhraseScreen.tsx
```diff
@@ -29,30 +27,79 @@ const BackupRecoveryPhraseScreen = () => {
...
+    hasCloudBackup,
     seedPhrase,
   } = useBackupRecoveryPhraseScreen();
 
   const backupStatus = useMemo(
-    () => <BackupStatus status={backedUp ? 'success' : 'missing'} />,
-    [backedUp]
+    () => (hasManualBackup ? 'success' : 'missing'),
+    [hasManualBackup]
+  );
```
paulinhapenedo: super nit: abstractions like this doesn't remove complexity. This variable is only used inside another memoized function, we could have left as it was.

cardstack/src/screens/Backup/BackupRecoveryPhraseScreen/strings.ts
```diff
@@ -1,9 +1,13 @@
 export const strings = {
-  title: 'Recovery Phrase',
+  missing: {
+    title: 'Recovery Phrase',
+  },
+  success: { title: 'Backup Successful' },
   description:
-    'These 12 words are the keys to your wallet. Back them up on the cloud or back them up manually. Do not share this with anyone.',
+    'These 12 words are the keys to your wallet, please manually backup them. Do not share this with anyone.',
```
paulinhapenedo: What do you think about reusing the same copy we have on the Manual Backup screen?

`These 12 words are the keys to your wallet. Write them down or copy them to a password manager. Do not share this with anyone.`

oidoug: Sure thing, I had actually done that but forgot to push.

cardstack/src/screens/Backup/BackupRecoveryPhraseScreen/BackupRecoveryPhraseScreen.tsx
oidoug: Agreed, in this case it's there just because `backupStatus` is used twice, so it avoids duplicating the ternary op.
useMemo can actually have more overhead than just doing light processing or ternaries in place.

cardstack/src/screens/Backup/BackupRecoveryPhraseScreen/BackupRecoveryPhraseScreen.tsx
```diff
@@ -29,30 +27,79 @@ const BackupRecoveryPhraseScreen = () => {
...
+          {strings.cloudBackupBtn(Device.cloudPlatform)}
+        </Button>
+      );
+    } else {
+      buttons.push(
+        <Button marginBottom={5} variant="red" onPress={handleDeleteOnPress}>
+          {strings.deleteBackupBtn(Device.cloudPlatform)}
+        </Button>
+      );
+    }
```
paulinhapenedo: What do you think about this approach?

```jsx
const backupToCloudConfig = {
    variant: 'linkWhite' as const,
    onPress: handleCloudBackupOnPress,
  };

  const deleteCloudBackup = {
    variant: 'red' as const,
    onPress: handleDeleteOnPress,
  };

  const cloudBackupButtonsConfig = useMemo(
    () => (hasCloudBackup ? deleteCloudBackup : backupToCloudConfig),
    [hasCloudBackup, backupToCloudConfig, deleteCloudBackup]
  );
  
 // render
<PageWithStackHeaderFooter>
  <CenteredContainer>
    {!hasManualBackup && (
      <Button onPress={handleManualBackupOnPress} marginBottom={4}>
        {strings.manualBackupBtn}
      </Button>
    )}
    <Button {...cloudBackupButtonsConfig}>
      {strings.cloudBackupBtn(Device.cloudPlatform, hasCloudBackup)}
    </Button>
  </CenteredContainer>
</PageWithStackHeaderFooter>
      
      
// in strings.ts
{
cloudBackupBtn: (cloud: string, hasBackup: boolean) =>
hasBackup ? `Delete ${cloud} Backup` : `Back up to ${cloud}`,
}

```

To me, it doesn't make sense to create a `PageWithStackHeaderFooter` component in order to have code repeating the order of things on screen and then move the footer logic to be before the component declaration.

oidoug: I like it, I'll apply your suggestion with some small changes.

cardstack/src/screens/Backup/BackupRecoveryPhraseScreen/BackupRecoveryPhraseScreen.tsx
```diff
@@ -29,30 +27,51 @@ const BackupRecoveryPhraseScreen = () => {
...
             {strings.title}
           </Text>
-          <Text color="grayText" letterSpacing={0.4}>
-            {strings.description}
-          </Text>
+          {!hasManualBackup && (
+            <Text color="grayText" letterSpacing={0.4}>
+              {strings.description}
+            </Text>
+          )}
```
paulinhapenedo, line 72: super nit, but I forgot to comment: I think we could always show this description.

paulinhapenedo (approved):
LGTM! 👏

cardstack/src/screens/Backup/BackupRecoveryPhraseScreen/BackupRecoveryPhraseScreen.tsx
oidoug: Ah sorry, missed your comment here. I'll do this change in my next PR.

PR #1058, opened 2022-10-03
[CS-4666] Separate manual backup flag from cloud

oidoug:
### Description

This PR separates manual backups into its own flag. Manual backup is a simple boolean decision, so I limited on saving just that. The PR also refactors the actions that save backup decisions, separating cloud and manual calls and refactors the relevant code.

On loading a wallet, we check if backed-up manually to assign the new flag. The flag is only saved to secure store if something else needs to be saved, for example, if the user backups to cloud.

The PR also fixes the cloud button styling on RecoveryPhrase screen if its the only button visible.

- [x] Completes #(CS-4666)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

### DIFF
--- a/cardstack/src/hooks/backup/useWalletManualBackup.js
+++ b/cardstack/src/hooks/backup/useWalletManualBackup.js
@@ -2,9 +2,8 @@ import { captureException } from '@sentry/react-native';
 import { useCallback } from 'react';
 import { useDispatch } from 'react-redux';
 
-import WalletBackupTypes from '@rainbow-me/helpers/walletBackupTypes';
 import { useWallets } from '@rainbow-me/hooks';
-import { setWalletBackedUp } from '@rainbow-me/redux/wallets';
+import { setWalletManualBackup } from '@rainbow-me/redux/wallets';
 import logger from 'logger';
 
 export const useWalletManualBackup = () => {
@@ -13,9 +12,7 @@ export const useWalletManualBackup = () => {
 
   const confirmBackup = useCallback(async () => {
     try {
-      await dispatch(
-        setWalletBackedUp(selectedWallet.id, WalletBackupTypes.manual)
-      );
+      await dispatch(setWalletManualBackup(selectedWallet.id));
     } catch (e) {
       logger.sentry(
         `Error setting manual backup for wallet ID: ${selectedWallet.id}`
--- a/src/components/settings-menu/BackupSection/AlreadyBackedUpView.js
+++ b/src/components/settings-menu/BackupSection/AlreadyBackedUpView.js
@@ -55,12 +55,17 @@ export default function AlreadyBackedUpView() {
 
   const walletStatus = useMemo(() => {
     let status = null;
-    if (wallets[walletId].backedUp) {
-      if (wallets[walletId].backupType === WalletBackupTypes.manual) {
-        status = WalletBackupStatus.MANUAL_BACKUP;
-      } else {
-        status = WalletBackupStatus.CLOUD_BACKUP;
-      }
+    if (
+      wallets[walletId].manuallyBackedUp ||
+      (wallets[walletId].backedUp &&
+        wallets[walletId].backupType === WalletBackupTypes.manual)
+    ) {
+      status = WalletBackupStatus.MANUAL_BACKUP;
+    } else if (
+      wallets[walletId].backedUp &&
+      wallets[walletId].backupType === WalletBackupTypes.cloud
+    ) {
+      status = WalletBackupStatus.CLOUD_BACKUP;
     } else {
       status = WalletBackupStatus.IMPORTED;
     }
--- a/src/redux/wallets.js
+++ b/src/redux/wallets.js
@@ -91,9 +91,6 @@ export const walletsLoadState = () => async (dispatch, getState) => {
       selectedWallet.backupType === WalletBackupTypes.manual &&
       !selectedWallet.manuallyBackedUp
     ) {
-      console.log('::: setting manuallyBackedUp', {
-        manual: selectedWallet.manuallyBackedUp,
-      });
       selectedWallet.manuallyBackedUp = true;
     }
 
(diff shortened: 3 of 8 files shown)


src/redux/wallets.js
```diff
@@ -119,7 +128,18 @@ export const walletsSetSelected = wallet => async dispatch => {
   });
 };
 
-export const setWalletBackedUp = (walletId, method, backupFile = '') => async (
+export const setWalletManualBackup = walletId => async (dispatch, getState) => {
+  const { wallets } = getState().wallets;
+  const newWallets = { ...wallets };
+  newWallets[walletId] = {
+    ...newWallets[walletId],
+    manuallyBackedUp: true,
```
paulinhapenedo, line 136: Don't we need to also set `backedUp` and `backupType` here? Or am I missing something?

cardstack/src/hooks/backup/useWalletManualBackup.js
```diff
@@ -13,9 +12,7 @@ export const useWalletManualBackup = () => {
 
   const confirmBackup = useCallback(async () => {
     try {
-      await dispatch(
-        setWalletBackedUp(selectedWallet.id, WalletBackupTypes.manual)
-      );
+      await dispatch(setWalletManualBackup(selectedWallet.id));
     } catch (e) {
       logger.sentry(
         `Error setting manual backup for wallet ID: ${selectedWallet.id}`
```
paulinhapenedo, line 18: nit: This file could be converted to `.ts`.

src/redux/wallets.js
oidoug: No, sorry I should have this info in the description.

These props will only hold cloud backup info now. Ideally `backedUp` and `backupType` should be renamed, but renaming impacts older wallets with these values already saved (wallets are saved as key-pair text values in secure store).

So to avoid having to migrate both manual and cloud (which is also more complex) I choose to keep them for cloud as is.

cardstack/src/hooks/backup/useWalletManualBackup.js
oidoug: wow thanks, I didn't notice that.

oidoug: I'll change the extension in my next PR, I was already working on this file.

paulinhapenedo (approved):
LGTM

src/components/settings-menu/BackupSection/AlreadyBackedUpView.js
```diff
@@ -55,12 +55,17 @@ export default function AlreadyBackedUpView() {
...
-      if (wallets[walletId].backupType === WalletBackupTypes.manual) {
-        status = WalletBackupStatus.MANUAL_BACKUP;
-      } else {
-        status = WalletBackupStatus.CLOUD_BACKUP;
-      }
+    if (
+      wallets[walletId].manuallyBackedUp ||
+      (wallets[walletId].backedUp &&
+        wallets[walletId].backupType === WalletBackupTypes.manual)
+    ) {
```
danibonilha, line 62: you could assign `wallets[walletId]` to a variable to shrink this a bit

src/components/settings-menu/BackupSection/AlreadyBackedUpView.js
```diff
@@ -55,12 +55,17 @@ export default function AlreadyBackedUpView() {
...
+      status = WalletBackupStatus.MANUAL_BACKUP;
+    } else if (
+      wallets[walletId].backedUp &&
+      wallets[walletId].backupType === WalletBackupTypes.cloud
+    ) {
```
danibonilha, line 67: ditto, nit: also maybe check .backedUp once on top

src/components/settings-menu/BackupSection/AlreadyBackedUpView.js
```diff
@@ -55,12 +55,17 @@ export default function AlreadyBackedUpView() {
...
-    if (wallets[walletId].backedUp) {
-      if (wallets[walletId].backupType === WalletBackupTypes.manual) {
-        status = WalletBackupStatus.MANUAL_BACKUP;
-      } else {
-        status = WalletBackupStatus.CLOUD_BACKUP;
-      }
+    if (
+      wallets[walletId].manuallyBackedUp ||
+      (wallets[walletId].backedUp &&
+        wallets[walletId].backupType === WalletBackupTypes.manual)
```
danibonilha, line 61: Does this check need to exist ? it seems by migrating on walletLoad, this condition will never be checked 🤔

src/redux/wallets.js
```diff
@@ -130,22 +150,20 @@ export const setWalletBackedUp = (walletId, method, backupFile = '') => async (
...
     backupDate: Date.now(),
     backupFile,
-    backupType: method,
+    backupType: WalletBackupTypes.cloud,
   };
 
   await dispatch(walletsUpdate(newWallets));
   if (selected.id === walletId) {
     await dispatch(walletsSetSelected(newWallets[walletId]));
   }
```
danibonilha, line 159: Do we need this piece here ? I mean, we should investigate to keep consistent with the manualBackup, if this is required here, we'll probably need it there, but if we don't, we can remove from here

src/components/settings-menu/BackupSection/AlreadyBackedUpView.js
oidoug: Good tip

src/components/settings-menu/BackupSection/AlreadyBackedUpView.js
oidoug: It does not, I changed this piece before commiting to the migration. Checking by `manuallyBackedUp` is enough 👍

src/redux/wallets.js
oidoug: Good catch. I think it's still needed because there are places we pick and use the property "selected" from the store right away. Not calling this in manual backup saving doesn't seem to affect runtime, because the selected redux property will be updated, but it will affect when re-launching, selected will not have the updated values until another action does that.

I'll dispatch the action on manual as well.

PR #1059, opened 2022-10-03
[CS-4681] Add user friendly error messages to the backup flow

paulinhapenedo:
### Description
This PR is a follow-up to the comment made https://github.com/cardstack/cardwallet/pull/1054#discussion_r984920843 . It adds user friendly error messages to the cloud backup flow, including a default message for empty errors.

Screenshot below shows the fallback error message.

- [x] Completes #CS-4681

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot


cardstack/src/hooks/backup/utils.ts
```diff
@@ -4,8 +4,11 @@ const userFriendlyErrorMsg = {
   [CLOUD_BACKUP_ERRORS.KEYCHAIN_ACCESS_ERROR]:
     'You need to authenticate to proceed with the Backup process',
   [CLOUD_BACKUP_ERRORS.USER_CANCELED_DRIVE_API_AUTH]: `Not authorized to interact with Google Drive.`,
+  [CLOUD_BACKUP_ERRORS.ERROR_DECRYPTING_DATA]: `Incorrect password! Please try again.`,
+  [CLOUD_BACKUP_ERRORS.SPECIFIC_BACKUP_NOT_FOUND]: `We couldn't find your previous backup!`,
+  [CLOUD_BACKUP_ERRORS.ERROR_GETTING_ENCRYPTED_DATA]: `We couldn't access your backup at this time. Please try again later.`,
+  [CLOUD_BACKUP_ERRORS.WALLET_BACKUP_STATUS_UPDATE_FAILED]: `Please, try again in a few seconds. Make sure you have internet access.`,
```
oidoug: These look similar to what we have in `rn-cloud.ts`. Looks redundant to have have the errors here too.
Is it possible to have these errors extracted and reused?

cardstack/src/hooks/backup/useWalletCloudBackup.ts
```diff
@@ -86,8 +89,10 @@ export const useWalletCloudBackup = () => {
       } catch (error) {
         const title = `Error while trying to backup wallet to ${Device.cloudPlatform}`;
 
-        console.log('error', error);
-        const message = getFriendlyErrorMessage(error.message);
+        const message = getFriendlyErrorMessage(
+          error.message ||
+            CLOUD_BACKUP_ERRORS.WALLET_BACKUP_STATUS_UPDATE_FAILED
+        );
```
danibonilha: Why not default the `getFriendlyErrorMessage` to return a generic message, instead of setting here ? My concern is in case `error.message` returns a key that doesn't exist on the mapping, the app will crash, bc it will be undefined

cardstack/src/hooks/backup/utils.ts
```diff
@@ -4,8 +4,11 @@ const userFriendlyErrorMsg = {
...
 };
 
-export const getFriendlyErrorMessage = (
-  errorMessage = CLOUD_BACKUP_ERRORS.WALLET_BACKUP_STATUS_UPDATE_FAILED
-) => userFriendlyErrorMsg[errorMessage];
+export const getFriendlyErrorMessage = (errorMessage: string) =>
+  userFriendlyErrorMsg[errorMessage];
```
danibonilha: So I suggested to add the default case here, and also add an optional chaining.

cardstack/src/hooks/backup/utils.ts
danibonilha: up

paulinhapenedo: I'll remove this `utils` file and use only the messages that we have on the `rn-cloud`.

cardstack/src/hooks/backup/utils.ts
paulinhapenedo: Added the optional chaining and a default error message to be used in case there's none (since we're showing that error to the user in the Alert). There's no `getFriendlyErrorMessage` anymore since that was being used to change some of the error messages to be more user friendly.

oidoug (approved):
LGTM ⛵

PR #1062, opened 2022-10-04
[CS-4669] Create hook to handle selected wallet backup state

paulinhapenedo:
### Description
This PR creates the `useSelectedWallet` hook. it exposes the seed phrase and the wallet status regarding its backup state (cloud, manual or none).

Other adjustments made on this PR worth explaining:
1. `browserlist` was triggering annoying warnings whenever you ran `yarn start`. It doesn't affect our project, so I updated it to get rid of the warnings.
2. Added `manuallyBackedUp` to the `RainbowWallet` interface.
3. Renamed `isBackedUpWallet` function inside `models/backup.ts` as `hasCloudBackup` to better reflect what is does.
4. `jest.FakeTimers` was added to one of the test assertions to get rid of the `You called act(async () => ...) without await` warnings. 

- [x] Completes #CS-4649

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots
You shouldn't see any visual changes ;)

### DIFF
--- a/cardstack/src/hooks/useSelectedWallet.ts
+++ b/cardstack/src/hooks/useSelectedWallet.ts
@@ -34,9 +34,10 @@ export const useSelectedWallet = () => {
     [selectedWallet]
   );
 
-  const hasManualBackup = useMemo(() => selectedWallet.manuallyBackedUp, [
-    selectedWallet,
-  ]);
+  const hasManualBackup = useMemo(
+    () => selectedWallet.manuallyBackedUp ?? false,
+    [selectedWallet]
+  );
 
   return {
     seedPhrase,
--- a/cardstack/src/hooks/useSelectedWallet.ts
+++ b/cardstack/src/hooks/useSelectedWallet.ts
@@ -1,6 +1,7 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
 
-import WalletBackupTypes from '@rainbow-me/helpers/walletBackupTypes';
+import { isBackedUpWallet } from '@cardstack/models/backup';
+
 import { useWallets } from '@rainbow-me/hooks';
 import { loadSeedPhrase } from '@rainbow-me/model/wallet';
 import { logger } from '@rainbow-me/utils';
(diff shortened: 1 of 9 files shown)


cardstack/src/hooks/useSelectedWallet.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const hasCloudBackup = useMemo(
+    () =>
+      selectedWallet.backedUp &&
+      selectedWallet.backupType === WalletBackupTypes.cloud,
+    [selectedWallet]
+  );
+
+  const hasManualBackup = useMemo(
+    () => selectedWallet.manuallyBackedUp ?? false,
```
oidoug: When selectedWallet first become available, manuallyBackedUp will be already set.
But if selectedWallet is not, you might get a undefined exception anyway. 

Suggestion:

```suggestion
    () => selectedWallet?.manuallyBackedUp ?? false,
```

src/model/wallet.ts
```diff
@@ -130,6 +130,7 @@ export interface RainbowWallet {
   backupDate?: string;
   backupType?: string;
   damaged?: boolean;
+  manuallyBackedUp?: boolean;
```
oidoug, line 133: Good catch!

cardstack/src/hooks/useSelectedWallet.ts
```diff
@@ -0,0 +1,47 @@
...
+      getSeedPhrase();
+    }
+  }, [walletReady, getSeedPhrase]);
+
+  const hasCloudBackup = useMemo(
+    () =>
+      selectedWallet.backedUp &&
+      selectedWallet.backupType === WalletBackupTypes.cloud,
+    [selectedWallet]
+  );
```
danibonilha: we could use the old `isBackedUpWallet`, the checks shouldn't be different.

cardstack/src/hooks/useSelectedWallet.ts
```diff
@@ -0,0 +1,47 @@
...
+      }
+    } catch (error) {
+      logger.log('Error getting seed phrase', error);
+    }
+  }, [selectedWallet]);
+
+  useEffect(() => {
+    if (walletReady) {
+      getSeedPhrase();
+    }
```
danibonilha: maybe we should skip this call if seedPhrase already exists

cardstack/src/hooks/useSelectedWallet.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const hasCloudBackup = useMemo(
+    () =>
+      selectedWallet.backedUp &&
+      selectedWallet.backupType === WalletBackupTypes.cloud,
+    [selectedWallet]
+  );
+
+  const hasManualBackup = useMemo(
+    () => selectedWallet?.manuallyBackedUp ?? false,
```
danibonilha: nit: if it doesn't exist, it already returns a falsy value, unless there's an specific case where we check for null or undefined, we may not need this nullish coalescing operator. If it was added just for testing purposes we could make the assertion with `tobeFalsy()`

oidoug (approved):
LGTM 👏

PR #1063, opened 2022-10-04
[CS-4667] Handle cloud backup deletion

paulinhapenedo:
### Description
This PR adds the cloud backup deletion to the `useWalletCloudBackup` hook.
It also adds a redirect when the cloud backup is successful. We were missing that.

~I'm still working on the tests, but I don't wanna that to block this PR from being reviewed.~
Tests done!

# Update
Video reference updated on Linear.

- [x] Completes #CS-4667

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/src/components/settings-menu/DeveloperSettings.js
+++ b/src/components/settings-menu/DeveloperSettings.js
@@ -3,7 +3,7 @@ import { useNavigation } from '@react-navigation/native';
 import React, { useCallback } from 'react';
 import { ScrollView } from 'react-native';
 import { ListFooter, ListItem } from '../list';
-import { deleteCloudBackups } from '@cardstack/hooks';
+import { deleteAllBackups } from '@cardstack/models/rn-cloud';
 import { Routes } from '@cardstack/navigation';
 import GanacheUtils, { restartApp } from '@cardstack/utils';
 import { resetWallet } from '@rainbow-me/model/wallet';
@@ -31,7 +31,7 @@ const DeveloperSettings = () => {
         testID="reset-keychain-section"
       />
       <ListItem label="🔄 Restart app" onPress={restartApp} />
-      <ListItem label="🗑️ Remove all backups" onPress={deleteCloudBackups} />
+      <ListItem label="🗑️ Remove all backups" onPress={deleteAllBackups} />
       <ListItem
         label="🤷 Restore default experimental config"
         onPress={() => AsyncStorage.removeItem('experimentalConfig')}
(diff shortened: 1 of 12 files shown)


src/redux/wallets.js
```diff
@@ -112,8 +112,11 @@ export const walletsLoadState = () => async (dispatch, getState) => {
   }
 };
 
-export const walletsUpdate = wallets => async dispatch => {
+export const walletsUpdate = wallets => async (dispatch, getState) => {
+  const { selected } = getState().wallets;
+
   await saveAllWallets(wallets);
+  await dispatch(walletsSetSelected(wallets[selected.id]));
```
oidoug: It makes sense to always update the selected wallet. But I think you missed removing the call to `walletsSetSelected` at line 144, as you did in 161 and 292.

oidoug (approved):
LGTM 🔥 
I only have one comment you can address before merging.

oidoug (changes requested):
While testing I noticed that after deleting the backup, the Recovery Phrase screen opened from settings doesn't update, it still shows the "delete icloud backup" button.

src/components/settings-menu/DeveloperSettings.js
```diff
@@ -1,46 +1,23 @@
 import AsyncStorage from '@react-native-async-storage/async-storage';
 import { useNavigation } from '@react-navigation/native';
 import React, { useCallback } from 'react';
-import { Alert, ScrollView } from 'react-native';
-import GanacheUtils from '../../../cardstack/src/utils/ganache-utils';
+import { ScrollView } from 'react-native';
 import { ListFooter, ListItem } from '../list';
-import { deleteAllBackups } from '@cardstack/models/rn-cloud';
+import { deleteCloudBackups } from '@cardstack/hooks';
```
danibonilha: This function don't seem to exist outside a hook

src/hooks/useManageCloudBackups.ts
```diff
@@ -55,7 +55,7 @@ function deleteWalletWithConfirmation(wallets: any, dispatch: AppDispatch) {
         await dispatch(walletsUpdate(newWallets));
 
         // Delete all backups (debugging)
-        await deleteAllBackups();
+        await deleteAllCloudBackups();
 
         Alert.alert('Backups Deleted Succesfully');
       }
```
danibonilha, line 61: nit: we don't seem to need the `deleteWalletWithConfirmation` on this hook anymore, it could be replaced with the new `deleteCloudBackups` from useWalletCloud hook

paulinhapenedo: This hook will be deleted.

oidoug (approved):
All good! 🛸

PR #1064, opened 2022-10-05
[CS-4682] Connecting new backup flow into onboarding

oidoug:
### Description

This PR adds the backup flow in onboarding by updating the logic in `useShowOnboarding` hook. It changes some auxiliary code but the essence is in the hook.

It also adds a new persisted preference flag in case uses skip the backup flow.

**Note:** I added a AsyncStore.clear() on resetting the wallet so preferences stored with redux-persist, like skip backup or profile, would also be cleared. Let me know if you remember something that needs to be maintained over resets.

- [x] Completes #(CS-4682)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Several recording are available in Linear [here.](https://linear.app/cardstack/issue/CS-4682/[logic]-connect-new-backup-flow-to-the-onboarding-experience)

### DIFF
--- a/cardstack/src/hooks/onboarding/__tests__/useShowOnboarding.test.ts
+++ b/cardstack/src/hooks/onboarding/__tests__/useShowOnboarding.test.ts
@@ -90,6 +90,21 @@ describe('useShowOnboarding', () => {
     expect(mockedNavigate).toBeCalledWith(Routes.PROFILE_SLUG);
   });
 
+  it('should NOT navigate to profile flow when safes are loading', () => {
+    mockPrimarySafeHelper({
+      hasProfile: false,
+      isLoadingOnInit: true,
+    });
+
+    const { result } = renderHook(useShowOnboarding);
+
+    act(() => {
+      result.current.navigateToNextOnboardingStep();
+    });
+
+    expect(mockedNavigate).not.toBeCalled();
+  });
+
   it('should NOT navigate to profile flow when profile has been fetched and it exists', () => {
     mockPrimarySafeHelper({
       hasProfile: true,
--- a/cardstack/src/hooks/onboarding/__tests__/useShowOnboarding.test.ts
+++ b/cardstack/src/hooks/onboarding/__tests__/useShowOnboarding.test.ts
@@ -1,4 +1,4 @@
-import { renderHook } from '@testing-library/react-hooks';
+import { renderHook, act } from '@testing-library/react-hooks';
 
 import { useShowOnboarding } from '@cardstack/hooks/onboarding/useShowOnboarding';
 import { Routes } from '@cardstack/navigation/routes';
@@ -66,7 +66,11 @@ describe('useShowOnboarding', () => {
   it('should NOT navigate to profile flow when profile hasnt been fetched', () => {
     mockPrimarySafeHelper({ hasProfile: false, isLoadingOnInit: true });
 
-    renderHook(useShowOnboarding);
+    const { result } = renderHook(useShowOnboarding);
+
+    act(() => {
+      result.current.navigateToNextOnboardingStep();
+    });
 
     expect(mockedNavigate).not.toBeCalled();
   });
@@ -77,7 +81,11 @@ describe('useShowOnboarding', () => {
       isLoadingOnInit: false,
     });
 
-    renderHook(useShowOnboarding);
+    const { result } = renderHook(useShowOnboarding);
+
+    act(() => {
+      result.current.navigateToNextOnboardingStep();
+    });
 
     expect(mockedNavigate).toBeCalledWith(Routes.PROFILE_SLUG);
   });
@@ -88,12 +96,16 @@ describe('useShowOnboarding', () => {
       isLoadingOnInit: false,
     });
 
-    renderHook(useShowOnboarding);
+    const { result } = renderHook(useShowOnboarding);
+
+    act(() => {
+      result.current.navigateToNextOnboardingStep();
+    });
 
     expect(mockedNavigate).not.toBeCalled();
   });
 
-  it('should NOT navigate to profile flow if user has not wallet', () => {
+  it('should NOT navigate to profile flow if no wallet', () => {
     (useAuthSelector as jest.Mock).mockReturnValueOnce({
       hasWallet: false,
     });
@@ -103,7 +115,11 @@ describe('useShowOnboarding', () => {
       isLoadingOnInit: false,
     });
 
-    renderHook(useShowOnboarding);
+    const { result } = renderHook(useShowOnboarding);
+
+    act(() => {
+      result.current.navigateToNextOnboardingStep();
+    });
 
     expect(mockedNavigate).not.toBeCalled();
   });
@@ -113,7 +129,11 @@ describe('useShowOnboarding', () => {
       hasSkippedProfileCreation: true,
     });
 
-    renderHook(useShowOnboarding);
+    const { result } = renderHook(useShowOnboarding);
+
+    act(() => {
+      result.current.navigateToNextOnboardingStep();
+    });
 
     expect(mockedNavigate).not.toBeCalled();
   });
@@ -121,7 +141,11 @@ describe('useShowOnboarding', () => {
   it('should navigate to backup flow if not backedup', () => {
     mockUseWallets({ selectedWallet: { manuallyBackedUp: false } });
 
-    renderHook(useShowOnboarding);
+    const { result } = renderHook(useShowOnboarding);
+
+    act(() => {
+      result.current.navigateToNextOnboardingStep();
+    });
 
     expect(mockedNavigate).toBeCalledWith(Routes.BACKUP_EXPLANATION);
   });
@@ -132,7 +156,11 @@ describe('useShowOnboarding', () => {
       hasSkippedBackup: true,
     });
 
-    renderHook(useShowOnboarding);
+    const { result } = renderHook(useShowOnboarding);
+
+    act(() => {
+      result.current.navigateToNextOnboardingStep();
+    });
 
     expect(mockedNavigate).not.toBeCalledWith(Routes.BACKUP_EXPLANATION);
   });
--- a/cardstack/src/hooks/onboarding/__tests__/useShowOnboarding.test.ts
+++ b/cardstack/src/hooks/onboarding/__tests__/useShowOnboarding.test.ts
@@ -36,7 +36,9 @@ jest.mock('@cardstack/redux/persistedFlagsSlice', () => ({
 }));
 
 describe('useShowOnboarding', () => {
-  const mockUseWallets = (overwriteParams: any = {}) => {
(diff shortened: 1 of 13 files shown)


src/model/wallet.ts
```diff
@@ -904,6 +905,9 @@ export const resetWallet = async () => {
       await wipeSecureStorage(allWallets);
       await keychain.wipeKeychain();
 
+      // clear preferences on redux persist properties
+      await AsyncStorage.clear();
+
```
paulinhapenedo: Not sure about this approach, because we don't seem to be interacting with Async Storage directly anymore, only through Redux.

That's why we have `store.dispatch(skipProfileCreation(false));` a few lines below this code and why the actions receive params (true/false).

Maybe we could create an action that resets both `persistedFlags` and call it here?

oidoug: I totally missed that!

Good idea, I'll add a "clearFlags" action then.

cardstack/src/screens/Backup/BackupExplanationScreen/BackupExplanationScreen.tsx
```diff
@@ -14,26 +10,31 @@ import {
...
-    navigate(Routes.BACKUP_MANUAL_BACKUP, params);
-  }, [navigate, params]);
+    navigate(Routes.BACKUP_MANUAL_BACKUP, { seedPhrase });
+  }, [navigate, seedPhrase]);
 
   const handleLaterOnPress = useCallback(() => {
-    navDispatch(StackActions.popToTop());
-  }, [navDispatch]);
+    triggerSkipBackup();
+  }, [triggerSkipBackup]);
```
paulinhapenedo: Right now, tapping on it does nothing on the screen. We could add either the `StackActions.popToTop()` like before (which flashes the Wallet screen before moving to the Profile Creation flow) or `navigate(Routes.PROFILE_SLUG)` directly.

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/__tests__/useBackupSeedPhraseConfirmationScreen.test.ts
```diff
@@ -30,10 +29,11 @@ const mockSeedPhrase =
   'loan velvet fall cluster renew animal trophy clinic adjust fix soon enrich';
 
 describe('BackupSeedPhraseConfirmationScreen', () => {
-  const mockParams = () => {
+  const mockParams = (overwrite: any = {}) => {
```
paulinhapenedo: Couldn't this be `BackupRouteParams` instead of `any`?

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/useBackupSeedPhraseConfirmationScreen.ts
```diff
@@ -54,8 +56,11 @@ export const useBackupSeedPhraseConfirmationScreen = () => {
 
   const handleConfirmPressed = useCallback(() => {
     confirmBackup();
-    navigate(Routes.WALLET_SCREEN);
-  }, [navigate, confirmBackup]);
+
+    if (popStackOnSuccess) {
+      navDispatch(StackActions.pop(popStackOnSuccess));
+    }
+  }, [confirmBackup, popStackOnSuccess, navDispatch]);
```
paulinhapenedo: Tapping the "Done" button does nothing on the screen. We have to add the navigation here again. 

Also, keep in mind that we are reusing this function on line 70 of this file, in the `handleBackupToCloudPress` function. The user only sees the "success" buttons (Done and Backup to Cloud) after successfully choosing the words. If they tap on the latter, we have to set the manual backup as `true` and also redirect them to the cloud password screen.

cardstack/src/hooks/onboarding/__tests__/useShowOnboarding.test.ts
```diff
@@ -29,51 +35,47 @@ jest.mock('@cardstack/redux/persistedFlagsSlice', () => ({
...
 
-const mockPrimarySafeHelper = (overwriteParams?: any) =>
-  (usePrimarySafe as jest.Mock).mockImplementation(() => ({
-    hasProfile: false,
-    isLoadingOnInit: true,
-    ...overwriteParams,
-  }));
-
 describe('useShowOnboarding', () => {
+  const mockUseWallets = (overwriteParams: any = {}) => {
```
paulinhapenedo: nit: we could replace the `any` with `Partial<ReturnType<typeof useWallets>>`

cardstack/src/hooks/onboarding/__tests__/useShowOnboarding.test.ts
```diff
@@ -29,51 +35,47 @@ jest.mock('@cardstack/redux/persistedFlagsSlice', () => ({
...
+    (useWallets as jest.Mock).mockImplementation(() => ({
+      selectedWallet: { manuallyBackedUp: true },
+      walletReady: true,
+      ...overwriteParams,
+    }));
+  };
+
+  const mockPrimarySafeHelper = (overwriteParams: any = {}) =>
```
paulinhapenedo: nit: we could replace the `any` with `Partial<ReturnType<typeof usePrimarySafe>>`

cardstack/src/screens/Backup/BackupExplanationScreen/BackupExplanationScreen.tsx
oidoug: Strange, this should be navigating to profile through the effect on `onShowOnboarding`, it watches `selectedWallet.manuallyBackedUp` and `hasSkippedBackup`, so if any changes it should navigate to profile.

Something else must be off, I'll ping you to discuss offline.

cardstack/src/screens/Backup/BackupSeedPhraseConfirmationScreen/__tests__/useBackupSeedPhraseConfirmationScreen.test.ts
oidoug: yes well observed, I'll update it.

cardstack/src/hooks/onboarding/__tests__/useShowOnboarding.test.ts
```diff
@@ -29,51 +35,47 @@ jest.mock('@cardstack/redux/persistedFlagsSlice', () => ({
...
-    // Second render
-    rerender();
-
-    // Called just on first render
-    expect(mockedNavigate).toBeCalledTimes(1);
-  });
-
-  it('should NOT navigate to profile flow when profile hasnt been fetched', () => {
-    mockPrimarySafeHelper({ hasProfile: false, isLoadingOnInit: true });
-
```
danibonilha, line 73: I'm confused on why these tests were switched, and we removed this second render test, it is important to test  rerendering on this case, because this hook leaves on the wallet root

cardstack/src/hooks/onboarding/useShowOnboarding.ts
```diff
@@ -6,15 +6,32 @@ import { useAuthSelector } from '@cardstack/redux/authSlice';
...
+    hasSkippedProfileCreation,
+    hasSkippedBackup,
+  } = usePersistedFlagsSelector();
 
   const { navigate } = useNavigation();
 
   const { hasWallet } = useAuthSelector();
 
   useEffect(() => {
+    // Wait until wallet and backup info has properly being loaded.
```
danibonilha: nit: wait until wallet and backup info is loaded

src/model/wallet.ts
oidoug: These dispatching for skipProfileCreation and skipBackup actions are not saving the values in local store in time before the wallet reset happens. Not sure why yet, but investigating.

cardstack/src/hooks/onboarding/__tests__/useShowOnboarding.test.ts
oidoug: The updated hook implementation has a similar test using the callback.

cardstack/src/redux/persistedFlagsSlice.ts
```diff
@@ -19,12 +21,23 @@ const slice = createSlice({
     skipProfileCreation(state, action: PayloadAction<boolean>) {
       state.hasSkippedProfileCreation = action.payload;
     },
+    skipBackup(state, action: PayloadAction<boolean>) {
+      state.hasSkippedBackup = action.payload;
+    },
+    clearFlags(state) {
+      state = initialState;
+
+      // Hack note: Returning the state in slice reducers is not necessary generally,
```
oidoug:
```suggestion
      // Hack note: Returning the state in slice reducers is generally not necessary,
```

cardstack/src/screens/Backup/BackupExplanationScreen/BackupExplanationScreen.tsx
```diff
@@ -13,27 +9,35 @@ import {
...
-    navigate(Routes.BACKUP_MANUAL_BACKUP, params);
-  }, [navigate, params]);
+    navigate(Routes.BACKUP_MANUAL_BACKUP, { seedPhrase });
+  }, [navigate, seedPhrase]);
 
   const handleLaterOnPress = useCallback(() => {
-    navDispatch(StackActions.popToTop());
-  }, [navDispatch]);
+    triggerSkipBackup();
+    navigateToNextOnboardingStep(Routes.PROFILE_SLUG);
```
paulinhapenedo, line 33: If I'm importing an account that already has a profile, we don't want to be sending the user to the profile creation flow. Maybe we should not pass a route here?

oidoug: Hum.. good point, problem is that this route here has a purpose. `triggerSkipBackup` will not update the `skipBackup` flag in time when `navigateToNextOnboardingStep` is called, so it will not navigate even when it needs to. 🤔

cardstack/src/redux/persistedFlagsSlice.ts
```diff
@@ -19,12 +21,23 @@ const slice = createSlice({
...
+      // but this state does not update when the action is dispatched
+      // from resetWallet in wallet model for some unknown reason.
+      return state;
```
paulinhapenedo: You can only return the `initialState` as mentioned in the [docs](https://redux-toolkit.js.org/usage/immer-reducers#resetting-and-replacing-state). There's no need to assign it to `state`.

paulinhapenedo (approved):
I believe we can merge and handle the navigation issues in another PR.

cardstack/src/redux/persistedFlagsSlice.ts
oidoug: Thanks for the pointer! I was wondering why in some places we had to return and others not.

PR #1065, opened 2022-10-06
[CS-4638, CS-4639] Profile purchase flow copy changes

oidoug:
### Description

This PR updates the copy for the screens in the profile purchase flow. It also renames screen `PurchaseCTAScreen` to `ProfilePurchaseScreen` so it's more consistent with the other screens in the same flow, and changes the IAP init hook a bit to also provide the products, used now in the first screen in the flow.

- [x] Completes #(CS-4638)
- [x] Completes #(CS-4639)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/194409872-d16df9ec-21a3-4210-a04f-c49b12709f3a.mp4


paulinhapenedo (approved):
LGTM!

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
```diff
@@ -80,6 +83,14 @@ export const useProfileSlugScreen = () => {
     [isFetching, isLoading, slugValidation.slugAvailable]
   );
 
+  const purchaseDisclaimer = useMemo(
+    () =>
+      strings.purchaseDisclaimer(
+        profileProduct?.localizedPrice || defaultProfilePrice
```
danibonilha: instead of doing this twice, why don't we return the localizedValue in `useInitIAPProducts`  which we could convert to `useIAPProducts` ? This way we can keep the defaultPrice scooped and just find the products once, or at least add a helper function as you did for finding the product .

cardstack/src/screens/Profile/ProfileSlugScreen/ProfileSlugScreen.tsx
```diff
@@ -51,6 +52,10 @@ const ProfileSlugScreen = () => {
               {strings.input.description}
             </Text>
           </Container>
+
```
danibonilha: nit: we could remove this line skip

cardstack/src/screens/Profile/ProfileSlugScreen/useProfileSlugScreen.ts
oidoug: I added a localizedPrice to both hooks (useInitIAPProducts and useProfilePurchase) so default price can be scoped. I prefer not to rename useInitIAPProducts as it indicates its purpose.

PR #1067, opened 2022-10-12
[CS-4717] Add view mode to SuffixedInput

paulinhapenedo:
### Description
This PR adds a new prop to the `SuffixedInput` component API that enables or disables the `editable` prop from RN `TextInput`, allowing us to use the component in "view mode".

- [x] Completes #CS-4717

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

### DIFF
--- a/cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
+++ b/cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
@@ -23,6 +23,7 @@ const styles = StyleSheet.create({
 interface SuffixedInputProps extends TextInputProps {
   suffixText?: string;
   suffixTextProps?: TextProps;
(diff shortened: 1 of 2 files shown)


cardstack/src/components/Input/SuffixedInput/SuffixedInput.tsx
```diff
@@ -23,6 +23,7 @@ const styles = StyleSheet.create({
 interface SuffixedInputProps extends TextInputProps {
   suffixText?: string;
   suffixTextProps?: TextProps;
+  viewOnly?: boolean;
```
oidoug: super nit: Wdyt about using `readOnly` instead?

oidoug (approved):
LGTM!

PR #1068, opened 2022-10-13
[Cleanup]: Remove a couple more of old backup related functions

danibonilha:
### Description

This PR removes some unused functions, it also removes the e2e flows that are totally outdated, I'm leaving the send-flow one as an example for when we re-enable the e2e tests.


paulinhapenedo (approved):
Thanks! I missed those.

oidoug:
@danibonilha , could you please also check if `src/hooks/useWalletCloudBackup.ts`? I think it's a leftover too.

oidoug (approved):
LGTM, also did a happy path test on iOS and Android, no issues found. 👍

PR #1069, opened 2022-10-14
[CS-4520] Toast component redesign

oidoug:
### Description

Redesigned Toast modal. Based on `LoadingOverlay`, uses a hook `useToast` that returns a `showToast` callback where you can set message and duration (default to 3 secs).

All uses of previous toasts implementations were replaced, except for the `OfflineToast` because the new toast doesn't fit visually in this case.  One idea is to move the offline message to the navigation bar, as we did with the network badge.

The approach of using a stacked modal for a toast is unconventional, but that was defined by how we want our toasts to look: as modals.

I'm opening the PR as a draft to get the dev and design teams feedbacks.

- [x] Completes #(CS-4520)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot

Screenshot


paulinhapenedo:
I'm not very familiar with the idea of showing a toast as a route. In other RN projects I was involved, the Toast component was imported in the root component (App.ts, for example), thus available to be called using the same hook approach we have. What are the pros and cons of the route approach?

cc @oidoug @danibonilha

cardstack/src/screens/SeedPhraseBackup/SeedPhraseBackup.tsx
```diff
@@ -26,7 +26,7 @@ const SeedPhraseBackup = () => {
   const { params } = useRoute<RouteType<SeedPhraseBackupParams>>();
   const { seedPhrases = [], onSuccess } = params;
 
-  const { CopyToastComponent, copyToClipboard } = useCopyToast({});
+  const { copyToClipboard } = useCopyToast({});
```
danibonilha: you could default the copyToast param to an empty objc, then you can just use `useCopyToast();`

cardstack/src/components/ToastOverlay/useToast.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  const [startTimeout, stopTimeout] = useTimeout();
+
+  const dismissToast = useDismissCurrentRoute(Routes.TOAST_OVERLAY);
+
+  const showToast = useCallback(
+    ({ message, duration = defaultDuration }: ToastOverlayParams) => {
+      navigate(Routes.TOAST_OVERLAY, { message });
+
+      if (duration) {
```
danibonilha: Is there a case for duration to be 0 ?  If so, what does it mean ?

src/components/AppVersionStamp.tsx
```diff
@@ -57,7 +69,7 @@ const AppVersionStamp = ({ showBetaUserDisclaimer = false }) => {
           </Text>
         )}
       </TouchableWithoutFeedback>
-      <ToastComponent />
+      {/* <ToastComponent /> */}
```
danibonilha: lefover?

src/components/send/SendSheet.js
```diff
@@ -92,7 +95,7 @@ export default function SendSheet({
         contacts={contacts}
         isValidAddress={isValidAddress}
         onChangeAddressInput={onChangeInput}
-        onInvalidPaste={onInvalidPaste}
+        onInvalidPaste={() => showToast({ message: strings.invalidPaste })}
```
danibonilha: nit: could wrap on useCallback

cardstack/src/components/ToastOverlay/ToastOverlay.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import { ToastOverlayParams } from './useToast';
+
+const styles = StyleSheet.create({
+  overlayWrapper: {
+    flex: 1,
+    // Note: Workaround for android gesture transparency issue.
+    backgroundColor: '#00000001',
+  },
+});
```
danibonilha: maybe this should be a layout ?

cardstack/src/components/ToastOverlay/ToastOverlay.tsx
```diff
@@ -0,0 +1,41 @@
...
+      borderRadius={20}
+      elevation={20}
+      shadowColor="overlay"
+      shadowOffset={{
+        width: 0,
+        height: 15,
+      }}
+      shadowRadius={30}
+      shadowOpacity={1}
+    >
```
danibonilha: or even better, this could be a component, `Overlay` with same custom tweaks on the conent and it receives children to render

danibonilha (commented):
I tend to agree that is a bit odd to have a toast as a route, giving that it might eventually be something intractable, for example, most of the toast components i've seen and built, the user can tap or move it to dismiss it, or interact with other items on the screen, which is harder to do if it's a route. The LoadingOverlay is set to block every user interaction while something is happening, so I'm not sure if a toast should have the same behavior, but it might be a design choice. if the idea is to completely block user interaction I think it's ok to leave it as a route, although I'm still not convinced this is a toast at all. I like this shopify definition of a [toast](https://polaris.shopify.com/components/toast).

Ps: we should make it dismissible by touching on it, or on the overlay idk, some users don't have patience to wait the auto dismissal, specially now that it blocks the entire app

oidoug:
@danibonilha @paulinhapenedo , I agree that is odd for a toast to be blocky, but from discussions with Gardela in the ticket, the toast should show a back-drop so it made sense to be presented in a modal navigation.

We can also have it as non blocking component instantiated in the app root as Paula suggested. I'll try that to see how it goes.

I'm also waiting on a feedback from Gardela in the ticket. will leave it as a draft until then.

cardstack/src/components/ToastOverlay/useToast.ts
oidoug: From an api point of view, I wanted to allow for permanent toasts if duration is 0. 
The `dismissToast` callback is exported so thats an option.

cardstack/src/components/ToastOverlay/ToastOverlay.tsx
oidoug: do you mean, be defined in utils/layouts?

cardstack/src/components/ToastOverlay/ToastOverlay.tsx
oidoug: Ah yeah, thought about that too. will do!

cardstack/src/navigation/hooks.tsx
```diff
@@ -28,19 +27,23 @@ export const useCardstackMainScreens = (Stack: StackType) =>
...
+
     if (Navigation.getActiveRouteName() === routeName) {
-      goBack();
-      InteractionManager.runAfterInteractions(() => {
-        // this fixes a bug where the overlay dismiss didn't work on create profile, will investigate more
-        if (Navigation.getActiveRouteName() === routeName) {
-          goBack();
-        }
-      });
+      // Theres a bug in our navigator implementation where in some unknown circunstances
```
oidoug, line 36:
```suggestion
      // Theres a bug in our navigation implementation where in some unknown circumstances
```

cardstack/src/navigation/hooks.tsx
```diff
@@ -28,19 +27,23 @@ export const useCardstackMainScreens = (Stack: StackType) =>
...
-      });
+      // Theres a bug in our navigation implementation where in some unknown circumstances
+      // the pushed screen has a lower stack index than it's parent, causing goBack
+      // to pop the parent and not the the active route. The validation
+      // below is a workaround the issue.
+      if (parent && parent.getState().index > getState().index) {
+        parent.goBack();
+      } else {
+        goBack();
+      }
```
oidoug, line 44: Previous workaround was popping screens twice in some cases (opening wallet address from settings, for example).
I couldn't find why this happens, but identified that indexes are in a "wrong order".

danibonilha, line 44: Can you try to use the previous simplified version ?  This workaround was added without much thought and this specific route and issue, doesn't exist anymore, so before adding this new check I'd recommend we try the "old" way.

```
  const checkAndDismissCurrentRoute = useCallback(() => {
    if (Navigation.getActiveRouteName() === routeName) {
      goBack();
    }
  }, [goBack, routeName]);
```

Also worth testing if `Navigation.getActiveRouteName()` should leave outside the memoized callback 🤔

oidoug: I did test that, as commented some times the parent has a higher index than the child screen, so goBack pops out the parent and not the modal.

Unfortunately I couldn't find why that happens, all `navigate` calls seams alright, so this workaround was the best solution I found (the double goBack we had was popping screens twice).

It can be reproduced from opening screens from settings, like "My Wallet Address -> copy address" to show the message overlay.

oidoug:
@danibonilha @paulinhapenedo , Michael Dang suggested a more "snappier" animation, so here's an update. Please take a look at this again 🙏 

https://user-images.githubusercontent.com/129619/196544518-76875813-5687-4ce4-ad1d-67f7ffdae7f2.mov

paulinhapenedo (commented):
Shouldn't we rename the `useCopyToast` since it's not a toast anymore? It could be misleading.

Other than that, I think things are ok. They could be better, but it's not entirely up to us so ¯\_(ツ)_/¯

oidoug:
Renamed useCopyToast to `useCopyWithFeedback` so it's clearer what it does while keeping it generic.

cardstack/src/navigation/hooks.tsx
danibonilha: It seems to be related with the settings nested navigator, it has all sorts of weird bugs in it.

cardstack/src/navigation/presetOptions.ts
```diff
@@ -152,6 +183,15 @@ export const sheetPreset = ({
...
 });
 
+export const messageOverlayPreset = ({
+  backgroundOpacity = 'half',
+}: SheetPresetOptions = {}): StackNavigationOptions => ({
+  ...sheetPreset({ backgroundOpacity }),
+  cardStyleInterpolator: messageOverlayStyleInterpolator(
+    sheetBackgroundOpacity[backgroundOpacity]
+  ),
+});
```
danibonilha, line 193: It seems this could be a simple object, I see no reason for now to have `backgroundOpacity` as a param

cardstack/src/navigation/hooks.tsx
danibonilha: It seems to be related with the settings nested navigator, it has all sorts of weird bugs in it. Thanks fro trying the old way

oidoug:
@danibonilha @paulinhapenedo 

I did some more investigation on the settings modal and navigation state changes and the closest of an explanation I have is that the screen index issue occurs because the Settings is a modal that is also a navigator. When on a second navigate call from this stack the useNavigation hook is not picking up the right navigator and pushing new screens from the primary root navigator instead of the settings one.

I believe the best way to fix this is by refactoring the settings navigator to be a stack group in the primary nav.

I created a ticket to address this, since I believe it's too out-of-scope for this PR:
https://linear.app/cardstack/issue/CS-4786/[debt]-refactor-settings-modal-navigator-into-a-stackgroup

Even though having the Message Overlay as a stacked screen might not be the best possible solution, I think it works well and keeps the api clean (don't need to use a global state).

I think this PR improves the app's UX by making our "toast" messages a more visible, so I appreciate if we can merge it 😊👍

paulinhapenedo (approved):
Nice work on this one 👏

danibonilha (approved):
🚀

PR #1070, opened 2022-10-14
[CS-4715] Restore wallet screen

paulinhapenedo:
### Description
This PR adds the first screen of Restore/Import wallet flow. 
I choose to use the same `screenGroup` as the rest of the backup flow to group by feature, if you will.

One thing to notice is that I've changed the screenGroup configuration. For some weird reason, enabling gestures doesn't conflict with the iCloud password generation anymore (on the `BackupCloudPasswordScreen`). 

- [x] Completes #CS-4715

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

## iOS (iPhone 11)
Screenshot

## Android (Moto G8 Plus)
Screenshot

### DIFF
--- a/cardstack/src/navigation/screenGroups/backup.tsx
+++ b/cardstack/src/navigation/screenGroups/backup.tsx
@@ -20,7 +20,7 @@ export const BackupScreenGroup = ({ Stack }: { Stack: StackType }) => (
     screenOptions={{
       ...horizontalNonStackingInterpolator,
       detachPreviousScreen: false,
-      gestureEnabled: false, // presentation as 'card' brings conflicts with iCloud password autofill.
+      gestureDirection: 'horizontal',
     }}
   >
     <Stack.Screen
--- a/cardstack/src/navigation/screenGroups/backup.tsx
+++ b/cardstack/src/navigation/screenGroups/backup.tsx
@@ -10,6 +10,7 @@ import {
   BackupRecoveryPhraseScreen,
   BackupSeedPhraseConfirmationScreen,
   BackupManualScreen,
+  BackupRestoreExplanationScreen,
 } from '@cardstack/screens';
 
 import { StackType } from '../types';
@@ -41,6 +42,10 @@ export const BackupScreenGroup = ({ Stack }: { Stack: StackType }) => (
     <Stack.Screen
       component={BackupManualScreen}
       name={Routes.BACKUP_MANUAL_BACKUP}
+    />
+    <Stack.Screen
+      component={BackupRestoreExplanationScreen}
+      name={Routes.BACKUP_RESTORE_EXPLANATION}
     />
   </Stack.Group>
 );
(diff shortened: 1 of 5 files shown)


cardstack/src/screens/Backup/BackupRestoreExplanationScreen/BackupRestoreExplanationScreen.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { strings } from './strings';
+
+const BackupRestoreExplanationScreen = () => {
+  const handleRestoreCloudOnPress = useCallback(() => {
+    // TBD
+  }, []);
+
+  const handleRecoveryPhraseOnPress = useCallback(() => {
+    // TBD
+  }, []);
```
danibonilha, line 21: superNit: you could create the hook, so the next interaction this file doesn't need to change at all

cardstack/src/screens/Backup/BackupRestoreExplanationScreen/strings.ts
```diff
@@ -0,0 +1,7 @@
+export const strings = {
+  title: 'Restore wallet',
+  description:
+    'Restore your wallet with the 12 word recovery phrase that you have written down.',
+  primaryBtn: 'Restore with recovery phrase',
+  secondaryBtn: (cloud: string) => `Restore with ${cloud}`,
```
danibonilha: nit: Device is exported globally, so you probably can use `Device.cloudPlatform` here, without using a function

cardstack/src/navigation/screenGroups/backup.tsx
```diff
@@ -19,7 +20,7 @@ export const BackupScreenGroup = ({ Stack }: { Stack: StackType }) => (
     screenOptions={{
       ...horizontalNonStackingInterpolator,
       detachPreviousScreen: false,
-      gestureEnabled: false, // presentation as 'card' brings conflicts with iCloud password autofill.
+      gestureDirection: 'horizontal',
```
danibonilha, line 23: so odd what was happening before, glad to know it's fixed now

danibonilha (approved):
LGTM, just nits. Also, it seems the bold font is not being applied on the android buttons labels.

cardstack/src/screens/Backup/BackupRestoreExplanationScreen/strings.ts
oidoug: Good suggestion, agreed.

oidoug (approved):
LGTM!

About the font difference, I think it's just a aliasing difference because Paula's android has a large screen, so it looks thinner. 

The primary button uses the default font variant (regular), so it should render the same on both OS'ses.

cardstack/src/screens/Backup/BackupRestoreExplanationScreen/BackupRestoreExplanationScreen.tsx
paulinhapenedo: I wasn't thinking about creating a hook for this screen since it will only have these two functions 🤔

PR #1071, opened 2022-10-14
[CS-4720] UI seed phrase input box

oidoug:
### Description

Adds a new component `PhraseInput` for seed phrase recovery.

Notes:
- The implementation does not follow exactly the design, adding the number is tough!
- On android, lineHeight is not respected, so the appearances vary.
- Since numbers are not showing and spacing between words is smaller, the number of lines to show a whole seed phrase is at least 1 line less, so it is looking a bit empty. I accept suggestions on how to make this better :)

Thanks!

- [x] Completes #(CS-4720)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot

Screenshot

### DIFF
--- a/cardstack/src/components/Input/PhraseInput/PhraseInput.tsx
+++ b/cardstack/src/components/Input/PhraseInput/PhraseInput.tsx
@@ -9,19 +9,19 @@ import { Device } from '@cardstack/utils/device';
 const InputHeight = Device.isIOS ? 139 : 110;
 
 interface PasswordInputProps extends BaseInputProps {
-  isValid: boolean;
+  showAsError?: boolean;
 }
 
 const PhraseInput = forwardRef((props: PasswordInputProps, ref) => {
-  const { onChangeText, value, isValid = true, ...inputProps } = props;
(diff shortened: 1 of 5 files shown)


cardstack/src/components/Input/PhraseInput/PhraseInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+// Note: This is necessary to account for lack of support for
+// lineHeight in Android;
+const InputHeight = Device.isIOS ? 139 : 110;
+
+interface PasswordInputProps extends BaseInputProps {
+  isValid: boolean;
+}
+
+const PhraseInput = forwardRef((props: PasswordInputProps, ref) => {
+  const { onChangeText, value, isValid = true, ...inputProps } = props;
```
danibonilha: once we implement this , i'd suggest to not default isValid to true, I think it's ok for now bc the parent is not handling it, but maybe it should

danibonilha (approved):
It's good for now, we can explore later some ways to add the numbers, but definitely it's just a nice to have.

PR #1072, opened 2022-10-17
[CS-4716] Create BackupRestoreCloudScreen UI

paulinhapenedo:
### Description
This PR adds a component for the Restore from Cloud screen. It's only the UI, so no logic was added except for the password input.

- [x] Completes #CS-4716

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot

### DIFF
--- a/cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
+++ b/cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
@@ -49,7 +49,6 @@ const BackupRestoreCloudScreen = () => {
             autoFocus
             onChangeText={onChangeText}
             value={password}
-            returnKeyType="done"
             onSubmitEditing={handleRestoreOnPress}
             placeholder={strings.inputPlaceholder}
           />
--- a/cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
+++ b/cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
@@ -9,7 +9,6 @@ import {
   PageWithStackHeader,
   PageWithStackHeaderFooter,
   PasswordInput,
-  ScrollView,
   Text,
 } from '@cardstack/components';
 import { Device } from '@cardstack/utils';
@@ -21,12 +20,6 @@ const styles = StyleSheet.create({
   keyboardAvoidView: {
     flex: 1,
     marginBottom: 50,
-  },
-  scrollView: {
-    marginBottom: 50,
-  },
-  passwordInput: {
-    marginBottom: 5,
   },
 });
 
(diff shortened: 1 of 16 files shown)


cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <Container width="80%" marginBottom={7}>
+            <Text variant="pageHeader">{strings.title}</Text>
+            <Text color="grayText" letterSpacing={0.4} marginTop={4}>
+              {strings.description}
+            </Text>
+          </Container>
+          <Container>
+            <PasswordInput
+              isValid={false} // to avoid showing the validation checkmark
+              validationMessage="" // to avoid showing the validation checkmark
```
oidoug: nit: It would look better if both were optional and could be left undefined.

cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { strings } from './strings';
+import { useBackupRestoreCloudScreen } from './useBackupRestoreCloudScreen';
+
+const styles = StyleSheet.create({
+  keyboardAvoidView: {
+    flex: 1,
+    marginBottom: 50,
+  },
+  scrollView: {
+    marginBottom: 50,
```
oidoug: nit: why both the keyboard avoiding view and scrollview need a 50 bottom margin? Can't that be on just the scrollview? If it needs to be separate, since it's the same amount, you could call it something like `sharedMargin` and add to both as a concatenated style. Like commented below.

cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return (
+    <PageWithStackHeader showSkip={false}>
+      <ScrollView
+        flex={1}
+        contentContainerStyle={styles.scrollView}
+        showsVerticalScrollIndicator={false}
+      >
+        <KeyboardAvoidingView
+          style={styles.keyboardAvoidView}
```
oidoug: nit: If margin needs to be set on both scrollview and keyboardavoidview, maybe add as a concatenated style:
```
          style={[styles.keyboardAvoidView, styles.sharedMargin]}
```

oidoug (approved):
LGTM, just minor nitpicks.

cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+          behavior="position"
+          keyboardVerticalOffset={Device.isIOS ? 100 : 0}
```
danibonilha: nit: This seems to be recurring now, maybe we could add to the Device object itself `defaultKeyboardVerticalOffset` or something

cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
danibonilha: +1 just checking if validationMessage exists in order to render it on PasswordInput should be enough.

cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+          </Container>
+          <Container>
+            <PasswordInput
+              isValid={false} // to avoid showing the validation checkmark
+              validationMessage="" // to avoid showing the validation checkmark
+              autoFocus
+              containerProps={styles.passwordInput}
+              onChangeText={onChangeText}
+              value={password}
+              returnKeyType="done"
```
danibonilha: supernit: we seem to set this a bunch of times, maybe we could default the `Input` component to has this as done, and just update it where it's needed

cardstack/src/screens/Backup/BackupRestoreCloudScreen/useBackupRestoreCloudScreen.ts
```diff
@@ -0,0 +1,20 @@
+import { useCallback } from 'react';
+
+import { usePasswordInput } from '@cardstack/components';
+
+export const useBackupRestoreCloudScreen = () => {
+  const { password, onChangeText } = usePasswordInput({
+    validation: (text: string) => !text,
+  });
```
danibonilha: Thought we were not validating, if we update the passwordInput to conditioning validate, we should make validation optional here too.

cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
paulinhapenedo: Actually we don't need to use a `Scrollview` here. I'll be simplifying this code a bit.

cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
paulinhapenedo: It happened twice on the backup project. I wouldn't called it recurring, but ok. Abstracted.

cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
paulinhapenedo: Both are optional now.

PR #1077, opened 2022-10-20
[CS-4718] Restore wallet with cloud backup screen

paulinhapenedo:
### Description
This PR connects the restore backup from cloud logic to its screen.
Opening the PR as draft because I'm gonna add tests to it.

- [x] Completes #[CS-4718]

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots
Will be on Linear.

### DIFF
--- a/cardstack/src/models/rn-cloud.ts
+++ b/cardstack/src/models/rn-cloud.ts
@@ -80,14 +80,6 @@ const getBackupDocumentByFilename = (
     filenameMatcher.includes(file.name)
   );
 
-  if (!document) {
-    logger.sentry('[BACKUP] No backup found with that name ', filename);
-    const error = new Error(CLOUD_BACKUP_ERRORS.SPECIFIC_BACKUP_NOT_FOUND);
-
-    captureException(error);
-    throw error;
-  }
-
   return document;
 };
 
@@ -109,40 +101,47 @@ export const getDataFromCloud = async (
   });
 
   if (!backups || !backups.files || !backups.files.length) {
-    logger.log('[BACKUP] No backups found');
+    logger.sentry('[BACKUP] No backups found');
 
-    throw 'No backups found';
+    return;
   }
 
   const document = getBackupDocumentByFilename(backups, filename);
+
+  if (!document) {
+    logger.sentry('[BACKUP] No backup found with that name: ', filename);
+
+    return;
+  }
 
   const encryptedData = Device.isIOS
     ? await RNCloudFs.getIcloudDocument(document.name)
     : await RNCloudFs.getGoogleDriveDocument(document.id);
 
-  if (encryptedData) {
-    logger.sentry('[BACKUP] Got cloud document ', filename);
-
-    const backedUpDataStringified = await encryptor.decrypt(
-      backupPassword,
-      encryptedData
+  if (!encryptedData) {
+    logger.sentry(
+      `[BACKUP] We couldn't get the encrypted data for: `,
+      document
     );
 
-    if (backedUpDataStringified) {
-      const backedUpData = JSON.parse(backedUpDataStringified);
-      return backedUpData;
-    } else {
-      logger.sentry(`[BACKUP] We couldn't decrypt the data`);
-      const error = new Error(CLOUD_BACKUP_ERRORS.ERROR_DECRYPTING_DATA);
-      captureException(error);
-      throw error;
-    }
+    return;
   }
 
-  logger.sentry(`[BACKUP] We couldn't get the encrypted data`);
-  const error = new Error(CLOUD_BACKUP_ERRORS.ERROR_GETTING_ENCRYPTED_DATA);
-  captureException(error);
-  throw error;
+  logger.sentry('[BACKUP] Got cloud document ', filename);
+
+  const decryptedBackupData = await encryptor.decrypt(
+    backupPassword,
+    encryptedData
+  );
+
+  if (!decryptedBackupData) {
+    logger.sentry(`[BACKUP] We couldn't decrypt the data `, filename);
+
+    return;
+  }
+
+  const backedUpData = JSON.parse(decryptedBackupData);
+  return backedUpData;
 };
 
 /**
--- a/cardstack/src/models/rn-cloud.ts
+++ b/cardstack/src/models/rn-cloud.ts
@@ -3,9 +3,11 @@ import RNCloudFs, { ListFilesResult } from 'react-native-cloud-fs';
 import { CARDWALLET_MASTER_KEY } from 'react-native-dotenv';
 import RNFS from 'react-native-fs';
 
+import { iCloudAlertConfig } from '@cardstack/hooks';
 import { BackupUserData, BackupSecretsData } from '@cardstack/types';
 import { Device } from '@cardstack/utils';
 
+import { Alert } from '@rainbow-me/components/alerts';
 import AesEncryptor from '@rainbow-me/handlers/aesEncryption';
 import { logger } from '@rainbow-me/utils';
 
@@ -103,6 +105,10 @@ export const getDataFromCloud = async (
     if (isAvailable) {
       logger.log('[BACKUP] syncing iCloud');
       await syncCloudIOS();
+    } else {
+      Alert(iCloudAlertConfig);
+
+      return;
     }
   }
 
--- a/cardstack/src/models/rn-cloud.ts
+++ b/cardstack/src/models/rn-cloud.ts
@@ -97,6 +97,15 @@ export const getDataFromCloud = async (
   backupPassword: string,
   filename: string
 ): Promise<BackupSecretsData | undefined> => {
+  if (Device.isIOS) {
+    const isAvailable = await isIOSCloudBackupAvailable();
+
+    if (isAvailable) {
+      logger.log('[BACKUP] syncing iCloud');
+      await syncCloudIOS();
+    }
+  }
+
   if (Device.isAndroid) {
     await RNCloudFs.loginIfNeeded();
   }
@@ -109,7 +118,7 @@ export const getDataFromCloud = async (
   if (!backups || !backups.files || !backups.files.length) {
     logger.log('[BACKUP] No backups found');
 
-    throw '[BACKUP] No backups found';
+    throw 'No backups found';
   }
 
   const document = getBackupDocumentByFilename(backups, filename);
--- a/cardstack/src/models/rn-cloud.ts
+++ b/cardstack/src/models/rn-cloud.ts
@@ -99,19 +99,6 @@ export const getDataFromCloud = async (
   backupPassword: string,
   filename: string
 ): Promise<BackupSecretsData | undefined> => {
-  if (Device.isIOS) {
-    const isAvailable = await isIOSCloudBackupAvailable();
-
-    if (isAvailable) {
-      logger.log('[BACKUP] syncing iCloud');
-      await syncCloudIOS();
-    } else {
-      Alert(iCloudAlertConfig);
-
-      return;
-    }
-  }
-
   if (Device.isAndroid) {
     await RNCloudFs.loginIfNeeded();
   }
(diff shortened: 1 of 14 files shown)


cardstack/src/models/rn-cloud.ts
```diff
@@ -97,6 +97,15 @@ export const getDataFromCloud = async (
...
   filename: string
 ): Promise<BackupSecretsData | undefined> => {
+  if (Device.isIOS) {
+    const isAvailable = await isIOSCloudBackupAvailable();
+
+    if (isAvailable) {
+      logger.log('[BACKUP] syncing iCloud');
+      await syncCloudIOS();
+    }
+  }
```
danibonilha: With the new flow, what happens if iCloud is not available ? Before we won't even show the option to backup from, if no user data was found, now the flow is different, we should probably add same feedback on the restore screen

danibonilha: Also not sure we need to sync every time we call `getDataFromCloud`, since we call `fetchUserDataFromCloud` which calls `getDataFromCloud` then we call `getDataFromCloud` again in a matter of minutes. Maybe this check should be moved to `fetchUserDataFromCloud`

cardstack/src/navigation/tabBarNavigator.tsx
```diff
@@ -237,7 +237,10 @@ export const StackNavigator = () => {
       )}
       {SharedScreens({ navigationKey: !hasWallet ? 'non-auth' : 'auth' })}
       {ProfileScreenGroup({ Stack })}
-      {BackupScreenGroup({ Stack })}
+      {BackupScreenGroup({
+        Stack,
+        navigationKey: !hasWallet ? 'non-auth' : 'auth',
```
danibonilha: nit: maybe this navigationKey could be a constant, so we could share between stacks

cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
```diff
@@ -33,27 +37,26 @@ const BackupRestoreCloudScreen = () => {
...
-      >
-        <Container width="80%" marginBottom={7}>
-          <Text variant="pageHeader">{strings.title}</Text>
-          <Text color="grayText" letterSpacing={0.4} marginTop={4}>
-            {strings.description}
-          </Text>
-        </Container>
-        <Container>
+      <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
+        <KeyboardAvoidingView
```
danibonilha, line 41: Something for us to think about on new PRs, should this be a pattern on screens with input ? Some of the new screens doesn't have it. like `BackupCloudPasswordScreen`, if so we can add a ticket to find and apply on screens that doesn't have it

cardstack/src/models/rn-cloud.ts
paulinhapenedo: Good catch.

cardstack/src/screens/Backup/BackupRestoreExplanationScreen/useBackupRestoreExplanationScreen.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const useBackupRestoreExplanationScreen = () => {
+  const { navigate } = useNavigation();
+  const { showLoadingOverlay, dismissLoadingOverlay } = useLoadingOverlay();
+
+  const handleRestoreCloudOnPress = useCallback(async () => {
+    try {
+      showLoadingOverlay({ title: 'Fetching backup information...' });
+
+      logger.log('[BACKUP] fetching backup info...');
```
danibonilha, line 20: it's best to use `.sentry` on this case, so we can see on sentry, if we use log, it shows just locally

cardstack/src/screens/Backup/BackupRestoreExplanationScreen/useBackupRestoreExplanationScreen.ts
```diff
@@ -0,0 +1,38 @@
...
+      if (data) {
+        dismissLoadingOverlay();
+        logger.log(`[BACKUP] Downloaded backup info`);
+
+        navigate(Routes.BACKUP_RESTORE_CLOUD, { userData: data });
+      }
+    } catch (e) {
+      logger.log('[BACKUP] Error getting userData', e);
+      Alert(strings.errorMessage);
+    }
```
danibonilha, line 32: Let's just double check if we don't need to call `dismissLoadingOverlay` in case of an error

src/hooks/useWalletManager.ts
```diff
@@ -236,38 +228,35 @@ export default function useWalletManager() {
...
-      };
+            if (wallet) {
+              initWalletResetNavState();
+            } else {
+              dismissLoadingOverlay();
+            }
+          } catch (e) {
+            // in case of creation error, we should redirect the user to the import via seed phrase
+            // if the user goes back in the stack, they are not presented with the PIN creation
+            navigate(Routes.IMPORT_SEED_SHEET);
```
danibonilha: Shouldn't we redirect to the new BackupRestoreScreen ? My guess it that `IMPORT_SEED_SHEET` will be removed.

cardstack/src/screens/Backup/BackupRestoreCloudScreen/useBackupRestoreCloudScreen.ts
```diff
@@ -1,13 +1,69 @@
...
+        const { restoredSeed, backedUpWallet } = restoredInfo;
+
+        if (isValidSeed(restoredSeed)) {
+          await importWallet({
+            seed: restoredSeed,
+            backedUpWallet,
+          });
+
+          return;
+        }
```
danibonilha: if the seed is not valid, something also went wrong while backing up, so we need a user feedback here too.

cardstack/src/screens/Backup/BackupRestoreCloudScreen/useBackupRestoreCloudScreen.ts
```diff
@@ -1,13 +1,69 @@
...
+      dismissKeyboardOnAndroid();
+
+      showLoadingOverlay({ title: WalletLoadingStates.RESTORING_WALLET });
 
-  const handleRestoreOnPress = useCallback(() => {
-    // TBD
-  }, []);
+      const restoredInfo = await restoreCloudBackup(password, params.userData);
+
+      if (restoredInfo) {
```
danibonilha: Same for not getting any restoredInfo, and we should log this different states to sentry. I mean, I know that getDatFromCloud throws some errors, alongside this `restoreCloudBackup`  but I have to say I find this pattern a bit confusing, I'd rather have "empty returns", since we need to check for restoreInfo, if data, etc. It's fine to keep as it is, but if want to know more about that ping me 😉

cardstack/src/screens/Backup/BackupRestoreExplanationScreen/useBackupRestoreExplanationScreen.ts
paulinhapenedo: Moved the `dismissLoadingOverlay ` to be out of the if loop above because of the iCloud error handling.

src/hooks/useWalletManager.ts
paulinhapenedo: Like I mentioned yesterday, I'm not sure how @oidoug will be handling the seed phrase import. I prefer to handle this in the next ticket (the one that's handling connecting flows and deleting old stuff).

cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
paulinhapenedo: `BackupCloudPasswordScreen` doesn't have that because the user has to interact with the checkbox in order to move forward (tapping Done on the keyboard doesn't trigger the action to backup). In this particular case, the user can just type their password and be good to go, but yeah, I believe that this a good pattern to have. 

I'll create a ticket for that, if the team agrees.

cardstack/src/screens/Backup/BackupRestoreExplanationScreen/useBackupRestoreExplanationScreen.ts
paulinhapenedo: For every log on this function or just this one?

cardstack/src/navigation/tabBarNavigator.tsx
oidoug: OMG! This is exactly what I was looking for! I was having issues navigating to 'auth' screens after importing the wallet. I didn't know about this navigationKey, but it's the way to go.

This was also an issue when integrating the onboarding, probably can simplify some stuff there too.

Thanks Paula! 👏

cardstack/src/screens/Backup/BackupRestoreCloudScreen/BackupRestoreCloudScreen.tsx
oidoug: I think it's case by case. Screens with a single input hardly need it. 
Anyway, it's a good idea to check if there are other cases where we should add it.

cardstack/src/screens/Backup/BackupRestoreCloudScreen/useBackupRestoreCloudScreen.ts
oidoug: Good point, we should have a "contact support" kind of alert here.

cardstack/src/screens/Backup/BackupRestoreExplanationScreen/useBackupRestoreExplanationScreen.ts
oidoug: I think this type of simple text logs don't need to be in sentry. Better report when there are issues or when something misbehaved. Like in the catch below, there I think it should report to sentry.

src/hooks/useWalletManager.ts
oidoug: I think redirecting the user to import with the seed directly will feel confusing. Maybe we should show an error alert, then return to the main screen in import, BACKUP_RESTORE_EXPLANATION. Even better if we have an extra message there after returning, saying something like, "please try importing with your written seed phrase".

danibonilha: it doesn't really matter what the next PR will handle because both of the flows have the same entry-point, so the correct behavior here, would be to return to that common entry-point, which I guess it's the `BACKUP_RESTORE_EXPLANATION `

cardstack/src/screens/Backup/BackupRestoreCloudScreen/useBackupRestoreCloudScreen.ts
paulinhapenedo: Added a default error message to show users that has a "contact support" copy.

cardstack/src/screens/Backup/BackupRestoreCloudScreen/useBackupRestoreCloudScreen.ts
paulinhapenedo: Thanks for the pairing on this one!

cardstack/src/screens/Backup/BackupRestoreExplanationScreen/useBackupRestoreExplanationScreen.ts
```diff
@@ -0,0 +1,38 @@
...
+export const useBackupRestoreExplanationScreen = () => {
+  const { navigate } = useNavigation();
+  const { showLoadingOverlay, dismissLoadingOverlay } = useLoadingOverlay();
+
+  const handleRestoreCloudOnPress = useCallback(async () => {
+    try {
+      showLoadingOverlay({ title: 'Fetching backup information...' });
+
+      logger.log('[BACKUP] fetching backup info...');
+      const data = await fetchUserDataFromCloud();
```
oidoug, line 21: nit: Noticed now that before this user data fetch was done right after the welcome screens opens, and now it's only done once the user taps the cloud backup button. Wouldn't it be better to do it as the explanation screens pops in?

oidoug (approved):
LGTM 👍 
Regarding my comment below, I think it's fine fetching user data on button press because it's usually quick, just pointed out since we did it preemptively before.

If you'll be adding tests, I suggest adding a new card in linear to handle that.

cardstack/src/screens/Backup/BackupRestoreExplanationScreen/useBackupRestoreExplanationScreen.ts
paulinhapenedo: The problem with this approach is that the user will open the explanation screen and be presented with either the native Android login modal to log-in into their account or an alert if they don't have iCloud yet available, and this will happen without them even doing anything. It will looks weird.

oidoug: I see, then it's fine the way it is.

The current implementation in welcome screen only checks for iOS, and usually iOS users have icloud on.

PR #1078, opened 2022-10-21
[CS-4721] Logic for restore wallet with seed phrase

oidoug:
### Description

This PR adds the option to import a wallet using a seed phrase. It uses parts of soon deprecated `useImportSeedSheet` and abstracts calls to derive wallet and ENS address to `ethers-wallet` model.

- [x] Completes #(CS-4721)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Vid available on linear.


cardstack/src/hooks/backup/useWalletSeedPhraseImport.ts
```diff
@@ -0,0 +1,76 @@
...
+      navigate(Routes.MODAL_SCREEN, {
+        actionType: 'Import',
+        asset: [],
+        isNewProfile: true,
+        onCloseModal: handleImportAccountOnCloseModal,
+        profile: { name: name || checkedEns.current },
+        type: 'wallet_profile',
+        withoutStatusBar: true,
+      });
+    },
```
danibonilha: Does it make sense to keep using this import modal ?

cardstack/src/models/ethers-wallet.ts
```diff
@@ -62,3 +67,56 @@ const getEthersWallet = async (accountAddress?: string) => {
...
+
+  try {
+    const ens =
+      (await mainnetProvider.lookupAddress?.(walletAddress)) || undefined;
+
+    return ens;
+  } catch (error) {
+    logger.log('Error looking up ENS for imported HD type wallet', error);
+  }
+};
```
danibonilha: not sure we need this ens stuff anymore, if we are never showing the user name, I'd suggest to once we hit the Done button on the seed screen, we start importing and remove on this old profile related stuff

cardstack/src/hooks/backup/useWalletSeedPhraseImport.ts
oidoug: Probably not. We moved away from these account customizations in favor of profiles.

cardstack/src/models/ethers-wallet.ts
oidoug: Ok, I agree. I'll not use the import sheet here then. 

Better also revert the changes I did on useImportSeedSheet, leave it as deprecated.

cardstack/src/hooks/backup/useWalletSeedPhraseImport.ts
```diff
@@ -0,0 +1,47 @@
...
+import { sanitizeSeedPhrase } from '@rainbow-me/utils/formatters';
+import logger from 'logger';
+
+export const useWalletSeedPhraseImport = (
+  seedPhrase: string,
+  ethWallet?: EthereumWalletFromSeed
+) => {
+  const { importWallet } = useWalletManager();
+
+  const checkedWallet = useRef<EthereumWalletFromSeed | undefined>(ethWallet);
```
danibonilha: checkedWallet is only used inside `handleImportWallet`. So there's no need to create a ref here, we can scope it inside the function.

cardstack/src/hooks/backup/useWalletSeedPhraseImport.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const sanitizedSeedPhrase = useMemo(() => sanitizeSeedPhrase(seedPhrase), [
+    seedPhrase,
+  ]);
```
danibonilha: same for this const, only used inside `handleImportWallet`

cardstack/src/hooks/backup/useWalletSeedPhraseImport.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const isSeedPhraseValid = useMemo(() => isValidSeedPhrase(seedPhrase), [
+    seedPhrase,
+  ]);
+
+  const handleImportWallet = useCallback(async () => {
+    if (!isSeedPhraseValid || !sanitizedSeedPhrase) {
```
danibonilha: what does a `not sanitizedSeedPhrase` mean ? Hm, maybe we should add this check within the `isSeedPhraseValid` . Oh I think `isValidSeedPhrase` already sanitizes it, so not sure we need to do this here

cardstack/src/models/ethers-wallet.ts
```diff
@@ -62,3 +65,26 @@ const getEthersWallet = async (accountAddress?: string) => {
...
+ * deriveWalletFromSeed function derive a eth account from a seed phrase input
+ * @param seedPhrase : string
+ * @returns Promise<EthereumWalletFromSeed | undefined>
+ */
+export const deriveWalletFromSeed = async (
+  seedPhrase: string
+): Promise<EthereumWalletFromSeed | undefined> => {
+  if (!seedPhrase || !isValidSeedPhrase(seedPhrase)) return undefined;
+
+  const cleanSeedPhrase = sanitizeSeedPhrase(seedPhrase);
```
danibonilha: we are sanitizing it before calling this funciton, and then we sanitize again here ?

cardstack/src/models/ethers-wallet.ts
```diff
@@ -62,3 +65,26 @@ const getEthersWallet = async (accountAddress?: string) => {
...
+/**
+ * deriveWalletFromSeed function derive a eth account from a seed phrase input
+ * @param seedPhrase : string
+ * @returns Promise<EthereumWalletFromSeed | undefined>
+ */
+export const deriveWalletFromSeed = async (
+  seedPhrase: string
+): Promise<EthereumWalletFromSeed | undefined> => {
+  if (!seedPhrase || !isValidSeedPhrase(seedPhrase)) return undefined;
+
```
danibonilha: I'm confused about this function, there's this check on the handle import, and then we are doing the same checks here again, also this is just used while importing the wallet

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
```diff
@@ -1,13 +1,59 @@
...
+    },
+    [isPhraseWrong]
+  );
+
+  const onDonePressed = useCallback(async () => {
+    setLoading();
+
+    if (isSeedPhraseValid) {
+      await handleImportWallet();
+    }
```
danibonilha: if we check when seed is valid here, maybe we don't need to check on `handleImportWallet`

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
```diff
@@ -1,13 +1,59 @@
...
+  const handlePhraseTextChange = useCallback(
+    (updatedPhrase: string) => {
+      setPhrase(updatedPhrase);
+
+      const words = updatedPhrase.split(' ');
+
+      // Phrase is deamed complete when it has 12 words
+      const isComplete =
+        words.length === 12 && words[words.length - 1].length > 0;
+
```
danibonilha: This is partially the same thing that is done on `isValidSeedPhrase`  so maybe we could add another function that partially validates the format

danibonilha: Or you could use this regex `updatedPhrase.match(/\w+/g).length` then, there's no need to trim, because \w already gets all words and ignores the spaces and punctuations.

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
```diff
@@ -1,13 +1,59 @@
...
+      setIsPhraseComplete(isComplete);
+
+      // Reset wrong state when on edit
+      if (isPhraseWrong) setIsPhraseWrong(false);
+    },
+    [isPhraseWrong]
+  );
+
+  const onDonePressed = useCallback(async () => {
+    setLoading();
```
danibonilha: What is this loading for ? Doesn't importWallet already shows the loading overlay ? 🤔

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
```diff
@@ -1,13 +1,59 @@
...
-  const handlePhraseTextChange = useCallback(() => {
-    // TBD
+  const [phrase, setPhrase] = useState('');
+  const [isPhraseComplete, setIsPhraseComplete] = useState(false);
+  const [isPhraseWrong, setIsPhraseWrong] = useState(false);
+  const [loading, setLoading, setLoadingDone] = useBooleanState();
+
+  const { handleImportWallet, isSeedPhraseValid } = useWalletSeedPhraseImport(
+    phrase
+  );
```
danibonilha: I'm even wondering if we need this to be a separated hook, technically the importWallet will only happen on this screen, so maybe we could move it's logic to here, since it will basically be this single `handleImportWallet` function

cardstack/src/hooks/backup/useWalletSeedPhraseImport.ts
paulinhapenedo: Why are we receiving a wallet object as param? I don't see us using it anywhere in the code right now. Am I missing something?

Removing this param could simplify the code by removing the `useRef` and using a regular `const` in line 34.

oidoug: My intention was to reuse this hook in `ImportSeedSheet` but went another way after a previous comment from you. I'll simplify this.

cardstack/src/hooks/backup/useWalletSeedPhraseImport.ts
oidoug: You're right, it's redundant, I'll clean this up.

cardstack/src/models/ethers-wallet.ts
oidoug: This was supposed to be reused, doesn't add value to what we already have in `ethereumUtils` export. I'll remove this extra function out 👍

cardstack/src/models/ethers-wallet.ts
oidoug: ditto

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
oidoug: It was useful before when I was checking for ens - it took a while an a loading state was needed for the done button. It's not needed anymore, I'll remove it.

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
oidoug: Yeah kinda overthinking on my end.
Not sure how we'll handle importing from settings, it will still be there for now I think, but in any case, it's not using this extra hook now anyways.

cardstack/src/hooks/backup/useWalletSeedPhraseImport.ts
oidoug: My intention was to reuse this hook in `ImportSeedSheet` but went another way after a comment from Dani. I'll simplify this.

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
danibonilha: I guess we don't need to import from settings anymore, it was just something useful for devs while we still had multi-wallet support, so i'm good with removing it if y'all agree to.

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
```diff
@@ -1,13 +1,77 @@
...
+
+    setIsPhraseWrong(!isSeedPhraseValid);
+
+    if (!isSeedPhraseValid) {
+      logger.error('Error: Seed phrase provided is invalid.');
+
+      return;
+    }
+
+    console.log(':::1');
```
paulinhapenedo: Ooops ✂️

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
```diff
@@ -1,13 +1,77 @@
...
+
+    try {
+      const sanitizedPhrase = sanitizeSeedPhrase(phrase);
+
+      console.log(':::2');
```
paulinhapenedo: Ooops ✂️

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
```diff
@@ -1,13 +1,77 @@
...
+
+      const checkedWallet = await ethereumUtils.deriveAccountFromWalletInput(
+        sanitizedPhrase
+      );
+
+      console.log(':::3', { sanitizedPhrase, checkedWallet });
```
paulinhapenedo: Ooops ✂️

oidoug: 🤦  my bad

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
```diff
@@ -1,13 +1,71 @@
...
+      const checkedWallet = await ethereumUtils.deriveAccountFromWalletInput(
+        sanitizedPhrase
+      );
+
+      await importWallet({
+        seed: sanitizedPhrase,
+        checkedWallet,
+      });
+    } catch (error) {
+      logger.error('Error deriving and importing wallet.', error);
```
paulinhapenedo, line 59: nit: shouldn't we be logging this into Sentry?

paulinhapenedo (approved):
Looks good! Just a nit comment that you can feel free to ignore.

cardstack/src/screens/Backup/BackupRestorePhraseScreen/useBackupRestorePhraseScreen.ts
oidoug: I wondered the same but was afraid what could eventually be returned in `error` and turn this into a privacy issue.

PR #1082, opened 2022-10-31
[CS-4794] Remote config for controlling Wyre

oidoug:
### Description

This PR adds a new remote config for controlling prepaid card issuing with Wyre, the flag is called `featureWyre`.
It also removes the Device flag since we can set the flag in firebase to be true for iOS only (use "iOS only" rule). 

`featureWyre` is set to true for iOS right now for testing, it should be switched to false before release.

- [x] Completes #(CS-4794)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

"Buy Prepaid Card" button will be hidden.

### DIFF
--- a/cardstack/src/utils/device.ts
+++ b/cardstack/src/utils/device.ts
@@ -14,17 +14,14 @@ const Device = {
   screenWidth,
   cloudPlatform: isIOS ? 'iCloud' : 'Google Drive',
   keyboardBehavior: 'padding' as const,
-  supportsNativeWyreIntegration: isIOS,
   supportsHapticFeedback: isIOS,
   scrollSheetOffset: isIOS ? -(screenHeight * 0.2) : 1,
-  tabBarHeightSize: screenHeight * 0.1,
   keyboardEventWillShow: isIOS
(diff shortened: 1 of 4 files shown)


cardstack/src/utils/device.ts
```diff
@@ -14,7 +14,6 @@ const Device = {
   screenWidth,
   cloudPlatform: isIOS ? 'iCloud' : 'Google Drive',
   keyboardBehavior: 'padding' as const,
-  supportsFiatOnRamp: isIOS,
   supportsNativeWyreIntegration: isIOS,
```
paulinhapenedo: nit: I don't think we are using this flag anymore. We could remove it too.

paulinhapenedo (approved):
LGTM

cardstack/src/utils/device.ts
oidoug: true, removing...

oidoug: There were a couple more unused attributes, I removed them too.

PR #1083, opened 2022-11-01
[CS-4760] Seed Phrase Table indicator while loading wallet

oidoug:
### Description

It was reported by QA that the seed phrase shows an empty state if shown right after the app opens. This is caused by the seedphrase not being available at the time its presented by SeedPhraseTable component.

This PR adds a loading indicator as an optional prop for SeedPhraseTable.

I took the chance to also rename the Screen `BackupRecoveryPhraseScreen` to `BackupSeedPhraseScreen` so it's better distinguished from the new `BackupRestorePhraseScreen` and it's closer to `BackupCloudPasswordScreen`.

- [x] Completes #(CS-4760)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/199267105-b3edc412-5342-493e-9fa7-ded6d1c196af.mp4


paulinhapenedo (approved):
LGTM!

PR #1084, opened 2022-11-07
[CS-4786] Settings screens group refactor

oidoug:
### Description

Settings screens had their own navigator instance which caused issues when screens that were from outside settings' navigation instance were called by settings screens - the parent screen was the wrong one.

Also, settings was using an older implementation pattern. Now it is inside a group and uses the same shared Stack as the rest of the app.

The hack for popping overlay screens that was in `useDismissCurrentRoute` hook is not necessary anymore and was removed.

There are still be some code readability improvements we can make, such as moving the `settings-menu` screens to TS, but might be better done in another PR. I tried to keep this one tightly scoped. Let me know what else you would want to see moved.

- [x] Completes #(CS-4786)

#### Recent Updates to PR

- Refactors SettingsSection to SettingsScreen
- Moves navigate calls from SettingsModal to SettingsScreen
- Removes SettingsModal
- Removes Onboarding tests from settings

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/200331212-18a21f11-41ad-4c24-ad3c-111c5277050e.mp4


paulinhapenedo (approved):
I think it works for now, but would be nice to revisit this in the near future since it's not using the same approach for the routes as we use for all the others.

cardstack/src/navigation/screenGroups/settings.tsx
```diff
@@ -0,0 +1,157 @@
...
+  Routes,
+} from '@cardstack/navigation';
+import { SecurityScreen } from '@cardstack/screens';
+import DesignSystemScreen from '@cardstack/screens/Dev/DesignSystemScreen';
+import WalletAddressScreen from '@cardstack/screens/WalletAddressScreen/WalletAddressScreen';
+import { palette } from '@cardstack/theme';
+
+import {
+  CurrencySection,
+  LanguageSection,
```
danibonilha: maybe it's not on the scope, but I don't see a reason to keep this languageSection on this new file

danibonilha (commented):
If we are taking the time to refactor it, I'd suggest we also change the implementation and apply the same patterns we have for newer code. Now if the intent of this PR is just to remove settings from its own navigator so the dismissRoute hack is not needed. I guess it's ok to merge it as is and revisit later. But i'm a fan of trying to make the code better while refactoring.

oidoug:
I think it's fine to refactor the routes in this PR too. I'll do that, thanks for the feedback 👍

oidoug:
Made some more changes. I appreciate your feedback!

paulinhapenedo:
Super nit but we could completely remove the `SettingsModal.js` and use only the `SettingsSection.js`.

And I don't know if right now we have more time to do this with rewards and other stuff getting traction, but I think it would be nice to also migrate `SettingsSection` to `tsx` and replace all the custom components (ListItem, ColumnWithDividers and ListItemArrowGroup) with `tsx` versions or delete them at all.

oidoug:
@paulinhapenedo  Removing `SettingsModal` makes sense, its not doing much, but refactoring `SettingsSection` IMO isn't worth the effort now.

oidoug:
@paulinhapenedo I tried to move `SettingsScreen` to TS but it's a more involved work. The inner components (list items) need some rework to fit here.

paulinhapenedo (approved):
LGTM 🚀

PR #1085, opened 2022-11-07
[CS-4826] Handle wallet navigation after dApp notification

paulinhapenedo:
### Description
This PR attempts to solve an issue where a notification from the dApp to the wallet for authorization wasn't being correctly handled if the app was in background mode and specially if the biometric auth failed.

~I'm opening as a draft because running the tests locally is giving me flaky results.~

- [x] Completes #CS-4826

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/redux/requests.ts
+++ b/cardstack/src/redux/requests.ts
@@ -1,6 +1,7 @@
 import { AnyAction } from '@reduxjs/toolkit';
 import { filter, get, omit, values } from 'lodash';
 
+import { Navigation, Routes } from '@cardstack/navigation';
 import { getRequestDisplayDetails } from '@cardstack/parsers/signing-requests';
 
 import {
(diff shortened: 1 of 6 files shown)


danibonilha (approved):
LGTM 🚀

cardstack/src/redux/requests.ts
```diff
@@ -24,6 +25,17 @@ const REQUESTS_CLEAR_STATE = 'requests/REQUESTS_CLEAR_STATE';
 // Requests expire automatically after 1 hour
 const EXPIRATION_THRESHOLD_IN_MS = 1000 * 60 * 60;
 
+export const handleWalletConnectRequests = (request?: any) => {
+  if (request) {
+    setTimeout(() => {
```
oidoug, line 30: I'm assuming calling the handler just after `InteractionManager.runAfterInteractions` isn't enough, right?

oidoug (approved):
LGTM 🚀

cardstack/src/redux/requests.ts
paulinhapenedo: We are not calling it after a `InteractionManager.runAfterInteractions` anymore 🤔

PR #1090, opened 2022-11-08
[CS-4842]: Use rn patch to fix android build

danibonilha:
### Description
The build was broken because the react team had publish a package reference which resulted on versions mismatches on maven, more info can be found [here](https://github.com/facebook/react-native/issues/35210). The team has publish a patch in order to fix that, this PR applies the patched version.
And here's a working [build](https://github.com/cardstack/cardwallet/commit/87e627fc55d0e476a06eed5e4f5a8a11f21b678f/checks)


oidoug (approved):
👍

PR #1093, opened 2022-11-09
[CS-4817] Update SDK and signatures implementation to comply with new SDK version

oidoug:
### Description

Updates cardpay-sdk and did-resolver to version 1.0.17;
And refactors signature calls to card pay SDK to pass on `rewardProgramId`.

It uses `rewardDefaultProgramId` as default program Id when required.

- [x] Completes #(CS-4817)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -116,17 +117,21 @@ export const fetchRewardPoolTokenBalances = async ({
...
+      );
 
   const rewardTokensWithPrice = await Promise.all(
     rewardTokens?.map(
       async ({
         tokenSymbol: symbol,
         balance,
         tokenAddress,
-        rewardProgramId,
+        rewardProgramId: tokenBalanceRewardProgramId,
```
danibonilha: if we are fetching the tokens based on a rewardProgramId, why do we need to rename this ?  Isn't it going to be the same id anyways ?

danibonilha (commented):
A small comment, also don't we have some sort of filters to match the tokens and balances to the programId? With this change, my guess is that we don't need some of it anymore, it can be done on a separate PR though, just think it's worth to map the places, maybe add some TODOS

cardstack/src/services/rewards-center/rewards-center-service.ts
oidoug: I renamed this response to avoid the name clashing with the `rewardProgramId` parameter that was also added.
It is the same ID, I think we can use the parameter value here.

oidoug:
@danibonilha , TBH I don't have yet enough context of the SDK additions to know if we can filter the rewards differently or not, at first glance I don't see it.

I'll keep this PR with the minimal changes to unblock the following tasks, ok?

paulinhapenedo (approved):
LGTM!
I think it's worthy merging this PR and then create additional if necessary. It also unblocks a refactor ticket.

PR #1095, opened 2022-11-11
[CS-4832] Remove all xdai references in favor of gnosis

paulinhapenedo:
### Description
Opening as a draft while I refactor some stuff:
- [x] check if we can remove `networkInfo.ts` in favor of using the same information from the SDK;

~change the `NetworkSection` component and hook to list the networks differently. Instead of showing Layer 1 and Layer 2, we show "CardPay supported" and "Others", with something to distinguish between mainnet and testnet.~ will be moving this into to scope of CS-4834 and CS-4835 because it makes more sense to change Layer 1 and Layer 2 approach in there.

----

- [x] Completes #CS-4832

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots
TBA


oidoug (approved):
LGTM, a lot better to use the SDK 👍

PR #1096, opened 2022-11-16
[CS-4829][CS-4830] Update reward center to list all claims and allow claiming one-by-one

oidoug:
### Description

This PR adds a new API resource to fetch individual reward proofs for given program and it's info from the SDK and refactor Reward Center Screen to display the program explainer and the individual proofs.

Notes:
Total claim available item is not being shown because it was too confusing, it's hard to indicate it's the total of the below cells, so I prefer to keep only the individual proofs visible.

The UI is WIP, I appreciate feedbacks. It's still not showing the proof explanation, coming in the next PR.

- [x] Completes #(CS-4829)
- [x] Completes #(CS-4830)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot

### DIFF
--- a/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
+++ b/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
@@ -23,27 +23,13 @@ const RewardsCenterScreen = () => {
   } = useRewardsCenterScreen();
 
   const mainPoolRowProps = useMemo(
-    () =>
-      hasRewards
-        ? {
-            primaryText: fullBalanceToken?.balance.display || '',
-            subText: fullBalanceToken?.native.balance.display || '',
-            coinSymbol: fullBalanceToken?.token.symbol || '',
-            isClaimable: !!fullBalanceToken?.isClaimable,
-          }
-        : undefined,
-    [hasRewards, fullBalanceToken]
-  );
-
-  const rewardsRowProps = useMemo(
-    () =>
-      rewards?.map(reward => ({
-        primaryText: reward?.balance.display || '',
-        subText: reward?.native.balance.display || '',
-        coinSymbol: reward?.token.symbol || '',
-        isClaimable: true,
-      })),
-    [rewards]
+    () => ({
+      primaryText: fullBalanceToken?.balance.display || '',
+      subText: fullBalanceToken?.native.balance.display || '',
+      coinSymbol: fullBalanceToken?.token.symbol || '',
+      isClaimable: !!fullBalanceToken?.isClaimable,
+    }),
+    [fullBalanceToken]
   );
 
   return (
@@ -56,12 +42,12 @@ const RewardsCenterScreen = () => {
         ) : (
           <Container flex={1}>
             {!isRegistered &&
-              (mainPoolRowProps ? (
+              (hasRewards ? (
                 <RegisterContent {...mainPoolRowProps} />
               ) : (
                 <NoRewardContent />
               ))}
-            {isRegistered && <ClaimContent claimList={rewardsRowProps} />}
+            {isRegistered && <ClaimContent rewards={rewards} />}
           </Container>
         )}
       </Container>
--- a/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
+++ b/cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
@@ -1,15 +1,13 @@
 import React, { useMemo } from 'react';
-import { StyleSheet } from 'react-native';
 
-import { Container, NavigationStackHeader, Image } from '@cardstack/components';
-
-import rewardBanner from '../../assets/rewards-banner.png';
+import { Container, NavigationStackHeader } from '@cardstack/components';
 
 import {
   RegisterContent,
   NoRewardContent,
   ClaimContent,
   RewardLoadingSkeleton,
+  RewardProgramHeader,
 } from './components';
 import { strings } from './strings';
 import { useRewardsCenterScreen } from './useRewardsCenterScreen';
@@ -18,49 +16,57 @@ const RewardsCenterScreen = () => {
   const {
     isRegistered,
     hasRewards,
+    rewards,
     fullBalanceToken,
     isLoading,
+    rewardProgramExplainer,
   } = useRewardsCenterScreen();
 
   const mainPoolRowProps = useMemo(
-    () => ({
-      primaryText: fullBalanceToken?.balance.display || '',
-      subText: fullBalanceToken?.native.balance.display || '',
-      coinSymbol: fullBalanceToken?.token.symbol || '',
-      isClaimable: !!fullBalanceToken?.isClaimable,
-    }),
-    [fullBalanceToken]
+    () =>
+      hasRewards
+        ? {
+            primaryText: fullBalanceToken?.balance.display || '',
+            subText: fullBalanceToken?.native.balance.display || '',
+            coinSymbol: fullBalanceToken?.token.symbol || '',
+            isClaimable: !!fullBalanceToken?.isClaimable,
+          }
+        : undefined,
+    [hasRewards, fullBalanceToken]
+  );
+
+  const rewardsRowProps = useMemo(
+    () =>
+      rewards?.map(reward => ({
+        primaryText: reward?.balance.display || '',
+        subText: reward?.native.balance.display || '',
+        coinSymbol: reward?.token.symbol || '',
+        isClaimable: true,
+      })),
+    [rewards]
   );
 
   return (
     <Container backgroundColor="white" flex={1}>
       <NavigationStackHeader title={strings.navigation.title} />
       <Container backgroundColor="white" flex={1}>
-        <Image source={rewardBanner} style={styles.headerImage} />
+        <RewardProgramHeader title={rewardProgramExplainer} />
         {isLoading ? (
           <RewardLoadingSkeleton />
         ) : (
           <Container flex={1}>
             {!isRegistered &&
-              (hasRewards ? (
+              (mainPoolRowProps ? (
                 <RegisterContent {...mainPoolRowProps} />
               ) : (
                 <NoRewardContent />
               ))}
-            {isRegistered && (
-              <ClaimContent
-                claimList={hasRewards ? [mainPoolRowProps] : undefined}
-              />
-            )}
+            {isRegistered && <ClaimContent claimList={rewardsRowProps} />}
           </Container>
         )}
       </Container>
     </Container>
   );
 };
-
-const styles = StyleSheet.create({
-  headerImage: { width: '100%' },
-});
 
 export default RewardsCenterScreen;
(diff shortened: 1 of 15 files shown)


cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
```diff
@@ -18,49 +16,57 @@ const RewardsCenterScreen = () => {
...
       <Container backgroundColor="white" flex={1}>
-        <Image source={rewardBanner} style={styles.headerImage} />
+        <RewardProgramHeader title={rewardProgramExplainer} />
         {isLoading ? (
           <RewardLoadingSkeleton />
         ) : (
           <Container flex={1}>
             {!isRegistered &&
-              (hasRewards ? (
+              (mainPoolRowProps ? (
```
paulinhapenedo: If we are taking the proposed approach, we should probably renamed this because `mainPoolRowProps` isn't conveying its purpose on a if statement.

oidoug: Yeah, the naming used all around rewards is mislanding. 

Checking for the full "pool balance" may not even be needed anymore, but better keep it for now because we aren't sure yet what will change in rewards.

paulinhapenedo:
I'm getting an error when trying to claim a reward that I have on Gnosis. The error is probably because it's dust as I already have a rewards safe.


https://user-images.githubusercontent.com/690904/202302369-fdda2b9e-1290-4a3d-a587-54264b76e2cd.mp4

cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
danibonilha: I'm confused about the usage of mainPoolRewards here too, `hasRewards` in the render gives a clearer idea why we have the condition, but just checking for `mainPoolRowProps` doesn't say much, and we need to go into the prop to see that is checking if it hasRewards

cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
```diff
@@ -18,49 +16,57 @@ const RewardsCenterScreen = () => {
...
+
+  const rewardsRowProps = useMemo(
+    () =>
+      rewards?.map(reward => ({
+        primaryText: reward?.balance.display || '',
+        subText: reward?.native.balance.display || '',
+        coinSymbol: reward?.token.symbol || '',
+        isClaimable: true,
+      })),
+    [rewards]
```
danibonilha: I think this is okay for now, but with the new UI we might need a new component, instead of reusing the rewardRow, in this case, i'd suggest that ClaimContent, gets the rewards and deals with the formatting within it

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
```diff
@@ -42,7 +42,7 @@ export const ClaimContent = ({ claimList }: ClaimContentProps) => {
     navigate,
   ]);
 
-  const renderClaimList = useCallback(
+  const ClaimList = useMemo(
     () =>
       claimList?.map((item, index) => (
         <RewardRow
```
danibonilha: Now that we can have a bunch of rows with information, this should be converted into a flatlist

cardstack/src/theme/textVariants.ts
```diff
@@ -79,4 +79,11 @@ export const textVariants = {
     textTransform: 'uppercase',
     ...fontFamilyVariants.bold,
   },
+  promoBannerTitle: {
+    fontSize: 12,
+    letterSpacing: 1.2,
+    textTransform: 'uppercase',
+    textAlign: 'center',
+    ...fontFamilyVariants.semiBold,
+  },
```
danibonilha: Is there an intention to reuse this ? If not, I don't think it makes much sense to add this kind of text variables

cardstack/src/services/rewards-center/rewards-center-types.ts
```diff
@@ -75,3 +80,28 @@ export interface RewardWithdrawParams
     Omit<RewardsPoolSignedBaseParams, 'rewardProgramId'> {}
 
 export type RewardWithdrawGasEstimateParams = RewardWithdrawBaseParams;
+
+export interface RewardsValidProofsParams {
+  accountAddress: string;
+  rewardProgramId: string;
+  nativeCurrency: NativeCurrency;
+}
```
danibonilha: I guess you could reuse `RewardsSafeQueryParams` giving that safeAddress is optional and you are reusing the data.query anyways on the component

cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -250,6 +282,14 @@ export const getWithdrawGasEstimate = async ({
   return gasEstimate.amount;
 };
 
+export const getRewardProgramInfo = async (rewardProgramId: string) => {
+  const rewardManager = await getRewardManagerInstance();
+
+  const programInfo = await rewardManager.getRewardProgramInfo(rewardProgramId);
+
+  return programInfo.programExplainer;
```
danibonilha, line 313: Maybe we should reflect that this function just returns the `programExplainer`, or we should return the entire `getRewardProgramInfo`  and use the programExplainer on the component

cardstack/src/services/rewards-center/rewards-center-types.ts
```diff
@@ -75,3 +80,28 @@ export interface RewardWithdrawParams
...
+    display: string;
+    wei: string;
+  };
+  native: {
+    balance: {
+      amount: number;
+      display: string;
+    };
+  };
+};
```
danibonilha: There's a `TokenType` to extend and a `TokenInfo`

cardstack/src/screens/RewardsCenterScreen/RewardsCenterScreen.tsx
oidoug: Seems fine to use RewardRow, I'm using it for now without changes. Using your suggestion of handling the reward list inside ClaimContent.

cardstack/src/theme/textVariants.ts
oidoug: I wanted to reuse in the promo banner, but it ended up being slightly different.. removing.

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
```diff
@@ -33,46 +29,36 @@ const tabs = [
...
           title={strings.register.gasInfoBanner.title}
           message={strings.register.gasInfoBanner.message}
         />
-        {renderClaimList()}
       </Container>
+    ),
+    [title]
+  );
+
+  const ListFooter = () => (
```
oidoug: Note: useMemo here did not work, the linter errors with `TypeError: Cannot read property 'parent' of null`. It looks like FlatList doesn't work with a memoized footer, but I can't find the specific reason.

Extracting this as a component looks like too much.
I appreciate your input.

cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -78,6 +79,55 @@ const getValidProofs = async ({
...
+  );
+
+  const claimableProofs = safeAddress
+    ? await rewardPoolInstance.getUnclaimedValidProofsWithoutDust(
+        accountAddress,
+        rewardProgramId,
+        safeAddress
+      )
+    : [];
+
```
danibonilha: isn't both of these proofs claimables?  Maybe we could rename to give better context

cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -78,6 +79,55 @@ const getValidProofs = async ({
...
+          balance: proof.amount,
+        } as unknown) as TokenInfo,
+        nativeCurrency,
+        true
+      );
+
+      const claimingInfo =
+        claimableProofs.find(
+          findProof => findProof.rootHash === proof.rootHash
+        ) || {};
```
danibonilha, line 119: What does this mean ? if it exists on withoutDust it's claimable ? Maybe we could add a comment to refer to that

cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -78,6 +79,55 @@ const getValidProofs = async ({
...
+
+      return {
+        ...proof,
+        ...tokenWithPrice,
+        ...claimingInfo,
+        isClaimable: !!claimingInfo,
```
danibonilha, line 125: I'm confused on spreading proof and claimingInfo, why do we need that ?

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
```diff
@@ -81,6 +67,33 @@ export const ClaimContent = ({ claimList }: ClaimContentProps) => {
...
+          }
+        />
+      </Container>
+    ),
+    [onClaimSingleRewardPress]
+  );
+
+  return (
+    <FlatList
+      data={rewards}
```
danibonilha, line 80: Let's add `showsVerticalScrollIndicator={false}` to avoid reintroducing the scroll bug

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
danibonilha: Not sure why the linter would error this, but it might be because our typescript is 4.5.5 and it supports <4.5.0 only, giving this is the footer and it's small, maybe we could assign the <Container> and it's content directly to the ListFooterComponent, since the Flatlist component is small and the footer is the last parameter

danibonilha (approved):
Just added some comments to be addressed before merging it

cardstack/src/services/rewards-center/rewards-center-service.ts
oidoug: Kinda, `proofs` might contain dust that we can't claim. `claimableProofs`'s proofs should all be claimable since it doesn't include dust.

cardstack/src/services/rewards-center/rewards-center-service.ts
oidoug: Thats was my reasoning, proofs without dust are claimable. I'll improve naming and add a comment 👍

cardstack/src/services/rewards-center/rewards-center-service.ts
oidoug: `claimingInfo` has all values `proof` has plus some extra values, like estimatedGas and `tokenSymbol`. It could replace proof, but spreading and replacing looks cleaner IMO.

cardstack/src/screens/RewardsCenterScreen/components/ClaimContent.tsx
oidoug: The scroll bug happens because the parent container had a padding. No the padding happens inside each FlatList's children and the scroll bar is presented in the correct place, so I preferred to show it.

PR #1097, opened 2022-11-16
[CS-4835] Remove Kovan support

paulinhapenedo:
### Description
This PR removes Kovan support from the wallet. It also handles unsupported chains from now on by using the `supportedChainsArray`method from the sdk. If the user was using the app with a network that we are not supporting anymore, the app will gracefully change the network by updating the `redux-persist` store.

- [x] Completes #CS-4835

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


oidoug (approved):
LGTM, seems like a safe refactor 👍

PR #1098, opened 2022-11-17
[Feat]: Initial wallet connect 2.0 support

danibonilha:
### Description

This PR adds the ability to connect a wallet using wallet connect 2.0 while keeping retro compatibility with v1. It's an initial setup only to connect the wallet, the upcoming PR's will contain request handling, listing connections etc.
For WC 2.0 there's this[ Dapp ](https://react-app.walletconnect.com) example where we can test the connection.

### Screenshots

The first is using walletConnect 2.0, and sushi swap uses v1:

https://user-images.githubusercontent.com/20520102/202568694-b7818c90-2142-44b3-b72e-0537e61ac80d.mov


paulinhapenedo (approved):
Tested locally on both android and iOS and seems to be working. I could see all the requests sent from the dapp.

PR #1099, opened 2022-11-18
[CS-4833, CS-4834] Add Goerli, Polygon and Mumbai support

paulinhapenedo:
### Description
Opening as draft because Polygon and Mumbai are throwing errors on the Hub API.

One thing to notice is a change in the `Podfile` that will handle the code signing error both Doug and I are getting when trying to build the app using xcode 14.1. The changes were copied from the [Expo repository](https://github.com/expo/expo/pull/19095). 

- [x] Completes #CS-4833, #CS-4834

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot


oidoug (commented):
Looking good so far.

Just a suggestion for when renaming or moving files, I prefer to use `git mv` instead of renaming the file directly, so history for the file is maintained.

cardstack/src/components/Transactions/TransactionBase.tsx
```diff
@@ -57,7 +57,27 @@ export const TransactionBase = (props: TransactionBaseProps) => {
...
+        break;
+
+      case NetworkType.polygon:
+      case NetworkType.mumbai:
+        return 'Polygonscan';
+        break;
+    }
+  };
+
+  const blockExplorerName = getBlockExporerName(network);
```
oidoug: I think it would be preferable to extract the network configuration into a "network config" file. Same for `poligonNode` defined in web3-provider below. Maybe having these as simple objects is better, clearear, even though it would have duplicated info.

cardstack/src/models/web3-provider.ts
```diff
@@ -26,13 +29,21 @@ const Web3WsProvider = {
...
+        hubConfigResponse,
+        currentNetwork
+      );
+
+      const polygonNode =
+        currentNetwork === NetworkType.polygon
+          ? 'wss://ws-matic-mainnet.chainstacklabs.com'
+          : currentNetwork === NetworkType.mumbai
+          ? 'wss://ws-matic-mumbai.chainstacklabs.com'
+          : rpcNodeWssUrl;
```
oidoug: Same as mentioned above for block explorer. I think a simpler config file would make it more obvious the different options for different networks.

Maybe there are more similar string configs we could pull off to a config file, for a future PR.

src/hooks/useInitializeAccountData.js
```diff
@@ -16,7 +16,7 @@ export default function useInitializeAccountData() {
       await Promise.all([fetchMainAssets, fetchCollectibles]);
       logger.sentry('Initialize account data and collectibles ');
     } catch (error) {
-      logger.sentry('Error initializing account data');
+      logger.sentry('Error initializing account data', error);
```
oidoug: Logging this error value along the message in sentry might be risky, concerning user privacy, unless we are certain it never contain any user-info.

src/hooks/useLoadAccountData.js
```diff
@@ -15,22 +15,16 @@ export default function useLoadAccountData() {
...
-      const promises = [];
-      if (isMainnet(network)) {
-        const p2 = dispatch(collectiblesLoadState());
-        promises.push(p2);
-      }
-      const p3 = dispatch(requestsLoadState());
-      const p4 = dispatch(walletConnectLoadState());
-
-      promises.push(p3, p4);
+      const p1 = dispatch(collectiblesLoadState());
```
oidoug, line 23: I think `collectiblesLoadState` should only be runned in ethereum main.

src/redux/fallbackExplorer.js
```diff
@@ -315,7 +315,8 @@ export const fetchAssetsBalancesAndPrices = async () => {
     fallbackExplorerBalancesHandle: currentBalancesTimeout,
   } = store.getState().fallbackExplorer;
 
-  const assets = isMainnet(network) ? currentAssets : testnetAssets[network];
+  const assets =
+    network !== NetworkType.sokol ? currentAssets : testnetAssets.sokol;
```
oidoug, line 319: I didn't understand the change here. Why use `currentAssets` for the other testnets except on sokol? Could you elaborate?

src/redux/fallbackExplorer.js
```diff
@@ -410,26 +410,21 @@ export const fetchAssetsBalancesAndPrices = async () => {
...
-    const newAssets = await findAssetsToWatch(
-      accountAddress,
-      latestTxBlockNumber,
-      dispatch
-    );
+  const newAssets = await findAssetsToWatch(
+    accountAddress,
+    latestTxBlockNumber,
+    dispatch
+  );
```
oidoug, line 420: Will `findAssetsToWatch` work for the other networks now?

oidoug (commented):
It's looking great, I'm loving the change from strict network checking to a more network agnostic code 👏 
No major change request, but I have some concerns and questions for you below.

cardstack/src/models/web3-provider.ts
paulinhapenedo: This is a workaround while Fadhlan was working on the hub to do these changes. I commited this just in case any of you were testing locally the app. It will be removed now.

To be clearer: the rpcNodeWssUrl for both Polygon and Mumbai in the hub are not working.

oidoug:
Found a change in behaviour that you might want to look into:

Previous network switcher screen would show testnets on open if user was on a testnet:

https://user-images.githubusercontent.com/129619/203556228-2401eea5-96ce-4e6c-9b92-018c5c5c3249.mp4

Current implementation does not, so it looks odd:

https://user-images.githubusercontent.com/129619/203556363-d9938164-203e-4cc1-b161-d385aa5233c8.mp4

cardstack/src/components/Transactions/TransactionBase.tsx
paulinhapenedo: This is not a good solution, I know, but I don't see any advantages into abstracting this into a file to be used just here. Looks a bit like premature optimization. I prefer to abstract things when I'm sure I'm gonna use it in another file.

danibonilha: Or we could just use a generic name, like `[Network] Blockexplorer`

src/hooks/useInitializeAccountData.js
paulinhapenedo: The only user info we use in these two functions (`fallbackExplorerInit` and `collectiblesRefreshState`) are the account address, which is public, but if you believe it's better to remove it, I'll do it.

cardstack/src/components/Transactions/TransactionBase.tsx
oidoug: The point here is not abstraction or optimization, but making config values more visible. Personal preference though, so I'm okay as is.

src/components/settings-menu/NetworkSection/useNetworkSection.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { settingsUpdateNetwork } from '../../../redux/settings';
+import { Container, FloatingTag, RadioItemProps } from '@cardstack/components';
+import { NetworkType } from '@cardstack/types';
+import { CARDPAY_SUPPORTED_NETWORKS, isMainnet } from '@cardstack/utils';
+
+export const useNetworkSection = () => {
+  const { network } = useAccountSettings();
+  const dispatch = useDispatch();
+  const [selectedNetwork, setSelectedNetwork] = useState(network);
+  const [isShowAllNetworks, setShowAllNetworks] = useState<boolean>(false);
```
danibonilha: nit: isShowingAll or showAll

src/redux/fallbackExplorer.js
paulinhapenedo: We need that for Sokol because of the "Other tokens" section.

src/hooks/useLoadAccountData.js
danibonilha: collectiblesLoadState only gets cached data from the storage, so the network really doesn't matter, and openSea is compatible with polygon too.

paulinhapenedo: But we weren't running only on Ethereum mainnet, we were running on all mainnets (ethereum, gnosis and would be running on Polygon as well).

src/hooks/useInitializeAccountData.js
oidoug: Looks improbable that fallbackExplorerInit will throw an error with the address in it, but better safe than sorry 🤷

src/redux/fallbackExplorer.js
paulinhapenedo: Tested locally with the account that I have more assets and it works.

src/hooks/useLoadAccountData.js
danibonilha: Which reminds me of the CollectibleHeader where we have a filter so mainnet only to show `View on OpenSea` which is also valid for polygon https://support.opensea.io/hc/en-us/articles/4404027708051-Which-blockchains-are-compatible-with-OpenSea-

cardstack/src/components/Transactions/TransactionBase.tsx
paulinhapenedo: > Or we could just use a generic name, like `[Network] Blockexplorer`

What do you think about this suggestion, @oidoug ?

src/redux/fallbackExplorer.js
danibonilha: Sokol is a funky network and unfortunately the api we use doesn't return the token information, being, the token address, coingeko id, etc for our .cpxd tokens, so we kind of need to mock this information in other to show it on Other Tokens balances

src/redux/fallbackExplorer.js
danibonilha: This was basically filtering out `sokol` if the sokol other tokens sections keeps working, the answer is yes.

src/hooks/useLoadAccountData.js
paulinhapenedo: Changed the CollectibleHeader here https://github.com/cardstack/cardwallet/pull/1099/commits/8d6d61253ec13989b915382a61208a98d7e0a377

paulinhapenedo:
@oidoug reverted the changes on the NetworkScreen.

oidoug (approved):
LGTM 🚢

PR #1100, opened 2022-11-22
[CS-4904]: List WalletConnect 2.0 Sessions and Pairings

danibonilha:
### Description

This PR adds a new item on settings in order to list the new wallet connect sessions. Now we have sessions and pairings, there's no design for it, so for now I'm simply listing all in the same place with a different label. It also removes some components that were not being used.
There's still some events, and listing the sessions details for example, but since there's no design yet, I'm keeping it simple, just to have it working.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Simulator Screen Recording - iPhone 12 mini - 2022-11-22 at 18 19 03


oidoug (commented):
@danibonilha , I see the new screen code looks a lot cleaner, but IMO, UI wise it would be better to have both v1 and v2 in the same screen and separated by sections.

danibonilha:
> @danibonilha , I see the new screen code looks a lot cleaner, but IMO, UI wise it would be better to have both v1 and v2 in the same screen and separated by sections.

@oidoug  I think it would be messier to create a section there, for two reasons, one there's no dApp right now that supports 2.0, besides the wallet connect example and it doesn't have proper naming etc, which makes us render the url instead of the name and takes more space, making it impossible to use the current Disconnect button, which was "replaced" on the new screen with a trash can. The second reason is that v1 will be totally removed once DApps are migrated Feb next year, so the old section will be gone too, so I don't think it's worth spending a lot of time now trying to make both of the code structures work together since they're different . Based on this facts it's best to have separate screens because it gives more speed and flexibility to us to hide/remove. For example, we might want to ship a new version next week, but the WC 2.0 integration is not entirely done, if we want to, we can just hide the item from the list, whereas having sections in the same screen would be more cumbersome to handle. Plus we might need a complete redesign to give more information on each session and pairing.

oidoug:
> 

Ok, that makes perfect sense 👍

oidoug (approved):
LGTM 🚀

PR #1101, opened 2022-11-22
[CS-4831] Add claim level explanations to reward rows

oidoug:
### Description

This PR adds explanability to individual reward proofs. It needs to transform and parse `explanationData` and `explanationTemplate` into a readable string.

It is known that the transformation is not scalable at this version, because if new values are needed, an update to the transformation is required. A better solution would be to simplify the explanation stored in the proof or go with an even more complex way to provide the transformation function along the proof's values.

NOTE: {to_block} is an error in the `explationTemplate`, it should be {end_block}. It need to be fixed where it's generated.

- [x] Completes #(CS-4831)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot


cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -101,29 +109,60 @@ export const fetchValidProofsWithToken = async ({
...
+        findProof => findProof.rootHash === currentProof.rootHash
+      );
+
+      const proof: RewardProofType = claimingInfo
+        ? {
+            ...claimingInfo,
+            ...tokenWithPrice,
+            isClaimable: true,
+          }
+        : { ...currentProof, ...tokenWithPrice, isClaimable: false };
```
danibonilha: how about 
```ts
     const claimingInfo =
        claimableProofs.find(
          findProof => findProof.rootHash === proof.rootHash
        ) || {};

      const proof: RewardProofType  = {
          ...currentProof,
          ...tokenWithPrice,
          ...claimingInfo,
          isClaimable: !!claimingInfo.rootHash,
      };
```

cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -77,6 +79,12 @@ const getValidProofs = async ({
   return validProofs;
 };
 
+const getTokenSymbolForAddress = async (tokenAddress: string) => {
+  const rewardPoolInstance = await getRewardsPoolInstance();
+
+  return rewardPoolInstance.addTokenSymbol([{ tokenAddress }]);
```
danibonilha: if we're gonna use for a single token, maybe we could already return the first value, instead of mapping in the fetchProofs function

cardstack/src/services/utils/reward-explanation.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * parseTemplateExplanation
+ * Taken from https://gist.github.com/smeijer/6580740a0ff468960a5257108af1384e?permalink_comment_id=3437778#gistcomment-3437778
+ * @param template string such as in explanationTemplate: Earned {amount} {token} for a one-time aidrop. {rollover_amount} tokens.
+ * @param data obejct such as explanationData: { "amount": "100.0", "token": "0xB0427e9F03Eb448D030bE3EBC96F423857ceEb2f" }
```
danibonilha: object

cardstack/src/services/utils/reward-explanation.ts
```diff
@@ -0,0 +1,14 @@
...
+ * @returns formatted string such as: Earned 100.0 CARD.CPXD for a one-time aidrop. {rollover_amount} tokens.
+ */
+export const parseTemplateExplanation = (template: string, data: any) => {
```
danibonilha: should data be at least an optional object ?  or Record<string, string>

cardstack/src/services/utils/reward-explanation.ts
```diff
@@ -0,0 +1,14 @@
...
+  return template.replace(
+    /\{([^}]+)}/g,
+    (m, p) =>
+      p.split('.').reduce((a: any, f: any) => (a ? a[f] : undefined), data) ?? m
+  );
```
danibonilha: What does m and p stands for ?

danibonilha: And a and f ? And for the typing, isn't it string ?

cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -101,29 +109,60 @@ export const fetchValidProofsWithToken = async ({
...
+      if (proof.explanationData?.amount) {
+        proof.explanationData.amount = parseFloat(
+          fromWei(proof.explanationData.amount)
+        ).toFixed(3);
+      }
+
+      if (proof.explanationData?.rollover_amount) {
+        proof.explanationData.rollover_amount = parseFloat(
+          fromWei(proof.explanationData.rollover_amount)
+        ).toFixed(3);
```
danibonilha: why 3 places ? if we were to use 2, we could transform this into a util that receives a string in wei, `(value: string) => parseFloat(fromWei(value)).toFixed(2)` and reuse on the other methods in this file

cardstack/src/services/utils/__tests__/reward-explanation.test.ts
```diff
@@ -0,0 +1,36 @@
+import { parseTemplateExplanation } from '../reward-explanation';
+
+describe('Rewards explanation', () => {
+  it('should parse a proof template with a single value data entry', () => {
+    const template = 'Template {a}';
+    const data = { a: 'Value A' };
+
+    expect(parseTemplateExplanation(template, data)).toMatch(
+      'Template Value A'
+    );
```
danibonilha: I believe using `toBe ` is a safer assertion for this test, because toMatch doesn't deeply checks it, it returns true if it contains the value, but if other unwanted values were added we wouldn't know

cardstack/src/services/rewards-center/rewards-center-service.ts
oidoug: Hum, looks better but then we get a lint error on `rootHash` not existing in `{}` 😅

cardstack/src/services/utils/reward-explanation.ts
oidoug: Kinda overkill, but since this was picked from the linked implementation, it already provides parsing of nested values, like `{ a: { b: "value" } }` in a template like `{a.b}`. So for that to work, I think data needs to be any (?)

cardstack/src/services/utils/reward-explanation.ts
oidoug: Yeah, nomenclature is week here, I didn't fully understand what it did until you asked. This is my best guess:

m is the string token with the brackets;
p is the token "key" without brackets.

It then reduces on the key so nested keys are also replaced, like `{ a: { b: "value" } }` in a template like `{a.b}`.

I'm adding a test to cover that too.

cardstack/src/services/rewards-center/rewards-center-service.ts
oidoug: I used 3 to keep the value displayed consistent with the value above the explanation, which shows 3 decimals. 2 looks better though.
I'll put in a utils as suggested.

cardstack/src/services/utils/__tests__/reward-explanation.test.ts
oidoug: `parseTemplateExplanation` function always returns a string, so using `toMatch` seems alright.

danibonilha: Here's a comparison using both assertions with a simple string 
image
 As you can see the second assertion passes, even though is not what we want, it just starts with the correct values, but has other things we don't want, when we use `toBe` we make sure that the value is the same, which is erroring in the last assertion, because it doesn't "match"

cardstack/src/services/utils/reward-explanation.ts
danibonilha: it would be nice to add meaningful names

cardstack/src/services/utils/reward-explanation.ts
danibonilha: Maybe we can try Record<string, unknown>, if it doesn't work it's ok to leave as is

cardstack/src/services/rewards-center/rewards-center-service.ts
danibonilha: then casting the empty object  to proof might work `({} as Proof);`

cardstack/src/services/utils/__tests__/reward-explanation.test.ts
oidoug: Ahhh got it... I thought it matched the string whole, thanks for taking the time to explain 👍

cardstack/src/services/rewards-center/rewards-center-service.ts
oidoug: Good tip, that works surelly.

danibonilha (approved):
🚀

PR #1102, opened 2022-11-24
[Refactor]: Bump sdk and replace functions

danibonilha:
### Description
This PR  bumps the sdk the updated network and requests functions.

- There was an issue with unsupported accounts, connecting to the webProvider, I noticed the `getNetwork` might be called  before fully updating the redux state, so I added the check inside `getNetwork` to avoid erroring.
- There was another issue with rewards explainer being called on previous networks and throwing an error about inexistent reward program, so I added the skip to the query to only fetch once the account is ready and it's cardpay.
-  Replaced CardPayCompatible with sdk function.
- Added the `noCardPayAccoun`t and `isOnCarPayNetwork` flags to the `useAccountSettings` hooks, since we were duplicating the skip on multiple hooks and the useAccount is already used on those places, but keep in mind I didn't replace all the functions from the sdk with the hook since  in some places makes more sense to use the function directly.


paulinhapenedo (approved):
Didn't know about `isOnCardPayNetwork`. It's way better! 👏

src/components/settings-menu/NetworkSection/useNetworkSection.tsx
```diff
@@ -56,17 +56,15 @@ export const useNetworkSection = () => {
         disabled: false,
         hasTags: networkTags({
           isTestnet: !isMainnet(network as NetworkType),
-          hasCardPay: CARDPAY_SUPPORTED_NETWORKS.includes(
-            network as NetworkType
-          ),
+          hasCardPay: isOnCardPayNetwork,
```
paulinhapenedo: This one needs to change because it depends on each one of the networks in the map, otherwise we will only show the `CardPay` tag when we are using the app on a CardPay supported network.

Simulator Screen Shot - iPhone 14 Pro - 2022-11-24 at 17 14 03

danibonilha: oh good catch! i'll fix it

PR #1104, opened 2022-11-25
[Refactor]: Add wc on account init and remove dead code 

danibonilha:
### Description
This PR moves the wc 2.0 initialization from appInit to account init, so it happens while the rest of the app is still loading. In order to do that, some refactoring was made: 
- Added new useInitializeAccount hook which combines two old hooks useLoadData and useInitializeAccountData. 
- Moved the contact loading to the sendButton, since the contacts are only showed while sending an asset
- Added missing typings on SendButton 
- Removed old navigation hook wrapper
- Removed a bunch of unused components and hooks
- Added `mapDispatchToActions` to create an array with dispatch(action) so it's easier to trigger multiple actions. Now we can stop using p1,p2,p3...The way that it accepts action args is a bit funky but since it's on a single place I think it's not worth to spend too much time adapting that for an object or something, if it gets confusing and there's need on other places we can refactor it.
- Added es2020 support to use Promise.allSettled and remove promiseAllWithFails.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


src/components/sheet/sheet-action-buttons/SendActionButton.tsx
```diff
@@ -1,11 +1,30 @@
+import { useNavigation } from '@react-navigation/native';
+//Rb side doesn't know how to handle globals
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { OptionalUnion } from 'globals';
```
oidoug, line 4: Maybe you could add this to Rainbow's `globalVariables.js`.

src/hooks/useInitializeAccount.ts
```diff
@@ -0,0 +1,60 @@
...
+      settingsLoadCurrency,
+      imageMetadataCacheLoadState,
+      requestsLoadState,
+      walletConnectLoadState,
+      ...(isOnCardPayNetwork ? [dataLoadState] : []), // Update for cardpay happens on fetchSafes
+    ];
+
+    const promises = mapDispatchToActions(dispatch, actions);
+
+    return Promise.allSettled(promises);
```
oidoug, line 41: interesting 👀

src/components/sheet/sheet-action-buttons/SendActionButton.tsx
```diff
@@ -15,28 +34,24 @@ export default function SendActionButton({
...
-      iconProps={{
-        iconSize: 'medium',
-        marginRight: 2,
-        name: 'send',
-        top: 2,
-      }}
+      iconProps={iconProps}
       onPress={handlePress}
-      {...variantProp}
+      variant={small ? 'small' : undefined}
```
paulinhapenedo: I think we can remove this since it seems like there's only one place where we are passing it (`TokenWithChartSheet`) and it's set as `false`.

oidoug (approved):
LGTM, interesting use of all-settled, I didn't know about it.

paulinhapenedo (approved):

src/components/sheet/sheet-action-buttons/SendActionButton.tsx
danibonilha: Good observation!

src/components/sheet/sheet-action-buttons/SendActionButton.tsx
danibonilha: Doesn't work 😞

PR #1105, opened 2022-11-25
[CS-4909] Enable Activities in non-CardPay compatible networks

paulinhapenedo:
### Description
This PR fixes the Activities tab in non-CardPay compatible networks.
We are using a workaround to not flicker the empty state while we wait to load the `TransactionList`. The problem is that for networks that are not CardPay compatible, the `isLoadingTransactions` is already `false` when we load the screen, which makes the `isLoadingFallback` never be changed to `false` on the  `useEffect`.

- [x] Completes #CS-4909

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Gnosis

https://user-images.githubusercontent.com/690904/204051961-d5d5a207-d10c-4c4f-9948-358f1b2eec1f.mp4

### Ethereum

https://user-images.githubusercontent.com/690904/204051981-48835b6b-f096-4a90-b766-d713e9909ac0.mp4

### DIFF
--- a/cardstack/src/components/TransactionList/TransactionList.tsx
+++ b/cardstack/src/components/TransactionList/TransactionList.tsx
@@ -1,5 +1,5 @@
 import { useFocusEffect } from '@react-navigation/native';
-import React, { memo, useCallback, useEffect, useRef } from 'react';
+import React, { memo, useCallback, useMemo } from 'react';
 import { RefreshControl, SectionList, ActivityIndicator } from 'react-native';
 
 import {
@@ -33,15 +33,6 @@ export const TransactionList = memo(({ Header }: TransactionListProps) => {
     refetchLoading,
   } = useFullTransactionList();
 
-  const isLoadingFallback = useRef(true);
-
-  useEffect(() => {
-    if (!isLoadingTransactions) {
-      // Once tx are loading we don't need to track anymore
-      isLoadingFallback.current = false;
-    }
-  }, [isLoadingTransactions]);
-
   const renderSectionHeader = useCallback(
     ({ section: { title } }: { section: { title: string } }) => (
       <Container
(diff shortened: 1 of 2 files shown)


danibonilha (commented):
It seems the problem might be a bit bigger than this. I can't seem to find where we are fetching the transactions on other networks, it seems we are not 🤔 it just tries to get from the local storage, but I don't see any action on init that saves the tx on the storage in the first place. So we need to decide if it makes sense to track the txs for other networks or remove it entirely

paulinhapenedo:
Updated the approach as discussed in the mobile meeting.
If the user has any transaction in the local storage, we will show it. If the user doesn't have any transactions and it's on a network that there's no CardPay support, we will show a custom message.

cardstack/src/components/TransactionList/TransactionList.tsx
```diff
@@ -31,7 +36,7 @@ export const TransactionList = memo(({ Header }: TransactionListProps) => {
   const isLoadingFallback = useRef(true);
 
   useEffect(() => {
-    if (isLoadingTransactions) {
+    if (!isLoadingTransactions) {
```
danibonilha: With this change the flick between empty component and loading is back, if you want to ignore this case, I guess it's okay but then we don't need any of this `isLoadingFallback`

cardstack/src/components/TransactionList/TransactionList.tsx
```diff
@@ -63,20 +68,18 @@ export const TransactionList = memo(({ Header }: TransactionListProps) => {
     []
   );
 
+  const title = isOnCardPayNetwork
+    ? strings.emptyComponent
+    : strings.nonCardPayNetwork(network);
```
danibonilha: nit: we could useMemo

danibonilha (approved):

PR #1107, opened 2022-11-28
[CS-4539] Migrate notification api calls to RTK-Query

oidoug:
### Description

This PR migrates notification queries from root hub-service file to a new hub/notifications injected endpoint group, refactors the notification settings screen to use the new hooks, and adds an optimistic update for notifications preferences mutation so the switcher doesn't jump between states.

It also renames services/hub-service to hub-token-service, since it's only providing the Hub token by authenticating within the SDK.

- [x] Completes #(CS-4539)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/204381156-fc8e510c-cda1-4422-947c-9cb980f2f65f.mp4


cardstack/src/hooks/notifications-preferences/useUpdateNotificationPreferences.ts
```diff
@@ -14,57 +13,28 @@ export enum NotificationsOptionsStrings {
 }
 
 export const useUpdateNotificationPreferences = () => {
-  const { authToken, isLoading, error } = useAuthToken();
+  const {
+    data: options = [],
+    error: getError,
```
danibonilha: nit: getError might be misleading, if we want to use this info just as a boolean value, we can use the isError property

cardstack/src/hooks/prepaid-card/useAuthToken.ts
```diff
@@ -1,30 +0,0 @@
...
-  }, [getAuthToken]);
-
-  useEffect(() => {
-    if (error) {
-      logger.log('Error getting auth token', error);
-    }
-  }, [error]);
-
-  return { authToken, isLoading, error };
-};
```
danibonilha, line 30: So happy to finally get rid of this 🙌

cardstack/src/services/hub-token-service.ts
```diff
@@ -0,0 +1,89 @@
...
+) => {
+  // expires in a day.
+  await saveLocal(
+    tokenStorageKey,
+    { data: { authToken } },
+    1000 * 3600 * 24,
+    undefined,
+    walletAddress
+  );
+};
```
danibonilha, line 53: Shouldn't we be storing the token on keychain ?

cardstack/src/services/hub/notifications/hub-notifications-api.ts
```diff
@@ -0,0 +1,95 @@
...
+        body: hubBodyBuilder('push-notification-registration', {
+          'push-client-id': fcmToken,
+        }),
+      }),
+    }),
+    unregisterFcmToken: builder.mutation<string, RegisterFCMTokenQueryParams>({
+      query: ({ fcmToken }) => ({
+        url: `${routes.registerFCMToken}/${fcmToken}`,
+        method: 'DELETE',
+        responseHandler: response => response.text(),
```
danibonilha, line 32: Could this one use the `appendFMC` flag ?

cardstack/src/hooks/notifications-preferences/useUpdateNotificationPreferences.ts
```diff
@@ -14,57 +13,28 @@ export enum NotificationsOptionsStrings {
...
-      const updateItem = updateList?.find(
-        o =>
-          o.attributes['notification-type'] ===
-          item.attributes['notification-type']
-      );
-
-      if (authToken && updateItem) {
-        await setNotificationsPreferences(authToken, updateItem);
-      }
+    async (item: NotificationsPreferenceDataType, value: boolean) => {
```
danibonilha: nit: does value mean `isEnabled` ?, maybe we can use this as the name

cardstack/src/hooks/notifications-preferences/useUpdateNotificationPreferences.ts
```diff
@@ -14,57 +13,28 @@ export enum NotificationsOptionsStrings {
...
 
-  useEffect(() => {
-    if (!isLoading && authToken) fetchNotificationPreferences();
-  }, [authToken, isLoading, fetchNotificationPreferences]);
-
   return {
     options,
     onUpdateOptionStatus,
-    error,
+    error: getError && 'Failed to get notificition preferences',
```
danibonilha: if the error is a default string maybe we could use directly on the component and just use tthe flag to conditional render it

danibonilha (commented):
It's looking good, great cleanup

cardstack/src/hooks/notifications-preferences/useUpdateNotificationPreferences.ts
oidoug: oops, I was going to clean up this error but forgot

cardstack/src/services/hub-token-service.ts
oidoug: Hum, now that the user don't need to re-authenticate to get keychain data, it makes sense to move this as well. Probably better suited for another PR because I don't know how much work that would be. I'm adding a card.

oidoug: https://linear.app/cardstack/issue/CS-4924/[debt]-move-hubs-auth-token-to-keychain

cardstack/src/services/hub/notifications/hub-notifications-api.ts
oidoug: For sure! Oversight from me.

danibonilha (approved):
🚢

PR #1108, opened 2022-11-29
[CS-4923]: Add support to .allSettled to fix wallet not loading

danibonilha:
### Description

This PR fixes the wallet not loading issue, which was caused by using `Promise.allSettled`, the release build couldn't find this function. `.allSettled` was introduced in the [promise](https://github.com/then/promise)` 8.2.0` version, and react native uses the latest version but there are other older packages which uses older versions like 7.x.x. On the release build there was a confusion between those packages, now by setting the resolution to the latest version we can make sure the promise version used will always be `8.3.0` which is compatible with .allSettled. 
As a bonus I've also added the wc2.0 init fix so we can ship a new version once this PR is merged

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/src/hooks/useInitializeAccount.ts
+++ b/src/hooks/useInitializeAccount.ts
@@ -1,5 +1,5 @@
 import { captureException } from '@sentry/react-native';
-import { useCallback } from 'react';
+import { useCallback, useEffect } from 'react';
 import { useDispatch } from 'react-redux';
 
 import { dataLoadState } from '../redux/data';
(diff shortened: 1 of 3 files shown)


src/hooks/useInitializeAccount.ts
```diff
@@ -22,11 +22,13 @@ export default function useInitializeAccount() {
...
 
+  useEffect(() => {
+    if (accountAddress) WalletConnect.init(accountAddress);
+  }, [accountAddress]);
+
   const loadAccountData = useCallback(async () => {
     logger.sentry('Load wallet account data');
 
-    await WalletConnect.init(accountAddress);
-
```
oidoug, line 29: Look like 'accountAddress' isn't needed in the dep. array anymore. Or does this callback need to be re-instantiated when accountAddress changes?

oidoug (approved):
LGTM! 
There's a lint warning you should check.

PR #1110, opened 2022-11-30
[CS-4920]: Handle WC 2.0 requests

danibonilha:
### Description

This PR handles the wallet connect 2.0 requests and refactors slightly the confirmation functionality. it also removes duplicated logic, deprecated methods and merges some simple hooks. While working on this I noticed a couple of sign message requests on v1 were broken for a while, with these changes it's working again. 
You can validate the signing requests for v2 [here](https://react-app.walletconnect.com/)  and v1 [here](https://example.walletconnect.org/) or use cardpay requests. For the `eth_sendTransaction` method you need to have enough balance.

FYI there's lots of refactor to be done still but I chose to keep it as simple as possible now, so we can have both versions working, and once 2.0 becomes the official one we can cleanup a bunch of things, also there are a couple of missing information such as dappName, chainId on the signing request, which will be added on another PR.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Example of one type of request on 2.0:

https://user-images.githubusercontent.com/20520102/204909432-aba2afa4-d10c-4dc2-8909-85d9da0d9d29.mov


VitaliKirpu:
This is cool!

oidoug (approved):
LGTM! Great cleanup 🔥

PR #1111, opened 2022-11-30
[CS-4930]: Explicitly set "promise" on globals

danibonilha:
### Description

This PR fixes the wallet load on dev for Android, from what I could gather it seems there's some ordering issue or some packaging is overriding the `Promise` variable, so now we check for allSettled, if it doesn't exists it means it's using the wrong type of promise and it resets to use the promise package.
From yarn why, we can see that there's only one version which is the latest and contains allSettled, so this approach seems reasonable 
image

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


paulinhapenedo (approved):
Tested locally and it works!

PR #1112, opened 2022-12-01
[CS-4910] Update OpenSea URL to work with Polygon too

paulinhapenedo:
### Description
This PR updates the OpenSea URL to work with Polygon and other mainnets too. From the [OpenSea Docs](https://docs.opensea.io/reference/api-overview), we should be either using the API without prefix for mainnets and with `testnets-` for testnets, not the with the network name.

- [x] Completes #CS-4910

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


danibonilha (approved):
This will work correctly for goerli, although by re-reading the docs I'm not sure this api supports polygon/mumbai. They state in the support that OpenSea itself does, and from what I could gather, their stream api supports it, and in their website they use graphQl, but since there's no info about the network in this request, I think it's only ethereum/goerli.

PR #1113, opened 2022-12-02
[CS-4932]: Show dapp name and transaction chain on wc request 2.0

danibonilha:
### Description
This PR shows the chain and it's native token balance on the transactionSheet, for that, we needed a new instance of web3 
to independently check the balances (Let's observe the app behavior if lots of disconnected happens or pollingTracker errors), then if it's troublesome we can revisit and improve the new instance creation, there are some refactors on a useAccountProfile,  and I've switched the etherWeb3Provider setting order, so it doesn't lock the user on gnosis when switching network fails, although I think we can add a better handling for that but tbd.

### Screenshots

https://user-images.githubusercontent.com/20520102/205321157-f884c724-3608-4854-92c8-4adc7ea7b0cf.mov


cardstack/src/components/NetworkBadge/NetworkBadge.tsx
```diff
@@ -8,19 +8,21 @@ import { useRainbowSelector } from '@rainbow-me/redux/hooks';
...
+    getConstantByNetwork('name', props.network || currentNetwork)
   ).toUpperCase();
 
   return (
     <Container flexDirection="row" {...props}>
       <Container
         backgroundColor="backgroundLightGray"
         paddingHorizontal={2}
-        style={{ paddingVertical: 1 }}
+        paddingVertical={1}
```
oidoug, line 25: Just to point it out, we were using style prop so we could pad just 1 pixel vertically, now the banner has a 4px padding. If you think it looks better this way, I don't mind the change.

cardstack/src/components/TransactionConfirmationSheet/displays/GenericDisplay.tsx
```diff
@@ -2,30 +2,53 @@ import { getConstantByNetwork } from '@cardstack/cardpay-sdk';
...
-};
+export const GenericDisplay = ({
+  txNetwork,
+  messageRequest,
+}: TransactionConfirmationDisplayProps) => (
+  <>
+    <FromSection txNetwork={txNetwork} />
+    <HorizontalDivider />
+    <Container width="100%">
+      <SectionHeaderText>MESSAGE</SectionHeaderText>
```
oidoug, line 23: supernit: extract string

src/handlers/web3.js
```diff
@@ -54,13 +54,13 @@ export const etherWeb3SetHttpProvider = async network => {
...
 export const getEtherWeb3Provider = async (network = undefined) => {
-  let wsConnected = web3Provider.provider?.connected;
+  let wsConnected = web3Provider?.provider?.connected;
 
   // check websocket state and reconnect if disconnected
   while (!wsConnected) {
     const currentNetwork = network || (await getNetwork());
     await etherWeb3SetHttpProvider(currentNetwork);
-    wsConnected = web3Provider.provider?.connected;
+    wsConnected = web3Provider?.provider?.connected;
```
oidoug, line 63: I wonder if web3Provider is still undefined at this point, shouldn't we try to (re)instantiate it?

oidoug (approved):
LGTM 🚀  
Really nice improvement, a step towards simultaneous network support.
Tested, all looking good.

cardstack/src/components/TransactionConfirmationSheet/displays/GenericDisplay.tsx
danibonilha: There are lot's of MESSAGE/FROM strings across the project we should do it on a diff task

src/handlers/web3.js
danibonilha: This file is .js, I've just added a couple of optional chaining to make sure nothing breaks, there is already a loop to check if it's connected and it tries to (re)instantiate

PR #1114, opened 2022-12-02
[CS-4725] UI: Screen for enabling push notifications and permission type selection

oidoug:
### Description

Adds `NotificationsPermissionScreen` that shows available notification types options for user to select as they are asked to enable notifications.

- [x] Completes #(CS-4725)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot

Screenshot

Screenshot

### DIFF
--- a/cardstack/src/models/firebase.ts
+++ b/cardstack/src/models/firebase.ts
@@ -25,6 +25,26 @@ type FCMTokenStorageType = {
 const getPermissionStatus = (): Promise<FirebaseMessagingTypes.AuthorizationStatus> =>
   messaging().hasPermission();
 
+export const needsToAskForNotificationsPermissions = async (): Promise<
+  boolean | undefined
+> => {
+  let permissionStatus = null;
+
+  try {
+    permissionStatus = await getPermissionStatus();
+
+    return (
+      permissionStatus !== messaging.AuthorizationStatus.AUTHORIZED &&
+      permissionStatus !== messaging.AuthorizationStatus.PROVISIONAL
+    );
+  } catch (error) {
+    logger.sentry(
+      'Error checking if a user has push notifications permission',
+      error
+    );
+  }
+};
+
 export const getFCMToken = async (): Promise<FCMTokenStorageType> => {
   try {
     const {
@@ -166,21 +186,7 @@ export const requestPermission = () =>
 
 export const checkPushPermissionAndRegisterToken = async () => {
   return new Promise(async resolve => {
-    let permissionStatus = null;
-
-    try {
-      permissionStatus = await getPermissionStatus();
-    } catch (error) {
-      logger.sentry(
-        'Error checking if a user has push notifications permission',
-        error
-      );
-    }
-
-    if (
-      permissionStatus !== messaging.AuthorizationStatus.AUTHORIZED &&
-      permissionStatus !== messaging.AuthorizationStatus.PROVISIONAL
-    ) {
+    if (await needsToAskForNotificationsPermissions()) {
       Alert({
         buttons: [
           {
--- a/cardstack/src/models/firebase.ts
+++ b/cardstack/src/models/firebase.ts
@@ -28,15 +28,14 @@ const getPermissionStatus = (): Promise<FirebaseMessagingTypes.AuthorizationStat
 export const needsToAskForNotificationsPermissions = async (): Promise<
   boolean | undefined
 > => {
-  let permissionStatus = null;
-
   try {
-    permissionStatus = await getPermissionStatus();
+    const { AUTHORIZED, PROVISIONAL } = messaging.AuthorizationStatus;
 
-    return (
-      permissionStatus !== messaging.AuthorizationStatus.AUTHORIZED &&
-      permissionStatus !== messaging.AuthorizationStatus.PROVISIONAL
-    );
+    const permissionStatus = await getPermissionStatus();
+
+    const isAuthorized = [AUTHORIZED, PROVISIONAL].includes(permissionStatus);
+
+    return !isAuthorized;
   } catch (error) {
     logger.sentry(
       'Error checking if a user has push notifications permission',
--- a/cardstack/src/screens/NotificationsPermissionScreen/NotificationsPermissionScreen.tsx
+++ b/cardstack/src/screens/NotificationsPermissionScreen/NotificationsPermissionScreen.tsx
@@ -41,17 +41,18 @@ const NotificationsPermissionScreen = () => {
         <>
           <HorizontalDivider backgroundColor="blueDarkest" />
           <Container>
-            <Checkbox checkboxPosition="left">
-              <Text color="white">{notificationType}</Text>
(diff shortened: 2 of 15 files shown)


cardstack/src/models/firebase.ts
```diff
@@ -25,6 +25,26 @@ type FCMTokenStorageType = {
...
+      permissionStatus !== messaging.AuthorizationStatus.PROVISIONAL
+    );
+  } catch (error) {
+    logger.sentry(
+      'Error checking if a user has push notifications permission',
+      error
+    );
+  }
+};
+
```
oidoug, line 46: To be used in `useWalletManager`.

cardstack/src/models/firebase.ts
```diff
@@ -25,6 +25,26 @@ type FCMTokenStorageType = {
 const getPermissionStatus = (): Promise<FirebaseMessagingTypes.AuthorizationStatus> =>
   messaging().hasPermission();
 
+export const needsToAskForNotificationsPermissions = async (): Promise<
+  boolean | undefined
+> => {
+  let permissionStatus = null;
+
+  try {
+    permissionStatus = await getPermissionStatus();
```
danibonilha: we could probably make this a const and remove the let on the top since this value is never reassigned, nor checked by null

cardstack/src/models/firebase.ts
```diff
@@ -25,6 +25,26 @@ type FCMTokenStorageType = {
...
+
+    return (
+      permissionStatus !== messaging.AuthorizationStatus.AUTHORIZED &&
+      permissionStatus !== messaging.AuthorizationStatus.PROVISIONAL
+    );
```
danibonilha: we can destructure the authStatus to make it less verbose, and even use the includes function to check since it's the same value.

cardstack/src/models/firebase.ts
danibonilha:
```suggestion
    try {
      const { AUTHORIZED, PROVISIONAL } = messaging.AuthorizationStatus;

      const permissionStatus = await getPermissionStatus();

      const isAuthorized = [AUTHORIZED, PROVISIONAL].includes(permissionStatus);

      return !isAuthorized;
    } catch (error) {
      logger.sentry(
        'Error checking if a user has push notifications permission',
        error
      );
    }
```

cardstack/src/screens/NotificationsPermissionScreen/useNotificationsPermissionScreen.ts
```diff
@@ -0,0 +1,26 @@
+import { useCallback } from 'react';
+
+import { useUpdateNotificationPreferences } from '@cardstack/hooks/notifications-preferences/useUpdateNotificationPreferences';
+
+export const useNotificationsPermissionScreen = () => {
+  const {
+    options,
+    isError,
+    // onUpdateOptionStatus,
```
danibonilha: letfover?

cardstack/src/screens/NotificationsPermissionScreen/useNotificationsPermissionScreen.ts
```diff
@@ -0,0 +1,26 @@
+import { useCallback } from 'react';
+
+import { useUpdateNotificationPreferences } from '@cardstack/hooks/notifications-preferences/useUpdateNotificationPreferences';
+
```
danibonilha, line 4: nit: maybe we can add this hook to the index, so the import path gets smaller

cardstack/src/screens/NotificationsPermissionScreen/NotificationsPermissionScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+  } = useNotificationsPermissionScreen();
+
+  const renderItem = useCallback(
+    ({ item }: ListRenderItemInfo<NotificationsPreferenceDataType>) => {
+      const notificationType =
+        NotificationsOptionsStrings[
+          item?.attributes[
+            'notification-type'
+          ] as keyof typeof NotificationsOptionsStrings
+        ];
```
danibonilha, line 38: Maybe we can come up with some sort of a helper for this, since we need to do the same on notificationSection

cardstack/src/screens/NotificationsPermissionScreen/NotificationsPermissionScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+            </Checkbox>
+            <NotificationBanner
+              paddingTop={6}
+              paddingRight={8}
+              paddingBottom={3}
+              paddingLeft={10}
+              title="Cardstack"
+              body={notificationType}
+            />
+          </Container>
```
danibonilha, line 56: my 2 cents is that the Notification should live inside the checkbox, it seems more intuitive to just tap the row, and have it selected

cardstack/src/screens/NotificationsPermissionScreen/NotificationsPermissionScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+      return (
+        <>
+          <HorizontalDivider backgroundColor="blueDarkest" />
+          <Container>
+            <Checkbox checkboxPosition="left">
+              <Text color="white">{notificationType}</Text>
+            </Checkbox>
+            <NotificationBanner
+              paddingTop={6}
+              paddingRight={8}
```
danibonilha: Maybe we should get rid of this padding, on smaller devices, it messes the payment received one:
image

Or calculate a fixed width for the notification banner based on device's screen width

cardstack/src/screens/NotificationsPermissionScreen/useNotificationsPermissionScreen.ts
oidoug: It is already, the import is wrong 👍

cardstack/src/screens/NotificationsPermissionScreen/NotificationsPermissionScreen.tsx
oidoug: TBH this string value should be stored remotely, but for now, I'm thinking that it should be returned "hydrated" by the service layer.
If it's ok with you, I'll leave this as is now to update this in the logic PR.

cardstack/src/screens/NotificationsPermissionScreen/NotificationsPermissionScreen.tsx
oidoug: Yeah, good suggestion.

cardstack/src/screens/NotificationsPermissionScreen/NotificationsPermissionScreen.tsx
oidoug: It needs some padding, I'll add a percentual width then.

danibonilha (approved):

PR #1115, opened 2022-12-05
[Feat]: Bump sdk and refactor coingcko mapping to get token price

danibonilha:
### Description

This PR bumps the sdk so it can use the new coingeckoNativeId, and also adds a refactor on the way we map the ids, to also support polygon tokens, this way we are able to fetch the correct token price 

image

PS: it would be nice to add tests for this, but first I'd like to investigate if there's a better way for us to get the prices, all this mapping seems too much.


paulinhapenedo (approved):
LGTM! 👏

VitaliKirpu:
Ace!

PR #1116, opened 2022-12-05
[CS-4936] Get gas estimates working on Polygon/Mumbai

paulinhapenedo:
### Description
This PR has some changes into how we are fetching the gas prices. The extensive logic that was in redux is being replaced by using the Hub's `/gas-station` endpoint. The data transformation is being done in the RTK `transformResponse`. 

Since we are using RTK, we started using its `polling` feature instead of manually refetching with a `setTimeout`. Once the screen isn't mounted anymore, the RTK stops polling automatically.

- [x] Completes #CS-4963

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/hooks/transaction-confirmation/use-close-screen.ts
+++ b/cardstack/src/hooks/transaction-confirmation/use-close-screen.ts
@@ -6,22 +6,19 @@ import { useDispatch } from 'react-redux';
 import { Routes } from '@cardstack/navigation';
 import { WCRedirectTypes } from '@cardstack/screens/sheets/WalletConnectRedirectSheet';
 
-import { useGas } from '@rainbow-me/hooks';
 import { useRainbowSelector } from '@rainbow-me/redux/hooks';
 import { walletConnectRemovePendingRedirect } from '@rainbow-me/redux/walletconnect';
 import { SEND_TRANSACTION } from '@rainbow-me/utils/signingMethods';
 
 import { useRouteParams } from './use-route-params';
 
-export const useCloseScreen = (isMessageRequest: boolean) => {
+export const useCloseScreen = () => {
   const dispatch = useDispatch();
   const { goBack, canGoBack, navigate } = useNavigation();
 
   const pendingRedirect = useRainbowSelector(
     ({ walletconnect }) => walletconnect.pendingRedirect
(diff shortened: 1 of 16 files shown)


oidoug (approved):
Really nice refactor! LGTM ✈️

cardstack/src/hooks/transaction-confirmation/use-close-screen.ts
```diff
@@ -38,10 +35,6 @@ export const useCloseScreen = (isMessageRequest: boolean) => {
         navigate(Routes.WALLET_SCREEN);
       }
 
-      if (!isMessageRequest) {
-        stopPollingGasPrices();
-      }
-
```
danibonilha, line 44: I totally forgot about the gas usage here, let me make some tests, my hunch is that we'll probably need to use some sort of lazy loading to avoid getting the gas prices on messageRequests

src/hooks/useGas.js
```diff
@@ -32,13 +39,18 @@ export default function useGas() {
     })
   );
 
-  const startPollingGasPrices = useCallback(
-    () => dispatch(gasPricesStartPolling()),
-    [dispatch]
+  const { data } = useGetGasPricesQuery(
```
danibonilha: Maybe we could rename this, to avoid confusion with the other functions that have a parameter called data

src/parsers/gas.js
```diff
@@ -78,15 +78,15 @@ export const parseLayer2GasPrices = data => ({
...
   return {
     estimatedTime: {
       amount: timeAmount,
       display: getMinimalTimeUnitStringForMs(timeAmount),
     },
     option,
     value: {
-      amount: weiAmount,
+      amount: value,
       display: `${parseInt(value, 10)} Gwei`,
```
danibonilha: I don't think this is Gwei anymore, since the value is the same as the amount

cardstack/src/hooks/transaction-confirmation/use-close-screen.ts
danibonilha: Giving the gas calculations complexity,  we can leave as it, and revisit on the london upgrade

PR #1118, opened 2022-12-07
[CS-4727] Logic and flow for push notifications permissions screen

oidoug:
### Description

This PR adds logic to show or skip the Notification Permissions Screen. After skipped once the screen will not show again, but users can go to notifications in Settings to be presented with the notification permission again if still pending. Skipping is important to avoid spamming users that don't want notifications at all.

The screen was added to the onboarding flow and before backup, as first intended for the onboarding flow.

- [x] Completes #(CS-4727)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/129619/206207559-9da34f5b-5ddf-4f91-a448-096201fcc253.mp4


cardstack/src/components/Checkbox/Checkbox.tsx
```diff
@@ -32,11 +32,11 @@ export const Checkbox = ({
   const [disabled] = useState(isDisabled);
```
danibonilha: I know it's not the purpose of this PR, nor it was you who wrote this, but I think we can remove this state, and use the disabled property directly

cardstack/src/components/Checkbox/Checkbox.tsx
```diff
@@ -32,11 +32,11 @@ export const Checkbox = ({
...
 
   const handleCall = useCallback(() => {
-    setSelected(!selected);
-
     if (onPress) {
-      onPress();
+      onPress(!selected);
     }
+
+    setSelected(!selected);
```
danibonilha: nit: we could assign `!selected ` to a variable to make sure we use the same value

cardstack/src/models/firebase.ts
```diff
@@ -203,7 +203,7 @@ export const checkPushPermissionAndRegisterToken = async () => {
           },
           {
             onPress: async () => {
-              resolve(true);
+              resolve(false);
             },
```
danibonilha: in this case, shouldn't the catch block return resolve as false too ?

cardstack/src/screens/NotificationsPermissionScreen/NotificationsPermissionScreen.tsx
```diff
@@ -24,40 +23,37 @@ const NotificationsPermissionScreen = () => {
   const {
     options,
     isError,
-    handleSkipOnPress,
+    handleOnUpdateOption,
```
danibonilha: supernit: I'd suggest `handleUpdateOption` or `onUpdateOption`

cardstack/src/screens/NotificationsPermissionScreen/NotificationsPermissionScreen.tsx
```diff
@@ -24,40 +23,37 @@ const NotificationsPermissionScreen = () => {
...
     handleEnableNotificationsOnPress,
+    handleSkipOnPress,
```
danibonilha: ditto: `handleSkipPress` or `onSkipPress`

cardstack/src/screens/NotificationsPermissionScreen/NotificationsPermissionScreen.tsx
```diff
@@ -105,6 +101,8 @@ const NotificationsPermissionScreen = () => {
     []
   );
 
+  const keyExtractor = (item: NotificationsOptionsType) => item.type;
```
danibonilha: nit: could wrap on useCallback

cardstack/src/screens/NotificationsPermissionScreen/useNotificationsPermissionScreen.ts
```diff
@@ -1,22 +1,49 @@
...
+    options,
+    isError,
+    onUpdateOptionStatus,
+  } = useUpdateNotificationPreferences();
+
+  const handleOnUpdateOption = useCallback(
+    async (option: NotificationsOptionsType, isEnabled: boolean) =>
+      onUpdateOptionStatus(option.type, isEnabled),
+    [onUpdateOptionStatus]
+  );
```
danibonilha: do we really need this function at all? if we are just calling `onUpdateOptionStatus` with the same params, except it's option, vs option.type, but it seems we could just use onUpdateOptionStatus on the component

cardstack/src/models/firebase.ts
```diff
@@ -203,7 +203,7 @@ export const checkPushPermissionAndRegisterToken = async () => {
...
             style: 'cancel',
             text: 'Dismiss',
```
danibonilha: also since we are messing with this function, could you replace the lang.t messages with simple strings ?

cardstack/src/types/NotificationsPreferenceDataType.ts
```diff
@@ -1,6 +1,18 @@
...
+  'customer_payment' = 'New Payment Received',
+  'prepaid_card_drop' = 'Prepaid Card Drop',
+}
+
+export interface NotificationsOptionsType {
+  type: keyof typeof NotificationsOptionsStrings;
+  description: string;
+  status: NotificationStatusType;
+}
+
```
danibonilha, line 14: 👏

cardstack/src/hooks/onboarding/useShowOnboarding.ts
```diff
@@ -46,6 +48,14 @@ export const useShowOnboarding = () => {
         return;
       }
 
+      const askForNotificationsPermissions = await needsToAskForNotificationsPermissions();
```
danibonilha: nit: maybe we could rename this function, this gives the impression that permission will be "asked"/requested, when in fact it just checking if permission was already granted. Maybe something like, `needsNotificationPermission`  or not negate the isAuthorized variable on the function, so then we can use something like `hasNotificationPermission` and negate it accordingly with usage

cardstack/src/hooks/onboarding/useShowOnboarding.ts
```diff
@@ -54,9 +64,16 @@ export const useShowOnboarding = () => {
 
       if (shouldShowProfileCreationFlow) {
```
danibonilha: let's add a comment here explaining why we need this step, it's a bit confusing that `navigateToNextOnboardingStep` has steps as optionals, so it would be nice to add some sort of explanation on why this is needed

cardstack/src/screens/NotificationsPermissionScreen/useNotificationsPermissionScreen.ts
```diff
@@ -1,22 +1,49 @@
...
+    if (!accepted) {
+      triggerSkipNotificationPermission();
+    }
+
+    navigateToNextOnboardingStep(Routes.BACKUP_EXPLANATION);
+  }, [triggerSkipNotificationPermission, navigateToNextOnboardingStep]);
+
+  const handleSkipOnPress = useCallback(() => {
+    triggerSkipNotificationPermission();
+    navigateToNextOnboardingStep();
```
danibonilha: Skipping shouldn't take the user to the profile creation ?

cardstack/src/screens/NotificationsPermissionScreen/useNotificationsPermissionScreen.ts
danibonilha: On second though, shouldn't we wait for the user to hit enable, and give permission, before setting the notification options ? What's the point of storing this info if the user doesn't have an fcm token

cardstack/src/screens/NotificationsPermissionScreen/useNotificationsPermissionScreen.ts
```diff
@@ -1,22 +1,49 @@
...
-  const handleEnableNotificationsOnPress = useCallback(() => {
-    // TBD
-  }, []);
+    if (!accepted) {
+      triggerSkipNotificationPermission();
+    }
+
+    navigateToNextOnboardingStep(Routes.BACKUP_EXPLANATION);
+  }, [triggerSkipNotificationPermission, navigateToNextOnboardingStep]);
+
```
danibonilha, line 25: it feels to me, that it doesn't matter if it's accepted or not, skipped or not, the next 'step' behavior can be the same, triggerSkipNotificationPermission then navigateToNextOnboardingStep(Routes.BACKUP_EXPLANATION); since we are not waiting to set the preferences after users permissions

cardstack/src/components/Checkbox/Checkbox.tsx
```diff
@@ -54,7 +54,13 @@ export const Checkbox = ({
     >
       <Container
         alignItems="center"
-        backgroundColor={disabled ? 'underlineGray' : 'buttonPrimaryBackground'}
+        backgroundColor={
+          disabled
+            ? 'underlineGray'
+            : selected
+            ? 'checkboxSelected'
+            : 'checkboxDeselected'
```
danibonilha: Doesn't this change the color on screens that we don't want too ? Or is it a new pattern for our checkbox ?  Also this nested ternary is a bit scary hehe, it seems we never use this disabled property, so maybe we can remove it ?

cardstack/src/models/firebase.ts
oidoug: Yes, that makes sense 👍

cardstack/src/screens/NotificationsPermissionScreen/useNotificationsPermissionScreen.ts
oidoug: I thought the same, but we don't actually use the fcm-token to save this info, just the hub auth token, and it seemed simpler to just save the user selection, than to somehow invalidate the selection if they skip, feels weird to not save the way the UI works. 
Not sure though.

cardstack/src/models/firebase.ts
oidoug: Maybe we don't need this confirmation Alert anymore... Because we already have the whole screen as a "push will be enabled" disclaimer. WDYT?

cardstack/src/hooks/onboarding/useShowOnboarding.ts
oidoug: I prefer `needsNotificationPermission`, negating the response seems odd.

cardstack/src/hooks/onboarding/useShowOnboarding.ts
oidoug: Yeah, it's confusing because we want to navigate based on an effect "sometimes" and some other times we need to navigate before the effect change takes place. I'm refactoring this function a bit and adding some comments, plz take a look to see if it's somewhat improved.

cardstack/src/screens/NotificationsPermissionScreen/useNotificationsPermissionScreen.ts
oidoug: Maybe theres some overlap but it's not really the same. If the user accepts (don't skip) and then later removes permissions, we'll ask again.

I changed how the navigation works a bit, see if it's any better.

cardstack/src/components/Checkbox/Checkbox.tsx
oidoug: We need to differentiate the "fill" color somehow. I tested in white backgrounds and it's not optimal but better than using the same color for sure.

I removed the disabled as suggested.

cardstack/src/screens/NotificationsPermissionScreen/useNotificationsPermissionScreen.ts
oidoug: The second screen in the flow is the backup one. I think the flow is correct.

cardstack/src/components/Checkbox/Checkbox.tsx
```diff
@@ -22,21 +22,21 @@ const CHECKBOX_SIZE = 22;
...
   checkboxPosition = 'right',
   verticalAlign = 'center',
   children,
 }: CheckboxProps) => {
   const [selected, setSelected] = useState(isSelected);
-  const [disabled] = useState(isDisabled);
 
   const handleCall = useCallback(() => {
-    setSelected(!selected);
+    const currentSelected = selected;
```
danibonilha: hehe, the suggestion was to make sure we use exact the same value with the negation, so 
`const currentSelected = !selected;` but then maybe the name should change,
 `const toggledCheck = !selected;`

cardstack/src/hooks/onboarding/useShowOnboarding.ts
```diff
@@ -33,33 +35,53 @@ export const useShowOnboarding = () => {
...
+    }
+  }, [
+    shouldShowBackupFlow,
+    shouldShowProfileCreationFlow,
+    hasSkippedNotificationPermission,
+  ]);
+
+  // Will navigate imperactively to provided route.
+  const navigateOnboardingTo = useCallback(
+    (route: string) => {
```
danibonilha: could we type it as route ?

cardstack/src/models/firebase.ts
danibonilha: Agreed for the initial flow, but maybe it's needed for the settings one

cardstack/src/screens/NotificationsPermissionScreen/useNotificationsPermissionScreen.ts
danibonilha: Got it, I know it doesn't use the fcmToken, but without it, it means the user never gets the notification, so we are storing related info on our db to never be used, I think it's ok for now giving the amount of users, but that might be cumbersome in the future, let's just keep that in mind, so then my first question takes place again, we may not need this handle function, it;s just wrapping the onUpdateOptions, if the intention is to "rename", we can just pass 
 `{  handleUpdateOption:  onUpdateOptionStatus }` as part of the hook return.

cardstack/src/components/Checkbox/Checkbox.tsx
oidoug: Ah I see 👍

cardstack/src/hooks/onboarding/useShowOnboarding.ts
oidoug: Sure, do you mean `RouteNames` from global, right?

danibonilha: Yes

src/components/settings-menu/NotificationsSection.tsx
```diff
@@ -1,12 +1,42 @@
...
-} from '@cardstack/hooks';
-import { NotificationsPreferenceDataType } from '@cardstack/types';
+  checkPushPermissionAndRegisterToken,
+  getPermissionStatus,
+} from '@cardstack/models/firebase';
+import { NotificationsOptionsType } from '@cardstack/types';
+
+import { Alert } from '@rainbow-me/components/alerts';
+
+const buildAlert = (
```
danibonilha: since the alert is triggered here,  we could rename it to showAlert

danibonilha (approved):
🚀

PR #1121, opened 2022-12-08
[CS-4969]: Updates UI when gasPrices changes 

danibonilha:
### Description

This PR adds a flag (shouldUpdateTxFee) to be used on the gasCalculations useEffects dependencies and trigger a re-render with updated values, even though the prices were being polled correctly, the UI was never updated since it only uses txFee from redux.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

It would be nice to add some sort if animation when the price changes, but it's lower priority right now

Simulator Screen Recording - iPhone 14 - 2022-12-08 at 14 18 49

### DIFF
--- a/src/hooks/useGas.js
+++ b/src/hooks/useGas.js
@@ -1,6 +1,7 @@
 import { getConstantByNetwork } from '@cardstack/cardpay-sdk';
 
-import { useCallback } from 'react';
+import { isEqual } from 'lodash';
+import { useCallback, useEffect, useRef } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import {
   gasUpdateCustomValues,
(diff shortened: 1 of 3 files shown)


src/hooks/useGas.js
```diff
@@ -11,9 +12,10 @@ import {
 } from '../redux/gas';
 
 import useAccountSettings from './useAccountSettings';
+import usePrevious from './usePrevious';
 import { useGetGasPricesQuery } from '@cardstack/services';
 
-const GAS_PRICE_POLLING_INTERVAL = 15000; // 15s
+const GAS_PRICE_POLLING_INTERVAL = 10000; // 10s
```
oidoug, line 18: Strange that in your demo video it looks like it's pooling more often than 10 secs. Maybe something else is triggering the refetch?

oidoug (approved):
LGTM

src/hooks/useGas.js
danibonilha: Nah, the video is a gif, so it's speed up

PR #1123, opened 2022-12-09
[CS-4961] Remove GasSpeed button

paulinhapenedo:
### Description
This PR removes the GasSpeed button from the `TransactionConfirmation` component. It's something that adds complexity to the screen logic and that doesn't add much value to the user. Chris was OK with removing it.

It solves CS-4961.


danibonilha (approved):
🚀

PR #1124, opened 2022-12-12
[CS-4924] Moves Hub's auth token to secure store

oidoug:
### Description

This PR refactors hub's token setter and getter calls to use Expo's Secure Store, also saving a timestamp along with the new token for a expiration check. The stored token is ignored and deleted if timestamp is older than a day.

The token setter/getter has been moved to `hub/hub-service` file, along with other hub-service functions, and the file `hub-token-service` was deleted.

There are also some new tests for secure-storage functions.

- [x] Completes #(CS-4924)

*no visual changes*


danibonilha (approved):
LGTM!

PR #1125, opened 2022-12-12
[CS-4947]: Refactor the way assets are built in the app

danibonilha:
### Description

This is the starter PR to remove the fallback explorer and change the way we handle assets, (i wanted to do everything on one PR, but it's getting too big, so it's best to split it up,  the sooner we start using, the faster we can identify bugs).
On this PR most of the fallback important functions were refactored and moved to typescript. A couple of parsers were removed to simplify the way we build the assets. It also removes some unused hooks, and packages and renames kebab props to camelCase. I've also added comments to help the future us. 
The chart data will also be removed, part of it, it's already gone but I'll remove the components etc on the next PR.
There's the initial api structure for EOA, although i'm not 100% sure that's going to be the path, but if it's not I'll make sure to remove it

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


cardstack/src/services/eoa-assets/eoa-assets-api.ts
```diff
@@ -0,0 +1,21 @@
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/dist/query';
```
paulinhapenedo, line 1: We seem to have different imports for these functions. Most of the places we import these from `@reduxjs/toolkit/query/react`. Here and in the hub-wyre-api file are the only places where we import like this. I'm not sure which is the correct, but I thought it would be good to point it out.

cardstack/src/services/eoa-assets/eoa-assets-services.ts
```diff
@@ -0,0 +1,273 @@
...
+
+import { parseAssetName, parseAssetSymbol } from '@rainbow-me/parsers';
+import { shitcoins } from '@rainbow-me/references';
+import migratedTokens from '@rainbow-me/references/migratedTokens.json';
+import { getTokenMetadata } from '@rainbow-me/utils';
+import logger from 'logger';
+
+import { EOABaseParams, EOATxListResponse } from './eoa-assets-types';
+
+// TODO: store lastestTXBlock on storage with assets
```
paulinhapenedo, line 20: tiny nit: typo `latest`

cardstack/src/services/eoa-assets/eoa-assets-api.ts
danibonilha: This dist doesn't quite work actually, we should use the other one, it's there though, because this api hasn't been implemented yet, will be fixed on next PR

cardstack/src/services/eoa-assets/eoa-assets-services.ts
```diff
@@ -0,0 +1,273 @@
...
+  );
+
+  const { status, result: newTxs, message } = await request.json();
+
+  const allTxs = [...previousTx, ...newTxs];
+
+  if (status === '1' && newTxs.length === offset) {
+    if (message.contains('rate limit')) {
+      await delay(5000);
+    }
```
paulinhapenedo, line 171: What do you think about adding a comment here to explain why we have this delay?

danibonilha: can do, but I thought for this case, the code kinda explains it, if message contains 'rate limit' we need to wait before fetching again

paulinhapenedo (approved):
Leaving as approved, but I can't test NFT's because I don't have any accounts with it.

oidoug (approved):
LGTM, there's a lot to test, but we can iterate more before next release.

PR #1127, opened 2022-12-14
[CS-4992] Removes deprecated checkHubAuth

oidoug:
### Description

Removes deprecated checkHubAuth. 

It was once used by `useRequestPrepaidCardScreen` before starting the request, but since that check was removed and hub-api now handles unauthorized requests better by renewing the token and retrying the request.

- [x] Completes #(CS-4992)


danibonilha (approved):
LGTM

PR #1128, opened 2022-12-15
Use decimals from token contract in reward explanations

IanCal:
### Description

This swaps over to using the decimals defined in the token contract instead of assuming 18. 18 is common but not universal, USDC is 6.

I was able to build and test on an android emulator but don't have access to a mac for building for iOS.

- [x] Completes #(CS-4953)

### Checklist

- [ ] Tested on a small device
- [ ] Tested on iOS
- [x] Tested on Android

### Screenshots

image

### DIFF
--- a/cardstack/src/services/rewards-center/rewards-center-service.ts
+++ b/cardstack/src/services/rewards-center/rewards-center-service.ts
@@ -82,12 +82,10 @@ const getValidProofs = async ({
   return validProofs;
 };
 
-const getTokenSymbolForAddress = async (tokenAddress: string) => {
-  const rewardPoolInstance = await getRewardsPoolInstance();
-
-  const token = await rewardPoolInstance.addTokenSymbol([{ tokenAddress }]);
-
-  return token?.[0]?.tokenSymbol;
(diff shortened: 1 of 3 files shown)


cardstack/src/services/rewards-center/rewards-center-service.ts
```diff
@@ -82,12 +82,10 @@ const getValidProofs = async ({
...
-const getTokenSymbolForAddress = async (tokenAddress: string) => {
-  const rewardPoolInstance = await getRewardsPoolInstance();
-
-  const token = await rewardPoolInstance.addTokenSymbol([{ tokenAddress }]);
-
-  return token?.[0]?.tokenSymbol;
+const getTokenDetailsForAddress = async (tokenAddress: string) => {
+  const [web3, signer] = await getWeb3ProviderWithEthSigner();
+  const assetsAPI = await getSDK('Assets', web3, signer);
+  return await assetsAPI.getTokenInfo(tokenAddress);
```
oidoug, line 88: Ok, this makes a lot more sense 👍

oidoug (approved):
Nice. LGTM!

PR #1130, opened 2022-12-15
[CS-4978] Create new useGas hook in Typescript

paulinhapenedo:
### Description
~Opening as a draft because the work done here may impact the refactor Dani is working on. There's a lot of cleanup and better ways of writing stuff, this is a first pass of making the gas estimation work in the SendSheet screen without Redux.~

This PR refactors how we get the gas prices and calculate the gas fee for the EOA transactions. We're fetching the gas price from the gas oracle in the Hub, doing all the calculations in the hook and, above all, we are not saving these values in Redux. 

When deleting the old files, we also removed two libraries that we were not going to use anymore.

- [x] Relates to #CS-4978

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots
You shouldn't see any visual changes ;)


cardstack/src/hooks/gas-prices/useGas.ts
```diff
@@ -0,0 +1,134 @@
...
+import { useGetGasPricesQuery } from '@cardstack/services';
+import { GasPricesQueryResults } from '@cardstack/services/hub/gas-prices/gas-prices-types';
+
+import { useAccountAssets, useAccountSettings } from '@rainbow-me/hooks';
+import { ethUnits } from '@rainbow-me/references';
+import { ethereumUtils, showActionSheetWithOptions } from '@rainbow-me/utils';
+
+import { ParseTxFeeParams, TxFee, UseGasParams } from './types';
+
+const GAS_PRICE_POLLING_INTERVAL = 60000; // 1min
```
paulinhapenedo: Changed this to avoid hitting the Hub every 10s while debugging stuff. It will return to 10s in the final version of the PR.

cardstack/src/hooks/gas-prices/useGas.ts
```diff
@@ -0,0 +1,134 @@
...
+    //   {
+    //     cancelButtonIndex: options.length,
+    //     options,
+    //   },
+    //   (buttonIndex: number) => {
+    //     console.log('Click Button', buttonIndex);
+    //   }
+    // );
+    return false;
+  }, []);
```
paulinhapenedo: Commented out because it's the ActionSheet with the different transactions speeds and fees. It's not working yet.

src/redux/data.js
```diff
@@ -33,7 +33,6 @@ const TXN_WATCHER_POLL_INTERVAL = 5000; // 5 seconds
 // -- Constants --------------------------------------- //
 
 const DATA_UPDATE_ASSETS = 'data/DATA_UPDATE_ASSETS';
-const DATA_UPDATE_GENERIC_ASSETS = 'data/DATA_UPDATE_GENERIC_ASSETS';
```
paulinhapenedo: We weren't using this anywhere in the code except when getting the native token price (we would map through `genericAssets` and `assets`). It was always empty because we were never setting it, so I removed it.

src/redux/gas.js
```diff
@@ -1,12 +1,11 @@
 import { fromWei, greaterThanOrEqualTo } from '@cardstack/cardpay-sdk';
 import { get, isEmpty } from 'lodash';
-import { getEstimatedTimeForGasPrice } from '@rainbow-me/handlers/gasPrices';
-import { defaultGasPriceFormat, parseTxFees } from '@rainbow-me/parsers';
+import { parseTxFees } from '@rainbow-me/parsers';
```
paulinhapenedo: This whole file will be deleted, so don't worry about it (the `redux/gas`, not the `parseTxFees`)

src/utils/gas.js
```diff
@@ -9,22 +9,20 @@ import {
 } from 'lodash';
 import showActionSheetWithOptions from './actionsheet';
 
-const CUSTOM = 'custom';
 const FAST = 'fast';
 const NORMAL = 'normal';
 const SLOW = 'slow';
 
-const GasSpeedOrder = [SLOW, NORMAL, FAST, CUSTOM];
+const GasSpeedOrder = [SLOW, NORMAL, FAST];
```
paulinhapenedo: This file will also be deleted once the ActionSheet with the speeds is implemented, so don't worry about the changes here.

cardstack/src/components/index.ts
```diff
@@ -15,7 +15,7 @@ export * from './BalanceCoinRow';
 export * from './OptionItem';
 export * from './Input';
 export * from './Sheet';
-export * from './ListItem';
+export * from './ListItem/ListEmptyComponent';
```
oidoug: nit: I recommend renaming this to `ListItemEmpty`, renaming the parent folder `ListItem` to `ListItemEmpty`, and add an index, to follow our conventions better.

cardstack/src/hooks/gas-prices/useGas.ts
```diff
@@ -0,0 +1,147 @@
...
+  const nativeTokenDecimals = getConstantByNetwork(
+    'nativeTokenDecimals',
+    network
+  );
+
+  const { assets, getAssetBalance, getAssetPrice } = useAssets();
+  const [txFees, setTxFees] = useState<TxFee>();
+
+  const [selectedGasSpeed, setSelectedGasSpeed] = useState<
+    keyof GasPricesQueryResults
```
oidoug, line 34: nit: I wonder if it isn't better define this type in `./types` as well. What `keyof GasPricesQueryResults` will yield?

cardstack/src/hooks/gas-prices/useGas.ts
```diff
@@ -0,0 +1,147 @@
...
+        keyof typeof gasPricesData
+      >;
+
+      const calculateGasFee = speeds.map(speed => {
+        const nativeTokenSymbol = getConstantByNetwork(
+          'nativeTokenSymbol',
+          network
+        );
+
+        const fee = gasLimit * (gasPricesData[speed].amount as number); // value in GWEI
```
oidoug: Isn't better to instead of casting as number, to create a Number instance, like `Number(gasPricesData[speed].amount)`?

src/screens/SendSheetEOA.js
```diff
@@ -207,47 +196,39 @@ const useSendSheetScreen = () => {
...
-      isValidAddress && amountDetails.isSufficientBalance && isSufficientGas;
-    if (!selectedGasPrice.txFee || !validTransaction || isAuthorizing) {
+      isValidAddress &&
+      amountDetails.isSufficientBalance &&
+      hasSufficientForGas;
+
+    if (!selectedFee || !validTransaction || isAuthorizing) {
       logger.sentry('preventing tx submit for one of the following reasons:');
-      logger.sentry('selectedGasPrice.txFee ? ', selectedGasPrice?.txFee);
+      logger.sentry('selectedFee ? ', selectedFee);
```
oidoug: super nit: These sentry logs could be cleaner... I would just write one logger.sentry with something like `logger.sentry('Preventing Tx submit given values:', { selectedFee, validTransaction, isAuthorizing } );`.

And `captureEvent` seems redundant.

cardstack/src/hooks/gas-prices/useGas.ts
```diff
@@ -0,0 +1,147 @@
...
+const GAS_PRICE_POLLING_INTERVAL = 10000; // 10s
+
+export const useGas = ({ network }: UseGasParams) => {
+  const { nativeCurrency } = useAccountSettings();
+  const chainId = getConstantByNetwork('chainId', network);
+
+  const nativeTokenDecimals = getConstantByNetwork(
+    'nativeTokenDecimals',
+    network
+  );
```
danibonilha: nit: this const could live inside the function which is used

cardstack/src/hooks/gas-prices/useGas.ts
```diff
@@ -0,0 +1,147 @@
...
+      network,
+      nativeCurrency,
+      gasPricesData,
+      nativeTokenDecimals,
+      assets,
+      getAssetPrice,
+    ]
+  );
+
+  const hasSufficientForGas = useMemo(() => {
```
oidoug, line 110: nit: `hasBalanceForGas`?

cardstack/src/hooks/gas-prices/useGas.ts
```diff
@@ -0,0 +1,147 @@
...
+
+import { useAssets } from '../assets/useAssets';
+
+import { ParseTxFeeParams, TxFee, UseGasParams } from './types';
+
+const GAS_PRICE_POLLING_INTERVAL = 10000; // 10s
+
+export const useGas = ({ network }: UseGasParams) => {
+  const { nativeCurrency } = useAccountSettings();
+  const chainId = getConstantByNetwork('chainId', network);
```
danibonilha: supernit: this could be moved to the query params since it's only used there

cardstack/src/hooks/gas-prices/useGas.ts
```diff
@@ -0,0 +1,147 @@
...
+import { ethUnits } from '@rainbow-me/references';
+import { ethereumUtils, showActionSheetWithOptions } from '@rainbow-me/utils';
+
+import { useAssets } from '../assets/useAssets';
+
+import { ParseTxFeeParams, TxFee, UseGasParams } from './types';
+
+const GAS_PRICE_POLLING_INTERVAL = 10000; // 10s
+
+export const useGas = ({ network }: UseGasParams) => {
```
danibonilha: nit: if we need to use the `useAccountSettings` anyways here, we may not need to get network by a parameter, but I guess it's ok, maybe with multichain we're going to need it anyway

src/utils/ethereumUtils.js
```diff
@@ -39,23 +39,12 @@ import logger from 'logger';
...
-  const { network } = store.getState().settings;
-  const nativeTokenAddress = getConstantByNetwork(
-    'nativeTokenAddress',
-    network
-  );
-  return getAsset(assets, nativeTokenAddress)?.price?.value || 0;
-};
-
 const getBalanceAmount = async (selectedGasPrice, selected) => {
   const network = store.getState().settings.network;
```
danibonilha: nit: we could receive network as a parameter, or even move this function to the hook that is used, since it's only used there

src/utils/ethereumUtils.js
```diff
@@ -39,23 +39,12 @@ import logger from 'logger';
...
   let amount = get(selected, 'balance.amount', 0);
   if (isNativeToken(selected?.symbol, network)) {
     if (!isEmpty(selectedGasPrice)) {
-      const txFeeRaw = get(selectedGasPrice, 'txFee.value.amount');
+      const txFeeRaw = get(selectedGasPrice, 'value.amount');
       const txFeeAmount = fromWei(txFeeRaw);
```
danibonilha: nit: `const txFeeAmount = fromWei(selectedGasPrice.value.amount);`

src/screens/SendSheetEOA.js
```diff
@@ -388,8 +347,10 @@ const useSendSheetScreen = () => {
         },
         network
       )
-        .then(gasLimit => updateTxFee(gasLimit))
-        .catch(() => updateTxFee(null));
+        .then(gasLimit => getTxFees({ gasLimit }))
+        .catch(e => {
+          logger.log('Error estimating gas limit for transaction', e);
+        });
```
danibonilha: nit: maybe this function could be moved inside the hook, so `getTxFees` don't need to be exposed, we could have a updateTxFees, which would call estimateGas then getTxFees, in  the hook itself

oidoug (approved):
Awesome refactor, so much cleaner 👏 
Left some nits, no blockers.

Did some tests, couldn't see the gas fee changing but it looks all good.

cardstack/src/hooks/gas-prices/useGas.ts
paulinhapenedo: Checking the code, you can see that `keyof GasPricesQueryResults` yields the 3 possible speeds: slow, standard and fast. Instead of creating an Enum or a `const` with these values, I prefer to use them as they come from the Hub.

If the team prefer, I can change all the types that are inside the `./types.ts` file into the types file created for the gas oracle API. I didn't do that because they are not directly related.

cardstack/src/components/index.ts
paulinhapenedo: The new conventions, you mean? Because this component is already inside the @cardstack space.
Sure. Will do that.

src/screens/SendSheetEOA.js
paulinhapenedo: `captureEvent` sends the event to Sentry. The `logger.sentry` saves locally the message and do not send to Sentry unless there's a `captureEvent` function call or a `captureException`, which then sends the breadcrumb of logs to Sentry.

danibonilha (approved):
Nice job! Way cleaner

PR #1131, opened 2022-12-15
Additional steps that may be required on linux

IanCal:
Just added a couple of things to the readme:

* File system watches issue on linux
* A very odd fastlane plugin issue that I created a branch for on the original repo
* A note that yarn start is required for android dev too

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,19 @@ Follow steps 1 through 4 from the iOS steps above, then:
 
 5. Run `bundle exec fastlane sync_google_service_info`
 
-6. Run `yarn android`
+6. Run `yarn start` to start the React Native Bundler.
 
-7. Launch an Android emulator or connect a test device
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,9 +45,19 @@ Follow steps 1 through 4 from the iOS steps above, then:
 
 5. Run `bundle exec fastlane sync_google_service_info`
 
-6. Run `yarn android`
+6. Run `yarn start` to start the React Native Bundler.
```
danibonilha: This may be needed for linux only, on mac at least the yarn android command already starts a metro bundler, so maybe we can add that info

danibonilha (approved):

README.md
IanCal: Oh good to know, I've moved this part to the troubleshooting instead then.

PR #1132, opened 2022-12-15
[CS-4987]: Refactor assets apis and remove fallback explorer 

danibonilha:
### Description
This PR adds eoa-assnts and coingeko-apis using RTK so we can leverage its cache and soon the polling interval:

- Adds new useAssets hook where we have assets, ids, and assets in a normalized way, and instead of modifying the assets structure, to get balances, prices etc we can just use its id with getters. There is still however the legacyAssets struct to be used on components that haven’t been migrated yet since safes also get the balances information appended. All queries data in this hook have the information mapped by id (address || tokenId):
- Bumps RTK-Query to use args on transform response
- Bumps sdk to use updated Sokol base api and correct polygon address
- Removes a bunch of dead code 
- Removes fallbackExplorer 🎉
- Improve how we can add a new api, by just adding it to an object, whereas before we had to manually add to reducers and middleware’s
- Bumps assets storage version to use new one
- Stores latestTxBlockNumber to discover new tokens from latest block instead of refetching all tokentxs everytime.

Things to look out for:  
- Make sure assets are appearing correctly per EOA address and network 
- All native tokens should have native currency and appear as asset even if balance is zero
- Tokens on mainnet if they exists on coingecko we should display native currency price 
- If a new token is added to your account and you refresh the app, it should appear.

Stuff to be covered on upcoming PRs:

- Removal of chart info  [CS-4998]
- Price handling on signing requests, and other TODO, it was not priorize on this one, because it works, it just doesn’t show the native currency value [CS-4999]
- I’ve noticed there are lots of websocket connecting logs, which I’m investigating [CS-4994]
- Add tests, it should be easier now to test [CS-5000]
- Migrate accountLocal.js to .ts  [CS-5001] this one is pending on this PR and chart removal

- Maybe add a ticket to refactor collectibles  eventually?


paulinhapenedo (approved):
Tested doing transactions in different networks and it works!
Thank you for doing it and for taking you time to mob-pair to explain the changes too.

PR #1133, opened 2022-12-16
[CS-4998]: Remove charts components 

danibonilha:
### Description
This PR removes the chart references and its components, it also applies some TODOS, and refactor the tokenWithChart to just have the token info, I've removed  the small coin icon from the balance section since it already appears on the sheet header. There is a little bit more of a cleanup and it also adds a condition to only allow send if a token has balance.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

image


src/screens/PinAuthenticationScreen.js
```diff
@@ -223,7 +222,14 @@ const PinAuthenticationScreen = () => {
...
+            color="white"
+            fontSize={18}
+            textAlign="center"
+            weight="extraBold"
+          >
+            {titleMap[actionType]}
+          </Text>
           <PinValue translateX={errorAnimation} value={value} />
         </ColumnWithMargins>
       </Centered>
```
danibonilha, line 235: We may not use this screen anymore, but I guess we should keep it for a while, since some users might not have migrated to new PIN yet ?

paulinhapenedo (approved):
Nice cleanup! 👏

oidoug (approved):
LGTM

PR #1134, opened 2022-12-16
[CS-4991] FCM token saving refactor

oidoug:
### Description

This PR simplifies to the bare minimum the data stored for FCM tokens. It also saves now on SecureStore, even though FCM tokens are fine being outside secure store, I think it doesn't hurt to be saved securely and makes now all tokens saved in the same place.

Also it's now one less place that uses react-native-store.

I refrained from creating a firebase hook, because there's not enough advantages and doesn't really exclude the need for imperative calls or the model.

- [x] Completes #(CS-4991)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

*no ui changes*

### DIFF
--- a/cardstack/src/models/firebase.ts
+++ b/cardstack/src/models/firebase.ts
@@ -46,7 +46,7 @@ export const removeFCMToken = async (walletAddress?: string) => {
       }
     }
   } catch (e) {
-    logger.sentry('Unregister FcmToken failed --', e);
+    logger.sentry('Unregister FcmToken failed', e);
   }
 };
 
--- a/cardstack/src/models/firebase.ts
+++ b/cardstack/src/models/firebase.ts
@@ -19,8 +19,10 @@ import {
   deleteSecureFCMToken,
 } from './secure-storage';
 
-export const getFCMToken = async (): Promise<string | undefined> => {
-  const keyAddress = (await loadAddress()) || '';
+export const getFCMToken = async (
+  walletAddress?: string
+): Promise<string | undefined> => {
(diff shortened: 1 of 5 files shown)


paulinhapenedo (approved):
LGTM!

cardstack/src/models/firebase.ts
```diff
@@ -180,6 +80,30 @@ export const storeRegisteredFCMToken = async () => {
   }
 };
 
+export const getPermissionStatus = (): Promise<FirebaseMessagingTypes.AuthorizationStatus> =>
```
danibonilha: nit: I think typescript can infer this type properly, so we could shrink this

cardstack/src/models/firebase.ts
```diff
@@ -180,6 +80,30 @@ export const storeRegisteredFCMToken = async () => {
...
+  messaging().hasPermission();
+
+export const needsNotificationPermission = async (): Promise<
+  boolean | undefined
+> => {
```
danibonilha: ditto

cardstack/src/models/firebase.ts
```diff
@@ -9,116 +9,47 @@ import {
...
-  } catch (error) {
-    logger.sentry(
-      'Error checking if a user has push notifications permission',
-      error
-    );
-  }
-};
+export const getFCMToken = async (
+  walletAddress?: string
+): Promise<string | undefined> => {
```
danibonilha: ditto

danibonilha (approved):
Nice, way cleaner and good point on not creating a hook! Just some nits

PR #1135, opened 2022-12-16
[Fix]: Not check for balances before storing assets

danibonilha:
### Description

We should store all possible tokens to check if there's a balance after, if we check before storing and there's an error on the balance function or something, we might end up with missing assets, so it's safer to just store all tokens. This PR also replaces hardcoded info with info from the sdk.


paulinhapenedo (approved):
LGTM!

PR #1138, opened 2022-12-20
[CS-5015]: Add a single provider for each chain

danibonilha:
### Description

This PR creates the model to add a web3 provider and an ethers one for each chain so we can have them working simultaneously. it also completes CS-4979 by taking into consideration the txNetwork in order to create the wallet to sign. Finally it cleans up a contract file which was not being used.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Here you can see the app was connected to gnosis, but received a goerli request and it was able to show the ETH balance and approve the request successfully. Also there's only one websocket connected for each network.

https://user-images.githubusercontent.com/20520102/208723312-07855106-078b-4997-bec5-bbe2f5cb3505.mov


paulinhapenedo (approved):
Tested using https://react-app.walletconnect.com/ and works for all testnets.

On mainnets, I couldn't test `sendTransaction` and `signTransaction` because the dapp was throwing errors on their interface, but all the other methods were showing the TransactionConfirmationSheet with the correct token balance.

PR #1139, opened 2022-12-21
[CS-5026] Fix selecting coin inside SendSheetEOA component

paulinhapenedo:
### Description
The bug was caused by not taking into consideration that, when opening the coin selector, the `selectedAsset` is reset to `{}`. Not having the optional operator on the `balance` was breaking the app. 

While on the area, we've decided to add types to the `updateMaxInputBalance` function arguments.

- [x] Completes #CS-5026

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/690904/208949919-141ffd0b-af6a-4ce1-a9d8-e67c4de8f46a.mp4

### DIFF
--- a/cardstack/src/hooks/gas-prices/types.ts
+++ b/cardstack/src/hooks/gas-prices/types.ts
@@ -1,11 +1,12 @@
+import { BalanceType } from '@cardstack/types';
+
 export interface ParseTxFeeParams {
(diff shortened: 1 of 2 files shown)


danibonilha (approved):
🚀

PR #1140, opened 2022-12-21
[CS-5025]:  Cleanup packages and dead code

danibonilha:
### Description
This PR removes packages and files that were not being used, or was only used in one place and was. easily replaceable. We shouldn't have much layout change besides the pressable animation which was scaling and now it's using opacity. 

it removes:
- [x] background-timer
- [x] react-native-text-size
- [x] use-memo-one
- [x] @uniswap/sdk
- [x] miscellaneous functions

it also completes CS-3900, by removing the references of shitcoins from sendAsset list alongside with the savings sections, although I've changed this file I've refrained from refactoring to make it better since it's a really complex build and it was not the point for this PR, and hopefully we'll render the list of coins in a different way soon.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android


src/components/send/SendAssetList.js
```diff
@@ -213,20 +144,11 @@ export default class SendAssetList extends React.Component {
         }
       }
       const smallBalanesheight =
```
oidoug: old typo:
smallBalanesheight -> smallBalancesHeight

src/screens/MaintenanceMode.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { memo } from 'react';
+
+import { CenteredContainer, Container, Text } from '@cardstack/components';
+import { CardwalletLogo } from '@cardstack/components/CardwalletLogo';
+
+const DEFAULT_MESSAGE =
+  'The app is going through scheduled maintenance, please try again later.';
```
oidoug, line 7: super nit: Rename DEFAULT_MESSAGE to `strings = { message }` so we can find it if we add i18n.

oidoug (approved):
Pretty nice 👏 

Hard to evaluate if nothing of value will be lost on `SendAssetList.js` removals, but looks like most is related to filtered spammy tokens.

As far as I could test it all looks good.

src/screens/MaintenanceMode.tsx
danibonilha: For this specific case I'll leave it as is, since this fallback will be used just in case firebase fails which should not be frequent

PR #1142, opened 2022-12-21
[CS-5034]: Fix gasPrice on send transaction

danibonilha:
### Description

To send tokens we should pass the price (in wei) **per unit of gas** the transaction will pay, we were passing the calculated fee, which is not what the provider expects.

image


oidoug (approved):
Back in business! 🚀

paulinhapenedo (approved):

PR #1143, opened 2022-12-22
[CS-5002]: Handle prices polling interval and remove safes from redux

danibonilha:
### Description

This PR migrates the redux usage with safes data, to use the rtk query. it also updates the useAsset hook to fetch information from time to time

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Tx loading correctly:

image

### DIFF
--- a/cardstack/src/hooks/assets/useAssets.ts
+++ b/cardstack/src/hooks/assets/useAssets.ts
@@ -25,7 +25,7 @@ const emptyState = {
   balance: { amount: '0', display: '' },
 };
 
-const pollintInterval = {
+const pollingInterval = {
   ten_seconds: 10000,
   one_minute: 60000,
 };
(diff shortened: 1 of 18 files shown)


cardstack/src/hooks/assets/useAssets.ts
```diff
@@ -25,6 +25,11 @@ const emptyState = {
   balance: { amount: '0', display: '' },
 };
 
+const pollintInterval = {
```
paulinhapenedo: `pollingInterval`, maybe?

paulinhapenedo (approved):
LGTM 👏

oidoug (commented):
LGTM, will make the typing issue I'm getting easier to fix 👏

PR #1145, opened 2022-12-27
[CS-5001] Safes saving and fetchSafes service call refactor

oidoug:
### Description

This PR refactors `accountLocal` to use types and `fetchSafes` service call on `gnosis-service` to convert safes to *SafeType hydrating token info, prices to conform with query types.

- [x] Completes #(CS-5001)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

*no visual change*


cardstack/src/redux/collectibles.ts
```diff
@@ -171,7 +176,7 @@ const fetchNFTsViaRpcNode = () => async (
   const { accountAddress, nativeCurrency, network } = getState().settings;
 
   const { assets } = (await getAssets(accountAddress, network)) as {
-    assets: Asset[];
+    assets: AssetType[];
```
danibonilha: I understand the usage of `AssetType` but I'd vouch for using `Asset` to avoid confusion with `AssetTypes` or renaming `AssetTypes` to something else

cardstack/src/services/gnosis-service.ts
```diff
@@ -50,51 +54,59 @@ export const fetchSafes = async (
...
+      merchantSafes.map(
+        async merchantSafe =>
+          ({
+            ...(await updateMerchantSafeWithCustomization(merchantSafe)),
+            ...(await updateSafeWithTokenPrices(merchantSafe, nativeCurrency)),
+            ...{
+              revenueBalances: await getRevenuePoolBalances(
+                merchantSafe.address,
+                nativeCurrency
+              ),
```
danibonilha: by combining this here we should probably add a try/ catch on `getRevenuePoolBalances`  and return some sort of default value, because if this call fails, no merchant will appear at all, before, we had already the merchant with prices, and customizations built, so in case we didn't get revenue, merchant data was send anyways

cardstack/src/services/gnosis-service.ts
```diff
@@ -50,51 +54,59 @@ export const fetchSafes = async (
...
-    const addMerchantSafesCustomization = Promise.all(
-      merchantSafes.map(updateMerchantSafeWithCustomization)
+    const extendAllDepotSafes = Promise.all(
+      depots.map(
+        async depot =>
+          ({
+            ...(await updateSafeWithTokenPrices(depot, nativeCurrency)),
+          } as DepotType)
+      )
     );
```
danibonilha:
```suggestion
   const extendAllDepotSafes = Promise.all(
      depots.map(depot => updateSafeWithTokenPrices(depot, nativeCurrency))
    );
```

danibonilha:
```suggestion
   const extendAllDepotSafes = Promise.all(
      depots.map(depot => updateSafeWithTokenPrices(depot, nativeCurrency))
    );
```

cardstack/src/services/gnosis-service.ts
```diff
@@ -50,51 +54,59 @@ export const fetchSafes = async (
...
+          ({
+            ...(await updateMerchantSafeWithCustomization(merchantSafe)),
+            ...(await updateSafeWithTokenPrices(merchantSafe, nativeCurrency)),
+            ...{
+              revenueBalances: await getRevenuePoolBalances(
+                merchantSafe.address,
+                nativeCurrency
+              ),
+            },
+          } as MerchantSafeType)
```
danibonilha: We shouldn't need to cast any of these types, if we are doing so, there's probably a mismatch we should handle

cardstack/src/services/gnosis-service.ts
```diff
@@ -50,51 +54,59 @@ export const fetchSafes = async (
...
-      addMerchantSafesCustomization,
+    const [
+      prepaidCardsTyped,
+      merchantSafesTyped,
+      depotsTyped,
+    ] = await Promise.all([
+      extendAllPrepaidCards,
+      extendAllMerchantSafes,
+      extendAllDepotSafes,
     ]);
```
danibonilha, line 83: wondering if we should use allSettled, to avoid not showing any safes in case of failure, but since it has always been like this, I guess it's ok to leave it as is and add a task to track it

cardstack/src/services/gnosis-service.ts
```diff
@@ -178,7 +190,7 @@ export const fetchGnosisSafes = async (address: string) => {
       });
     });
 
-    const { depots, prepaidCards, merchantSafes } = normalizeSafesByType(safes);
+    const { depots, prepaidCards, merchantSafes } = groupSafesByType(safes);
```
danibonilha: This `fetchGnosisSafes` function doesn't seem to be used anywhere besides one test

cardstack/src/utils/validators.ts
```diff
@@ -49,3 +49,9 @@ export const matchMinLength = (text: string, minLength: number): boolean =>
   text.length >= minLength;
 
 export const hasAtLeastOneDigit = (text: string): boolean => /\d/.test(text);
+
+// Util function to be used as a type safe substitute to .filter(Boolean)
+// usage: `array.filter(nonNullable)`
+export const nonNullable = <T>(value: T): value is NonNullable<T> => {
+  return value !== null && value !== undefined;
+};
```
danibonilha, line 61: Boolean matchs all falsy values for javascript like empty string and boolean itself, we should probably add this check too

src/handlers/localstorage/accountLocal.ts
```diff
@@ -0,0 +1,196 @@
...
+const transactionsVersion = '0.2.6';
+const collectiblesVersion = '0.2.4';
+
+const ACCOUNT_INFO = 'accountInfo';
+const ASSETS = 'assets';
+const PREPAID_CARDS = 'prepaidCards';
+const DEPOTS = 'depots';
+const MERCHANT_SAFES = 'merchantSafes';
+const PURCHASE_TRANSACTIONS = 'purchaseTransactions';
+const SAVINGS = 'savings';
```
danibonilha: Saving is not used anymore

src/handlers/localstorage/accountLocal.ts
```diff
@@ -0,0 +1,196 @@
...
+const TRANSACTIONS = 'transactions';
+const COLLECTIBLES = 'collectibles';
+
+export const accountLocalKeys = [
+  ACCOUNT_INFO,
+  ASSETS,
+  PURCHASE_TRANSACTIONS,
```
danibonilha: ditto for purchase_transactions

cardstack/src/services/prepaid-cards/prepaid-card-service.ts
```diff
@@ -36,6 +40,19 @@ export const addPrepaidCardCustomization = async (card: PrepaidCardSafe) => {
...
+  prepaidCard: PrepaidCardSafe,
+  nativeCurrency: NativeCurrency
+) => ({
+  // The order matters, first add new property then modify tokens
+  // otherwise tokens with prices will be overwritten by old tokens
+  ...(await addPrepaidCardCustomization(prepaidCard)),
+  ...((await updateSafeWithTokenPrices(
+    prepaidCard,
+    nativeCurrency
+  )) as PrepaidCardType),
```
danibonilha: I think we can remove this cast as well

cardstack/src/services/gnosis-service.ts
```diff
@@ -50,51 +54,59 @@ export const fetchSafes = async (
...
+      )
     );
 
-    const [extendedPrepaidCards, extendedMerchantSafes] = await Promise.all([
-      addPrepaidCardsCustomization,
-      addMerchantSafesCustomization,
+    const [
+      prepaidCardsTyped,
+      merchantSafesTyped,
+      depotsTyped,
```
danibonilha: Nit: I guess `extended[SafeType]`  or [SafeType]extended  makes more sense than [SafeType]Typed, since everything on typescript is typed

cardstack/src/services/gnosis-service.ts
```diff
@@ -50,51 +54,59 @@ export const fetchSafes = async (
...
 
-    const safesWithTokenPrices = await Promise.all(
-      safes?.map(safe => updateSafeWithTokenPrices(safe, nativeCurrency))
-    );
+    const { depots, prepaidCards, merchantSafes } = groupSafesByType(safes);
 
-    const { depots, prepaidCards, merchantSafes } = normalizeSafesByType(
-      (safesWithTokenPrices as unknown) as Safe[]
+    const extendAllPrepaidCards = Promise.all(
+      prepaidCards.map(async prepaidCard =>
```
danibonilha: I think we don't need the keyword `async` here

cardstack/src/services/gnosis-service.ts
```diff
@@ -50,51 +54,59 @@ export const fetchSafes = async (
...
+        extendPrepaidCard(prepaidCard, nativeCurrency)
```
danibonilha, line 54: nice reuse of `extendPrepaidCard`

cardstack/src/services/gnosis-service.ts
oidoug: PrepaidCardType doesn't really need the casting, removing.
MerchantSafeType works if some attributes are optional.

I'm having issues with `DepotType` thought, TS still thinks DepotSafe has a "prepaid-card" type 🤷 , casting for now.

cardstack/src/services/gnosis-service.ts
oidoug: I'm removing the function and test for now, I'll try add a test for fetchSafe before merging or add a card to add to do so.

cardstack/src/services/gnosis-service.ts
oidoug: Well, just tried and it complicates things a bit 😅 . I'll add a card to rework this later.

cardstack/src/services/gnosis-service.ts
oidoug: Good point 👍

cardstack/src/redux/collectibles.ts
oidoug: Using any other name to `AssetTypes` is odd, so I'm reverting back to `Asset`.

cardstack/src/services/gnosis-service.ts
danibonilha: In this case we can add a generic type to `updateSafeWithTokenPrices` which extends to Safe, then typescript will know the return of the function is the current received safe param which would be depot in this case plus the native price info. 

```ts
export const updateSafeWithTokenPrices = async <S extends Safe>(
  safe: S,
  nativeCurrency: NativeCurrency
) => {
```

image

cardstack/src/services/prepaid-cards/prepaid-card-service.ts
```diff
@@ -36,6 +39,16 @@ export const addPrepaidCardCustomization = async (card: PrepaidCardSafe) => {
...
+  prepaidCard: PrepaidCardSafe,
+  nativeCurrency: NativeCurrency
+) => ({
+  // The order matters, first add new property then modify tokens
+  // otherwise tokens with prices will be overwritten by old tokens
+  ...(await addPrepaidCardCustomization(prepaidCard)),
+  ...(await updateSafeWithTokenPrices(prepaidCard, nativeCurrency)),
+});
+
 export const getPrepaidCardByAddress = async (
```
danibonilha: This function is also not used anywhere, this was used when we had prepaidcards on redux, but we don't use that info anymore

danibonilha (approved):
Nice refactoring, just two small adjustments left and it's good to go! 

- Add [generics](https://github.com/cardstack/cardwallet/pull/1145#discussion_r1058372074) to updateSafeWithTokenPrices 
- Remove getPrepaidCardByAddress function

cardstack/src/services/gnosis-service.ts
oidoug: Nice, that does the trick.

PR #1146, opened 2022-12-27
[CS-5003] Add custom MATIC icon

paulinhapenedo:
### Description
This PR updated the MATIC token icon manually by adding it to the `CustomCoinIcons.ts` file.

A better approach would be to use the `tokenList` from the SDK and completely remove the `react-coin-icon` dependency, but at the moment we don't have a tokenlist file for Gnosis and Sokol. I'll be pairing with Luke to get the token information and create the tokenlist file for these chains.

- [x] Completes #CS-5003

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot


oidoug (approved):
LGTM 👍

danibonilha (approved):

PR #1150, opened 2022-12-29
[CS-5057]: Improve CI test and lint 

danibonilha:
### Description

This PR adds a new way to cache and flags yarn to use cache when possible.
- Fixes tests warning about act, erros, etc 
- Removes packages we don't need 
- Reorder deps to devDeps
- Removes styleling and cspell which were not being used

Test time on local machine should also be improved , mine was taking 300+ seconds now it's between 35-60seconds

### Screenshots

Tests on develop were around :
image

Now they've dropped to:
image

~~I'm running a release build to make sure everything is working as expected since there were changes to production stuff~~

Release was successful, PS: the release was triggered before rebasing with develop, so the pay tab crash is still present

Also iOS build dropped 15 min from build time


oidoug (approved):
Pretty good improvement 👏

package.json
```diff
@@ -18,20 +18,19 @@
...
+    "ios:small": "react-native run-ios --simulator='iPhone 12 mini'",
+    "lint": "eslint . --ext .ts,.tsx,.js --cache",
     "nodeify": "rn-nodeify --install --hack 'crypto,buffer,react-native-randombytes,vm,stream,http,https,os,url,net,fs,process'",
     "ts-check": "yarn tsc --skipLibCheck --noEmit",
     "prestorybook": "rnstl",
     "postinstall": "./scripts/postinstall.sh",
     "start": "yarn prestorybook && react-native start",
     "start:clean": "watchman watch-del-all && rm -rf $TMPDIR/react-* && rm -rf $TMPDIR/metro-* && rm -rf $TMPDIR/haste-map-* && react-native start -- --reset-cache",
-    "test": "APP_ENV=test jest --coverage",
+    "test": "APP_ENV=test jest --maxWorkers=50% --coverage",
```
oidoug, line 30: minor: maxWorkers=50% is good when using jest watch, but we could have another test command without limitations to be used when running tests before pushing code.

The pre-push commands are defined in `.husky/pre-push`

danibonilha: It improved on prepush for me significantly, not sure why, if it has any bad outcomes for you I can add a cmd locally just for me,  considering I'm the only one using intel

oidoug: No need, it's fine as is.

PR #1152, opened 2023-01-03
[CS-5058]: Remove js global variables and theme

danibonilha:
### Description

This PR removes the global variables, simplify eslint config and remove the theme context/provider, in favor of using colors directly. 

The app seems to be working fine, ~but for some weird reason, the rn-cloud tests are broken, so I'm investigating~


cardstack/src/screens/sheets/Collectible/components/CollectibleImageWrapper.tsx
```diff
@@ -33,7 +33,7 @@ const CollectibleImageWrapper = ({
         backgroundColor={collectible.background}
         imageUrl={imageUrl}
         item={collectible}
-        resizeMode="contain"
+        resizeMode={'contain' as any} // Casting to be used on js file
```
oidoug, line 36: Hum, just out of curiosity, what changed that this casting is now needed? BC it had the default to "cover" already and wasn't erroring before.

oidoug (commented):
PR looking good!
Seems like the file `buildTextStyles.js` could use a refactor :)

danibonilha:
> PR looking good! Seems like the file `buildTextStyles.js` could use a refactor :)

buildTextStyles is rainbow leftover, as soon as we are able to remove the components that uses it, we should get rid of this file. It's best to remove the old Text components than refactoring this file

oidoug:
> 

True, I didn't meant you should do it in this PR, just pointed it looks out-of-place. Would be nice to get rid of styled-components too.

cardstack/src/screens/sheets/Collectible/components/CollectibleImageWrapper.tsx
danibonilha: `CollectibleImage` was wrapped on magicMemo, which was not auto-inferring any types, once that was replace with `memo`, the type inference happened, since it defaults to `cover` typescript thinks the only allowed value is `cover` and doesn't accept `contain`

oidoug (approved):
LGTM, currency conversion seems to work fine too.

oidoug (commented):
Still all good

src/components/expanded-state/ContactProfileState.js
```diff
@@ -43,16 +44,21 @@ const ContactProfileState = ({ address, color: colorProp, contact }) => {
   const inputRef = useRef(null);
   const { network } = useAccountSettings();
 
+  const dismissKeyBoardOnAndroid = useCallback(() => {
+    Device.isAndroid && Keyboard.dismiss();
+  }, []);
+
```
paulinhapenedo: nit: couldn't we import the `dismissKeyBoardOnAndroid` we have on `presetOptions`?

danibonilha: Good call, totally forgot about that!

paulinhapenedo (approved):
Another great cleanup! 🧹🧼

PR #1153, opened 2023-01-05
[CS-5054]: Fix flicker on auto polling

danibonilha:
### Description

This PR adds the same behavior for other tokens as we have on safes while refreshing by showing a timestamp of the last updated value, or a loading indicator while refreshing. I've removed the prices from the refetch loading flag since it updates every 10s, and it doesn't seem valuable for the user to have a feedback about that, but I'm open to discussions.  I've also add timestamp handling functions, to avoid any confusions, we mostly use unix timestamp because the sdk uses it, so we need a way to handle js-to-unix and unix-to-js. Anyways, these implementations fixes the flickering since now the refresh feedback is different. It also fixes a refresh error I've found on non cardpay networks.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/20520102/210862054-953de470-c10b-482b-abf3-01428e732270.mp4

### DIFF
--- a/src/hooks/useAssetListData.tsx
+++ b/src/hooks/useAssetListData.tsx
@@ -51,6 +51,7 @@ const usePrepaidCardSection = (
     data: prepaidCards,
     timestamp,
     Component: PrepaidCard,
+    type: 'safe',
   };
 };
 
@@ -64,6 +65,7 @@ const useDepotSection = (
   data: depots,
   timestamp,
   Component: Depot,
+  type: 'safe',
 });
 
 const useMerchantSafeSection = (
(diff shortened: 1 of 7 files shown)


src/hooks/useAssetListData.tsx
```diff
@@ -51,6 +51,7 @@ const usePrepaidCardSection = (
     data: prepaidCards,
     timestamp,
     Component: PrepaidCard,
+    type: 'safe',
```
oidoug, line 54: nit: create a enum for these, could also reuse on `AssetList/types/AssetListSectionItem`

oidoug (approved):
LGTM, working as expected.

src/hooks/useAssetListData.tsx
danibonilha: It's already a type inside `AssetListSectionItem`:
image
 Not sure if an enum would bring much  more value, so I'll leave as is for now

PR #1157, opened 2023-01-09
[CS-3965] Remove language settings and libraries

paulinhapenedo:
### Description
This PR continues the cleanup work we're doing on the app by removing i18n libraries that we are not using at the moment. If and when the time comes for us to do i18n on the app, we can revisit this topic.

It also removes an unused file `/handlers/gnosis`.

- [x] Completes #CS-3965

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots
No visual changes


src/model/wallet.ts
```diff
@@ -182,14 +181,13 @@ export const sendTransaction = async ({
...
-      Alert.alert(lang.t('wallet.transaction.alert.failed_transaction'));
-      logger.sentry('Failed to SEND transaction, alerted user');
+      Alert.alert('Failed to send transaction');
+      logger.sentry('Failed to send transaction', error);
       captureException(error);
       return null;
     }
   } catch (error) {
-    Alert.alert(lang.t('wallet.transaction.alert.authentication'));
+    Alert.alert('Authentication Failed');
```
danibonilha: nit: maybe we could add a strings files for these "new" strings, specially because some of these errors are the  same

danibonilha (approved):
Cleanup looking good


Just a small nit

src/model/wallet.ts
paulinhapenedo: I'm not sure where to put this file. Is it ok to put it in the same folder as `wallet.ts` and use it to store the strings from both `wallet.ts` and `keychain.ts` (they both have a lot of strings)?

cc @danibonilha

oidoug (approved):
Looks good, and I agree with Dani, better move the strings still used to a separate file to conform with our current standard.

src/model/wallet.ts
oidoug: Maybe just extracting the strings to a const at the top is good enough.

danibonilha: > I'm not sure where to put this file. Is it ok to put it in the same folder as `wallet.ts` and use it to store the strings from both `wallet.ts` and `keychain.ts` (they both have a lot of strings)?
> 
> cc @danibonilha

@paulinhapenedo 
I guess so, they are both models, so it makes sense to have a strings files there to be shared between the models, we could split the context into the object itself like. 

```tsc
const strings = {
    shared: {
        authFailed: 'Authentication failed' 
    },
   wallet: {
     someStringOnlyforWalletModel: 'foobbar'
   }, 
  keychain: {
    someStringOnlyForKeychains: 'barFoo'
  }
}
```

PR #1158, opened 2023-01-09
[CS-5099] Use rewards parsed explanation from SDK

oidoug:
### Description

This PR refactor the app to use the parsed reward explanation that is now returned in each proof, when available.

SDK changes were merged and published today so this is ready to be merged as well.

- [x] Completes #(CS-5099)

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

Screenshot


danibonilha (approved):

paulinhapenedo (approved):
Nice!
Worth mentioning here for future reference that this SDK update also removes Mumbai from the network list as it's not being supported anymore.

PR #1160, opened 2023-01-10
[CS-5106]: Use single eslint config

danibonilha:
### Description

This PR merges the eslint  config in to a single one, instead of using external packages we add our own rules, unfortunately there were some any casting I needed to make in order for this to work, but that's a soon to be re-enabled rule so we can remove all the `any` from the project. 
It also removes immer and it completes CS-5113

image


paulinhapenedo (approved):
Wow 👏

PR #1162, opened 2023-01-12
[Cleanup]: Remove wyre auth widget

danibonilha:
### Description
Removing wyre leftover and webview package that was used only in it.


paulinhapenedo (approved):
Oops, my bad! 👏

oidoug (approved):

PR #1165, opened 2023-01-16
[CS-5114]: Opt-out React 18 and create single wallet instance

danibonilha:
### Description

React 18 changed how the useEffect hook works which seemed to have made the app a bit laggy sometimes, there's even an [issue](https://github.com/facebook/react-native/issues/35778) on react native's repo,  I'm not sure if the problem is an external package or the way we are handling some stuff in the app, But until we figure it out we can opt-out from react 18. If the new architecture is not enable the app continues to use React 17 as stated [here](https://reactnative.dev/docs/react-18-and-react-native#react-18-enabled-by-default). Android already has concurrent mode disabled, this PR disables it for iOS too. In order to avoid some expensive calls it also caches the wallet using the network and address as key, this way  there's no need for multiple storage calls and finally it removes a deprecated package that's been hanging, the react-navigation one, since it has been replaced by other individual packages.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### DIFF
--- a/cardstack/src/models/ethers-wallet.ts
+++ b/cardstack/src/models/ethers-wallet.ts
@@ -1,7 +1,9 @@
+import { supportedChainsArray } from '@cardstack/cardpay-sdk';
 import { ethers, utils, Wallet } from 'ethers';
 
 import {
   DEFAULT_HD_PATH,
+  loadAddress,
   loadPrivateKey,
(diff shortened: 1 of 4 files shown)


oidoug (approved):
LGTM, nice catch on the react-navigation v4 lost dependency.

PR #1168, opened 2023-01-17
[CS-5121]: Fix crash on wallet connect

danibonilha:
### Description

With the colors cleanup, we introduced a bug where dapps without icons were crashing, there was a function using an inner function that didn't exist, if the file was .ts we would catch this sooner, so this PR fixes the issue and migrates the colors file to typescript to prevent future mistakes.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/20520102/212938716-b25a987a-0264-4837-9a31-1bf1a6a2b614.mov


paulinhapenedo (approved):
Nice catch!

PR #1172, opened 2023-01-17
[CS-5130]: Fix pay merchant flow when amount is not set

danibonilha:
### Description
This PR completes CS-5131 and CS-5130, these are non-blocking issues but it's nice to include it in the release candidate.


oidoug (approved):
LGTM

PR #1173, opened 2023-01-18
[CS-4967] Improved build actions

oidoug:
### Description

This PR improves the DX of our Github workflows.

Now to bump a version we can confidently use Release workflow; it will push a PR with all version changes.
  
- Fixes version bump on both android and ios
- Sets tag for each platform with the correct build number (easier to find later)
- Simplifies release creation (old tag)
- Improves origin branch selection 
    - the drop-down defines both where the action runs from and the base branch it will use

#### Notes

* The action installs a more recent fastlane version than what we were using, this has been happening already without us noticing because the result of the action was never pushed back into develop.

- [x] Completes #(CS-4967)


oidoug:
This is finally working @danibonilha , @lukemelia.

danibonilha (approved):
Nice job!

PR #1177, opened 2023-01-20
[CS-5158]:  Add support for v4 signData and decode hubAuth

danibonilha:
### Description

While integrating safe-tools-app with wallet connect 2.0 I noticed the message was not appearing, nor it was being decoded. This PR refactors the signing-methods utils to .ts and add supports for v3 and v4 and also re-enable the decoding for messages that doesn't contain verifying contracts.

### Checklist

- [x] Tested on iOS
- [x] Tested on Android

### Screenshots

https://user-images.githubusercontent.com/20520102/213784582-f903ebed-034d-4f34-a0be-b1518b1ceb55.mov


oidoug (approved):
LGTM 👏

PR #1180, opened 2023-01-23
[Chore]: Fix ios build number release

danibonilha:
### Description

With the release's action title renaming the build number was reset for iOS and testflight doesn't accept older build numbers, this PR adds the latest build number as a constant so we can iterate. it also adds a function to build and release  that we can reuse since alpha and beta have the same config.

### DIFF
--- a/ios/fastlane/Fastfile
+++ b/ios/fastlane/Fastfile
@@ -1,19 +1,20 @@
 
 default_platform(:ios)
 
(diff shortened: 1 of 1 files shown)


oidoug (approved):
Ah, a silly mistake by me, I thought that bumping the version would mean the build number could be reset as well. Your solution is good.

